PR #9, opened 2019-05-24
Use Utf8 by default

damccorm:
We might decide we want to get rid of this option entirely, but I think it should be on by default regardless (aka doesn't need message from server to be on)


piyaphartp (commented):
Run

Issue #157, opened 2019-10-30
Extra output for every debug output


ericsciple:
When an action writes debug, every debug line produces 4 lines of output.

Here is sample log output from one debug statement from an action:

```
::debug::RUNNER_WORKSPACE = '/Users/eric/repos/actions/runner/_layout/_work/testing'
##[debug]Processing command 'debug'
##[debug]RUNNER_WORKSPACE = '/Users/eric/repos/actions/runner/_layout/_work/testing'
##[debug]Processed command 'debug' successfully
```

Out of the four lines, only the third one should exist.

PR #158, opened 2019-10-30
add-mask is leaking a secret in master if debug or ::echo::on is set

### LINKED ISSUE #159: Add Mask Command leaks secrets the secret being registered for masking
The `add-mask` command echoes the input, then processes the command, which means that when the input is echoed the secret is not registered. This causes the secret to leak.

thboop:
If debug variable is set or echo::on is used, add-mask will echo before the secret is registered and it will leak the secret.  This bug did not ship and was caught in the release branch

Resolves #159, #157


lurbanski-equinix:
The bug is live in situation when the ::add-mask:: is the first command in a step, then the command is showed. If the command is in second and so on lines, then it is correctly not shown.

PR #212, opened 2019-12-05
Remove old "v1" artifact download/publish code

juliobbv:
Remove "v1" (Build2) artifact download/publish API code from v1.0 action

Also, remove the entire Build2 REST API SDK, as it's no longer used by the runner with this PR.


juliobbv:
Merging PR, so changes will be incorporated with the next runner deployment in January.

PR #249, opened 2019-12-21
Treat warnings as errors

### LINKED ISSUE #179: Consider setting TreatWarningsAsError for all projects
After a few of the currently-open PRs are merged (#174, #178), all compiler and code analyzer warnings should be fixed. It would be good to set `TreatWarningsAsErrors` to true for all projects to ensure that contributors are forced to fix warnings before submitting a PR.

I recommend this setting be placed in the new `Directory.Builds.props` file that I included in this PR: https://github.com/actions/runner/pull/173/files#diff-6483fde6c45c90fdfe98ca72eadfa651

That way the setting will automatically be shared by all projects in that folder and sub-folders. Individual projects can override that setting if they want to (but they shouldn't!).

casz:
fixes #179


src/Test/L0/ProcessInvokerL0.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
casz: Argh these are encoding with UTF-8 BOM. Properly best to add an editorconfig 🤔

TingluoHuang:
Thanks! @casz

Issue #254, opened 2020-01-02
Upload unfinished logs from runner when worker doesn't handle cancellation and get killed


TingluoHuang:
Thank you 🙇‍♀ for wanting to create a feature in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions or issues with the service, writing workflows or actions, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

If you have a feature request that is relevant to this repository, the runner, then please include the information below:

**Describe the enhancement**

When user cancel a workflow run, the worker process has about 5 mins to response the cancellation and finish what it is doing and exit, otherwise the runner will kill the worker process.

When the process kill happen, we will lost some job/step logs since worker may not finish upload those.

It would be great if the runner can finish upload those logs, since those logs may help us to debug why the cancellation time out and what is hanging.

**Code Snippet**
If applicable, add a code snippet.

**Additional information**
Add any other context about the feature here.

NOTE: if the feature request has been agreed upon then the assignee will create an ADR.   See docs/adrs/README.md

PR #255, opened 2020-01-02
Upload unfinished job/step logs from runner on cancellation timeout

TingluoHuang:
#254

### DIFF
--- a/src/Runner.Common/Logging.cs
+++ b/src/Runner.Common/Logging.cs
@@ -24,7 +24,6 @@ namespace GitHub.Runner.Common
 
         private Guid _timelineId;
         private Guid _timelineRecordId;
-        private string _pageId;
         private FileStream _pageData;
         private StreamWriter _pageWriter;
         private int _byteCount;
@@ -40,7 +39,6 @@ namespace GitHub.Runner.Common
         {
             base.Initialize(hostContext);
             _totalLines = 0;
-            _pageId = Guid.NewGuid().ToString();
             _pagesFolder = Path.Combine(hostContext.GetDirectory(WellKnownDirectory.Diag), PagingFolder);
             _jobServerQueue = HostContext.GetService<IJobServerQueue>();
             Directory.CreateDirectory(_pagesFolder);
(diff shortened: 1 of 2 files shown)


src/Runner.Common/Logging.cs
```diff
@@ -40,7 +39,6 @@ public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _totalLines = 0;
-            _pageId = Guid.NewGuid().ToString();
```
TingluoHuang: we don't really need a new guid for each page, we can leverage the timeline record id for each steps, those are already unique per step.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -712,6 +716,116 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
...
+
+            var logs = Directory.GetFiles(logFolder);
+            if (logs.Length == 0)
+            {
+                return;
+            }
+
+            try
+            {
+                var systemConnection = message.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection));
```
TingluoHuang: pull the cred from the job message, we will use this cred to talk to actions service and upload the log.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -712,6 +716,116 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
...
+
+                        if (!int.TryParse(logName.Substring(logPageSeperator + 1), out pageNumber))
+                        {
+                            Trace.Warning($"log file '{log}' doesn't follow naming convension 'GUID_INT'.");
+                            continue;
+                        }
+                    }
+
+                    var record = timeline.Records.FirstOrDefault(x => x.Id == logRecordId);
+                    if (record != null)
```
TingluoHuang: We only upload log files to timeline records for this job.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -712,6 +716,116 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
...
+            var logs = Directory.GetFiles(logFolder);
+            if (logs.Length == 0)
+            {
+                return;
+            }
+
+            try
+            {
+                var systemConnection = message.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection));
+                ArgUtil.NotNull(systemConnection, nameof(systemConnection));
```
thboop: Should we throw here, or should we just return because its best effort?

PR #263, opened 2020-01-09
proxy support ADR

bryanmacfarlane:
self hosted runners use proxies


bryanmacfarlane:
@TingluoHuang - copying this over so the oss community can see the background and decisions.

Issue #262, opened 2020-01-09
Setting and targeting based on custom labels


nodesocket:
There is mention of labels (https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-self-hosted-runners-in-a-workflow). How can I set a custom label on a actions runner instance and target it? For example we have a staging and production actions runner. If the branch = master we want to use the production actions runner, else we want to use the staging actions runner.

R2ZER0:
A perhaps easier starting point could be to add the runner name as a label, so you could do something like:
```yaml
runs-on: [self-hosted, my-runner-name]
```

bryanmacfarlane:
Custom labels are on the short backlog.  Hopefully coming soon.

jonchang:
Homebrew has forked the GitHub Actions runner to support custom labels targeting by macOS version. It'd be great if this feature gets added to the official runner!

https://github.com/Homebrew/gha-runner

rubencwi:
This is a critical functionality to be able to differentiate runners between environments or geo locations

timharris777:
> Homebrew has forked the GitHub Actions runner to support custom labels targeting by macOS version. It'd be great if this feature gets added to the official runner!
> 
> https://github.com/Homebrew/gha-runner

This is great!! I just looked at the code changes. I think I'll be using this fork . Hoping github does this soon as it is critical functionality for us to adopt github actions.

TingluoHuang:
@MalloZup I think the feature is out there and you can take a try.
https://help.github.com/en/actions/hosting-your-own-runners/using-labels-with-self-hosted-runners

MalloZup:
@TingluoHuang thx a lot! :cupid:

davidkarlsen:
```
Unrecognized command-line input arguments: 'labels'. For usage refer to: .\config.cmd --help or ./config.sh --help
./config.sh --version
2.169.1
```

davidkarlsen:
Nothing in cmdline help:
```
runner@89ba235fdc6c:~$ ./config.sh --help

Commands:
 ./config.sh         Configures the runner
 ./config.sh remove  Unconfigures the runner
 ./run.sh            Runs the runner interactively. Does not require any options.

Options:
 --help     Prints the help for each command
 --version  Prints the runner version
 --commit   Prints the runner commit

Config Options:
 --unattended     Disable interactive prompts for missing arguments. Defaults will be used for missing options
 --url string     Repository to add the runner to. Required if unattended
[...]

PR #270, opened 2020-01-13
Add the runner name as a label

R2ZER0:
This PR adds the runner name as a label, so that it can be selected by it's name in a workflow, e.g.:
```yaml
runs-on: [self-hosted, my-runner-name]
```
Use cases for this being any time when you have multiple runners attached to a single repo, but need to distinguish between them. For example, you could have a runner with a GPU for testing machine learning models, and another without a GPU for general loads - this would enable you to run your ML tests specifically on the runner with the GPU.

This could be thought of as a first-step, "minimum viable solution" for some of the use cases for custom labels that would allow more people to get started sooner - https://github.com/actions/runner/issues/262

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -472,6 +472,7 @@ namespace GitHub.Runner.Listener.Configuration
             agent.Labels.Add("self-hosted");
             agent.Labels.Add(VarUtil.OS);
             agent.Labels.Add(VarUtil.OSArchitecture);
+            agent.Labels.Add(agent.Name);
 
             return agent;
         }
@@ -492,6 +493,7 @@ namespace GitHub.Runner.Listener.Configuration
             agent.Labels.Add("self-hosted");
             agent.Labels.Add(VarUtil.OS);
             agent.Labels.Add(VarUtil.OSArchitecture);
+            agent.Labels.Add(agent.Name);
 
             return agent;
         }


bryanmacfarlane:
We already discussed with product earlier.  Product wants custom labels and doesn't to automatically create any labels (including machine name) outside of os and arch.  It can cause unintended collisions for folks that don't want to necessarily route.  @chrispat from product

bryanmacfarlane:
Custom labels are coming soon and that will allow folks that do want to route at the machine level to do that (or even create a label for two machines etc).

chrispat:
@R2ZER0 for your scenarios I would think some sort of capability labeling would be more appropriate than a name of the runner.  I think it would be useful to enable you to provide a lit of labels you want to add to the runner as part of the registration.

bryanmacfarlane:
Yeah, we could pull from a file / args like we do with envvars

R2ZER0:
I agree that proper custom labels would be a much better solution, that would support this plus a wider set of use cases. This PR is an interim solution that at least enables some runner selection until full support for custom labels arrives. Unfortunately I'm not much of a C-sharper.

A suggestion, how about adding custom labels as a command-line argument, 
e.g. `./config.sh --label my-custom-label --url ...`

bryanmacfarlane:
Yes.  I think that's where @chrispat and were going - an interim solution in the middle which only requires runner changes.  I would suggest command line arg, pull from file combo similar to what we do with envvars.

bryanmacfarlane:
Per the contribution guide, we should proceed with an ADR with a quick write up.  Let me know if you want to write it ...

R2ZER0:
I'm happy to write it if it can wait until tomorrow morning

bryanmacfarlane:
That's fine.  Looking forward to it.  No rush but ❤️ contributions especially in direction ;)

R2ZER0:
Okay I've made a seperate PR for the ADR https://github.com/actions/runner/pull/275
Let me know any thoughts :)
I guess we can close this PR.

PR #274, opened 2020-01-14
adr: step outcome

ericsciple:
👉 👉 [Rendered](https://github.com/actions/runner/blob/users/ericsciple/m164stepadr/docs/adrs/0274-step-outcome.md) 👈 👈 

related to https://github.com/actions/checkout/issues/105

### DIFF
--- a/docs/adrs/0274-step-outcome-and-conclusion.md
+++ b/docs/adrs/0274-step-outcome-and-conclusion.md
@@ -1,0 +1,62 @@
+# ADR 0274: Step outcome and conclusion
+
+**Date**: 2020-01-13
+
+**Status**: Accepted
+
+## Context
+
+This ADR proposes adding `steps.<id>.outcome` and `steps.<id>.conclusion` to the steps context.
+
+This allows downstream a step to run based on whether a previous step succeeded or failed.
+
+Reminder, currently the steps contains `steps.<id>.outputs`.
+
+## Decision
+
+For steps that have completed, populate `steps.<id>.outcome` and `steps.<id>.conclusion` with one of the following values:
+
+- `success`
+- `failure`
+- `cancelled`
+- `skipped`
+
+When a continue-on-error step fails, the outcome will be `failure` even though the final conclusion is `success`.
+
+### Example
+
+```yaml
+steps:
+
+  - id: experimental
+    continue-on-error: true
+    run: ./build.sh experimental
+
+  - if: ${{ steps.experimental.outcome == 'success' }}
+    run: ./publish.sh experimental
+```
+
+### Terminology
+
+The runs API uses the term `conclusion`.
+
+Therefore we use a different term `outcome` for the value prior to continue-on-error.
+
+The following is a snippet from the runs API response payload:
+
+```json
+      "steps": [
+        {
+          "name": "Set up job",
+          "status": "completed",
+          "conclusion": "success",
+          "number": 1,
+          "started_at": "2020-01-09T11:06:16.000-05:00",
+          "completed_at": "2020-01-09T11:06:18.000-05:00"
+        },
+```
+
+## Consequences
+
+- Update runner
+- Update [docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions#steps-context)


docs/adrs/0274-step-conclusion.md
```diff
@@ -0,0 +1,43 @@
+# ADR 0274: Step conclusion
+
+**Date**: 2020-01-13
+
+**Status**: Accepted
+
+## Context
```
ericsciple: @chrispat do you have any good scenarios offhand to add to the ADR?

chrispat: The main scenario is running another step if a specific step passes or fails.

ericsciple: do you have a good example? i'm looking to add a good example in the adr

chrispat: How about a step that generates an artifact if if succeedes but doesn't if it fails and then you want to condition the upload artifact on that conclusion.

ericsciple: @chrispat wouldnt the job be in a failed state at that point? so downstream upload artifact step wouldnt run anyway

ericsciple:
marking as draft... need to think a little more. was thinking about exposing errors since continue-on-error flips conclusion to success but that has different problem since error annotation doesnt necessarily mean failure. may be fine... just want to think if there is a different approach... also following up on a few different adrs, will come back to this one later (lower priority).

docs/adrs/0274-step-conclusion.md
TingluoHuang: what about?
```
steps:
- id: genpkg
  uses: myorg/genpkg@v1
- id: testpkg
  uses: myorg/testpkg@v1
- id: uploadpkg
  if: steps.genpkg.conclusion == success
  uses: actions/upload-artifact@v1
```

ericsciple: @TingluoHuang Why even bother with the condition then? Without a condition, if `genpkg` fails then downstream steps without a condition will not execute.

In order to meet the scenarios, due to continue-on-error flipping failed to success, I think exposing issues is the key.

So this scenario would work:

```yaml
- id: experimental-proj
  continue-on-error: true
  run: ./build.sh experimental-proj
- if: ${{ !steps.experimental-proj.issues[0] }}
  run: ./publish.sh experimental-proj
```

chrispat: What if I want to explicitly run a step if a previous step was skipped?

ericsciple: Good point. Sorry I misspoke. I think exposing conclusion _and_ issues is the key to covering these types of scenarios.

ericsciple: Although technically i guess you could use the same condition the previous step did, if you also want to skip :P

TingluoHuang: `skipped` only caused by `if` return `false`, in that case you can just add `!` to the condition that skipped the step?

```
- run: printenv
  if: whatever condition
- run: ls -l
  if: !(whatever condition)
```

ericsciple: Pushed updates to the ADR with `outcome` since the continue-on-error value is more useful.

Ting and I played with what adding issues to the context would look like. It adds a lot of complexity for users to deal with.

We feel `outcome` (pre continue-on-error value) solves the most common scenarios.

ericsciple:
removed from draft

docs/adrs/0274-step-conclusion.md
ericsciple: @chrispat for thoughts about this latest proposal ^

ericsciple:
@chrispat for thoughts about this latest proposal

chrispat:
Why not add conclusion in addition to outcome so can potentially make a determination on the failure of a specific step?

ericsciple:
I like that. Will update the ADR.

xt0rted:
I have two workflows that need the ability to exit the job with a neutral or skipped state if an early step doesn't run to completion. Failing the job creates unneeded notifications or incorrect statuses (email, web, and slack), and adding checks to every step after them is very verbose. Will this proposal let me mark the outcome or conclusion as cancelled/skipped so the later steps won't run unless explicitly set using an `if` condition?

One workflow is checking comments for `/commands` and later steps only need to run if the right value is found and the user has the right permission level. The other is for linting where if fixes were made they need to be committed and pushed back.

ericsciple:
got an approval from ting offline

ericsciple:
@xt0rted unfortunately no this proposal is only for making a step's conclusion available via the [`steps` expression context](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions)

PR #275, opened 2020-01-14
Add ADR for config time custom label support

R2ZER0:
Add a proposed ADR for adding custom label support

The runner already adds labels during configuration.   By adding a simple cli arg, automating the addition of self hosted runners with the proper labels

It also avoids a race condition between registering, potential jobs starting and adding labels.  It ensures, when it comes online, it's with the proper configuration.

## Scenarios:

**Creating your own containers with runners in it for your org**:  If an org is going to create a container with the runner in it (or a VM for that matter) to standardize the use of self hosted runners, not only the runner, tools and config can be packaged up, but also how it advertises itself when it comes online.   You wouldn't want to have to go to the web ui or require the script that's starting the container to have to know about that internals.

**Outside / in automation of setting up runners**: if you're scripting adding runners, it allows a simple arg addition to ensure it comes online with the right labels.

### DIFF
--- a/docs/adrs/0262-custom-label-support.md
+++ b/docs/adrs/0262-custom-label-support.md
@@ -1,0 +1,43 @@
+# ADR 262: Custom Label Support
+**Date**: 2020-01-14
+
+**Status**: Proposed
+
+## Context
+
+In the case of a single repo having multiple self-hosted runners, actions do not currently support any way of selecting which runner or kind of runner a job should run on, other than by architecture or OS.
+
+Example use cases for this could be:
+
+* Your project involves some machine learning models. You want to add a second runner with a GPU for testing your models, and to have your model tests job use this runner, but other jobs to use the runner without a GPU. 
+* You want to use GitHub actions to deploy your app, and for security you use a self-hosted runner inside your isolated network. If you have a separate isolated network for each environment (e.g. testing, staging, production), you would need to run your deploy job on a different runner depending on which environment you are deploying to.
+* Any other situation where you would like to select which self-hosted runner to use for a job based on the capabilities or characteristics of the runner.
+
+GitHub actions already support selecting a job runner based on labels, and automatically give runners labels based on OS and arch, so a natural extension would be to allow operators of self-hosted runners to add their own labels, so that users can select job runners based on any operator-defined characteristic.
+
+See Issue: https://github.com/actions/runner/issues/262
+
+## Decision
+
+This ADR proposes that we add a `--label` option to `config.sh`, which could be used to add custom additional labels to the configured runner.
+
+For example, to add a single extra label the operator could run:
+```bash
+./config.sh --label my-extra-label
+```
+Which would add the label `my-extra-label` to the runner, and enable users to select the runner in their workflow using this label:
+```yaml
+runs-on: [self-hosted, my-extra-label]
+```
+
+To add multiple labels:
+```bash
+./config.sh --label my-extra-label --label yet-another-label
+```
+This would add both labels `my-extra-label` and `yet-another-label`.
+
+## Consequences
+
+The ability to add custom labels to a self-hosted runner would enable most scenarios where job runner selection based on runner capabilities or characteristics are required.
+
+There could be unexpected behavior if an operator adds inappropriate labels, e.g. they add the label `linux` to a Windows runner, but this would be considered operator error.
--- a/docs/adrs/0275-custom-label-support.md
+++ b/docs/adrs/0275-custom-label-support.md
@@ -1,4 +1,4 @@
-# ADR 262: Custom Label Support
+# ADR 275: Custom Label Support
 **Date**: 2020-01-14
 
 **Status**: Proposed
--- a/docs/adrs/0275-support-adding-custom-labels-during-config.md
+++ b/docs/adrs/0275-support-adding-custom-labels-during-config.md
@@ -1,4 +1,4 @@
-# ADR 275: Custom Label Support
+# ADR 275: Support adding custom labels during runner config
 **Date**: 2020-01-14
 
 **Status**: Proposed
--- a/docs/adrs/0275-support-adding-custom-labels-during-config.md
+++ b/docs/adrs/0275-support-adding-custom-labels-during-config.md
@@ -36,6 +36,8 @@ To add multiple labels:
 ```
 This would add both labels `my-extra-label` and `yet-another-label`.
 
+It would not be possible to remove labels from an existing runner using `config.sh`, instead labels would have to be removed using the GitHub UI.
+
 ## Consequences
 
 The ability to add custom labels to a self-hosted runner would enable most scenarios where job runner selection based on runner capabilities or characteristics are required.


R2ZER0:
There's a couple of things I'm not sure about:
* What should loading labels from a file look like? Maybe a `--labels-file <file>` option, and what would the format of that file be?
* Do we need a `--remove-label <label>` option for reconfiguring existing runners?

docs/adrs/0262-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+
+Example use cases for this could be:
+
+* Your project involves some machine learning models. You want to add a second runner with a GPU for testing your models, and to have your model tests job use this runner, but other jobs to use the runner without a GPU. 
+* You want to use GitHub actions to deploy your app, and for security you use a self-hosted runner inside your isolated network. If you have a separate isolated network for each environment (e.g. testing, staging, production), you would need to run your deploy job on a different runner depending on which environment you are deploying to.
+* Any other situation where you would like to select which self-hosted runner to use for a job based on the capabilities or characteristics of the runner.
+
+GitHub actions already support selecting a job runner based on labels, and automatically give runners labels based on OS and arch, so a natural extension would be to allow operators of self-hosted runners to add their own labels, so that users can select job runners based on any operator-defined characteristic.
+
```
bryanmacfarlane: Another scenario for config time labels is someone automating setting up self-hosted runners (VMs, K8s, etc.).  When it's automated, there should be a way to automate applying labels during the automated configuration.  That means that it's more than an interim solution - even if we have the full blown way to add, edit, delete from the web UI (coming) this would still be needed.

docs/adrs/0262-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+./config.sh --label my-extra-label
+```
+Which would add the label `my-extra-label` to the runner, and enable users to select the runner in their workflow using this label:
+```yaml
+runs-on: [self-hosted, my-extra-label]
+```
+
+To add multiple labels:
+```bash
+./config.sh --label my-extra-label --label yet-another-label
```
bryanmacfarlane: @TingluoHuang do you think this should be one option with a delimited list.  I just want to make sure that our cli handling code doesn't have issues with multiple args with the same name.

docs/adrs/0262-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+```
+This would add both labels `my-extra-label` and `yet-another-label`.
+
```
bryanmacfarlane: Address remove label.  I think it should only be a web UI action since config is typically a one time operation.  That does leave a gap until we have web UI support.  In that gap, the (not ideal) workaround would be to unconfigure and reconfigure.  On configure, I think we should rewrite the set.  That means there's no cli arg to remove.

bryanmacfarlane:
Overall, LGTM.  I left a few comments to address.

docs/adrs/0262-custom-label-support.md
chrispat: I actually prefer the model of multiple args for the automation scenario.  It is easier to generate and requires less escaping.

docs/adrs/0262-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
+# ADR 262: Custom Label Support
```
bryanmacfarlane: rename this and file to 275.  ADR number is PR number so easy to find and no collisions. :)

docs/adrs/0262-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+* You want to use GitHub actions to deploy your app, and for security you use a self-hosted runner inside your isolated network. If you have a separate isolated network for each environment (e.g. testing, staging, production), you would need to run your deploy job on a different runner depending on which environment you are deploying to.
+* Any other situation where you would like to select which self-hosted runner to use for a job based on the capabilities or characteristics of the runner.
+
+GitHub actions already support selecting a job runner based on labels, and automatically give runners labels based on OS and arch, so a natural extension would be to allow operators of self-hosted runners to add their own labels, so that users can select job runners based on any operator-defined characteristic.
+
+See Issue: https://github.com/actions/runner/issues/262
+
+## Decision
+
+This ADR proposes that we add a `--label` option to `config.sh`, which could be used to add custom additional labels to the configured runner.
```
chrispat: Should we add a file option as well like we have for ENV?

bryanmacfarlane: Yeah, I like that as well.  If we do it should be by convention like env.  That means create a file name labels (or is it .labels?) - whatever env is - and it just works.  We should apply envvars first, and then process each line in the file for envvars (${HOSTNAME}) and trim.

docs/adrs/0262-custom-label-support.md
TingluoHuang: we need to make code change to the runner to support multiple args with same name.

docs/adrs/0262-custom-label-support.md
TingluoHuang: Another way is to use file, user put labels into `.labels` file under runner root, the runner pick them up during runner config and session creation, so if user want to change labels after config, they just need to change the file and restart the runner process.

docs/adrs/0262-custom-label-support.md
TingluoHuang: Should we consider only have one way of doing this? I prefer file over cmdline args. 😄

thejoebourneidentity: I really like the file idea. Can you point me to an example of this working for envvars?

docs/adrs/0262-custom-label-support.md
thejoebourneidentity: That would get interesting depending on how the UI will work. I need to get some design progress on that. Hopefully by next week I can have something more fleshed out.

docs/adrs/0262-custom-label-support.md
R2ZER0: Good point! In fact, automated deployment of the runner to kubernetes is exactly what I'm doing :)

docs/adrs/0262-custom-label-support.md
R2ZER0: The inspiration for using with multiple args with came from docker, where you use the style
```bash
docker run --label my-label-1 --label my-label-2 ...
```
and it allows for easier scripting if you are drawing your args from multiple places, e.g. it's easy to add more labels by just concatenating the args:
```bash
./config.sh --label hardcoded-label "$PROJECT_ARGS" "$@"
```
If you have to put all the labels in a single list, you'd have to keep track of them separately. I don't think it's a massive issue though.

docs/adrs/0262-custom-label-support.md
R2ZER0: From my perspective in docker/kubernetes-land, it's easier to automate config using commandline options than using files. When deploying a container, changing the commandline args is just a change in the deployment config, but changing a file involves either rebuilding your container image and baking-in the settings or writing a startup wrapper script that generates the file from env vars/commandline.
Not so sure about hardcoding the path to the file, but I like having the option to load from a file. Docker similarly has a `docker run --label-file path/to/labels.txt ...` option which loads a newline-seperated set of labels. Using this, an organisation could make a VM/container image with their different labels files in it (say `labels.testing.txt`, `labels.staging.txt`, `labels.prod.txt`) and select which one to use with `--label-file` at deploy time.

Penagwin: @bryanmacfarlane 
If you want to support other ways to set the settings then I'd model it after something like https://github.com/spf13/viper - not that you'd use that library but you might want to look at how it loads settings. 

A command line argument is kinda a necessity IMO. It would be bizarre not to have that option.

bryanmacfarlane: A bit of background on why we have a file for envvars.  When the runner is run as a service the service file is generated (linux or mac) which runs svc.sh and we didn't want users to have to edit the file (post config).  There's also scenarios to modify and just restart the svc.  All that said, it's post config modifications to the env which really doesn't apply here.  So, if we couple that with Joe's principle of just having one, then that means we can just have cmd line config option and that's it.  There's no strong scenario like there is for runtime envvars and restart since config is a one time only operation.  The container point is also a good one where it can just be passed through (instead of having to modify files).  So I think taking all that into consideration means only cmd line is good.

docs/adrs/0275-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+```bash
+./config.sh --label my-extra-label --label yet-another-label
+```
+This would add both labels `my-extra-label` and `yet-another-label`.
+
+## Consequences
+
+The ability to add custom labels to a self-hosted runner would enable most scenarios where job runner selection based on runner capabilities or characteristics are required.
+
+There could be unexpected behavior if an operator adds inappropriate labels, e.g. they add the label `linux` to a Windows runner, but this would be considered operator error.
```
thboop: We should consider not allowing the user to add the well-known labels to a runner as a custom label. I don't think users will intentionally do this, but in the case of copy-paste error or a user running the wrong script will result in some confusion.

docs/adrs/0275-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
+# ADR 275: Custom Label Support
```
TingluoHuang: Might want to change the title and file name to something like: `Support adding custom labels during runner config`
`Custom Label Support` is the overall name for custom runner labels feature we are doing which also include UI experience in github.com and documentation.

TingluoHuang:
@bryanmacfarlane for final review.

thejoebourneidentity:
This is shaping up well @TingluoHuang. I just posted new product plan additions here: https://github.com/github/c2c-actions/pull/578

Let's review and see how that impacts the ADR, if at all.

docs/adrs/0275-custom-label-support.md
```diff
@@ -0,0 +1,43 @@
...
+./config.sh --label my-extra-label
+```
+Which would add the label `my-extra-label` to the runner, and enable users to select the runner in their workflow using this label:
+```yaml
+runs-on: [self-hosted, my-extra-label]
+```
+
+To add multiple labels:
+```bash
+./config.sh --label my-extra-label --label yet-another-label
```
TingluoHuang: I just realized we also support pass args via env to avoid command line invocation been log to system log in the certain OS/Setting.
https://github.com/actions/runner/blob/master/src/Runner.Listener/CommandSettings.cs#L83

We might need to make `--label` read the env as JSON array. 😄

docs/adrs/0275-custom-label-support.md
R2ZER0: Sure thing, i'll change it

docs/adrs/0275-custom-label-support.md
R2ZER0: There could be situations where it is valid to add these well-known labels - for example: you have a runner on x86 system, but the system has binfmt setup to allow seamless execution of arm binaries via qemu, so that you can run unit tests on applications built for resource-constrained arm devices like the raspberry pi, where running the tests on actual hardware would be very slow. In this case the system is capable of executing both x86 and arm binaries. You can have a similar setup in windows subsystem for linux, where both windows and linux binaries are executable.

docs/adrs/0262-custom-label-support.md
R2ZER0: I've added a line about removing labels: 
`It would not be possible to remove labels from an existing runner using `config.sh`, instead labels would have to be removed using the GitHub UI.
`

docs/adrs/0275-custom-label-support.md
R2ZER0: Are labels potentially considered secret, and that's why they shouldn't be logged, or is there another reason?

docs/adrs/0275-custom-label-support.md
bryanmacfarlane: r2zero is correct, this was an explicit goal of the initial label support.  An org level admin to explicitly overriding the use of a hosted pool in his org was a scenario.  @chrispat

docs/adrs/0262-custom-label-support.md
bryanmacfarlane: We should be consistent within the runner and offer cmd line args and picking up from a file similar to our `.env`.  Please add to the ADR.

docs/adrs/0275-custom-label-support.md
thboop: That makes sense to me  👍

bryanmacfarlane:
I reviewed the code and I think we should take

```bash
--labels one,two
```

as opposed to 
```bash
--label one --label two
```

The way the runner currently works is it allows envvar overrides and the parser parses args to string, string dictionary.

The first option is consistent with the current behavior and is virtually free to implement.  The second would have us overhaul the parser and current arg patterns and it would be inconsistent with the envvar override current mechanisms (not even sure how we would do it).

bryanmacfarlane:
☝️ I also followed up with product @thejoebourneidentity and @TingluoHuang - we're going to go with `--labels one,two` which also aligns with the envvar override value of `one,two`.  Because of the latter, we have to handle splitting, trimming and potentially escaping commas if that is in fact a scenario.  I'll follow up on whether labels are alpha numeric or what symbol set we want to support.

docs/adrs/0275-custom-label-support.md
bryanmacfarlane: I don't think so.  We do however have a general pattern of allowing envvars for other args which may be secret (some OSes log cli invocations in system and event logs).  The only potential secret here would be the token and even that is time bombed.

We wouldn't remove that capability for this one arg - it's a general runner capability to be able to supply any input as an envvar.

docs/adrs/0262-custom-label-support.md
bryanmacfarlane: I'm also not proposing or referring to other ways to potentially set settings.  I'm outlining how the existing runner applies settings from file, envvars, and then override with args or prompt.  e.g. we're not rethinking the settings system here - we're just adding one more arg.

bryanmacfarlane:
@R2ZER0 - our crews are now at the point where we're implementing these features.   Any chance you can take another pass and update with all the comments?  If you're busy that's OK and we can fork / copy or move ahead.  

I think it's just a matter of:
- adding scenarios
- update to `--labels`
- mention that we will be consistent with existing arg patterns ( envvar override, file) 

Let me know how we can help.

EricWF:
@bryanmacfarlane Would you accept a different version of this ADR + implementation?
(Would it speed up the process?)

bryanmacfarlane:
Closing in favor of: https://github.com/actions/runner/pull/397 

Retaining discussions and history here.  I referenced this PR in that one.

PR #276, opened 2020-01-14
ADR: problem matchers

ericsciple:


bryanmacfarlane:
I know it will be a pain :grimacing: , but can we have a PR for each with the PR == ADR# so we're consistent?

ericsciple:
copied ADR to this repo

PR #277, opened 2020-01-14
ADR: Run action shell options

ericsciple:


ericsciple:
copied ADR to this repo

PR #278, opened 2020-01-14
adr: add env context

ericsciple:


ericsciple:
copied ADR to this repo

PR #279, opened 2020-01-14
adr: hashFiles expression function

ericsciple:


ericsciple:
copied ADR to this repo

PR #284, opened 2020-01-16
Update Base64 Encoders to deal with suffixes

### LINKED ISSUE #291: Base64 Masking of Secrets should account for appending strings
The base 64 encoder handles the cases of base64(secret) and base64(prefix + secret) pretty well. However, the case of base64(secret+suffix) does not get masked in some cases.

For example, consider the workflow
```
name: Node CI

on: [pull_request,push]

jobs:
  testsecret:
    runs-on: ubuntu-latest
    steps:
    - run: |
        echo ${{secrets.abc}} | base64
        echo ${{secrets.abc}}
        echo hello
```
We set the secret `abc` to `1234`

In our logs we see:
```
MTIzNAo=
***
hello
```
`MTIzNAo=` is the string `1234\n` converted to base 64. It appears in the log because we register the string `MTIzNA==` as a secret. We should register `MTIzN` as a secret in this case, so this will get partially masked.

We should:
- Trim `=` from base64 encoded versions of secrets, as it may prevent us from getting matches when we should. (`=` is a base64 padding character that gets removed when you append something to the string)
- Trim the last character from a base64 string that contains `=` as it may change

thboop:
Resolves #291

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -83,6 +83,7 @@ namespace GitHub.Runner.Common
             _loadContext.Unloading += LoadContext_Unloading;
 
             this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscape);
+            this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscapeTrimmed);
             this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscapeShift1);
             this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscapeShift2);
             this.SecretMasker.AddValueEncoder(ValueEncoders.ExpressionStringEscape);
--- a/src/Sdk/DTLogging/Logging/ValueEncoders.cs
+++ b/src/Sdk/DTLogging/Logging/ValueEncoders.cs
@@ -16,6 +16,11 @@ namespace GitHub.DistributedTask.Logging
         {
             return Convert.ToBase64String(Encoding.UTF8.GetBytes(value));
         }
+        
+        public static String Base64StringEscapeTrimmed(String value)
+        {
+            return TrimBase64End(Convert.ToBase64String(Encoding.UTF8.GetBytes(value)));
+        }
 
         // Base64 is 6 bits -> char
         // A byte is 8 bits
@@ -67,11 +72,11 @@ namespace GitHub.DistributedTask.Logging
             {
                 var shiftArray = new byte[bytes.Length - shift];
                 Array.Copy(bytes, shift, shiftArray, 0, bytes.Length - shift);
-                return Convert.ToBase64String(shiftArray);
+                return TrimBase64End(Convert.ToBase64String(shiftArray));
             }
             else
             {
-                return Convert.ToBase64String(bytes);
+                return TrimBase64End(Convert.ToBase64String(bytes));
             }
         }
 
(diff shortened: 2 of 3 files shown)


src/Runner.Common/HostContext.cs
```diff
@@ -83,6 +83,7 @@ public HostContext(string hostType, string logFile = null)
             _loadContext.Unloading += LoadContext_Unloading;
 
             this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscape);
+            this.SecretMasker.AddValueEncoder(ValueEncoders.Base64StringEscapeTrimmed);
```
thboop: I made the decision to treat this as a separate encoder, which will slightly affect runtime performance. 

We want to ensure as a contract to the user that if they register a secret, we will mask it and various encodings of it (base64, JSON, XML). We also want to best effort mask what we can if the secret is base64 encoded as a part of a larger string. We will leak a few bits a bit of information, but the majority of the secret will be masked.

src/Test/L0/HostContextL0.cs
```diff
@@ -93,12 +93,21 @@ public void DefaultSecretMaskers()
                 Assert.Equal("123***123", _hc.SecretMasker.MaskSecrets("123Pass%20word%20123%21123"));
                 Assert.Equal("123***123", _hc.SecretMasker.MaskSecrets("123Pass&lt;word&gt;123!123"));
                 Assert.Equal("123***123", _hc.SecretMasker.MaskSecrets("123Pass''word''123!123"));
-                Assert.Equal("OlBh***", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($":Password123!"))));
-                Assert.Equal("YTpQ***", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($"a:Password123!"))));
+                Assert.Equal("OlBh***Q==", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($":Password123!"))));
```
thboop: Results from trimming the last character cause a 2-4 bits of information at the end to be leaked.

TingluoHuang:
@thboop any linked issue?

thboop:
@TingluoHuang Added #291

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -103,5 +108,26 @@ private static string Base64StringEscapeShift(String value, int shift)
 
             return result.ToString();
         }
+
+        private static String TrimBase64End(String value)
+        {
+            if (String.IsNullOrEmpty(value))
+            {
+                return String.Empty;
```
TingluoHuang: will this cause all empty string become ***? will `value` ever be empty?

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -16,6 +16,11 @@ public static String Base64StringEscape(String value)
         {
             return Convert.ToBase64String(Encoding.UTF8.GetBytes(value));
         }
+        
+        public static String Base64StringEscapeTrimmed(String value)
```
TingluoHuang: port this back to server side.

thboop:
Going to file a small ADR for this change to be sure we are all in agreement

bryanmacfarlane:
@thboop - this seems more like a bug and the expected behavior (masking a secret) is not being affected by this bug fix.  Is there anything contentious or ambiguous where we need an ADR here?  The contribution guide says to create an ADR for features or behavior changes that aren't straight forward.

thboop:
> @thboop - this seems more like a bug and the expected behavior (masking a secret) is not being affected by this bug fix. Is there anything contentious or ambiguous where we need an ADR here? The contribution guide says to create an ADR for features or behavior changes that aren't straight forward.

@bryanmacfarlane , There is some fallout as a result of these changes where secrets with a prefix or suffix of certain lengths may have more information revealed. Due to the sensitivity of secrets, and wanting to preserve why these decisions were made, I thought it would be good to file a short ADR and make sure we were in agreement before merging.

src/Sdk/DTLogging/Logging/ValueEncoders.cs
thboop: We short circuit if the value is empty 
https://github.com/actions/runner/blob/master/src/Sdk/DTLogging/Logging/SecretMasker.cs#L111
https://github.com/actions/runner/blob/master/src/Sdk/DTLogging/Logging/SecretMasker.cs#L172

PR #285, opened 2020-01-16
Add Temporary Step to fix build on windows

thboop:
Temporary Workaround for https://github.com/actions/virtual-environments/issues/263, we should remove this as soon as we safely can

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,14 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    # Set Path workaround for https://github.com/actions/virtual-environments/issues/263
+    - run: |
+        echo "::add-path::C:\Program Files\Git\mingw64\bin"
+        echo "::add-path::C:\Program Files\Git\usr\bin"
+        echo "::add-path::C:\Program Files\Git\bin"
+      if: matrix.os == "windows-latest"
+      name: "Temp step to Set Path for Windows"
+
     # Build runner layout
     - name: Build & Layout Release
       run: |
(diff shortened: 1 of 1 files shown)


thboop:
I did not add this to the release script, we will need to add it if we plan on doing a release before the root issue is resolved.

.github/workflows/build.yml
```diff
@@ -43,6 +43,14 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    # Set Path workaround for https://github.com/actions/virtual-environments/issues/263
+    - run: |
+        echo "::add-path::C:\Program Files\Git\mingw64\bin"
```
juliobbv: Do we have a guarantee that git is installed in this path? Should we build the path based on `%PROGRAMFILES%` instead?

thboop: The PR that broke this was here: https://github.com/actions/virtual-environments/pull/211/files

So I just manually added these paths back. I don't envision this being a long term solution so I feel something a little hacky like this should be okay.

juliobbv: Ah, the virtual environment makes the same assumptions:

https://github.com/actions/virtual-environments/pull/211/files

So we should be good with these changes then.

eine:
@thboop, might this fix actions/toolkit#312? It is broken since a few hours ago.

DaanDeMeyer:
Is there anything that prevents you from running the dev script directly with `shell: bash` since the only thing dev.cmd seems to be doing is searching for sh.exe and running dev.sh.

Issue #283, opened 2020-01-15
Runner logs infrastructure details


bryanmacfarlane:
**Describe the bug**
consider this log output:
```bash
Run actions/setup-ruby@v1
##[error]Version 2.6.3 not found
##[error]Node run failed with exit code 1
```
The fact that the action is run with node and the return code protocol is not relevant to the user.  the user workflow did not run node - they are setting up ruby.  the `failure` is that 2.6.3 is not found, that's it.  The second line should go to trace / _diag

**To Reproduce**
See setup ruby example above.

**Expected behavior**
The return code of the handler execution should not go to user logs.

## Runner Version and Platform
2.163.1

bryanmacfarlane:
fyi @TingluoHuang

philipengberg:
Experiencing the same since last night

dakale:
Would it be useful to just update the error to something like: `Action has exited with code N` instead? With no error at all it relies on the action itself printing a ##error so sometimes users wont get any message at all depending on the action. We pass/fail the step based on the exit code so that contract is still in place. Could the toolkit expose an exit function that prints and message and exits the process with eg 1? That way people arent left to fend for themselves and have a consistent contract with the actions runner

`run:` steps also print "##[error]Process completed with exit code 1.", should we consider removing that too? There may be a better way to get consistency across various step types here

bryanmacfarlane:
> Would it be useful to just update the error to something like: Action has exited with code N instead?

The return code is an internal protocol so.

> With no error at all it relies on the action itself printing a ##error so sometimes users wont get any message at all depending on the action.

All the actions and examples so try, catch, error with a message.  If a message isn't outputted the step will still fail and there won't be an error annotation which is what the action chose to do by not outputting a message.  This will also encourage actions to output a useful message.

> run: steps also print "##[error]Process completed with exit code 1.", should we consider removing that too? 

No, in that case the **_user_** ran something and we're outputting the result of what they ran.

dakale:
Thought about it more, and I agree now. Small PR here https://github.com/actions/runner/pull/290

PR #290, opened 2020-01-17
Trace javascript action exit code instead of user logs

dakale:
Fixes https://github.com/actions/runner/issues/283

Consuming actions from the graph should not imply knowledge of their runtime be it javascript, container etc

We use the exit code to determine step failure/success so tracing that info out is still useful for diagnostics

### DIFF
--- a/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
+++ b/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
@@ -122,7 +122,7 @@ namespace GitHub.Runner.Worker.Handlers
                 else
                 {
                     var exitCode = await step;
-                    Trace.Info($"Node Action run completed with exit code {exitCode}");
+                    ExecutionContext.Debug($"Node Action run completed with exit code {exitCode}");
                     if (exitCode != 0)
                     {
                         ExecutionContext.Result = TaskResult.Failed;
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -122,9 +122,9 @@ public async Task RunAsync(ActionRunStage stage)
                 else
                 {
                     var exitCode = await step;
+                    Trace.Info($"Node Action run completed with exit code {exitCode}");
```
TingluoHuang: Debug or Info? Info means we need to check runner diag log, Debug goes to user log when ACTIONS_STEP_DEBUG=true

dakale: Good point. I do think Debug makes more sense there. I went ahead and also updated it for the container actions since the same reasoning should apply there as wel

bryanmacfarlane:
debug was meant for users to debug their workflows (not action authors to debug return code).  trace wasn't meant as a deeper debug.  but as long as it's not console out, I'm happy.   I can see both sides of the argument.

dakale:
I think return code is also useful for action consumers who may have provided bad inputs or otherwise misused the action

Debug is way more accessible than trace for this imo, if we consider trace as the runners log (vs the users workflow log) which should probably only be checked if theres suspicion of a bug in the runner itself as opposed to a workflow level error 

Ill merge later if everyones happy with this

Issue #292, opened 2020-01-19
Post-cleanup action doesn't run when main action failed


TingluoHuang:
**Describe the bug**
Wrapper action's post execution should always get run by default.

**To Reproduce**
- uses: actions/checkout@v2
   with:
      repository: notexist/notexist

**Expected behavior**
Even checkout failed, the `post-job: checkout` should get run to cleanup.

PR #293, opened 2020-01-19
default post-job action's condition to always().

TingluoHuang:
fix: #292

### DIFF
--- a/src/Runner.Worker/ActionManifestManager.cs
+++ b/src/Runner.Worker/ActionManifestManager.cs
@@ -341,7 +341,7 @@ namespace GitHub.Runner.Worker
                             EntryPoint = entrypointToken?.Value,
                             Environment = envToken,
                             Cleanup = postEntrypointToken?.Value,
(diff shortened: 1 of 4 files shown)


ericsciple:
shouldnt it only run if the main ran?

ericsciple:
or, i guess sometimes there might not be a main?

TingluoHuang:
Post get added to step list only when its main pass condition check and about to run.

ericsciple:
Cool. I think this change is good then.

One more thing to think about, which i think is OK... When a run is queued, the server adjust all job/step conditions when none of the status functions are mentioned. So if the condition is `github.event == 'foo'` then it gets adjusted to `success() && (github.event == 'foo')`.

I don't think that type of adjustment is needed for the POST, in fact i think it would be opposite of what we want. So i think they way you have the code in this PR is good. Just something to think through since it's a little tricky - make sure i didnt make a mistake in my mental logic.

PR #295, opened 2020-01-20
Fix page log name isn't unqiue.

TingluoHuang:
I forgot timelinerecord id are reused for same pipeline+job+step name, previous change caused the page file name is not unique.

### DIFF
--- a/src/Runner.Common/Logging.cs
+++ b/src/Runner.Common/Logging.cs
@@ -100,7 +100,7 @@ namespace GitHub.Runner.Common
         {
             EndPage();
             _byteCount = 0;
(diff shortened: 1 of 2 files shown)


src/Runner.Common/Logging.cs
```diff
@@ -24,6 +24,7 @@ public class PagingLogger : RunnerService, IPagingLogger
 
         private Guid _timelineId;
         private Guid _timelineRecordId;
+        private string _pageId;
```
TingluoHuang: reverted from https://github.com/actions/runner/pull/255 to make sure page log name unique on disk.

ericsciple: tested on my env and it works

PR #298, opened 2020-01-22
Set both http_proxy and HTTP_PROXY env for runner/worker processes.

TingluoHuang:


src/Runner.Sdk/RunnerWebProxy.cs
```diff
@@ -71,6 +71,10 @@ public RunnerWebProxy()
             {
                 _httpProxyAddress = proxyHttpUri.AbsoluteUri;
 
+                // Set both environment variables since there are tools support both casing (curl, wget) and tools support only one casing (docker)
```
ericsciple: would it better to set one before the other, so that on Windows the env var propagated to downstream processes is always upper case?

PR #300, opened 2020-01-22
Include step's env into `env` context

TingluoHuang:
ADR: #299
Fixing: github/c2c-actions-runtime#195

### DIFF
--- a/src/NuGet.Config
+++ b/src/NuGet.Config
@@ -3,8 +3,6 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="dotnet-core" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
-    <add key="dotnet-buildtools" value="https://www.myget.org/F/dotnet-buildtools/api/v3/index.json" />    
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/src/Runner.Worker/ActionRunner.cs
+++ b/src/Runner.Worker/ActionRunner.cs
@@ -179,17 +179,15 @@ namespace GitHub.Runner.Worker
             ExecutionContext.Debug("Loading env");
             var environment = new Dictionary<String, String>(VarUtil.EnvironmentVariableKeyComparer);
 
-            // Apply environment set using ##[set-env] first since these are job level env
-            foreach (var env in ExecutionContext.EnvironmentVariables)
+#if OS_WINDOWS
+            var envContext = ExecutionContext.ExpressionValues["env"] as DictionaryContextData;
+#else
+            var envContext = ExecutionContext.ExpressionValues["env"] as CaseSensitiveDictionaryContextData;
+#endif
+            // Apply environment from env context, env context contains job level env and action's evn block
+            foreach (var env in envContext)
             {
-                environment[env.Key] = env.Value ?? string.Empty;
-            }
-
-            // Apply action's env block later.
-            var actionEnvironment = templateEvaluator.EvaluateStepEnvironment(Action.Environment, ExecutionContext.ExpressionValues, VarUtil.EnvironmentVariableKeyComparer);
-            foreach (var env in actionEnvironment)
-            {
-                environment[env.Key] = env.Value ?? string.Empty;
+                environment[env.Key] = env.Value.ToString();
             }
 
             // Apply action's intra-action state at last
--- a/src/Runner.Worker/StepsRunner.cs
+++ b/src/Runner.Worker/StepsRunner.cs
@@ -76,10 +76,33 @@ namespace GitHub.Runner.Worker
                 // Start
                 step.ExecutionContext.Start();
 
-                // Set GITHUB_ACTION
+                // Populate env context for each step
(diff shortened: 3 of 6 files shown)


src/NuGet.Config
```diff
@@ -3,8 +3,6 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="dotnet-core" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
```
TingluoHuang: don't need these anymore.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -179,17 +179,15 @@ public async Task RunAsync()
...
 
-            // Apply environment set using ##[set-env] first since these are job level env
-            foreach (var env in ExecutionContext.EnvironmentVariables)
+#if OS_WINDOWS
+            var envContext = ExecutionContext.ExpressionValues["env"] as DictionaryContextData;
+#else
+            var envContext = ExecutionContext.ExpressionValues["env"] as CaseSensitiveDictionaryContextData;
+#endif
+            // Apply environment from env context, env context contains job level env and action's evn block
+            foreach (var env in envContext)
```
TingluoHuang: I moved the `step.env` to when we start the step, so I can include the `step.env` as part for the `env` context before evaluate displayname, if and step.inputs.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -76,10 +76,33 @@ public async Task RunAsync(IExecutionContext jobContext)
...
 
-                // Set GITHUB_ACTION
+                // Populate env context for each step
+                Trace.Info("Initialize Env context for step");
+#if OS_WINDOWS
+                var envContext = new DictionaryContextData();
+#else
+                var envContext = new CaseSensitiveDictionaryContextData();
+#endif
+                step.ExecutionContext.ExpressionValues["env"] = envContext;
```
TingluoHuang: every step will has a new `env` context copy that include's `env` from job and `::set-env::`

src/Runner.Worker/ActionRunner.cs
ericsciple: Doesn't display name get calculated at the beginning of the job.

When calculating display name at beginning of a job, should check if script contains `env` and delay. We have logic on server to check for referenced contexts... Might be the key to solving hashFiles issue too.

ericsciple: Should we evaluate `if` before `env`? Or is it because we need `env` to calculate display name, even when skipped?

src/Runner.Worker/ActionRunner.cs
```diff
@@ -179,17 +179,15 @@ public async Task RunAsync()
             ExecutionContext.Debug("Loading env");
             var environment = new Dictionary<String, String>(VarUtil.EnvironmentVariableKeyComparer);
 
-            // Apply environment set using ##[set-env] first since these are job level env
-            foreach (var env in ExecutionContext.EnvironmentVariables)
+#if OS_WINDOWS
+            var envContext = ExecutionContext.ExpressionValues["env"] as DictionaryContextData;
+#else
+            var envContext = ExecutionContext.ExpressionValues["env"] as CaseSensitiveDictionaryContextData;
```
ericsciple: is env on the runner side case sensitive today on linux? i thought expressions always are case insentive dictionaries

ericsciple: could go either way, but inconsistent for env defined above job level if case sensitive within job

ericsciple: i wonder if should be case sensitive all the way up, and deserialize into the correct type. Runner would need to reconstruct case-insensitive on windows

src/Runner.Worker/ActionRunner.cs
TingluoHuang: we re-evaluate displayname in `stepsrunner` before each step run: `actionRunner.TryEvaluateDisplayName(step.ExecutionContext.ExpressionValues, step.ExecutionContext)`, we need `env` to evaluate `if`, we don't calculate display name on skip.

src/Runner.Worker/ActionRunner.cs
TingluoHuang: i made it case sensitive because we was trying to put machine env into the `env` context, now I am thinking maybe we just make it case-insensitive and lines up with job/workflow env.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -179,17 +179,15 @@ public async Task RunAsync()
...
             {
-                environment[env.Key] = env.Value ?? string.Empty;
-            }
-
-            // Apply action's env block later.
-            var actionEnvironment = templateEvaluator.EvaluateStepEnvironment(Action.Environment, ExecutionContext.ExpressionValues, VarUtil.EnvironmentVariableKeyComparer);
-            foreach (var env in actionEnvironment)
-            {
-                environment[env.Key] = env.Value ?? string.Empty;
+                environment[env.Key] = env.Value.ToString();
```
ericsciple: isnt there an implicit operator to string?

PR #307, opened 2020-01-29
Fix windows directions in release notes

thechrisjohnson:
If you follow the directions explicitly they fail. I'm confident that people can figure this out on their own, but it just makes it nicer when copying and pasting :-)

### DIFF
--- a/releaseNote.md
+++ b/releaseNote.md
@@ -24,7 +24,7 @@
   - Treat warnings as errors during compile (#249) 
 
 ## Windows x64
-We recommend configuring the runner under "<DRIVE>:\actions-runner". This will help avoid issues related to service identity folder permissions and long file path restrictions on Windows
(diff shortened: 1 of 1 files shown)


releaseNote.md
```diff
@@ -32,7 +32,7 @@ mkdir \actions-runner ; cd \actions-runner
 Invoke-WebRequest -Uri https://github.com/actions/runner/releases/download/v<RUNNER_VERSION>/actions-runner-win-x64-<RUNNER_VERSION>.zip -OutFile actions-runner-win-x64-<RUNNER_VERSION>.zip
```
TingluoHuang: do you mind fixing the `<DRIVE>:\`? it looks weird after it gets rendered.
`We recommend configuring the runner under ":\actions-runner"`

thechrisjohnson: Done

PR #314, opened 2020-02-03
use /usr/bin/env to find bash in scripts

mmahut:
This changes uses the ```/usr/bin/env``` utility to find the location of bash in scripts. 

Right now, the scripts are assuming the location of ```bash``` which might not be true on some systems such as NixOS and others.

For more information, please see https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability


bryanmacfarlane:
Is there an issue logged for this?

mmahut:
@bryanmacfarlane I do not think so.

zimbatm:
@mmahut semi out of topic, I have a WIP PR to make the actions runner on nixos: https://github.com/nix-community/infra/pull/5

Mic92:
With the following shell.nix I was able to build it:

```
with import <nixpkgs> {};
let
  fhs = pkgs.buildFHSUserEnv {
    name = "runner-env";
    targetPkgs = pkgs: with pkgs; [
      dotnet-sdk_3
      icu
    ];
    extraOutputsToInstall = [ "dev" ];
  };
in fhs.env
```

```
nix-shell -p --run 'dotnet msbuild -t:Build -p:PackageRuntime=linux-x64 -p:BUILDCONFIG=Release -p:RunnerVersion=$(cat src/runnerversion) src/dir.proj'
```

Mic92:
It also produces a `_layout/bin/Sdk.deps.json` which could be used to download all nuget dependencies.

domenkozar:
@Mic92 could we also just run actions in that FHS?

Mic92:
> @Mic92 could we also just run actions in that FHS?

Yes. All that should work.

nvtkaszpir:
> Is there an issue logged for this?

#565

zimbatm:
This is still a valid fix

Mic92:
> Is there an issue logged for this?

I created an issue in case this is required for processing the pull request: https://github.com/actions/runner/issues/848

TingluoHuang (approved):

PR #325, opened 2020-02-11
Switch to use token service instead of SPS for exchanging oauth token.

TingluoHuang:
https://github.com/github/c2c-actions-runtime/pull/348

Goal: Migrate existing runner to use Authorization url from Token service instead of SPS.

FF:
1. For newly configured runner, service base on FF to send initial authorization url from Token or SPS back during config.
2. Start migration all self-hosted runner for a single repository.

### DIFF
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -78,8 +78,10 @@ namespace GitHub.Runner.Common
         bool IsServiceConfigured();
         bool HasCredentials();
         CredentialData GetCredentials();
+        CredentialData GetV2Credentials();
         RunnerSettings GetSettings();
         void SaveCredential(CredentialData credential);
+        void SaveV2Credential(CredentialData credential);
         void SaveSettings(RunnerSettings settings);
         void DeleteCredential();
         void DeleteSettings();
@@ -90,9 +92,11 @@ namespace GitHub.Runner.Common
         private string _binPath;
         private string _configFilePath;
         private string _credFilePath;
+        private string _credV2FilePath;
         private string _serviceConfigFilePath;
 
         private CredentialData _creds;
+        private CredentialData _credsV2;
         private RunnerSettings _settings;
 
         public override void Initialize(IHostContext hostContext)
@@ -114,6 +118,9 @@ namespace GitHub.Runner.Common
             _credFilePath = hostContext.GetConfigFile(WellKnownConfigFile.Credentials);
             Trace.Info("CredFilePath: {0}", _credFilePath);
 
+            _credV2FilePath = hostContext.GetConfigFile(WellKnownConfigFile.CredentialsV2);
+            Trace.Info("CredV2FilePath: {0}", _credV2FilePath);
+
             _serviceConfigFilePath = hostContext.GetConfigFile(WellKnownConfigFile.Service);
             Trace.Info("ServiceConfigFilePath: {0}", _serviceConfigFilePath);
         }
@@ -123,7 +130,7 @@ namespace GitHub.Runner.Common
         public bool HasCredentials()
         {
             Trace.Info("HasCredentials()");
-            bool credsStored = (new FileInfo(_credFilePath)).Exists;
+            bool credsStored = (new FileInfo(_credFilePath)).Exists || (new FileInfo(_credV2FilePath)).Exists;
             Trace.Info("stored {0}", credsStored);
             return credsStored;
         }
@@ -152,6 +159,16 @@ namespace GitHub.Runner.Common
             }
 
             return _creds;
+        }
+
+        public CredentialData GetV2Credentials()
+        {
+            if (_credsV2 == null && File.Exists(_credV2FilePath))
+            {
+                _credsV2 = IOUtil.LoadObject<CredentialData>(_credV2FilePath);
+            }
+
+            return _credsV2;
         }
 
         public RunnerSettings GetSettings()
@@ -188,6 +205,21 @@ namespace GitHub.Runner.Common
             File.SetAttributes(_credFilePath, File.GetAttributes(_credFilePath) | FileAttributes.Hidden);
         }
 
+        public void SaveV2Credential(CredentialData credential)
+        {
+            Trace.Info("Saving {0} v2 credential @ {1}", credential.Scheme, _credV2FilePath);
+            if (File.Exists(_credV2FilePath))
+            {
+                // Delete existing credential file first, since the file is hidden and not able to overwrite.
+                Trace.Info("Delete exist runner credential v2 file.");
+                IOUtil.DeleteFile(_credV2FilePath);
+            }
+
+            IOUtil.SaveObject(credential, _credV2FilePath);
+            Trace.Info("v2 Credentials Saved.");
+            File.SetAttributes(_credV2FilePath, File.GetAttributes(_credV2FilePath) | FileAttributes.Hidden);
+        }
+
         public void SaveSettings(RunnerSettings settings)
         {
             Trace.Info("Saving runner settings.");
@@ -206,6 +238,7 @@ namespace GitHub.Runner.Common
         public void DeleteCredential()
         {
             IOUtil.Delete(_credFilePath, default(CancellationToken));
+            IOUtil.Delete(_credV2FilePath, default(CancellationToken));
         }
 
         public void DeleteSettings()
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -78,10 +78,10 @@ namespace GitHub.Runner.Common
         bool IsServiceConfigured();
         bool HasCredentials();
         CredentialData GetCredentials();
-        CredentialData GetV2Credentials();
+        CredentialData GetMigratedCredentials();
         RunnerSettings GetSettings();
         void SaveCredential(CredentialData credential);
-        void SaveV2Credential(CredentialData credential);
+        void SaveMigratedCredential(CredentialData credential);
         void SaveSettings(RunnerSettings settings);
         void DeleteCredential();
         void DeleteSettings();
@@ -92,11 +92,11 @@ namespace GitHub.Runner.Common
         private string _binPath;
         private string _configFilePath;
         private string _credFilePath;
-        private string _credV2FilePath;
+        private string _migratedCredFilePath;
         private string _serviceConfigFilePath;
 
         private CredentialData _creds;
-        private CredentialData _credsV2;
+        private CredentialData _migratedCreds;
         private RunnerSettings _settings;
 
         public override void Initialize(IHostContext hostContext)
@@ -118,8 +118,8 @@ namespace GitHub.Runner.Common
             _credFilePath = hostContext.GetConfigFile(WellKnownConfigFile.Credentials);
             Trace.Info("CredFilePath: {0}", _credFilePath);
 
-            _credV2FilePath = hostContext.GetConfigFile(WellKnownConfigFile.CredentialsV2);
-            Trace.Info("CredV2FilePath: {0}", _credV2FilePath);
+            _migratedCredFilePath = hostContext.GetConfigFile(WellKnownConfigFile.MigratedCredentials);
+            Trace.Info("MigratedCredFilePath: {0}", _migratedCredFilePath);
 
             _serviceConfigFilePath = hostContext.GetConfigFile(WellKnownConfigFile.Service);
             Trace.Info("ServiceConfigFilePath: {0}", _serviceConfigFilePath);
@@ -130,7 +130,7 @@ namespace GitHub.Runner.Common
         public bool HasCredentials()
         {
             Trace.Info("HasCredentials()");
-            bool credsStored = (new FileInfo(_credFilePath)).Exists || (new FileInfo(_credV2FilePath)).Exists;
+            bool credsStored = (new FileInfo(_credFilePath)).Exists || (new FileInfo(_migratedCredFilePath)).Exists;
             Trace.Info("stored {0}", credsStored);
             return credsStored;
         }
@@ -161,14 +161,14 @@ namespace GitHub.Runner.Common
             return _creds;
         }
 
-        public CredentialData GetV2Credentials()
+        public CredentialData GetMigratedCredentials()
         {
-            if (_credsV2 == null && File.Exists(_credV2FilePath))
+            if (_migratedCreds == null && File.Exists(_migratedCredFilePath))
             {
-                _credsV2 = IOUtil.LoadObject<CredentialData>(_credV2FilePath);
+                _migratedCreds = IOUtil.LoadObject<CredentialData>(_migratedCredFilePath);
             }
 
-            return _credsV2;
+            return _migratedCreds;
         }
 
         public RunnerSettings GetSettings()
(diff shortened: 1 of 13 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -296,304 +299,312 @@ private async Task RunOnceAsync(Pipelines.AgentJobRequestMessage message, Worker
...
-            term.WriteLine($"{DateTime.UtcNow:u}: Running job: {message.JobDisplayName}");
-
-            // first job request renew succeed.
-            TaskCompletionSource<int> firstJobRequestRenewed = new TaskCompletionSource<int>();
-            var notification = HostContext.GetService<IJobNotification>();
-
-            // lock renew cancellation token.
-            using (var lockRenewalTokenSource = new CancellationTokenSource())
-            using (var workerProcessCancelTokenSource = new CancellationTokenSource())
+            Busy = true;
```
TingluoHuang: The diff is really bad.
Here is what the change looks like:
```
Busy = true;
try
{
   <Everything in the original method>
}
finally
{
    Busy = false;
}
```

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -45,52 +48,60 @@ public override void Initialize(IHostContext hostContext)
 
         public async Task<bool> SelfUpdate(AgentRefreshMessage updateMessage, IJobDispatcher jobDispatcher, bool restartInteractiveRunner, CancellationToken token)
         {
-            if (!await UpdateNeeded(updateMessage.TargetVersion, token))
+            Busy = true;
```
TingluoHuang: The diff is really bad.
Here is what the change looks like:
```
Busy = true;
try
{
   <Everything in the original method>
}
finally
{
    Busy = false;
}
```

src/Runner.Common/ConfigurationStore.cs
```diff
@@ -78,8 +78,10 @@ public interface IConfigurationStore : IRunnerService
         bool IsServiceConfigured();
         bool HasCredentials();
         CredentialData GetCredentials();
+        CredentialData GetV2Credentials();
```
TingluoHuang: we keep the existing `.credentials` file on disk, migration will create `.v2Credentials` file, if anything goes wrong with the new authorization setting in `v2Credentials`, we will fallback to `.credentials`

src/Runner.Common/RunnerServer.cs
```diff
@@ -334,5 +337,14 @@ public Task<TaskAgent> UpdateAgentUpdateStateAsync(int agentPoolId, int agentId,
...
             return _genericTaskAgentClient.UpdateAgentUpdateStateAsync(agentPoolId, agentId, currentState);
         }
+
+        //-----------------------------------------------------------------
+        // Runner Auth Url
+        //-----------------------------------------------------------------
+        public Task<string> GetRunnerAuthUrlAsync(int runnerPoolId, int runnerId)
+        {
+            CheckConnection(RunnerConnectionType.MessageQueue);
+            return _messageTaskAgentClient.GetAgentAuthUrlAsync(runnerPoolId, runnerId);
```
TingluoHuang: There is a FF on server to control the return value from the endpoint. When FF is off, the endpoint always return empty. When FF is on, the endpoint will return authorization url from Token service and this will trigger runner migration.

src/Runner.Listener/Configuration/CredentialManager.cs
```diff
@@ -40,7 +40,7 @@ public ICredentialProvider GetCredentialProvider(string credType)
             return creds;
         }
 
-        public VssCredentials LoadCredentials()
+        public VssCredentials LoadCredentials(bool preferV2 = true)
```
TingluoHuang: this is used during runner startup, we will prefer the new authorization url from Token service.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -18,6 +18,7 @@ namespace GitHub.Runner.Listener
     [ServiceLocator(Default = typeof(JobDispatcher))]
     public interface IJobDispatcher : IRunnerService
     {
+        bool Busy { get; }
```
TingluoHuang: There are 3 different parts in the runner having Http connection with service.
- Message queue loop for pull job from service.
- Job Dispatcher handle renew lock for running job.
- Runner self-updater handle download latest runner and post update progress to service.

I want to make sure there is no job and self-update running when migrate to use new authorization during message queue loop.

src/Runner.Listener/MessageListener.cs
```diff
@@ -32,18 +35,33 @@ public sealed class MessageListener : RunnerService, IMessageListener
...
+        // Whether load credentials from .v2credentials file
+        internal bool _useV2Credentials;
+
+        // Whether .credentials existing, we will fallback to this if .v2credential doesn't work.
+        internal bool _v1CredentialsExists;
+
+        // need to check auth url if there is only .credentials and auth schema is OAuth
+        internal bool _needToCheckAuthorizationUrlUpdate;
+        internal Task<VssCredentials> _authorizationUrlMigrationBackgroundTask;
+        internal Task _authorizationUrlRollbackReattemptDelayBackgroundTask;
```
TingluoHuang: when auth from Token service doesn't work, we will fallback to SPS and wait for 2 days before trying Token service again.

src/Runner.Listener/MessageListener.cs
```diff
@@ -74,6 +92,25 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
+            if (v1Creds != null)
+            {
+                _v1CredentialsExists = true;
+            }
+
+            if (v2Creds == null)
+            {
+                if (v1Creds.Scheme == Constants.Configuration.OAuth)
+                {
+                    _needToCheckAuthorizationUrlUpdate = true;
```
TingluoHuang: Hosted runner use OAuthAccessToken instead of OAuth, and we don't need any migration for hosted runner.

src/Runner.Listener/MessageListener.cs
```diff
@@ -397,5 +510,69 @@ private bool IsSessionCreationExceptionRetriable(Exception ex)
...
+                {
+                    var v2AuthorizationUrl = await _runnerServer.GetRunnerAuthUrlAsync(_settings.PoolId, _settings.AgentId);
+                    if (!string.IsNullOrEmpty(v2AuthorizationUrl))
+                    {
+                        var credData = _configStore.GetCredentials();
+                        var clientId = credData.Data.GetValueOrDefault("clientId", null);
+                        var currentAuthorizationUrl = credData.Data.GetValueOrDefault("authorizationUrl", null);
+                        Trace.Info($"Current authorization url: {currentAuthorizationUrl}, new authorization url: {v2AuthorizationUrl}");
+
+                        if (string.Equals(currentAuthorizationUrl, v2AuthorizationUrl, StringComparison.OrdinalIgnoreCase))
```
TingluoHuang: this means when user configure the runner, service already make the runner using Token service, so there is no needed for migration.

TingluoHuang: the authorization url is original return from service during runner config

src/Runner.Listener/MessageListener.cs
```diff
@@ -74,6 +92,25 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
+            var v1Creds = _configStore.GetCredentials();
+            var v2Creds = _configStore.GetV2Credentials();
+            if (v1Creds != null)
+            {
+                _v1CredentialsExists = true;
+            }
+
+            if (v2Creds == null)
+            {
+                if (v1Creds.Scheme == Constants.Configuration.OAuth)
```
ericsciple: would v2 ever be null and v1 scheme not oauth?

src/Runner.Listener/MessageListener.cs
```diff
@@ -74,6 +92,25 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
             string errorMessage = string.Empty;
             bool encounteringError = false;
 
+            var v1Creds = _configStore.GetCredentials();
+            var v2Creds = _configStore.GetV2Credentials();
+            if (v1Creds != null)
+            {
+                _v1CredentialsExists = true;
```
ericsciple: nit: looks like some unnecessary local vars adding extra complexity. It looks like v1creds, v2creds, and useLegacyAuthUrl are all that is needed. For instance, instead of managing _useV2Credentials, could set v2creds=null if need to fallback.

ericsciple: nevermind now i see the motivation (2 day delay)

src/Runner.Listener/MessageListener.cs
TingluoHuang: Hosted only has v1 and not oauth.

src/Runner.Listener/MessageListener.cs
ericsciple: perhaps only _v1CredentialsExists is unnecessary

src/Runner.Listener/MessageListener.cs
```diff
@@ -120,8 +163,20 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
...
                     if (!IsSessionCreationExceptionRetriable(ex))
                     {
-                        _term.WriteError($"Failed to create session. {ex.Message}");
-                        return false;
+                        if (_useV2Credentials && _v1CredentialsExists)
+                        {
+                            // v2 credentials might cause lose permission during permission check,
+                            // we will force to use v1 credential and try again
+                            _useV2Credentials = false;
+                            _authorizationUrlRollbackReattemptDelayBackgroundTask = HostContext.Delay(TimeSpan.FromDays(2), token); // retry v2 creds in 2 days.
```
ericsciple: what is the motivation for such a long delay?

src/Runner.Listener/MessageListener.cs
ericsciple: Another example, if env var GITHUB_ACTIONS_RUNNER_LEGACYAUTHURL  then could v2creds just be set to null. Would it simplify the logic to eliminate local variable `useLegacyAuthUrl`?

src/Runner.Listener/MessageListener.cs
```diff
@@ -120,8 +163,20 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
 
                     if (!IsSessionCreationExceptionRetriable(ex))
                     {
-                        _term.WriteError($"Failed to create session. {ex.Message}");
-                        return false;
+                        if (_useV2Credentials && _v1CredentialsExists)
```
ericsciple: why not just flip back and forth every time? why the 2 day delay?

src/Runner.Listener/MessageListener.cs
ericsciple: in that case, wouldnt there just be v2 creds?

TingluoHuang: I think the way I name this causes confusion, I should just call them `cred` and `cred.migrate`
A runner that has both `cred` and `cred.migrate` will require to be:
1. self-host runner
2. original configured OAuth with SPS

hosted runner and newly configured runner should always just has `cred`

src/Runner.Listener/MessageListener.cs
TingluoHuang: `_v1CredentialsExists` is unnecessary, v1Cred should always exist.

src/Runner.Listener/MessageListener.cs
TingluoHuang: During the initial rollout, I am trying to avoid wasting time trying Token service if there is a service bug and we need to do a binary deployment to fix the problem in Token service.
I can short the back-off to 12-24 hours

src/Runner.Common/HostContext.cs
```diff
@@ -282,6 +282,12 @@ public string GetConfigFile(WellKnownConfigFile configFile)
                         ".credentials");
                     break;
 
+                case WellKnownConfigFile.MigratedCredentials:
+                    path = Path.Combine(
+                        GetDirectory(WellKnownDirectory.Root),
+                        ".credentials_migrated");
```
TingluoHuang: only self-hosted runners that currently authed with SPS will get is file created as part of the migration.
once we have all FF turn on and shutdown SPS, any newly configured self-hosted runner will only have `.credentials` file.

src/Runner.Listener/MessageListener.cs
```diff
@@ -58,8 +73,8 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
 
             // Create connection.
             Trace.Info("Loading Credentials");
-            var credMgr = HostContext.GetService<ICredentialManager>();
-            VssCredentials creds = credMgr.LoadCredentials();
+            _useMigratedCredentials = !StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_SPSAUTHURL"));
```
thboop: What sets this env variable?

src/Runner.Listener/MessageListener.cs
```diff
@@ -397,5 +501,80 @@ private bool IsSessionCreationExceptionRetriable(Exception ex)
...
+                    {
+                        var credData = _configStore.GetCredentials();
+                        var clientId = credData.Data.GetValueOrDefault("clientId", null);
+                        var currentAuthorizationUrl = credData.Data.GetValueOrDefault("authorizationUrl", null);
+                        Trace.Info($"Current authorization url: {currentAuthorizationUrl}, new authorization url: {migratedAuthorizationUrl}");
+
+                        if (string.Equals(currentAuthorizationUrl, migratedAuthorizationUrl, StringComparison.OrdinalIgnoreCase))
+                        {
+                            // We don't need to update credentials.
+                            Trace.Info("No needs to update authorization url");
```
thboop: NIT `No need` rather then `No needs`

src/Runner.Listener/MessageListener.cs
```diff
@@ -205,7 +295,21 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                     }
                     else if (!IsGetNextMessageExceptionRetriable(ex))
                     {
-                        throw;
+                        if (_useMigratedCredentials)
```
thboop: Same concern as above

src/Runner.Listener/MessageListener.cs
```diff
@@ -397,5 +501,69 @@ private bool IsSessionCreationExceptionRetriable(Exception ex)
...
+                        return migratedRunnerCredential;
+                    }
+                    else
+                    {
+                        Trace.Verbose("No authorization url updates");
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Trace.Error("Fail to get/test new authorization url.");
```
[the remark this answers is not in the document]
TingluoHuang: `Trace.Error()` only goes to diag log which should be fine.

src/Runner.Listener/Configuration/CredentialManager.cs
```diff
@@ -50,6 +50,16 @@ public VssCredentials LoadCredentials()
             }
 
             CredentialData credData = store.GetCredentials();
+
```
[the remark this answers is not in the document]
TingluoHuang: I was trying to avoid touch existing code since all the migration code can be deleted after migration finished.

src/Runner.Listener/MessageListener.cs
TingluoHuang: End user.

src/Runner.Listener/MessageListener.cs
```diff
@@ -120,8 +152,21 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
 
                     if (!IsSessionCreationExceptionRetriable(ex))
                     {
-                        _term.WriteError($"Failed to create session. {ex.Message}");
-                        return false;
+                        if (_useMigratedCredentials)
+                        {
+                            // migrated credentials might cause lose permission during permission check,
```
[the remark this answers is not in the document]
TingluoHuang: `IsSessionCreationExceptionRetriable()` is already narrow down to only permission related exception.

src/Runner.Listener/MessageListener.cs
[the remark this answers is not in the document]
thboop: Its pretty nitpicky but we do doc those files

src/Runner.Listener/MessageListener.cs
thboop: Is this something we will tell users to set to fix issues they are encountering then?

PR #329, opened 2020-02-12
Set Minimum Length on base64 trimming to > 2 pre-encoded characters

### LINKED ISSUE #326: 2.165.1 runner over mask log when using secrets with JSON/newline
**Describe the bug**
Log get over masked when using JSON with newline as secret

**To Reproduce**
1. add a secret variable (has new line):
```
TestJson=
{
}
```
2. Run a workflow which `echo ${{secrets.TestJson}}`
3. all `e` and `f` in the log get masked, which causes the log un-readable.

**Expected behavior**

2.165.1

## What's not working?
https://github.com/actions/canary/runs/439952952?check_suite_focus=true

thboop:
Resolves #326

### DIFF
--- a/src/Sdk/DTLogging/Logging/ValueEncoders.cs
+++ b/src/Sdk/DTLogging/Logging/ValueEncoders.cs
@@ -118,7 +118,8 @@ namespace GitHub.DistributedTask.Logging
             if (value.EndsWith('='))
             {
                 var trimmed = value.TrimEnd('=');
-                if (trimmed.Length > 1)
+                // Minimum length to make sure we don't register 1 or 2 characters as a secret.
+                if (trimmed.Length > 3)
                 {
                     // If a base64 string ends in '=' it indicates that the base 64 character is only using 2 or 4 of the six bytes and will change if another character is added
                     // For example 'ab' is 'YWI=' in base 64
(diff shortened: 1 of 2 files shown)


src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -118,7 +118,8 @@ private static String TrimBase64End(String value)
             if (value.EndsWith('='))
             {
                 var trimmed = value.TrimEnd('=');
-                if (trimmed.Length > 1)
+                // Minimum length to make sure we don't register 1 or 2 characters as a secret.
+                if (trimmed.Length > 3)
```
thboop: For original strings (not base 64 encoded versions) < 4 characters we will not trim (like we did in the past before the recent changes in #284 .
This includes strings that are 4 or 5 characters, but when shifted via the base 64 encoders then end up being < 4 characters. The 4 or 5 character encoders will trim, but the shifted versions that end up being length < 4 will not.

src/Test/L0/HostContextL0.cs
```diff
@@ -108,6 +110,12 @@ public void DefaultSecretMaskers()
                 Assert.Equal("OlBh***To=", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($":Password123!:"))));
                 Assert.Equal("YTpQ***E6YQ==", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($"a:Password123!:a"))));
                 Assert.Equal("YWJjOlBh***Tph", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($"abc:Password123!:a"))));
+                Assert.Equal("YWJjOlBh***Tph", _hc.SecretMasker.MaskSecrets(Convert.ToBase64String(Encoding.UTF8.GetBytes($"abc:Password123!:a"))));
```
TingluoHuang: we tested `abc:Password123!:a` twice. 😄

src/Test/L0/HostContextL0.cs
```diff
@@ -85,6 +85,8 @@ public void DefaultSecretMaskers()
                 _hc.SecretMasker.AddValue("Pass word 123!");
                 _hc.SecretMasker.AddValue("Pass<word>123!");
                 _hc.SecretMasker.AddValue("Pass'word'123!");
+                _hc.SecretMasker.AddValue("3ch");
+                _hc.SecretMasker.AddValue("{");
```
TingluoHuang: add tests for both `{` and `}` 
make sure `{` won't mask `e` and `}` won't mask `f`

PR #331, opened 2020-02-12
commands translate file path from container action

ericsciple:


src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -306,9 +307,9 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
             }
 
             // Translate file path back from container path
-            if (context.Container != null)
+            if (container != null)
             {
-                file = context.Container.TranslateToHostPath(file);
+                file = container.TranslateToHostPath(file);
```
ericsciple: add L0

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using GitHub.DistributedTask.Pipelines;
 using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Common.Util;
+using GitHub.Runner.Worker.Container;
```
ericsciple: despite the diff, the actual change is pretty small. The diff looks bad due to the signature change of IActionCommandExtension.ProcessCommand()

src/Runner.Worker/Handlers/OutputManager.cs
```diff
@@ -84,7 +84,7 @@ public void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
             {
                 // This does not need to be inside of a critical section.
                 // The logging queues and command handlers are thread-safe.
-                if (_commandManager.TryProcessCommand(_executionContext, line))
+                if (_commandManager.TryProcessCommand(_executionContext, line, _container))
```
ericsciple: OutputManager is the only place that calls commandManager.TryProcessCommand

In the OutputManager ctor, _container is set to \"step container ?? job container\". The change here is to pass the correct container for the command manager to use when translating paths.

PR #339, opened 2020-02-19
Update Runner Register GitHub API URL to Support Org-level Runner

lokesh755:
Draft PR

Updated GitHub API request as per this -> https://github.com/github/c2c-actions-runtime/issues/373#issuecomment-588394924

Sample request

```
POST https://api.github.com/actions/runner-registration HTTP/1.1
Host: api.github.com
Authorization: RemoteAuth AAO2GDSAMPLEIQLJBOI2BZ26JXCTA
User-Agent: GitHubActionsRunner-win-x64/2.165.0
Content-Type: application/json; charset=utf-8
Content-Length: 50

{"url":"https://github.com/bbq-beets/logopu-test"}
```

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -522,13 +522,12 @@ namespace GitHub.Runner.Listener.Configuration
         private async Task<GitHubAuthResult> GetTenantCredential(string githubUrl, string githubToken)
         {
             var gitHubUrlBuilder = new UriBuilder(githubUrl);
-            var githubApiUrl = $"https://api.{gitHubUrlBuilder.Host}/actions/register-runner";
-            using (var httpClientHandler = HostContext.CreateHttpClientHandler())
+            var githubApiUrl = $"https://api.{gitHubUrlBuilder.Host}/actions/runner-registration";
+           using (var httpClientHandler = HostContext.CreateHttpClientHandler())
             using (var httpClient = new HttpClient(httpClientHandler))
             {
                 httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("RemoteAuth", githubToken);
                 httpClient.DefaultRequestHeaders.UserAgent.Add(HostContext.UserAgent);
-                httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/vnd.github.shuri-preview+json"));
 
                 var bodyObject = new Dictionary<string, string>()
                 {
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -1,8 +1,12 @@
 using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Common;
 using GitHub.Runner.Common.Util;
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -522,15 +524,20 @@ private async Task<bool> IsHostedServer(string serverUrl, VssCredentials credent
 
         private async Task<GitHubAuthResult> GetTenantCredential(string githubUrl, string githubToken)
         {
-            var gitHubUrl = new UriBuilder(githubUrl);
-            var githubApiUrl = $"https://api.{gitHubUrl.Host}/repos/{gitHubUrl.Path.Trim('/')}/actions-runners/registration";
+            var githubApiUrl = "https://api.github.com/actions/register-runner";
```
TingluoHuang: you may want to persist `api.{gitHubUrl.Host}` as before, since in local dotcom dev env it's github.localhost

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -522,15 +524,20 @@ private async Task<bool> IsHostedServer(string serverUrl, VssCredentials credent
...
             using (var httpClientHandler = HostContext.CreateHttpClientHandler())
             using (var httpClient = new HttpClient(httpClientHandler))
             {
                 httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("RemoteAuth", githubToken);
                 httpClient.DefaultRequestHeaders.UserAgent.Add(HostContext.UserAgent);
                 httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/vnd.github.shuri-preview+json"));
-                var response = await httpClient.PostAsync(githubApiUrl, new StringContent("", null, "application/json"));
+
+                var bodyObject = new JObject
```
TingluoHuang: create `Dictionary<string, string>` and use `StringUtil.ConvertToJson()`

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -522,15 +521,21 @@ private async Task<bool> IsHostedServer(string serverUrl, VssCredentials credent
 
         private async Task<GitHubAuthResult> GetTenantCredential(string githubUrl, string githubToken)
         {
-            var gitHubUrl = new UriBuilder(githubUrl);
-            var githubApiUrl = $"https://api.{gitHubUrl.Host}/repos/{gitHubUrl.Path.Trim('/')}/actions-runners/registration";
+            var gitHubUrlBuilder = new UriBuilder(githubUrl);
+            var githubApiUrl = $"https://api.{gitHubUrlBuilder.Host}/actions/register-runner";
```
TingluoHuang: is this new api ready for production usage?
if not, then you probably want to keep the `GetTenantCredential()` as it is and add a new method, like: `GetTenantCredentialPreview()`, and flip between these 2 bases on some env variable (GITHUB_NEW_REGISTERENDPOINT=true) to test the new endpoint.

lokesh755: It'll be in production tomorrow. I just created this little early, so that you can have a look. Will merge it once the api is in production.

PR #345, opened 2020-02-25
Use the input url scheme in the register runner url

joshmgross:
### Issue
The current runner registration URL defaults to `HTTPS`, which can make it difficult to test locally when using `HTTP` URLs.

This change uses the input URL's scheme for the runner registration API URL.

### Testing

Tested with a runner connected to `http://github.localhost` and verified that the runner was successfully registered.


lokesh755:
Looks good to me

PR #352, opened 2020-03-02
Add event type to credential call

joshmgross:
To support auditing, we need to know which event is occurring when retrieving the tenant credential. Sending this event type will also allow us to split up the permissions needed for adding and removing a runner (which currently uses the same scopes).


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -552,5 +553,11 @@ private async Task<GitHubAuthResult> GetTenantCredential(string githubUrl, strin
                 }
             }
         }
+
+        private static class RunnerCredentialEvents
```
TingluoHuang: move this to `Runner.Common.Constants.cs` and just name it `RunnerEvent.Register` and `RunnerEvent.Remove`

joshmgross: Moved to constants

PR #361, opened 2020-03-08
ADR for wrapper action

TingluoHuang:
👉[Rendered](https://github.com/actions/runner/blob/users/tihuang/wrapperadr/docs/adrs/0361-wrapper-action.md) 👈 

https://github.com/github/c2c-actions-runtime/issues/141

### DIFF
--- a/docs/adrs/0361-wrapper-action.md
+++ b/docs/adrs/0361-wrapper-action.md
@@ -57,15 +57,15 @@ Setting `pre` to `always()` will make sure no matter what condition evaluate res
 > Currently `post` works for both repository action (`org/repo@v1`) and local action (`./my-action`)
 
 Valid action:
-- only hss `main`
-- only has `pre`
+- only has `main`
 - has `pre` and `main`
-- has `pre` and `post`
 - has `main` and `post`
 - has `pre`, `main` and `post`
 
 Invalid action:
+- only has `pre`
 - only has `post`
+- has `pre` and `post`
 
 Potential downside of introducing `pre`:
 
--- a/docs/adrs/0361-wrapper-action.md
+++ b/docs/adrs/0361-wrapper-action.md
@@ -27,12 +27,6 @@ Node Action Example:
    pre: 'setup.js'
    pre-if: 'success()' // Optional
    main: 'index.js'
-
- name: 'My action with post'
- description: 'My action with post'
- runs:
-   using: 'node12'
-   main: 'index.js'
    post: 'cleanup.js'
    post-if: 'success()' // Optional
 ```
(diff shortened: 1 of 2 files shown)


docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,73 @@
...
+`pre` executes in order of how the steps are defined and `post` runs in the opposite order.
+`pre` will always be added to job steps list during job setup.  
+> Action referenced from local repository (`./my-action`) won't get `pre` setup correctly since the repository haven't checkout during job initialize.  
+> We can't use GitHub api to download the repository since there is a about 3 mins delay between `git push` and the new commit available to download using GitHub api.
+
+`post` will be added to steps list lazily when the action's main execution passed `if` condition check and about to run.
+> Currently `post` works for both repository action (`org/repo@v1`) and local action (`./my-action`)
+
+When you use an action in the job, it can be a stand alone action which doesn't need `pre` and `post`, or it can be an action that serves steps before it which might need `pre` to initialize or it can be an action that serves steps after it which might need `post` to cleanup.
+It doesn't really make sense to me that an action needs to have `pre`, `main` and `post` all together.
```
TingluoHuang: I can't think any scenario to use `pre`, `post` and `main` all together.
Can anyone find one for me? 😄

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,73 @@
...
+ name: 'My action with post'
+ description: 'My action with post'
+ runs:
+   using: 'node12'
+   main: 'index.js'
+   post: 'cleanup.js'
+   post-if: 'success()' // Optional
+```
+
+Container Action Example:
```
chrispat: In the container case will the runtime use the same container instance for each of the phases?

TingluoHuang: no, since all container action runs with `docker run`

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,73 @@
+# ADR 361: Wrapper Action
+
+**Date**: 2020-03-06
+
+**Status**: Pending
+
+## Context
+
+In addition to action's regular execution, action author may wants their action has a chance to participate in:
+- Job initialize
```
ericsciple: Why participate in job initialize?

- If for docker login, we can solve that a different way
- If for running stuff on host (for container job), conflicts with future policy scenarios.

We should do the `reserve-this-agent-across-multiple-jobs` feature instead.

docs/adrs/0361-wrapper-action.md
ericsciple: 🤔 ... nope!

docs/adrs/0361-wrapper-action.md
chrispat: That is what I figured.  This needs to be clear as we document because it has impacts on runtime state.

docs/adrs/0361-wrapper-action.md
chrispat: Probably not but I will confess that history shows me I can't possible anticipate all of the things our users will want to do.

TingluoHuang: then, the order of all posts get to run will not be the reverse order of steps, since we lazily adding post only if the step's `pre`/`main` is going to run (to make local action with `post` work), it will be the reverse order of step's pre or main.
ex:
`action_foo` has pre/main/post
`action_bar` has main/post

workflow:
```yaml
steps:
- use: actions/action_bar@master
- use: actions/action_foo@master
```

Steps get run:
- foo.pre
- bar.main
- foo.main
- bar.post
- foo.post

docs/adrs/0361-wrapper-action.md
TingluoHuang: @chrispat for any other scenarios. 😄

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,75 @@
...
+   using: 'docker'
+   image: 'mycontainer:latest'
+   pre-entrypoint: 'setup.sh'
+   pre-if: 'success()' // Optional
+   entrypoint: 'entrypoint.sh'
+   post-entrypoint: 'cleanup.sh'
+   post-if: 'success()' // Optional
+```
+
+Both `pre` and `post` will has default `pre-if/post-if` sets to `always()`.
```
ericsciple: for `pre` what about `succeeded() || <MAIN_CONDITION>`

ericsciple: Actually why not just follow the condition the user specifies in their yaml?

ericsciple: ...for pre

TingluoHuang: what about the main condition is base on another step's outputs, in that case the `pre` won't run since the output doesn't exists.

chrispat: Why wouldn't pre run?  It can't be conditional based on the output of other steps.

TingluoHuang: if we do what Eric suggest `succeeded() || <MAIN_CONDITION>`, then there is a chance the `pre` won't run.

ericsciple: It's a tradeoff, but I think that makes more sense, than pre always running

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,75 @@
...
+- has `pre` and `main`
+- has `main` and `post`
+- has `pre`, `main` and `post`
+
+Invalid action:
+- only has `pre`
+- only has `post`
+- has `pre` and `post`
+
+Potential downside of introducing `pre`:
```
ericsciple: Personally i'm not in favor of adding `pre`. Will review and give technical feedback.

Thanks for adding the "downsides" section.

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,75 @@
...
+   post-entrypoint: 'cleanup.sh'
+   post-if: 'success()' // Optional
+```
+
+Both `pre` and `post` will has default `pre-if/post-if` sets to `always()`.  
+Setting `pre` to `always()` will make sure no matter what condition evaluate result the `main` gets at runtime, the `pre` has always run already.   
+`pre` executes in order of how the steps are defined.  
+`pre` will always be added to job steps list during job setup.  
+> Action referenced from local repository (`./my-action`) won't get `pre` setup correctly since the repository haven't checkout during job initialize.  
+> We can't use GitHub api to download the repository since there is a about 3 mins delay between `git push` and the new commit available to download using GitHub api.
```
ericsciple: Also customers are using `./` to enable private actions, by first checking-out the private repo.

So it also means `pre` wont work for people doing the "private action" workaround today.

docs/adrs/0361-wrapper-action.md
```diff
@@ -0,0 +1,75 @@
...
+Invalid action:
+- only has `pre`
+- only has `post`
+- has `pre` and `post`
+
+Potential downside of introducing `pre`:
+
+- Extra magic wrt step order. Users should control the step order. Especially when we introduce templates.  
+- Eliminates the possibility to lazily download the action tarball, since `pre` always run by default, we have to download the tarball to check whether action defined a `pre`  
+- `pre` doesn't work with local action, we suggested customer use local action for testing their action changes, ex CI for their action, to avoid delay between `git push` and GitHub repo tarball download api.
```
ericsciple: This means `pre` also doesnt work for "private actions" (today customers are checking out private repo then using `./`)

chrispat: Yes that is correct and it is a tradeoff that I am willing to take.

docs/adrs/0361-wrapper-action.md
chrispat: What if the main condition can't be satisfied in the `pre` which could easily be the case.

TingluoHuang: that's why i said pre always run. :)

PR #363, opened 2020-03-09
Change problem matchers output to debug

konradpabjan:
Our setup actions use problem matchers to display information if something goes wrong.

It is a little annoying though that information always shows up about adding problem matchers. Distracts from the main purpose of our actions.

`setup-python` output
image

`setup-dotnet` output
image

`setup-node` output
image

IMO this seems like debug information that should only show up if `ACTIONS_STEP_DEBUG` is set to `true`

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -736,7 +736,7 @@ namespace GitHub.Runner.Worker
                 var owners = config.Matchers.Select(x => $"'{x.Owner}'");
                 var joinedOwners = string.Join(", ", owners);
                 // todo: loc
-                this.Output($"Added matchers: {joinedOwners}. Problem matchers scan action output for known warning or error strings and report these inline.");
+                this.Debug($"Added matchers: {joinedOwners}. Problem matchers scan action output for known warning or error strings and report these inline.");
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -736,7 +736,7 @@ public void AddMatchers(IssueMatchersConfig config)
                 var owners = config.Matchers.Select(x => $"'{x.Owner}'");
                 var joinedOwners = string.Join(", ", owners);
                 // todo: loc
-                this.Output($"Added matchers: {joinedOwners}. Problem matchers scan action output for known warning or error strings and report these inline.");
+                this.Debug($"Added matchers: {joinedOwners}. Problem matchers scan action output for known warning or error strings and report these inline.");
```
cclauss:
```suggestion
                this.Debug($"Added problem matcher(s): {joinedOwners}");
```

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -778,7 +778,7 @@ public void RemoveMatchers(IEnumerable<string> owners)
                 owners = removedMatchers.Select(x => $"'{x.Owner}'");
                 var joinedOwners = string.Join(", ", owners);
                 // todo: loc
-                this.Output($"Removed matchers: {joinedOwners}");
+                this.Debug($"Removed matchers: {joinedOwners}");
```
cclauss:
```suggestion
                this.Debug($"Removed problem matchers: {joinedOwners}");
```

ericsciple:
@juliobbv i thought this followed the echo setting?

ericsciple:
seems fine to me

PR #366, opened 2020-03-10
Expose job name as $GITHUB_JOB

TingluoHuang:
https://github.com/github/c2c-actions-runtime/pull/387


src/Runner.Worker/StepsRunner.cs
```diff
@@ -95,7 +95,10 @@ public async Task RunAsync(IExecutionContext jobContext)
                     if (step is IActionRunner actionStep)
                     {
                         // Set GITHUB_ACTION
-                        step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
+                        if (!string.IsNullOrEmpty(actionStep.Action.ContextName))
```
ericsciple: Lets touch base tomorrow. The rest of the PR seems fine. Not sure what to do about this. Would need scope name to disambiguate. Might be better to just leave it as refname? I dont know

ericsciple: but yes contextName is null unless explicitly defined by customer so this change is better than it was

ericsciple: feel free to merge since this is net improvement, we can touch base later

PR #369, opened 2020-03-13
support workflow.defaults.run and jobs.<jobid>.defaults.run

TingluoHuang:
https://github.com/github/c2c-actions-runtime/pull/413

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -58,12 +58,21 @@ namespace GitHub.Runner.Worker.Handlers
             string shellCommandPath = null;
             bool validateShellOnHost = !(StepHost is ContainerStepHost);
             string prependPath = string.Join(Path.PathSeparator.ToString(), ExecutionContext.PrependPath.Reverse<string>());
-            Inputs.TryGetValue("shell", out var shell);
+            string shell = null;
+            if (!Inputs.TryGetValue("shell", out shell) || string.IsNullOrEmpty(shell))
+            {
+                // TODO: figure out how defaults interact with template later
+                // for now, we won't check job.defaults if we are inside a template.
+                if (string.IsNullOrEmpty(ExecutionContext.ScopeName) && ExecutionContext.JobDefaults.TryGetValue("run", out var runDefaults))
+                {
+                    runDefaults.TryGetValue("shell", out shell);
+                }
+            }
             if (string.IsNullOrEmpty(shell))
             {
 #if OS_WINDOWS
                 shellCommand = "pwsh";
-                if(validateShellOnHost)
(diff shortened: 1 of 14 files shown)


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -58,12 +58,19 @@ public override void PrintActionDetails(ActionRunStage stage)
             string shellCommandPath = null;
             bool validateShellOnHost = !(StepHost is ContainerStepHost);
             string prependPath = string.Join(Path.PathSeparator.ToString(), ExecutionContext.PrependPath.Reverse<string>());
-            Inputs.TryGetValue("shell", out var shell);
+            string shell = null;
+            if (!Inputs.TryGetValue("shell", out shell))
+            {
+                if(ExecutionContext.JobDefaults.TryGetValue("run", out var runDefaults))
```
ericsciple: formatting :) `if (`

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -58,12 +58,19 @@ public override void PrintActionDetails(ActionRunStage stage)
             string shellCommandPath = null;
             bool validateShellOnHost = !(StepHost is ContainerStepHost);
             string prependPath = string.Join(Path.PathSeparator.ToString(), ExecutionContext.PrependPath.Reverse<string>());
-            Inputs.TryGetValue("shell", out var shell);
+            string shell = null;
```
ericsciple: different cases to think through. Not sure what behavior makes sense:

- Default defined, shell input defined but empty.

Does this fallback to job default? Or does it fallback to the handler default?

ericsciple: Consider from a template, how do you specify use the handler default.

ericsciple: Since we dont have templates yet, should we add a TODO to figure out?

ericsciple: Do we know if we're within a scope. Can we fail if inside template? That way we are forced to reconcile later rather than accidentally choosing a behavior :)

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -58,12 +58,19 @@ public override void PrintActionDetails(ActionRunStage stage)
...
+                {
+                    runDefaults.TryGetValue("shell", out shell);
+                }
+            }
             if (string.IsNullOrEmpty(shell))
             {
 #if OS_WINDOWS
                 shellCommand = "pwsh";
-                if(validateShellOnHost)
+                if (validateShellOnHost)
```
ericsciple: nice

PR #372, opened 2020-03-16
add step outcome and step conclusion

TingluoHuang:
https://github.com/actions/runner/pull/274

### DIFF
--- a/src/Runner.Worker/StepsRunner.cs
+++ b/src/Runner.Worker/StepsRunner.cs
@@ -351,6 +351,7 @@ namespace GitHub.Runner.Worker
 
                 if (continueOnError)
                 {
+                    step.ExecutionContext.Outcome = step.ExecutionContext.Result;
                     step.ExecutionContext.Result = TaskResult.Succeeded;
                     Trace.Info($"Updated step result (continue on error)");
                 }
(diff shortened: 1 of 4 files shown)


src/Runner.Worker/StepsRunner.cs
```diff
@@ -458,10 +459,10 @@ private void CompleteStep(IStep step, IStep nextStep, TaskResult? result = null,
                         return;
                     }
 
+                    var parentScopeName = scope.ParentName;
```
ericsciple: i dont think this is necessary. Code above should handle already. Outputs is special because leaving a scope. Outcome and conclusion should always be set already when step finishes.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -470,9 +471,15 @@ private void CompleteStep(IStep step, IStep nextStep, TaskResult? result = null,
                             executionContext.Debug($"{reference}='{outputValue}'");
                         }
                     }
+
+                    stepsContext.SetOutcome(parentScopeName, contextName, (step.ExecutionContext.Outcome ?? result ?? step.ExecutionContext.Result ?? TaskResult.Succeeded).ToActionResult().ToString());
+                    stepsContext.SetConclusion(parentScopeName, contextName, (result ?? step.ExecutionContext.Result ?? TaskResult.Succeeded).ToActionResult().ToString());
                 }
             }
 
+            stepsContext.SetOutcome(executionContext.ScopeName, executionContext.ContextName, (step.ExecutionContext.Outcome ?? result ?? step.ExecutionContext.Result ?? TaskResult.Succeeded).ToActionResult().ToString());
```
ericsciple: consider abstract within execution context code if that makes more sense

Issue #368, opened 2020-03-08
The CI environment variable is not set in GitHub Actions


eregon:
**Describe the bug**

The CI environment variable is not set in GitHub Actions. A lot of software relies on that to detect whether it's running in CI.
Example: https://github.com/jruby/jruby/issues/6113#issuecomment-596196601

**Area for Triage**: 

**Question, Bug, or Feature?**: 
Feature

**Virtual environments affected**
- [x] macOS 10.15
- [x] Ubuntu 16.04 LTS
- [x] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

**Expected behavior**
The environment variable CI should be set to `true`, as it is in most CI solutions out there (e.g., TravisCI).

**Actual behavior**
It's not set: https://github.com/ruby/setup-ruby/runs/493307400?check_suite_focus=true

TingluoHuang:
@maxim-lobanov we have set `GITHUB_ACTIONS=true` for this.
https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

eregon:
I think `CI` is a fairly standard environment variable to be sets in CIs. Requiring `GITHUB_ACTIONS` means it's quite verbose in the code and not portable.

TingluoHuang:
GitHub action is not only a CI system, the workflow run can be triggered on any GitHub event, like Issue create/comment, so I am not sure setting `CI=true` is appropriate for those scenarios. 😄 

@chrispat to comment on this.

xt0rted:
You can set environment variables [at the workflow level](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#env) so it's easy to add this where needed.

```yml
on: pull_request
env:
  CI: true
jobs:
  do-things:
```

chrispat:
We could even include this in our starter workflows that are more CI oriented.

hross:
talked to @chrispat and we agree that making this easier is the right thing to do since this seems to reduce the friction.

eregon:
I agree it's easy enough to set it, and starter workflows can help there.
However it's still a potential gotcha when migrating to GitHub Actions which doesn't sounds worth the effort debugging, when I think almost all other CI solutions set it.

I understand GitHub Actions is not just a CI solution, but I'd argue any code running in GitHub Actions is meant to be run "unattended, automated, CI-like" and so I doubt it would hurt to set it.

[...]

PR #374, opened 2020-03-17
set CI=true when launch process in actions runner.

TingluoHuang:
#368


thorn0:
Would be good if you announced such changes somehow. Yesterday this broke our CI. As it turned out, our tests contained code that depended on this variable. It took quite a while until I found out the reason of the failures. See https://github.com/prettier/prettier/pull/7980

TingluoHuang:
@chrispat for an idea on where could we announce this kind of change. :)

chrispat:
There should have been a changelog entry as this went out.  I will look and see if it didn't publish for some reason.  I apologize for the inconvenience.

PR #381, opened 2020-03-22
cache actions on premises and download from .com

ericsciple:
This PR enables downloading actions for GHES alpha. The goal is `actions/checkout` works in GHES alpha.

For alpha only, the following changes
- Remove auth header when downloading actions from github.com
- Do not clear the actions cache between jobs (avoid rate limiting for unauthenticated requests)

For beta, a different solution will be used (actions shipped in box).

### DIFF
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -15,6 +15,9 @@ namespace GitHub.Runner.Common
     [DataContract]
     public sealed class RunnerSettings
     {
+        [DataMember(Name = "IsHostedServer", EmitDefaultValue = false)]
+        private bool? _isHostedServer;
+
         [DataMember(EmitDefaultValue = false)]
         public int AgentId { get; set; }
 
@@ -42,6 +45,21 @@ namespace GitHub.Runner.Common
         [DataMember(EmitDefaultValue = false)]
         public string MonitorSocketAddress { get; set; }
 
+        [IgnoreDataMember]
+        public bool IsHostedServer
+        {
+            get
+            {
+                // Old runners do not have this property. Hosted runners likely don't have this property either.
+                return _isHostedServer ?? true;
+            }
+
+            set
+            {
+                _isHostedServer = value;
+            }
+        }
+
         /// <summary>
         // Computed property for convenience. Can either return:
         // 1. If runner was configured at the repo level, returns something like: "myorg/myrepo"
@@ -67,6 +85,15 @@ namespace GitHub.Runner.Common
                 }
 
                 return repoOrOrgName;
+            }
+        }
+
+        [OnSerializing]
+        private void OnSerializing(StreamingContext context)
+        {
+            if (_isHostedServer.HasValue && _isHostedServer.Value)
+            {
+                _isHostedServer = null;
             }
         }
     }
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -58,8 +58,14 @@ namespace GitHub.Runner.Worker
                 executionContext.Warning("The 'PREVIEW_ACTION_TOKEN' secret is depreciated. Please remove it from the repository's secrets");
             }
 
-            // Clear the cache (local runner)
-            IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
+            // Clear the cache (for self-hosted runners)
+            // Note, temporarily avoid this step for the on-premises product, to avoid rate limiting.
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
+            var isHostedServer = configurationStore.GetSettings().IsHostedServer;
+            if (isHostedServer)
+            {
+                IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
+            }
 
             foreach (var action in actions)
             {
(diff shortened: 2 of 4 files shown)


src/Runner.Common/ConfigurationStore.cs
```diff
@@ -15,6 +15,9 @@ namespace GitHub.Runner.Common
     [DataContract]
     public sealed class RunnerSettings
     {
+        [DataMember(Name = "IsHostedServer")]
```
ericsciple: @TingluoHuang  let me know what you think about this approach. I left the PR in draft for now... looking for early feedback

src/Runner.Worker/ActionManager.cs
```diff
@@ -634,6 +641,19 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+        {
+            if (!File.Exists(watermarkFile))
+            {
+                return false;
+            }
+
+            var content = File.ReadAllText(watermarkFile);
+            return !string.IsNullOrEmpty(content) &&
+                DateTimeOffset.TryParse(content, out DateTimeOffset created) &&
+                DateTimeOffset.Now - created < TimeSpan.FromDays(1);
```
ericsciple: i might change this to something very large, like 10 years since this is just a temporary change. I dont really want actions re-downloading after 1 day.

src/Runner.Worker/ActionManager.cs
```diff
@@ -58,8 +58,14 @@ public async Task<List<JobExtensionRunner>> PrepareActionsAsync(IExecutionContex
                 executionContext.Warning("The 'PREVIEW_ACTION_TOKEN' secret is depreciated. Please remove it from the repository's secrets");
             }
 
-            // Clear the cache (local runner)
-            IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
+            // Clear the cache (for self-hosted runners)
+            // Note, temporarily avoid this step for the on-premises product, to avoid rate limiting.
+            var store = HostContext.GetService<IConfigurationStore>();
+            var isHostedServer = store.GetSettings().IsHostedServer;
+            if (isHostedServer)
```
ericsciple: only do this on hosted service

src/Runner.Common/ConfigurationStore.cs
TingluoHuang: i am wonder whether we should just fix the system.servertype variable since it use to set to `onprem` or `hosted`?

src/Runner.Worker/ActionManager.cs
TingluoHuang: how do we handle tag get updates? ex: checkout@v2 suppose to point to latest version of v2 checkout.
i know when you use the api to download the tarball, you will get a nested zip/tar.gz, the root folder is actually the short SHA of the ref it resolve to, i think we can also keep tracking that. So we will not redownload same action only if the underline resolved SHA is not changed between 2 runs.

src/Runner.Worker/ActionManager.cs
```diff
@@ -634,6 +641,19 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             }
         }
 
+        private bool CheckWatermark(string watermarkFile)
```
ericsciple: delete this function

src/Runner.Worker/ActionManager.cs
```diff
@@ -448,7 +454,8 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             ArgUtil.NotNullOrEmpty(repositoryReference.Ref, nameof(repositoryReference.Ref));
 
             string destDirectory = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Actions), repositoryReference.Name.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar), repositoryReference.Ref);
-            if (File.Exists(destDirectory + ".completed"))
+            string watermarkFile = destDirectory + ".completed";
+            if (CheckWatermark(watermarkFile))
```
ericsciple: revert this line

bryanmacfarlane:
We might need to also not supply the token in this case.  Not sure.  We know that not having a token from dotcom will rate limit us (anonymous).  However, if we supply the wrong token (the one from ghes), the first request might fail.

src/Runner.Common/ConfigurationStore.cs
ericsciple: Setting up GHES now to test. Otherwise will only persist this setting for on-prem, so we can drop it later after alpha.

src/Test/L0/Worker/ActionManagerL0.cs
```diff
@@ -111,6 +111,57 @@ public async void PrepareActions_DownloadActionFromGraph()
             }
         }
 
+        [Fact]
```
ericsciple: finish tests

src/Runner.Worker/ActionManager.cs
```diff
@@ -498,24 +505,33 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
-                                if (string.IsNullOrEmpty(authToken))
-                                {
-                                    // TODO: Depreciate the PREVIEW_ACTION_TOKEN
-                                    authToken = executionContext.Variables.Get("PREVIEW_ACTION_TOKEN");
-                                }
-
-                                if (!string.IsNullOrEmpty(authToken))
+                                var configurationStore = HostContext.GetService<IConfigurationStore>();
+                                var isHostedServer = configurationStore.GetSettings().IsHostedServer;
+                                if (isHostedServer)
```
ericsciple: the diff looks bad, but adding auth header just moved inside if block

Issue #367, opened 2020-03-11
Provide runner as a Docker Image


WtfJoke:
**Describe the enhancement**
Is it possible for you to provide official docker images for letting a runner run?

**Additional information**
We want to use self hosted runner, but we dont want to maintain machines/vms so it would be cool to have a ready to run official docker image in order to use them in services like aws fargate/azure container instances.

If im wrong here and I should have opened a topic in the github actions community forum please let me know and feel free to close the issue/feature request.

vincentbrison:
Some example of runner as a Docker image:
- https://github.com/tcardonne/docker-github-runner
- https://github.com/myoung34/docker-github-actions-runner
- https://github.com/joeyx22lm/docker-github-actions-runner

Additionally, Runner running in Docker is not able to use Docker itself (https://github.com/actions/runner/blob/f9b5d626c5486d1f02b3bd45879f61834b7aeecc/src/Runner.Worker/ContainerOperationProvider.cs#L68), which is a major downside to be able to manage self hosted runners with Kubernetes for example, without losing time playing with VMs or real machines.

WtfJoke:
Thank you @vincentbrison for providing these examples, I found a couple of them as well, but I am not sure if we are allowed to use them in our company (we have a restriction to use only official images). Thats the reason why I asked here. :)

The hint about docker in docker is useful (a couple of application suffer of this problem).  👍

ppamorim:
I don't understand this limitation, what's the difference between this and a VM running in the machine? We are using M1 machines as our hosts and since the native runner on macOS does not support containers, we decided to use Docker, but when executing it, we get this error.

scolastico:
Also its possible with docker to run a docker conatiner in a docker container.
You only need to mount the docker directory `/var/lib/docker` in the normal file system so it doesnt gets mounted as an AUFS mountpoint.

This page i found describes a bit better than me: https://www.docker.com/blog/docker-can-now-run-within-docker/

pdomagala-tr:
@scolastico did you manage to do this?

scolastico:
not in this way but in another: https://github.com/scolastico-dev/s.RunnerManager

pdomagala-tr:
are you able to run docker inside the docker container?

nikola-jokic:
Hey everyone,

There is a bare-minimum runner image available now: ghcr.io/actions/actions-runner:latest
The Dockerfile is available [here](https://github.com/actions/runner/blob/main/images/Dockerfile). 

[...]

PR #383, opened 2020-03-23
Allow using Docker actions when running within a Docker container

jpb:
Add `--allow-docker-in-docker` flag to optionally disable the `Container feature is not supported when runner is already running inside container` constraint.

This resolves an issue described in https://github.com/actions/runner/issues/367#issuecomment-597742895.

I am successfully using this change on my self-hosted runner infrastructure which mounts the host's docker socket into the container where the actions runner executes.

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -118,6 +118,7 @@ namespace GitHub.Runner.Common
                 //validFlags array as well present in the CommandSettings.cs
                 public static class Flags
                 {
+                    public static readonly string AllowDockerInDocker = "allow-docker-in-docker";
                     public static readonly string Commit = "commit";
                     public static readonly string Help = "help";
                     public static readonly string Replace = "replace";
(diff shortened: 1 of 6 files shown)


bryanmacfarlane:
@TingluoHuang 

https://github.com/actions/runner/blob/master/docs/contribute.md#enhancements-and-feature-requests

TingluoHuang:
@jpb thanks for your contribution, do you mind follow [contribute.md](https://github.com/actions/runner/blob/master/docs/contribute.md#enhancements-and-feature-requests) to create a feature request issue and provide an ADR for the proposed changes.
the main concern I have is around volum mount and network setting, those might become different when the runner is running inside a container. We can discuss more in the Feature request issue. 😆

jpb:
Yes – I have created a feature request: https://github.com/actions/runner/issues/406.

DavidGOrtega:
@jpb I have forked the runner to do a quick implement of labels and support docker in docker. I dont need the flags so I just removed the exception. The final thing does not work.

```yaml
  runs-on: [gpu]
    container:
      image: docker://davidgortega/gpu:runner
      options: --gpus "device=0" -v "/var/run/docker.sock":"/var/run/docker.sock"

    steps:
      - uses: actions/checkout@v2
      
      - name: run
        run: |
          nvidia-smi
```

Fails the checkout action
image

Did it work for you or do you have any advice before I get deeper?

Issue #322, opened 2020-02-06
Container/service volumes arent getting added to generated docker commands


dakale:
**Describe the bug**

Give a workflow like:

```yaml
test:
    runs-on: ubuntu-latest
    container:
      image: ubuntu
      volumes: 
      - src:dst
```

The appropriate docker option, `-v src:dst` is sometimes not generated when creating the container

PR #384, opened 2020-03-23
Add expanded volumes strings to container mounts

dakale:
We track the "user" vs "actual", eg UserMountVolumes and MountVolumes for the values given in a workflow vs their expanded values, and the expanded values are what are actually sent to docker when starting the container. 

We lost some history when we forked (https://github.com/actions/runner/blob/c8afc848403652f31f84eb362fc0696ee23cffca/src/Runner.Worker/JobRunner.cs#L124) 
so its unclear why this was removed in the first place

This does not just add it directly back because we moved where the expansion happens, bit the functionality is the same. Now, the values are expanded bu the time we create the ContainerInfo object 

Fixes https://github.com/actions/runner/issues/322

### DIFF
--- a/src/Runner.Worker/Container/ContainerInfo.cs
+++ b/src/Runner.Worker/Container/ContainerInfo.cs
@@ -61,6 +61,7 @@ namespace GitHub.Runner.Worker.Container
                 foreach (var volume in container.Volumes)
                 {
                     UserMountVolumes[volume] = volume;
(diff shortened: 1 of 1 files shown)


TingluoHuang:
@dakale FYI, we remove those code because there isn't a Variables concept in GitHub Actions. 😄

dakale:
@TingluoHuang Makes sense. Any other concerns here? Ill merge before end of day if not

PR #386, opened 2020-03-25
add github.url and github.api_url for ghes alpha

ericsciple:
For GHES alpha release, expose GHES server URL and API URL for actions (e.g. where checkout should clone from)

### DIFF
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -10,6 +10,7 @@ namespace GitHub.Runner.Worker
         {
             "action",
             "actor",
+            "api_url", // temp for GHES alpha release
             "base_ref",
             "event_name",
             "event_path",
@@ -21,6 +22,7 @@ namespace GitHub.Runner.Worker
             "run_id",
             "run_number",
             "sha",
+            "url", // temp for GHES alpha release
             "workflow",
             "workspace",
         };
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Runtime.Serialization;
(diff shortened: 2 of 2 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -127,6 +128,17 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     context.SetRunnerContext("workspace", Path.Combine(_workDirectory, trackingConfig.PipelineDirectory));
                     context.SetGitHubContext("workspace", Path.Combine(_workDirectory, trackingConfig.WorkspaceDirectory));
 
+                    // Temporary hack for GHES alpha
+                    var configurationStore = HostContext.GetService<IConfigurationStore>();
+                    var runnerSettings = configurationStore.GetSettings();
+                    if (!runnerSettings.IsHostedServer && !string.IsNullOrEmpty(runnerSettings.GitHubUrl))
```
brcrista: I think this needs to be `string.IsNullOrEmpty(runnerSettings.GitHubUrl)`.  This will be non-null if you're talking to hosted GitHub -- see [where this is set](https://github.com/actions/runner/blob/dff1024cd3ab8f5b852750be7e1ddc605c3a45da/src/Runner.Listener/Configuration/ConfigurationManager.cs#L98).

brcrista: Or, is that supposed to be an `||` in the middle?

src/Runner.Worker/GitHubContext.cs
```diff
@@ -10,6 +10,7 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
         {
             "action",
             "actor",
+            "api_url", // temp for GHES alpha release
```
brcrista: I'm imagining these are consumed from actions somehow, but I'm having trouble seeing where 🧐
https://github.com/actions/checkout/pull/199

src/Runner.Worker/JobExtension.cs
ericsciple: good catch

@TingluoHuang when `!runnerSettings.IsHostedServer` should we fallback to `runnerSettings.ServerUrl` when `GitHubUrl` is not defined?

TingluoHuang: runnerSettings.ServerUrl is the pipelines url, pipelines.actions.githubusercontent.com/xxxxx

When GitHubUrl is not defined, it means we connected to Devfabric.

src/Runner.Worker/GitHubContext.cs
ericsciple: url-helper.ts

src/Runner.Worker/JobExtension.cs
ericsciple: ok i think the change here might be fine then, but additional change to figure out for registration against the appliance

ericsciple: discussed offline w/ ting, leaving this way for now. When config against appliance works, will need an additional change to persist githuburl for that scenario.

PR #396, opened 2020-03-30
Report git sha

jsoref:
Currently if someone uses a github action by tag (as opposed to sha), there's no way to know what precise version of the action was used.

I'm not sure if this works as written, but I'd like something like this to be implemented.


TingluoHuang:
Please log an issue for the problem you described, I will fix it. The PR you have might not work as you want. 😄

jsoref:
#398

Yeah, I didn't precisely expect it to work. It was just a pointer as well as an indication that I'm willing to write the code if necessary.

TingluoHuang:
i am going to close this pr and work on the issue.

PR #397, opened 2020-03-30
ADR 397: Configuration time custom labels

bryanmacfarlane:
👉 [rendered](https://github.com/actions/runner/blob/reg-labels-adr/docs/adrs/0397-runner-registration-labels.md) 👈 

Since configuring self-hosted runners is commonly automated via scripts, the labels need to be able to be created during configuration.  The runner currently registers the built-in labels (os, arch) during registration but does not accept labels via command line args to extend the set registered.


docs/adrs/0397-runner-registration-labels.md
```diff
@@ -0,0 +1,51 @@
...
+This would add the label `mylabel` and `anotherlabel` to the runner, and enable users to select the runner in their workflow using this label:
+```yaml
+runs-on: [self-hosted, mylabel, anotherlabel]
+```
+
+It would not be possible to remove labels from an existing runner using `config.sh`, instead labels would have to be removed using the GitHub UI.
+
+## Overriding built-in labels
+
+Note that it is possible to register "built-in" hosted labels like `ubuntu-latest` and is not considered an error.  This is an effective way for the org / runner admin to dictate by policy through registration that this set of runners will be used without having to edit all the workflow files not and in the future.
```
joshmgross:
```suggestion
Note that it is possible to register "built-in" hosted labels like `ubuntu-latest` and is not considered an error.  This is an effective way for the org / runner admin to dictate by policy through registration that this set of runners will be used without having to edit all the workflow files now and in the future.
```

docs/adrs/0397-runner-registration-labels.md
```diff
@@ -0,0 +1,51 @@
...
+
+If a self-hosted runner built-in label such as os or arch, in the case that it is valid (`linux` on a linux os), it will have no effect.  If it's invalid (supplying `windows` on a linux machine) it will error.
```
joshmgross: Will this validation be done by the runner or the backend?

bryanmacfarlane: I think it has to be done by the runner since it can only say definitively what the current os and arch is.

lokesh755: Do we really want to throw error  when user supplies `windows` label on linux machine ? It will provide more flexibility to the user.

TingluoHuang: Feels like if we allow user tag Windows-latest to any runner then we don’t need to restrict windows label.

joshmgross: I don't think we're currently planning to have a similar restriction in the UI

bryanmacfarlane: OK, I'll remove and clarify we won't restrict. (I was carrying this over from the previous signed off ADR)

docs/adrs/0397-runner-registration-labels.md
```diff
@@ -0,0 +1,50 @@
...
+
+This is another version of [ADR275](https://github.com/actions/runner/pull/275)
+
+## Decision
+
+This ADR proposes that we add a `--labels` option to `config`, which could be used to add custom additional labels to the configured runner.
+
+For example, to add a single extra label the operator could run:
+```bash
+./config.sh --labels mylabel
```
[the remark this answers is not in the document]
bryanmacfarlane: If the action is to replace the runner, it should replace the labels.  I'll add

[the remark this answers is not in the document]
bryanmacfarlane: pushed.

docs/adrs/0397-runner-registration-labels.md
```diff
@@ -0,0 +1,50 @@
...
+> Note: the current runner command line parsing and envvar override algorithm only supports a single argument (key).
+
+This would add the label `mylabel` to the runner, and enable users to select the runner in their workflow using this label:
+```yaml
+runs-on: [self-hosted, mylabel]
+```
+
+To add multiple labels the operator could run:
+```bash
+./config.sh --labels mylabel,anotherlabel
```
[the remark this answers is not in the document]
bryanmacfarlane: pushed

Issue #398, opened 2020-03-30
Report SHA when retrieving GitHub actions


jsoref:
If you have a feature request that is relevant to this repository, the runner, then please include the information below:

**Describe the enhancement**
Currently if someone uses a github action by tag (as opposed to sha), there's no way to know what precise version of the action was used.

Please report the SHA used for each docker action – specifically in cases when actions aren't pinned to SHAs (they're often pinned to symbolic names – branches/tags)

**Code Snippet**
(Any workflow that specifies a docker action by a symbolic name.)

**Additional information**

NOTE: if the feature request has been agreed upon then the assignee will create an ADR.   See docs/adrs/README.md

jsoref:
@TingluoHuang: Is it possible to actually report this now with the introduction of `ActionDownloadInfo.ResolvedSha` in #515? I can't see any sign that this is actually surfaced, and I'd still very much like to have this information...

TingluoHuang:
@jsoref YES. I will create a PR.

TingluoHuang:
PR merged and the change will get release soon.

jsoref:
@TingluoHuang: Don't suppose you have a definition of `soon`? It hasn't arrived yet. How often are releases?

thboop:
> @TingluoHuang: Don't suppose you have a definition of `soon`? It hasn't arrived yet. How often are releases?

Hey @jsoref , this should be available for all customers by the end of this week. Its found in release [2.280.1](https://github.com/actions/runner/releases/tag/v2.280.1).

jsoref:
Is there a way for a workflow/action to know the version of the runtime?
I can't see a way to identify it in:
https://docs.github.com/en/actions/reference/environment-variables

I understand I could programmatically retrieve the logs and parse them by hand, but....

PR #400, opened 2020-03-31
print SHA for referenced action.

TingluoHuang:
https://github.com/actions/runner/issues/398

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -633,7 +633,7 @@ namespace GitHub.Runner.Worker
                         // we will print the short SHA when action is referenced via branch/tag
                         if (repositoryReference.Ref.StartsWith(shortSha, StringComparison.OrdinalIgnoreCase))
                         {
-                            // actions is already referenced by SHA
+                            // action is already referenced by SHA
                             shortSha = null;
                         }
                     }
@@ -909,4 +909,3 @@ namespace GitHub.Runner.Worker
         public string ActionRepository { get; set; }
     }
 }
-
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -465,7 +465,7 @@ namespace GitHub.Runner.Worker
                 // make sure we get a clean folder ready to use.
                 IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
                 Directory.CreateDirectory(destDirectory);
-                executionContext.Output($"Download action repository '{repositoryReference.Name}@{repositoryReference.Ref}'");
+                executionContext.Debug($"Downloading action repository '{repositoryReference.Name}@{repositoryReference.Ref}'");
             }
 
 #if OS_WINDOWS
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -613,14 +613,35 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
                 }
 #endif
 
-                // repository archive from github always contains a nested folder
+                // repository archive from github always contains a nested folder, the nested folder contains the short SHA for the ref
+                // ex: actions/checkout@master the nested folder looks like actions-checkout-01aeccc
+                // However, when you reference action using tag, the SHA in the folder name is not really the commit SHA, it's the SHA for the tag itself.
```
TingluoHuang: I didn't find an ideal solution for handle reference actions by tag, since github api only supports `GET: /repos/org/repo/tags`, it doesn't allow query by tagname, so i have to make multiple api calls to get the tag object then find out which commit the tag is pointing to.

src/Runner.Worker/ActionManager.cs
```diff
@@ -613,14 +613,35 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                    shortSha = subDirectories[0].Name;
+                    var splitIndex = shortSha.LastIndexOf('-');
+                    if (splitIndex > 0 && splitIndex < shortSha.Length - 1)
+                    {
+                        shortSha = shortSha.Substring(splitIndex + 1);
+
+                        // we will print the short SHA when action is referenced via branch/tag
+                        if (repositoryReference.Ref.StartsWith(shortSha, StringComparison.OrdinalIgnoreCase))
+                        {
+                            // actions is already referenced by SHA
```
jsoref:
```suggestion
                            // action is already referenced by SHA
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -613,14 +613,35 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                // repository archive from github always contains a nested folder, the nested folder contains the short SHA for the ref
+                // ex: actions/checkout@master the nested folder looks like actions-checkout-01aeccc
+                // However, when you reference action using tag, the SHA in the folder name is not really the commit SHA, it's the SHA for the tag itself.
+                var shortSha = "";
                 var subDirectories = new DirectoryInfo(stagingDirectory).GetDirectories();
                 if (subDirectories.Length != 1)
                 {
                     throw new InvalidOperationException($"'{archiveFile}' contains '{subDirectories.Length}' directories");
                 }
                 else
```
jsoref: Not related to this PR, but is the style to actually have an `else` after `throw`/`return`?

jsoref:
@TingluoHuang: what's the process for this being integrated?

TingluoHuang:
@jsoref my change doesn't work for reference action using tag. the ref that prints out is for the tag itself and not the commit that the tag points to. 😢

src/Runner.Worker/ActionManager.cs
jsoref: https://api.github.com/repos/Microsoft/terminal/git/ref/tags/1810.02002

https://api.github.com/repos/:org/:repo/git/ref/tags/:tagname (can include `/`) -- https://developer.github.com/v3/git/refs/#get-a-single-reference

jsoref:
@TingluoHuang: Is that API good enough? (There is the minor detail that I am not sure how you'd know if you're dealing w/ a tag or something else... I hope there's some standard priority for tags / branches...)

TingluoHuang:
@jsoref that is the problem, when I see a string `FOO`, I have to try twice in order to figure out whether `FOO` is a tag or a branch. So I have to make 2 extra network requests just for getting trace info.

jsoref:
Presumably, something went through the work of resolving the symbol to a thing, is that information entirely lost? :-(

PR #401, opened 2020-04-01
Fix runner config IsHostedServer detection for GHES alpha

ericsciple:
This PR fixes the IsHostedServer detection for GHES. Proper detection during config, is needed in order to set GITHUB_URL and GITHUB_API_URL during runtime

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -115,7 +115,7 @@ namespace GitHub.Runner.Listener.Configuration
                 try
                 {
                     // Determine the service deployment type based on connection data. (Hosted/OnPremises)
-                    runnerSettings.IsHostedServer = await IsHostedServer(runnerSettings.ServerUrl, creds);
+                    runnerSettings.IsHostedServer = IsHostedServer(new UriBuilder(runnerSettings.GitHubUrl));
 
                     // Validate can connect.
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -246,14 +246,6 @@ public async Task ConfigureAsync(CommandSettings command)
             {
                 UriBuilder configServerUrl = new UriBuilder(runnerSettings.ServerUrl);
                 UriBuilder oauthEndpointUrlBuilder = new UriBuilder(agent.Authorization.AuthorizationUrl);
-                if (!runnerSettings.IsHostedServer && Uri.Compare(configServerUrl.Uri, oauthEndpointUrlBuilder.Uri, UriComponents.SchemeAndServer, UriFormat.Unescaped, StringComparison.OrdinalIgnoreCase) != 0)
```
ericsciple, line 249: changing the urls interferred so i had to delete

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -495,34 +487,18 @@ private TaskAgent CreateNewAgent(string agentName, RSAParameters publicKey)
             return agent;
         }
 
-        private async Task<bool> IsHostedServer(string serverUrl, VssCredentials credentials)
+        private bool IsHostedServer(UriBuilder gitHubUrl)
```
ericsciple, line 490: hack for alpha. Flags was always "Hosted"

PR #405, opened 2020-04-03
Better Kusto Tracing for self-hosted runner.

TingluoHuang:
Adding following telemetry for self-hosted runner

- Add OrchestrationId when self-hosted runner renew lock.
- Add OAuth ClientId into useragent so we can distinguish hosted vs. self-hosted

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -24,7 +24,7 @@ namespace GitHub.Runner.Common
         CancellationToken RunnerShutdownToken { get; }
         ShutdownReason RunnerShutdownReason { get; }
         ISecretMasker SecretMasker { get; }
-        ProductInfoHeaderValue UserAgent { get; }
+        List<ProductInfoHeaderValue> UserAgents { get; }
         RunnerWebProxy WebProxy { get; }
         string GetDirectory(WellKnownDirectory directory);
         string GetConfigFile(WellKnownConfigFile configFile);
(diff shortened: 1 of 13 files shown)


src/Runner.Common/HostContext.cs
```diff
@@ -189,6 +189,17 @@ public HostContext(string hostType, string logFile = null)
...
             }
+
+            var credFile = GetConfigFile(WellKnownConfigFile.Credentials);
+            if (File.Exists(credFile))
+            {
+                var credData = IOUtil.LoadObject<CredentialData>(credFile);
+                if (credData != null &&
+                    credData.Data.TryGetValue("clientId", out var clientId))
+                {
+                    _userAgents.Add(new ProductInfoHeaderValue($"RunnerId", clientId));
```
TingluoHuang, line 200: Only self-hosted runner has the oauth app `clientid`

src/Sdk/DTWebApi/WebApi/TaskAgentHttpClient.cs
```diff
@@ -104,7 +105,30 @@ public class TaskAgentHttpClient : TaskAgentHttpClientBase
                 LockedUntil = expiresOn,
             };
 
-            return UpdateAgentRequestAsync(poolId, requestId, lockToken, request, userState, cancellationToken);
+            var additionalHeaders = new Dictionary<string, string>();
+            if (!string.IsNullOrEmpty(orchestrationId))
+            {
+                additionalHeaders["X-VSS-OrchestrationId"] = orchestrationId;
```
TingluoHuang, line 111: this Header will make everything on the service for this request has the same orchestrationid (activitylog/producttrace)

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -86,15 +87,29 @@ public void Run(Pipelines.AgentJobRequestMessage jobRequestMessage, bool runOnce
...
             }
 
+            var orchestrationId = string.Empty;
+            var systemConnection = jobRequestMessage.Resources.Endpoints.SingleOrDefault(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
+            var accessToken = systemConnection?.Authorization?.Parameters["AccessToken"];
+            if (!string.IsNullOrEmpty(accessToken))
+            {
+                var jwt = JsonWebToken.Create(accessToken);
+                var claims = jwt.ExtractClaims();
+                orchestrationId = claims.FirstOrDefault(x => string.Equals(x.Type, "orchid", StringComparison.OrdinalIgnoreCase))?.Value;
```
TingluoHuang, line 97: orchestrationid is part of the token's claims

PR #414, opened 2020-04-08
Added support for custom labels

lokesh755:
Adds support for command line arg `labels` to add custom labels.

Eg.
```
./config.sh --url https://github.com/actions --token FUGFUYTYTIYUTIGIUYH --labels label1,label2
```

Interactive experience
```
--------------------------------------------------------------------------------
|        ____ _ _   _   _       _          _        _   _                      |
|       / ___(_) |_| | | |_   _| |__      / \   ___| |_(_) ___  _ __  ___      |
|      | |  _| | __| |_| | | | | '_ \    / _ \ / __| __| |/ _ \| '_ \/ __|     |
|      | |_| | | |_|  _  | |_| | |_) |  / ___ \ (__| |_| | (_) | | | \__ \     |
|       \____|_|\__|_| |_|\__,_|_.__/  /_/   \_\___|\__|_|\___/|_| |_|___/     |
|                                                                              |
|                       Self-hosted runner registration                        |
|                                                                              |
--------------------------------------------------------------------------------

# Authentication

What is the URL of your repository? https://github.com/orgname/reponame
What is your runner register token? *****************************

√ Connected to GitHub

# Runner Registration

Enter the name of runner: [press Enter for LOKESHGOPU] lokesh-test-runner

This runner will have the following labels: 'self-hosted', 'Windows', 'X64'
Enter any additional labels (ex. label-1,label-2): [press Enter to skip] test-label1,test-label2

√ Runner successfully added
√ Runner connection is good

# Runner settings

Enter name of work folder: [press Enter for _work]

√ Settings Saved.

Would you like to run the runner as service? (Y/N) [press Enter for N]
```

[ADR 0397: Support adding custom labels during runner config](https://github.com/actions/runner/blob/master/docs/adrs/0397-runner-registration-labels.md)


src/Runner.Listener/CommandSettings.cs
```diff
@@ -249,6 +250,19 @@ public string GetStartupType()
             return GetArg(Constants.Runner.CommandLine.Args.StartupType);
         }
 
+        public ISet<string> GetLabels()
+        {
+            var labelSet = new HashSet<string>();
```
TingluoHuang, line 255: OrdinalIgnoreCase

TingluoHuang:
@lokesh755 LGTM, let me know what's the decision for interactive config. I can take another look then.

src/Runner.Listener/CommandSettings.cs
```diff
@@ -249,6 +250,24 @@ public string GetStartupType()
             return GetArg(Constants.Runner.CommandLine.Args.StartupType);
         }
 
+        public ISet<string> GetLabels(string description)
+        {
+            var labelSet = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            string labels = GetArgOrPrompt(
+                name: Constants.Runner.CommandLine.Args.Labels,
+                description: description,
+                defaultValue: string.Empty,
```
TingluoHuang, line 259: do we need to have both `defaultValue` and `isOptional`? the `defaultValue` suppose to cover the optional scenario.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -160,12 +160,20 @@ public async Task ConfigureAsync(CommandSettings command)
...
 
             TaskAgent agent;
+            ISet<string> userLabels = null;
             while (true)
             {
                 runnerSettings.AgentName = command.GetRunnerName();
 
                 _term.WriteLine();
 
+                if (userLabels == null)
```
TingluoHuang, line 170: should we also ask for labels? I don't think we ever reuse input on error.

src/Runner.Listener/CommandSettings.cs
lokesh755: I would like to target scenarios where the default value is `string.Empty` and the value is optional, so that we can display `[Enter to skip]` which speaks out it's optional parameter.

src/Runner.Listener/Configuration/ConfigurationManager.cs
lokesh755: I feel most of the times the error is raised when there exists an agent with the same name. I feel it might be tedious for user to re-enter the labels which he already entered. But I'm fine to change the behavior.

src/Runner.Listener/CommandSettings.cs
```diff
@@ -249,6 +250,24 @@ public string GetStartupType()
             return GetArg(Constants.Runner.CommandLine.Args.StartupType);
         }
 
+        public ISet<string> GetLabels(string description)
```
TingluoHuang, line 253: do we need to take the `description` as input? feels like we have the full context of generating description here.

src/Runner.Listener/CommandSettings.cs
TingluoHuang: cool. 👍

src/Runner.Listener/Configuration/ConfigurationManager.cs
TingluoHuang: let's follow the existing pattern just re-ask for everything. 😄

src/Runner.Listener/CommandSettings.cs
lokesh755: To display this -> `This runner will have the following labels: 'self-hosted', 'Windows', 'X64'`, which has dynamic labels info.

TingluoHuang: it only use `{VarUtil.OS}', '{VarUtil.OSArchitecture}` which are static property?

PR #416, opened 2020-04-08
launch middleman process on macOS to workaround SIP limit

TingluoHuang:
When `DYLD_INSERT_LIBRARIES` set on macOS, SIP will remove it when launch process that under `/usr/bin`, ex: `bash`

We will use `node` as a proxy process on macOS when `DYLD_INSERT_LIBRARIES` is set to execute `runs` script.

More detail:
https://github.com/github/c2c-actions-runtime/issues/447

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -259,6 +259,16 @@ namespace GitHub.Runner.Worker.Handlers
 
             // dump out the command
             var fileName = isContainerStepHost ? shellCommand : commandPath;
+#if OS_OSX
+            if (Environment.ContainsKey("DYLD_INSERT_LIBRARIES"))  // We don't check `isContainerStepHost` because we don't support container on macOS
+            {
+                // launch `node macOSRunInvoker.js shell args` instead of `shell args` to avoid macOS SIP remove `DYLD_INSERT_LIBRARIES` when launch process
+                string node12 = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node12", "bin", $"node{IOUtil.ExeExtension}");
+                string macOSRunInvoker = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "macos-run-invoker.js");
+                arguments = $"\"{macOSRunInvoker.Replace("\"", "\\\"")}\" \"{fileName.Replace("\"", "\\\"")}\" {arguments}";
+                fileName = node12;
+            }
+#endif
             ExecutionContext.Debug($"{fileName} {arguments}");
 
             using (var stdoutManager = new OutputManager(ExecutionContext, ActionCommandManager))
--- a/src/Misc/layoutbin/macos-run-invoker.js
+++ b/src/Misc/layoutbin/macos-run-invoker.js
@@ -1,0 +1,13 @@
+const { spawn } = require('child_process');
+// argv[0] = node
+// argv[1] = macos-run-invoker.js
+var shell = process.argv[2];
+var args = process.argv.slice(3);
+console.log(`::debug::macos-run-invoker: ${shell}`);
+console.log(`::debug::macos-run-invoker: ${JSON.stringify(args)}`);
+var launch = spawn(shell, args, { stdio: 'inherit' });
+launch.on('exit', function (code) {
+    if (code !== 0) {
+        process.exit(code);
+    }
+});


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -259,6 +259,16 @@ public async Task RunAsync(ActionRunStage stage)
 
             // dump out the command
             var fileName = isContainerStepHost ? shellCommand : commandPath;
+#if OS_OSX
+            if (Environment.ContainsKey("DYLD_INSERT_LIBRARIES"))
+            {
+                // launch `node macOSRunInvoker.js shell args` instead of `shell args` to avoid macOS SIP remove `DYLD_INSERT_LIBRARIES` when launch process
+                string node12 = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node12", "bin", $"node{IOUtil.ExeExtension}");
+                string macOSRunInvoker = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "macOSRunInvoker.js");
+                arguments = $"\"{macOSRunInvoker.Replace("\"", "\\\"")}\" \"{fileName.Replace("\"", "\\\"")}\" {arguments}";
```
aibaars, line 268: `Replace("\"", "\\\"")` is not enough to escape an arbitrary string for the command line. For a unix-like OS, I think you should at least also escape `\` , `$` and ` . 

Isn't there a utility function somewhere for escaping an argument string in an OS specific way?

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -267,6 +277,7 @@ public async Task RunAsync(ActionRunStage stage)
                 StepHost.OutputDataReceived += stdoutManager.OnDataReceived;
                 StepHost.ErrorDataReceived += stderrManager.OnDataReceived;
 
+
```
aibaars, line 280:
```suggestion
```

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -259,6 +259,16 @@ public async Task RunAsync(ActionRunStage stage)
 
             // dump out the command
             var fileName = isContainerStepHost ? shellCommand : commandPath;
+#if OS_OSX
+            if (Environment.ContainsKey("DYLD_INSERT_LIBRARIES"))
```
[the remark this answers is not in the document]
TingluoHuang: we don't need to check that since `isContainerStepHost` can only be true on Linux. :)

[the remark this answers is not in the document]
aibaars: Fair enough, although someone might add support for container-based actions for  OSX some day and be really confused why it is trying to run the external node binary in the container :)

src/Runner.Worker/Handlers/ScriptHandler.cs
TingluoHuang: we don't need to escape those since those are special character for bash/sh, the runner launch process directly, so we only need to worry about `"` and ` `(space).

src/Runner.Worker/Handlers/ScriptHandler.cs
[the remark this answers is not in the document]
TingluoHuang: i will leave a comment in the code. :)

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -259,6 +259,16 @@ public async Task RunAsync(ActionRunStage stage)
 
             // dump out the command
             var fileName = isContainerStepHost ? shellCommand : commandPath;
+#if OS_OSX
+            if (Environment.ContainsKey("DYLD_INSERT_LIBRARIES"))  // We don't check `isContainerStepHost` because we don't support container on macOS
+            {
+                // launch `node macOSRunInvoker.js shell args` instead of `shell args` to avoid macOS SIP remove `DYLD_INSERT_LIBRARIES` when launch process
+                string node12 = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Externals), "node12", "bin", $"node{IOUtil.ExeExtension}");
+                string macOSRunInvoker = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "macOSRunInvoker.js");
```
ericsciple, line 267: consider hypens instead of camel case `macos-run-invoker`

src/Runner.Worker/Handlers/ScriptHandler.cs
ericsciple: Should we pass the command line using an environment variable instead? Then we avoid escaping complexity

src/Misc/layoutbin/macOSRunInvoker.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+// Copyright (c) GitHub. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+const { spawn } = require('child_process');
+var shell = process.argv[2];
+var args = process.argv.slice(3);
+var launch = spawn(shell, args, { stdio: 'inherit' });
+launch.on('exit', function (code) {
```
ericsciple, line 9: make sure you got the right one (exit vs closed). I think closed is when streams have been released. So i think exit is correct in this case.

src/Misc/layoutbin/macOSRunInvoker.js
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
```
ericsciple, line 1: is this used?

src/Misc/layoutbin/macOSRunInvoker.js
```diff
@@ -0,0 +1,13 @@
...
+// Copyright (c) GitHub. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+const { spawn } = require('child_process');
+var shell = process.argv[2];
+var args = process.argv.slice(3);
+var launch = spawn(shell, args, { stdio: 'inherit' });
+launch.on('exit', function (code) {
+    if (code != 0) {
```
ericsciple, line 10: should this be `!==` ? i used to like `==` but linters have turned me :(

src/Runner.Worker/Handlers/ScriptHandler.cs
aibaars: Alright, I see you have `_proc.StartInfo.UseShellExecute = false` in `ProcessInvoker.cs`. 

In that case you probably still need to worry about `\` . Otherwise an argument containing `\"` (a backslash and a quote) might cause some trouble. After the `Replace` you'd end-up with `\\"`, which I think would be interpreted as an escaped backslash and an unescaped `"`.

Wouldn't it be easier to use `ProcessStartInfo.ArgumentList` instead of `ProcessStartInfo.Arguments` ? That looks more convenient.

src/Misc/layoutbin/macOSRunInvoker.js
TingluoHuang: removed

src/Misc/layoutbin/macOSRunInvoker.js
TingluoHuang: fixed.

src/Runner.Worker/Handlers/ScriptHandler.cs
TingluoHuang: fixed.

src/Runner.Worker/Handlers/ScriptHandler.cs
TingluoHuang: fixed to using ENV.

TingluoHuang: we don't want to be in a business of parse customer's arg string into arg array. :)

Tombow916 (commented):
I fix the malfunction so stop bullying people

PR #419, opened 2020-04-09
Revoke GITHUB_TOKEN at the end of the job.

TingluoHuang:
To prevent accidental exposure of GITHUB_TOKEN in the log.
We will revoke the token as soon as we finish the job on the runner.

### DIFF
--- a/src/Runner.Common/GitHubServer.cs
+++ b/src/Runner.Common/GitHubServer.cs
@@ -1,0 +1,67 @@
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Services.Common;
+
+namespace GitHub.Runner.Common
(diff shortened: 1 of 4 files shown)


src/Runner.Common/GitHubServer.cs
```diff
@@ -0,0 +1,67 @@
...
+                        {
+                            requestMessage.Headers.Add("Accept", "application/vnd.github.gambit-preview+json");
+                            var response = await httpClient.SendAsync(requestMessage, CancellationToken.None);
+                            result.StatusCode = response.StatusCode;
+                            result.Headers = response.Headers;
+                            result.Message = await response.Content.ReadAsStringAsync();
+                            return result;
+                        }
+                    }
+                    catch (Exception ex) when (count++ < 3)
```
ericsciple, line 56: if you get 401 should we assume success :)

TingluoHuang:
Abandon this, I am going to revoke the token from service side.

PR #427, opened 2020-04-15
Sample scripts to automate scaleable runners 

bryanmacfarlane:
👉 [overview here](https://github.com/actions/runner/blob/automate/docs/automate.md) 👈

### DIFF
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -2,9 +2,9 @@
 
 ## Latest Release as Service on Nix VMs
 
-[Run or copy this script for your use](../scripts/latest-svc.sh) to automate configuring a runner as a service on Linux or Mac
+[Run or copy this script for your use](../scripts/create-latest-svc.sh) to automate configuring a runner as a service on Linux or Mac
 
-### Step 1: export PAT
+### Export PAT
 
 Create a GitHub PAT and export it before running the script
 
@@ -12,15 +12,16 @@ Create a GitHub PAT and export it before running the script
 export RUNNER_CFG_PAT=yourPAT
 ```
 
-### Step 2: config one liner
+### Ceate service one liner
 
-Repo level one liner; replace yourorg/yourrepo
+Repo level one liner; replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
-curl https://raw.githubusercontent.com/actions/runner/automate/scripts/latest-svc.sh | bash -s yourorg/yourrepo
+curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
-Org level one liner; replace yourorg
+### Uninstall service one liner
 
+Repo level one liner; replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
-curl https://raw.githubusercontent.com/actions/runner/automate/scripts/latest-svc.sh | bash -s yourorg
+curl https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-svc.sh | bash -s yourorg/yourrepo
 ```
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -1,0 +1,26 @@
+# Automate Configuring Self-Hosted Runner Pools
+
+## Latest Release as Service on Nix VMs
+
+[Run or copy this script for your use](../scripts/latest-svc.sh) to automate configuring a runner as a service on Linux or Mac
+
+### Step 1: export PAT
+
+Create a GitHub PAT and export it before running the script
+
+```bash
+export RUNNER_CFG_PAT=yourPAT
+```
+
+### Step 2: config one liner
+
+Repo level one liner; replace yourorg/yourrepo
+```bash
+curl https://github.com/actions/runner/blob/automate/scripts/latest-svc.sh | bash -s yourorg/yourrepo
+```
+
+Org level one liner; replace yourorg
+
+```bash
+curl https://github.com/actions/runner/blob/automate/scripts/latest-svc.sh | bash -s yourorg
+```
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -1,6 +1,5 @@
-# Automate Configuring Self-Hosted Runner Pools
+# Automate Configuring Self-Hosted Runners
 
-## Latest Release as Service on Nix VMs
 
 ## Export PAT
 
@@ -10,7 +9,7 @@ Before running any of these sample scripts, create a GitHub PAT and export it be
 export RUNNER_CFG_PAT=yourPAT
 ```
 
-### Create running as a service
+## Create running as a service
 
 **Scenario**: Run on a machine or VM (not container) which automates:
 
@@ -20,14 +19,14 @@ export RUNNER_CFG_PAT=yourPAT
  - Configure the runner
  - Run as a systemd (linux) or Launchd (osx) service
 
-[Sample script here](../scripts/create-latest-svc.sh) 
+:point_right: [Sample script here](../scripts/create-latest-svc.sh) :point_left:
 
 Run as a one-line. NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
 curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
-### Uninstall running as service 
+## Uninstall running as service 
 
 **Scenario**: Run on a machine or VM (not container) which automates:
 
@@ -35,7 +34,7 @@ curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-la
  - Acquires a removal token
  - Removes the runner
 
-[Sample script here](../scripts/remove-svc.sh) 
+:point_right: [Sample script here](../scripts/remove-svc.sh) :point_left:
 
 Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
@@ -50,7 +49,7 @@ curl https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-sv
  - Resolves id from name
  - Deletes the runner
 
-[Sample script here](../scripts/delete.sh) 
+:point_right: [Sample script here](../scripts/delete.sh) :point_left:
 
 Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) and replace runnername
 ```bash
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -16,11 +16,11 @@ export RUNNER_CFG_PAT=yourPAT
 
 Repo level one liner; replace yourorg/yourrepo
 ```bash
-curl https://github.com/actions/runner/blob/automate/scripts/latest-svc.sh | bash -s yourorg/yourrepo
+curl https://raw.githubusercontent.com/actions/runner/automate/scripts/latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
 Org level one liner; replace yourorg
 
 ```bash
-curl https://github.com/actions/runner/blob/automate/scripts/latest-svc.sh | bash -s yourorg
+curl https://raw.githubusercontent.com/actions/runner/automate/scripts/latest-svc.sh | bash -s yourorg
 ```
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -23,7 +23,7 @@ export RUNNER_CFG_PAT=yourPAT
 
 Run as a one-line. NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
-curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
+curl -s https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
 ## Uninstall running as service 
@@ -38,7 +38,7 @@ curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-la
 
 Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
-curl https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-svc.sh | bash -s yourorg/yourrepo
+curl -s https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-svc.sh | bash -s yourorg/yourrepo
 ```
 
 ### Delete an offline runner
@@ -53,5 +53,5 @@ curl https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-sv
 
 Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) and replace runnername
 ```bash
-curl https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-svc.sh | bash -s yourorg/yourrepo runnername
+curl -s https://raw.githubusercontent.com/actions/runner/automate/scripts/delete.sh | bash -s yourorg/yourrepo runnername
 ```
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -2,26 +2,57 @@
 
 ## Latest Release as Service on Nix VMs
 
-[Run or copy this script for your use](../scripts/create-latest-svc.sh) to automate configuring a runner as a service on Linux or Mac
+## Export PAT
 
-### Export PAT
-
-Create a GitHub PAT and export it before running the script
+Before running any of these sample scripts, create a GitHub PAT and export it before running the script
 
 ```bash
 export RUNNER_CFG_PAT=yourPAT
 ```
 
-### Ceate service one liner
+### Create running as a service
 
-Repo level one liner; replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
+**Scenario**: Run on a machine or VM (not container) which automates:
+
+ - Resolving latest released runner
+ - Download and extract latest
+ - Acquire a registration token
+ - Configure the runner
+ - Run as a systemd (linux) or Launchd (osx) service
+
+[Sample script here](../scripts/create-latest-svc.sh) 
+
+Run as a one-line. NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
 ```bash
 curl https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
-### Uninstall service one liner
+### Uninstall running as service 
 
-Repo level one liner; replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
+**Scenario**: Run on a machine or VM (not container) which automates:
+
+ - Stops and uninstalls the systemd (linux) or Launchd (osx) service
+ - Acquires a removal token
+ - Removes the runner
+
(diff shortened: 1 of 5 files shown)


docs/automate.md
```diff
@@ -0,0 +1,26 @@
+# Automate Configuring Self-Hosted Runner Pools
+
+## Latest Release as Service on Nix VMs
+
+[Run or copy this script for your use](../scripts/latest-svc.sh) to automate configuring a runner as a service on Linux or Mac
+
+### Step 1: export PAT
+
+Create a GitHub PAT and export it before running the script
```
joshmgross, line 9: We might want to clarify the scopes required for the PAT so that users don't have to look at the API docs to know. I also don't think we need to specify a "GitHub PAT".

```suggestion
Create a PAT with `repo` scopes for repo runners and `admin:org` scopes for org runners. Export this PAT as an environment variable before running the script.
```

docs/automate.md
```diff
@@ -0,0 +1,26 @@
...
+
+```bash
+export RUNNER_CFG_PAT=yourPAT
+```
+
+### Step 2: config one liner
+
+Repo level one liner; replace yourorg/yourrepo
+```bash
+curl https://raw.githubusercontent.com/actions/runner/automate/scripts/latest-svc.sh | bash -s yourorg/yourrepo
```
joshmgross, line 19: Users might be reluctant to pipe a script straight to bash, should we include instructions for downloading it and then running it as well?

scripts/latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+
+set -e
+
+#
+# Downloads latest releases (not pre-release) runner
+# Configures as a service
+#
+# Examples:
+# RUNNER_CFG_PAT=<yourPAT> ./nix-svc.sh myuser/myrepo
+# RUNNER_CFG_PAT=<yourPAT> ./nix-svc.sh myorg
```
joshmgross:
```suggestion
# RUNNER_CFG_PAT=<yourPAT> ./latest-svc.sh myuser/myrepo
# RUNNER_CFG_PAT=<yourPAT> ./latest-svc.sh myorg
```

joshmgross: This was fixed as I commented

docs/automate.md
bryanmacfarlane: Done.  Added a link to sample script

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+# 
+# Notes:
+# PATS over envvars are more secure
+# Should be used on VMs and not containers
+# Works on OSX and Linux 
+# Assumes x64 arch
+#
+
+runner_scope=${1}
+runner_name=${2:-$(hostname)}
```
joshmgross, line 29: If I'm spinning up a lot of VMs to run this script and create runners, are we guaranteed to have a unique hostname for each one?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+#---------------------------------------------------
+# extract to runner directory in this directory
+#---------------------------------------------------
+echo
+echo "Extracting ${runner_file} to ./runner"
+
+tar xzf "./${runner_file}" -C runner
+
+# export of pass
+sudo chown -R $svc_user ./runner
```
joshmgross, line 111: Does this assume passwordless sudo? Should we have an option for a user to pass in a password?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+
+pushd ./runner
+
+#---------------------------------------
+# Unattend config
+#---------------------------------------
+runner_url="https://github.com/${runner_scope}"
+echo
+echo "Configuring ${runner_name} @ $runner_url"
+echo "./config.sh --unattended --url $runner_url --token $RUNNER_TOKEN --name $runner_name"
```
joshmgross, line 121: `$RUNNER_TOKEN` is a secret value (works for up to an hour). Do we need to echo it out?

scripts/create-latest-svc.sh
bryanmacfarlane: That's why we take it as arg2.  But the vast majority of the time, hostname being the default is correct.

scripts/create-latest-svc.sh
bryanmacfarlane: That's on the user.   Most init script and VMs should run with passwordless sudo, if not, they will get prompted per design.  I think elevating needs to be explicit.

scripts/create-latest-svc.sh
joshmgross: I missed that, was looking at the Examples in this script file and didn't look at usage.

Should we have flags for those inputs? I.e. what if I want to specify User but not Name

Also thinking ahead to labels inputs, relying on the order of the arguments might not hold up

scripts/create-latest-svc.sh
bryanmacfarlane: I think in the script case it runs in seconds and docs cover that so I'm fine not.

scripts/create-latest-svc.sh
bryanmacfarlane: It's simple and customary for bash scripting.  I thought about the user issue - it's a real edge and in that case you put $(hostname) in front of it.  If we did flags I think the sample script would get bogged down with a bunch of parsing logic and folks won't be able to see the forest through the weeds of the script.  That is, unless someone knows of a trivial (works everywhere) cmd line parsing bash technique ...

docs/automate.md
bryanmacfarlane: Cool.  Is there somewhere in docs that clarifies the minimum scopes?  Ideally, I would just link to docs.  But agreed, good addition.

docs/automate.md
```diff
@@ -0,0 +1,57 @@
...
+
+**Scenario**: Run on a machine or VM (not container) which automates:
+
+ - Stops and uninstalls the systemd (linux) or Launchd (osx) service
+ - Acquires a removal token
+ - Removes the runner
+
+:point_right: [Sample script here](../scripts/remove-svc.sh) :point_left:
+
+Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level)
```
TingluoHuang, line 39: `Run as a one-line` instead of `Repo level one liner.` to match config ☝️ ☝️ ?

docs/automate.md
```diff
@@ -0,0 +1,57 @@
...
+ - Removes the runner
+
+:point_right: [Sample script here](../scripts/remove-svc.sh) :point_left:
+
+Repo level one liner.  NOTE: replace with yourorg/yourrepo (repo level) or just yourorg (org level) 
+```bash
+curl -s https://raw.githubusercontent.com/actions/runner/automate/scripts/remove-svc.sh | bash -s yourorg/yourrepo
+```
+
+### Delete an offline runner
```
TingluoHuang, line 44: do we need this? we are going to clean up dormant runners if the runner offline for 30 days.

joshmgross: Users might not want that offline runner showing up in the settings UI for 30 days

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+# TODO: validate systemd or osx svc installer
+
+#--------------------------------------
+# Get a config token
+#--------------------------------------
+echo
+echo "Generating a registration token..."
+
+# if the scope has a slash, it's an repo runner
+base_api_url="https://api.github.com/orgs"
```
TingluoHuang, line 70: do we want to support GHES?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+
+#---------------------------------------
+# Download latest released and extract
+#---------------------------------------
+echo
+echo "Downloading latest runner ..."
+
+latest_version_label=$(curl -s -X GET 'https://api.github.com/repos/actions/runner/releases/latest' | jq -r '.tag_name')
+latest_version=$(echo ${latest_version_label:1})
+runner_file="actions-runner-${runner_plat}-x64-${latest_version}.tar.gz"
```
TingluoHuang, line 87: what about ARM and ARM64? 😄

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+# Works on OSX and Linux 
+# Assumes x64 arch
+#
+
+runner_scope=${1}
+runner_name=${2:-$(hostname)}
+svc_user=${3:-$USER}
+
+echo "Configuring runner @ ${runner_scope}"
+sudo echo
```
TingluoHuang, line 33: why do we need to `sudo echo`, to avoid prompt downstream?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+pushd ./runner
+
+#---------------------------------------
+# Unattend config
+#---------------------------------------
+runner_url="https://github.com/${runner_scope}"
+echo
+echo "Configuring ${runner_name} @ $runner_url"
+echo "./config.sh --unattended --url $runner_url --token $RUNNER_TOKEN --name $runner_name"
+sudo -E -u ${svc_user} ./config.sh --unattended --url $runner_url --token $RUNNER_TOKEN --name $runner_name
```
TingluoHuang, line 122: do we need `--replace`?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,130 @@
...
+
+#---------------------------------------
+# Configuring as a service
+#---------------------------------------
+echo
+echo "Configuring as a service ..."
+./svc.sh install ${svc_user}
```
TingluoHuang: do we need to `sudo` for Linux?

scripts/remove-svc.sh
```diff
@@ -0,0 +1,67 @@
...
+
+echo $REMOVE_TOKEN
+
+#---------------------------------------
+# Stop and uninstall the service
+#---------------------------------------
+echo
+echo "Uninstall the service ..."
+pushd ./runner
+./svc.sh stop
```
TingluoHuang: do we need `sudo`?

bryanmacfarlane:
@TingluoHuang - Josh looked and approved.  Can you have a look?  Or suggest someone else?  I want to make sure these are ready to go with org level runner release.

bryanmacfarlane:
I verified on linux (Debian flavors) and osx with repo and org level ...

docs/automate.md
TingluoHuang: my question is more like when should we expect the customer to use this?
I can imagine in an autoscale VM/Kubernetes scenario, the customer will use the config script to set up a runner when creating VM/container and use the un-config script to un-config runner before we delete the VM/container.
If the VM/container crash or shutdown unexpected, we will use the delete offline script for cleanup. 

@bryanmacfarlane is my understand right?

scripts/create-latest-svc.sh
```diff
@@ -0,0 +1,135 @@
...
+echo
+echo "Generating a registration token..."
+
+# if the scope has a slash, it's an repo runner
+base_api_url="https://api.github.com/orgs"
+if [[ "$runner_scope" == *\/* ]]; then
+    base_api_url="https://api.github.com/repos"
+fi
+
+export RUNNER_TOKEN=$(curl -s -X POST ${base_api_url}/${runner_scope}/actions/runners/registration-token -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" | jq -r '.token')
```
TingluoHuang, line 75: in the error case, like my PAT doesn't have enough permission, will the customer have enough info to debug?

scripts/create-latest-svc.sh
TingluoHuang: RUNNER_TOKEN is the pool admin token, it has permission to remove other runners.
I don't think we should print it in any log.

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
...
+# Examples:
+# RUNNER_CFG_PAT=<yourPAT> ./delete.sh myuser/myrepo
+# RUNNER_CFG_PAT=<yourPAT> ./delete.sh myorg
+#
+# Usage:
+#     export RUNNER_CFG_PAT=<yourPAT>
+#     ./delete.sh scope name
+#
+#      scope required  repo (:owner/:repo) or org (:organization)
+#      name  optional  defaults to hostname.  name to delete
```
TingluoHuang, line 18: name is required too, i think.

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
+#/bin/bash
+
+set -e
+
+#
+# Force deletes a runner from the service
+# The caller should have already ensured the runner is gone and/or stopped
+#
+# Examples:
+# RUNNER_CFG_PAT=<yourPAT> ./delete.sh myuser/myrepo
```
TingluoHuang, line 10: the example doesn't contain runner_name input

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
...
+base_api_url="https://api.github.com/orgs"
+if [[ "$runner_scope" == *\/* ]]; then
+    base_api_url="https://api.github.com/repos"
+fi
+
+
+#--------------------------------------
+# Ensure offline
+#--------------------------------------
+runner_status=$(curl -s -X GET ${base_api_url}/${runner_scope}/actions/runners  -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" \
```
TingluoHuang, line 53: what about paging?
can we query `actions/runners/runner1`?

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
...
+        | jq -M -j ".runners | .[] | [select(.name == \"${runner_name}\")] | .[0].status")
+
+if [ -z "${runner_status}" ]; then 
+    fatal "Could not find runner with name ${runner_name}"
```
TingluoHuang, line 57: do we have to fetal? should we just return since the runner has gone?

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
...
+echo "Status: ${runner_status}"
+
+if [ "${runner_status}" != "offline" ]; then 
+    fatal "Runner should be offline before removing"
+fi
+
+#--------------------------------------
+# Get id of runner to remove
+#--------------------------------------
+runner_id=$(curl -s -X GET ${base_api_url}/${runner_scope}/actions/runners  -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" \
```
TingluoHuang, line 69: any way to get runner_id and runner_status in one curl request?

scripts/delete.sh
```diff
@@ -0,0 +1,83 @@
...
+if [ -z "${runner_id}" ]; then 
+    fatal "Could not find runner with name ${runner_name}"
+fi 
+
+echo "Removing id ${runner_id}"
+
+#--------------------------------------
+# Remove the runner
+#--------------------------------------
+curl -s -X DELETE ${base_api_url}/${runner_scope}/actions/runners/${runner_id} -H "authorization: token ${RUNNER_CFG_PAT}"
```
TingluoHuang, line 81: do we need any preview HTTP header like `application/vnd.github.everest-preview+json`?

scripts/remove-svc.sh
```diff
@@ -0,0 +1,76 @@
...
+echo
+echo "Uninstall the service ..."
+pushd ./runner
+prefix=""
+if [ "${runner_plat}" == "linux" ]; then 
+    prefix="sudo "
+fi 
+${prefix}./svc.sh stop
+${prefix}./svc.sh uninstall
+./config.sh remove --token $REMOVE_TOKEN
```
TingluoHuang, line 76: should we `sudo ./config.sh remove` like what we did during config?
if the runner folder belongs to a different svc user, config.sh may fail without sudo because the current user doesn't have permission.

scripts/create-latest-svc.sh
bryanmacfarlane: fixing.  accidentally left in when I was debugging a problem.

docs/automate.md
bryanmacfarlane: I fixed them to all be consistent.  `one-liner`

docs/automate.md
bryanmacfarlane: It's used by the same people that need the delete api (as opposed to the cli to remove).  This is used for outside/in automation where the runner goes away.  The customers I've interacted with so far for org level runners were using it for container / K8s runners where there's just an entry point.  The container is stopped outside in and then they to delete it after it goes away.

scripts/create-latest-svc.sh
bryanmacfarlane: Maybe later.  These are also sample scripts so they can copy and alter per instructions.  I want to walk a fine line between the samples being useful / illustrative but not too complex where it addresses all scenarios.

scripts/create-latest-svc.sh
bryanmacfarlane: Later.  We also need to do windows.  This first set of sample scripts illustrate linux and osx.  The sample should help others getting started as well.  See my point above

scripts/create-latest-svc.sh
bryanmacfarlane: Yeah, I was trying to prompt early so they don't kick it off, look away only to find a prompt later.

scripts/create-latest-svc.sh
bryanmacfarlane: Yeah.  Fixed.

scripts/remove-svc.sh
bryanmacfarlane: Yeah,  Fixed.  only on Linux.  (osx does a user level launchd)

scripts/delete.sh
bryanmacfarlane: Fixed.

scripts/create-latest-svc.sh
bryanmacfarlane: I decided not too as it's dangerous.  Seems like an interactive choice.  If the script is explicit and it already exists, then they can fix it before running again or fix their outside / in automation.

scripts/create-latest-svc.sh
bryanmacfarlane: It returns what the API returns since it's a simple script.  I think it's fine unless you have something more elegant in mind.  I don't think we can tell why it's access denied - it just is.

scripts/delete.sh
bryanmacfarlane: fixed

scripts/delete.sh
bryanmacfarlane: made the simple sample script work with 100.  Added note and TODO that more complex cases can loop on pages.

scripts/delete.sh
bryanmacfarlane: I added fatal.  It prints an error and returns exiting the script with 1

scripts/delete.sh
bryanmacfarlane: I could potentially refactor to save off contents of get and then apply different jq filters independently however I think this is more readable as a simple script.

scripts/remove-svc.sh
bryanmacfarlane: fixed

bryanmacfarlane:
@TingluoHuang I addressed most feedback and left some comments and TODO for further work later.  I think the scripts as is help get folks going (I know it would have helped with a few customers I interacted with directly this week).

scripts/delete.sh
bryanmacfarlane: I think we're supposed to doing preview periods.  https://developer.github.com/changes/2/#--creating-a-repository-dispatch-event-now-official-part-of-github-api

TingluoHuang: the reason I asked is that you have the preview header in the get runner endpoint and get registration token endpoint call.

Issue #428, opened 2020-03-31
verify configuration directive names


vladak:
I happened to setup the `checkout@v2` like so:
```yml
    - name: Checkout gh-pages branch                                            
      uses: actions/checkout@v2                                                 
      with:                                                                     
        path: gh-pages                                                          
        refs: gh-pages
```
however instead of complaining that I used `refs` instead of `ref` it happily checkout out the `master` branch.

ericsciple:
Sorry for the delay, i guess i forgot to hit transfer before.

TingluoHuang:
@ericsciple I guess we should at latest check inputs from action.yml and inputs from the workflow, then raise a warning when they do not match.

PR #429, opened 2020-04-15
Raise warning when action input does not match action.yml.

TingluoHuang:
Fix: #428


src/Runner.Worker/ActionRunner.cs
```diff
@@ -143,29 +143,41 @@ public async Task RunAsync()
...
                         inputs[key] = manifestManager.EvaluateDefaultInput(ExecutionContext, key, input.Value);
                     }
                 }
             }
 
+            foreach (var input in userInputs)
+            {
+                if (!validInputs.Contains(input))
+                {
+                    ExecutionContext.Warning($"Unrecognized input '{input}', valid inputs are ['{string.Join("', '", validInputs)}']");
```
ericsciple, line 177: nit: consider, "Unexpected input ..."

ericsciple: I wonder how hard it would be to plumb the line column file info through.

ericsciple: Should we drive this change into manifestManager.ValidateInputs or something? And it could output a list of warnings?

PR #433, opened 2020-04-17
Print node version in debug instead of output.

TingluoHuang:
The customer doesn't care what node version the infrastructure is using, so we should not put this info into every user-facing log.


bryanmacfarlane:
Yes!

PR #437, opened 2020-04-20
Use the API_URL and munge action URLs for GHES

pjquirk:
## Background
The runner downloads actions by creating a URL based on the name and ref specified in the workflow.  For example, for the step `uses: foo/bar@v1` the runner will attempt to download `https://api.github.com/repos/foo/bar/zipball/v1`.  

This has two issues related to GHES:
1. Runners connected to a GHES instance should retrieve all of their actions from GHES and not dotcom.
1. Since we're planning to allow the user to import certain community actions into GHES, we need to munge the URL to point to these imported actions.

## Solution

The API root URL is currently stored in the context as `api_url`, so the runner can use that to point to the GHES instance.  

Munging the URL is simple, but the retry logic needed to change a bit since `uses: foo/bar@v1` could be referencing an org/repo in GHES or an imported action.  In this PR, I changed the action download logic to consider if the runner is connected to GHES or dotcom.  If GHES, first look for the action by name as usual (from the GHES instance).  If that 404s, then munge the URL to look for an imported action.  In both GHES and hosted cases, 404s are detected and no retry is performed.

TODO: Testing with an actual GHES instance

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -614,9 +614,7 @@ namespace GitHub.Runner.Common
         public static HttpClientHandler CreateHttpClientHandler(this IHostContext context)
         {
             var handlerFactory = context.GetService<IHttpClientHandlerFactory>();
-            var clientHandler = handlerFactory.CreateClientHandler();
-            clientHandler.Proxy = context.WebProxy;
-            return clientHandler;
+            return handlerFactory.CreateClientHandler(context.WebProxy);
         }
     }
 
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -1,19 +1,18 @@
-﻿using GitHub.Runner.Common.Util;
-using System;
+﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Reflection;
 using System.Runtime.Loader;
 using System.Threading;
 using System.Threading.Tasks;
-using System.Diagnostics;
-using System.Net.Http;
-using System.Diagnostics.Tracing;
 using GitHub.DistributedTask.Logging;
-using System.Net.Http.Headers;
 using GitHub.Runner.Sdk;
 
 namespace GitHub.Runner.Common
@@ -614,7 +613,8 @@ namespace GitHub.Runner.Common
     {
         public static HttpClientHandler CreateHttpClientHandler(this IHostContext context)
         {
-            HttpClientHandler clientHandler = new HttpClientHandler();
+            var handlerFactory = context.GetService<IHttpClientHandlerFactory>();
+            var clientHandler = handlerFactory.CreateClientHandler();
             clientHandler.Proxy = context.WebProxy;
             return clientHandler;
         }
--- a/src/Runner.Common/HttpClientHandlerFactory.cs
+++ b/src/Runner.Common/HttpClientHandlerFactory.cs
@@ -1,0 +1,22 @@
+using System.Net.Http;
+
+namespace GitHub.Runner.Common
+{
+    [ServiceLocator(Default = typeof(HttpClientHandlerFactory))]
+    public interface IHttpClientHandlerFactory : IRunnerService
+    {
+        HttpClientHandler CreateClientHandler();
+    }
+
+    public class HttpClientHandlerFactory : IHttpClientHandlerFactory
+    {
+        public HttpClientHandler CreateClientHandler()
+        {
+            return new HttpClientHandler();
+        }
+
+        public void Initialize(IHostContext context)
+        {
+        }
+    }
+}
--- a/src/Runner.Common/HttpClientHandlerFactory.cs
+++ b/src/Runner.Common/HttpClientHandlerFactory.cs
@@ -1,18 +1,19 @@
 using System.Net.Http;
+using GitHub.Runner.Sdk;
 
 namespace GitHub.Runner.Common
 {
     [ServiceLocator(Default = typeof(HttpClientHandlerFactory))]
     public interface IHttpClientHandlerFactory : IRunnerService
     {
-        HttpClientHandler CreateClientHandler();
+        HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy);
     }
 
     public class HttpClientHandlerFactory : IHttpClientHandlerFactory
     {
-        public HttpClientHandler CreateClientHandler()
+        public HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy)
         {
-            return new HttpClientHandler();
+            return new HttpClientHandler() { Proxy = webProxy };
         }
 
         public void Initialize(IHostContext context)
--- a/src/Runner.Common/HttpClientHandlerFactory.cs
+++ b/src/Runner.Common/HttpClientHandlerFactory.cs
@@ -9,15 +9,11 @@ namespace GitHub.Runner.Common
         HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy);
     }
 
-    public class HttpClientHandlerFactory : IHttpClientHandlerFactory
+    public class HttpClientHandlerFactory : RunnerService, IHttpClientHandlerFactory
     {
         public HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy)
         {
             return new HttpClientHandler() { Proxy = webProxy };
-        }
-
-        public void Initialize(IHostContext context)
-        {
         }
     }
 }
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -504,6 +504,7 @@ namespace GitHub.Runner.Worker
                 string archiveLink = BuildLinkToActionArchive(apiUrl, repositoryReference.Name, repositoryReference.Ref);
                 Trace.Info($"Download archive '{archiveLink}' to '{destDirectory}'.");
                 await DownloadRepositoryActionAsync(executionContext, archiveLink, destDirectory);
+                return;
             }
             else
             {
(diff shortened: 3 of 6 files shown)


src/Runner.Common/HostContext.cs
```diff
@@ -614,9 +613,8 @@ public static class HostContextExtension
     {
         public static HttpClientHandler CreateHttpClientHandler(this IHostContext context)
         {
-            HttpClientHandler clientHandler = new HttpClientHandler();
-            clientHandler.Proxy = context.WebProxy;
-            return clientHandler;
+            var handlerFactory = context.GetService<IHttpClientHandlerFactory>();
```
ericsciple, line 616: changed for unit testing

src/Runner.Common/HttpClientHandlerFactory.cs
```diff
@@ -0,0 +1,23 @@
+using System.Net.Http;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Common
```
ericsciple, line 4: for unit testing

src/Runner.Worker/ActionManager.cs
```diff
@@ -504,27 +502,83 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                    {
+                        Trace.Info($"Failed to find the action at {ex.ActionUri}");
+                        continue;
+                    }
+                }
+                throw new NotSupportedException($"Failed to find the action '{repositoryReference.Name}'.  Paths attempted: {string.Join(", ", archiveLinks)}");
+            }
+        }
+
+        private string GetApiUrl(IExecutionContext executionContext, bool forceDotCom = false)
```
ericsciple, line 544: Eventually this will be set for hosted too. I would just fallback to `https://api.github.com` for back compat.

src/Runner.Worker/ActionManager.cs
```diff
@@ -504,27 +502,83 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                await DownloadRepositoryActionAsync(executionContext, archiveLink, destDirectory);
+            }
+            else
+            {
+                string apiUrl = GetApiUrl(executionContext);
+
+                // URLs to try:
+                var archiveLinks = new List<string> {
+                    // API_URL/repos/{repository}/zipball/{ref}  - A built-in action or an action the user has created, on their GHES instance
+                    BuildLinkToActionArchive(apiUrl, repositoryReference.Name, repositoryReference.Ref),
```
ericsciple, line 521: For example: https://my-ghes/api/v3/repos/actions/checkout/tarball/v2

src/Runner.Worker/ActionManager.cs
```diff
@@ -504,27 +502,83 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                    // API_URL/repos/actions/community-{repository.replace(/,_)}/zipball/{ref}  - A community action, synced to their GHES instance
+                    BuildLinkToActionArchive(apiUrl, $"actions/community-{repositoryReference.Name.Replace("/", "_")}", repositoryReference.Ref)
```
ericsciple, line 523: For example: https://my-ghes/api/v3/repos/community-actions/some-org_some-action/tarball/v1

src/Runner.Worker/ActionManager.cs
```diff
@@ -504,27 +502,83 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                        await DownloadRepositoryActionAsync(executionContext, archiveLink, destDirectory);
+                        return;
+                    }
+                    catch (ActionNotFoundException ex)
+                    {
+                        Trace.Info($"Failed to find the action at {ex.ActionUri}");
+                        continue;
+                    }
+                }
+                throw new NotSupportedException($"Failed to find the action '{repositoryReference.Name}'.  Paths attempted: {string.Join(", ", archiveLinks)}");
```
ericsciple, line 540: Action not found might be better

src/Runner.Common/HostContext.cs
```diff
@@ -1,19 +1,18 @@
-﻿using GitHub.Runner.Common.Util;
-using System;
+﻿using System;
```
ericsciple, line 1: You need to delete where the actions are cached for GHES.... let me find...

ericsciple: Check w/ Ting. Make sure this wont interfere with alpha (e.g. already branched)

ericsciple:
```
            // Clear the cache (for self-hosted runners)
            // Note, temporarily avoid this step for the on-premises product, to avoid rate limiting.
            var isHostedServer = configurationStore.GetSettings().IsHostedServer;
            if (isHostedServer)
            {
                IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
            }
```

ericsciple:
Thanks! This is great! Please make sure @TingluoHuang is OK with the changes too, before merging.

src/Runner.Common/HttpClientHandlerFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    [ServiceLocator(Default = typeof(HttpClientHandlerFactory))]
+    public interface IHttpClientHandlerFactory : IRunnerService
+    {
+        HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy);
+    }
+
+    public class HttpClientHandlerFactory : IHttpClientHandlerFactory
```
TingluoHuang, line 12: `public HttpClientHandlerFactory : RunnerService, IHttpClientHandlerFactory`

check other services for example.

src/Runner.Common/HttpClientHandlerFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public HttpClientHandler CreateClientHandler(RunnerWebProxy webProxy)
+        {
+            return new HttpClientHandler() { Proxy = webProxy };
+        }
+
+        public void Initialize(IHostContext context)
```
TingluoHuang, line 19: move `Initialize()` up

src/Runner.Worker/ActionManager.cs
TingluoHuang: +1, remove the `forceDotCom`, just failback.

src/Runner.Worker/ActionManager.cs
```diff
@@ -504,27 +502,83 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
             }
 
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
+            var isHostedServer = configurationStore.GetSettings().IsHostedServer;
+            if (isHostedServer)
+            {
+                string apiUrl = GetApiUrl(executionContext, forceDotCom: true);
+                string archiveLink = BuildLinkToActionArchive(apiUrl, repositoryReference.Name, repositoryReference.Ref);
+                Trace.Info($"Download archive '{archiveLink}' to '{destDirectory}'.");
+                await DownloadRepositoryActionAsync(executionContext, archiveLink, destDirectory);
```
TingluoHuang, line 512: `return;` after the `await DownloadRepositoryActionAsync()`

src/Runner.Worker/ActionManager.cs
```diff
@@ -571,34 +625,55 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                using (var response = await httpClient.GetAsync(link))
                                 {
-                                    await result.CopyToAsync(fs, _defaultCopyBufferSize, actionDownloadCancellation.Token);
-                                    await fs.FlushAsync(actionDownloadCancellation.Token);
-
-                                    // download succeed, break out the retry loop.
-                                    break;
+                                    if (response.IsSuccessStatusCode)
+                                    {
+                                        using (var result = await httpClient.GetStreamAsync(link))
```
TingluoHuang, line 632: why we need to GetAsync() and GetStreamAsync()? looks like we are making the same request twice?

src/Runner.Worker/ActionManager.cs
pjquirk: Great catch

src/Runner.Worker/ActionManager.cs
pjquirk: Yep, just fixed that, it should be `response.Content.ReadAsStreamAsync()`

src/Runner.Worker/ActionManager.cs
```diff
@@ -571,34 +625,55 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                        {
+                                            await result.CopyToAsync(fs, _defaultCopyBufferSize, actionDownloadCancellation.Token);
+                                            await fs.FlushAsync(actionDownloadCancellation.Token);
+
+                                            // download succeed, break out the retry loop.
+                                            break;
+                                        }
+                                    }
+                                    else if (response.StatusCode == HttpStatusCode.NotFound)
```
TingluoHuang, line 641: are we sure 404 means the repo/ref doesn't exist? will it happen for a bad cache on dotcom? ex, we were hitting 404 when query commit

pjquirk: You mean if the user put the wrong ref in, would we get a 404, and is it appropriate to say the action doesn't exist? I think so in that case, though we could add the ref to the error message to make that clearer.  

404 was the best way I could think of to detect this.  We don't want to retry in this case, since that would slow down every run where the action actually is in the fallback URL. Open to other ideas though.

wdelmail001:
> ## Background
> The runner downloads actions by creating a URL based on the name and ref specified in the workflow. For example, for the step `uses: foo/bar@v1` the runner will attempt to download `https://api.github.com/repos/foo/bar/zipball/v1`.
> 
> This has two issues related to GHES:
> 
> 1. Runners connected to a GHES instance should retrieve all of their actions from GHES and not dotcom.
> 2. Since we're planning to allow the user to import certain community actions into GHES, we need to munge the URL to point to these imported actions.
> 
> ## Solution
> The API root URL is currently stored in the context as `api_url`, so the runner can use that to point to the GHES instance.
> 
> Munging the URL is simple, but the retry logic needed to change a bit since `uses: foo/bar@v1` could be referencing an org/repo in GHES or an imported action. In this PR, I changed the action download logic to consider if the runner is connected to GHES or dotcom. If GHES, first look for the action by name as usual (from the GHES instance). If that 404s, then munge the URL to look for an imported action. In both GHES and hosted cases, 404s are detected and no retry is performed.
> 
> TODO: Testing with an actual GHES instance

PR #440, opened 2020-04-21
add secret masker for trimming double qoutes.

TingluoHuang:
https://github.com/github/c2c-actions-runtime/issues/451

### DIFF
--- a/src/Sdk/DTLogging/Logging/ValueEncoders.cs
+++ b/src/Sdk/DTLogging/Logging/ValueEncoders.cs
@@ -60,6 +60,20 @@ namespace GitHub.DistributedTask.Logging
             return SecurityElement.Escape(value);
         }
 
+        public static String TrimDoubleQuotes(String value)
+        {
+            var trimmed = string.Empty;
+            if (!string.IsNullOrEmpty(value) &&
+                value.Length > 8 &&
+                value.StartsWith('"') &&
+                value.EndsWith('"'))
+            {
+                trimmed = value.Substring(1, value.Length - 2);
+            }
+
+            return trimmed;
+        }
+
         private static string Base64StringEscapeShift(String value, int shift)
         {
             var bytes = Encoding.UTF8.GetBytes(value);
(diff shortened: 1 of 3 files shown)


src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -60,6 +60,20 @@ public static String XmlDataEscape(String value)
             return SecurityElement.Escape(value);
         }
 
+        public static String TrimDoubelQuoteDataEscape(String value)
+        {
+            var trimmed = string.Empty;
+            if (!string.IsNullOrEmpty(value) &&
+                value.StartsWith('"') &&
+                value.EndsWith('"') &&
+                value.Length > 2)
```
TingluoHuang, line 69: do we need min length for the string?

ericsciple: I would. If it's less than N then it's not a very good secret... maybe 5 or 6?

ericsciple: So 5 or 6, plus 2 more for the surrounding quotes.

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -60,6 +60,20 @@ public static String XmlDataEscape(String value)
             return SecurityElement.Escape(value);
         }
 
+        public static String TrimDoubelQuoteDataEscape(String value)
```
ericsciple, line 63: How about `TrimQuotes` or `TrimDoubleQuotes`?

"Data" is terminology specific to URI, to distinguish between a segment within the URI vs the whole thing.

ericsciple: And "Escape" terminology doesnt apply here either

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -60,6 +60,20 @@ public static String XmlDataEscape(String value)
...
+            var trimmed = string.Empty;
+            if (!string.IsNullOrEmpty(value) &&
+                value.StartsWith('"') &&
+                value.EndsWith('"') &&
+                value.Length > 2)
+            {
+                trimmed = value.Substring(1, value.Length - 2);
+            }
+
+            return string.IsNullOrEmpty(trimmed) ? string.Empty : trimmed;
```
ericsciple, line 74: why not just return `trimmed` :)

src/Sdk/DTLogging/Logging/ValueEncoders.cs
TingluoHuang: 👍

PR #450, opened 2020-04-23
Help trace worker crash in Kusto.

TingluoHuang:
There was a bug in the Actions hosted runner virtual environment that causes the available disk space reduced. http://github.com/actions/virtual-environments/issues/709

Some customer's workflow used all disk space and cause the runner crash as well.

We want to get alert when the number of worker crash increase, so we can proactively investigate these issue instead of waiting for customer report.

Each worker crash means there is a failed workflow run and the customer is not happy with it.

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -952,7 +952,7 @@ namespace GitHub.Runner.Listener
                 TimelineRecord jobRecord = timeline.Records.FirstOrDefault(x => x.Id == message.JobId && x.RecordType == "Job");
                 ArgUtil.NotNull(jobRecord, nameof(jobRecord));
                 var unhandledExceptionIssue = new Issue() { Type = IssueType.Error, Message = errorMessage };
-                unhandledExceptionIssue.Data[Constants.Runner.WorkerCrashIssueDataKey] = string.Empty;
+                unhandledExceptionIssue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.WorkerCrash;
                 jobRecord.ErrorCount++;
                 jobRecord.Issues.Add(unhandledExceptionIssue);
                 await jobServer.UpdateTimelineRecordsAsync(message.Plan.ScopeIdentifier, message.Plan.PlanType, message.Plan.PlanId, message.Timeline.Id, new TimelineRecord[] { jobRecord }, CancellationToken.None);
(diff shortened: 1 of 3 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -952,8 +951,10 @@ private async Task LogWorkerProcessUnhandledException(Pipelines.AgentJobRequestM
                 ArgUtil.NotNull(timeline, nameof(timeline));
                 TimelineRecord jobRecord = timeline.Records.FirstOrDefault(x => x.Id == message.JobId && x.RecordType == "Job");
                 ArgUtil.NotNull(jobRecord, nameof(jobRecord));
+                var unhandledExceptionIssue = new Issue() { Type = IssueType.Error, Message = errorMessage };
+                unhandledExceptionIssue.Data["_worker_crash_unhandled_exception"] = string.Empty;
```
TingluoHuang, line 955: service will check the issue data to decide whether fire product trace event.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -952,8 +951,10 @@ private async Task LogWorkerProcessUnhandledException(Pipelines.AgentJobRequestM
                 ArgUtil.NotNull(timeline, nameof(timeline));
                 TimelineRecord jobRecord = timeline.Records.FirstOrDefault(x => x.Id == message.JobId && x.RecordType == "Job");
                 ArgUtil.NotNull(jobRecord, nameof(jobRecord));
+                var unhandledExceptionIssue = new Issue() { Type = IssueType.Error, Message = errorMessage };
+                unhandledExceptionIssue.Data[Constants.Runner.WorkerCrashIssueDataKey] = string.Empty;
```
TingluoHuang, line 955: service will check the issue data to decide whether fire product trace event.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -486,7 +486,10 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
 
             foreach (var property in command.Properties)
             {
-                issue.Data[property.Key] = property.Value;
+                if (!string.Equals(property.Key, Constants.Runner.WorkerCrashIssueDataKey, StringComparison.OrdinalIgnoreCase))
```
ericsciple, line 489: dont we already filter everything other than specific properties?

src/Runner.Listener/JobDispatcher.cs
ericsciple: Think about:

Should we make this more generic? Do we have any other cases that warrant a more generic telemetry mechanism, e.g. `[...].Data["_telemetry"] = WORKER_CRASH;<MESSAGE>`

^ That way we can start adding TYPE;MESSAGE more generically to issues?

PR #451, opened 2020-04-24
post-alpha fixes for github.url github.api_url and github.graphql_url

ericsciple:
Post-alpha updates for handling github.url, github.api_url, and github.graphql_url:
- Propagate github.graphql_url to env vars
- Temporary back compat, check whether github.url is defined before adding value when not hosted

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -131,7 +131,7 @@ namespace GitHub.Runner.Worker
                     // Temporary hack for GHES alpha
                     var configurationStore = HostContext.GetService<IConfigurationStore>();
                     var runnerSettings = configurationStore.GetSettings();
-                    if (!runnerSettings.IsHostedServer && !string.IsNullOrEmpty(runnerSettings.GitHubUrl))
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -131,7 +131,7 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     // Temporary hack for GHES alpha
                     var configurationStore = HostContext.GetService<IConfigurationStore>();
                     var runnerSettings = configurationStore.GetSettings();
-                    if (!runnerSettings.IsHostedServer && !string.IsNullOrEmpty(runnerSettings.GitHubUrl))
+                    if (string.IsNullOrEmpty(context.GetGitHubContext("url")) && !runnerSettings.IsHostedServer && !string.IsNullOrEmpty(runnerSettings.GitHubUrl))
                     {
                         var url = new Uri(runnerSettings.GitHubUrl);
                         var portInfo = url.IsDefaultPort ? string.Empty : $":{url.Port.ToString(CultureInfo.InvariantCulture)}";
```
TingluoHuang, line 137: do we need?
```
context.SetGitHubContext("graphql_url", xxx);	                        
```

ericsciple: nah this is just alpha back compat. After the real URLs are sent in the payload we can delete this code.

PR #452, opened 2020-04-24
Adapt create-latest-svc.sh to work with GHES

brcrista:
With Actions on GHES under development, we'd like to be able to automate setting up runners while testing.

I've adapted this script to take an optional `ghe_domain` parameter.  If this is set, the script will use the given domain name for the GHES deployment.

**Testing**
* Configured a runner on a github.com repo
* Configured a runner on a GHES repo

### DIFF
--- a/scripts/create-latest-svc.sh
+++ b/scripts/create-latest-svc.sh
@@ -68,9 +68,9 @@ sudo -u ${svc_user} mkdir runner
 echo
 echo "Generating a registration token..."
 
-base_api_url="https://api.github.com/"
-if [ -z "${ghe_hostname}" ]; then
-    base_api_url="https://${ghe_hostname}/api/v3/"
+base_api_url="https://api.github.com"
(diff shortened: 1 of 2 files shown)


scripts/create-latest-svc.sh
```diff
@@ -7,27 +7,29 @@ set -e
...
 # Should be used on VMs and not containers
-# Works on OSX and Linux 
+# Works on OSX and Linux
 # Assumes x64 arch
 #
 
 runner_scope=${1}
-runner_name=${2:-$(hostname)}
-svc_user=${3:-$USER}
+ghe_hostname=${2}
```
TingluoHuang, line 30: do we need to worry about compat? since we are changing the argument order. 😄 
feels like we don't need to since this is just a sample script?

brcrista: I don't think so -- just a sample.

PR #453, opened 2020-04-24
Revert "Use the API_URL and munge action URLs for GHES (#437)"

pjquirk:
This reverts commit f798f5606bc5d619ea4581a56e22dcc1e79e33b7.  After talking in Slack, we're going to wait until the core actions are available in the box.


TingluoHuang:
@pjquirk @ericsciple what do we need to do with this PR? merge the revert?

pjquirk:
There was discussion in slack about whether this change would affect the Alpha, which relies on pulling actions from dotcom.  I wasn't sure which runner they use so I prepped this just in case.  What do you think?

pjquirk:
Closing, no longer relevant.

PR #468, opened 2020-05-06
Use an env var to point to an Actions Service dev instance

brcrista:
When running Actions Service L2 tests against GHES, the instance won't have a `codedev.ms` URL.   This adds the ability to specify the URL through the `USE_DEV_ACTIONS_SERVICE_URL` environment variable.

Unless someone wants to name their GHES instance `codedev.ms`, let's leave the old backdoor there for compat.

**Testing**
- We've been using this while hacking on L2s through a [special release](https://github.com/brcrista/runner/releases/tag/v2.169.0-l2.1).

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -92,9 +92,11 @@ namespace GitHub.Runner.Listener.Configuration
             _term.WriteSection("Authentication");
             while (true)
             {
-                // Get the URL
-                var inputUrl = command.GetUrl();
-                if (inputUrl.Contains("codedev.ms", StringComparison.OrdinalIgnoreCase))
+                // For testing against a dev deployment of Actions Service, set this environment variable to the deployment URL
+                var actionsServiceTestUrl = Environment.GetEnvironmentVariable("DEV_ACTIONS_SERVICE_TEST_URL");
+                var inputUrl = actionsServiceTestUrl ?? command.GetUrl();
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -92,9 +92,11 @@ public async Task ConfigureAsync(CommandSettings command)
...
             while (true)
             {
-                // Get the URL
-                var inputUrl = command.GetUrl();
-                if (inputUrl.Contains("codedev.ms", StringComparison.OrdinalIgnoreCase))
+                // For testing against a dev deployment of Actions Service, set this environment variable to the deployment URL
+                var actionsServiceTestUrl = Environment.GetEnvironmentVariable("DEV_ACTIONS_SERVICE_TEST_URL");
+                var inputUrl = actionsServiceTestUrl ?? command.GetUrl();
+                if (inputUrl.Contains("codedev.ms", StringComparison.OrdinalIgnoreCase)
+                    || actionsServiceTestUrl != null)
```
TingluoHuang, line 99: `DEV_ACTIONS_SERVICE_TEST_URL` might be better set as bool since I assume the L2 test are still going to pass server URL on the commandline args.
```C#
var devActionsService = Environment.GetEnvironmentVariable("DEV_ACTIONS_SERVICE_TEST_URL");
var inputUrl = command.GetUrl();
if (inputUrl.Contains("codedev.ms", StringComparison.OrdinalIgnoreCase) || devActionsService != null)
```

brcrista: Good point, that would be simpler 😆

PR #471, opened 2020-05-07
docker build using -f instead of implied default 

dakale:
This lets you specify `image: './path/to/dockerfile.Dockerfile` instead of requiring the filename to be just `Dockerfile`.

Passes the existing filename to `build` explicitly with `-f` but continues passing the directory containing the dockerfile to indicate the docker context, rather than assuming it will contain "Dockerfile" by default.

e.g. in an action you can now have (under image):

```yaml
# action.yml
name: 'Hello World'
runs:
  using: 'docker'
  image: './anyNameIWant.Dockerfile'
...
```

### DIFF
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -17,7 +17,7 @@ namespace GitHub.Runner.Worker.Container
         string DockerInstanceLabel { get; }
         Task<DockerVersion> DockerVersion(IExecutionContext context);
         Task<int> DockerPull(IExecutionContext context, string image);
-        Task<int> DockerBuild(IExecutionContext context, string workingDirectory, string dockerFile, string tag);
+        Task<int> DockerBuild(IExecutionContext context, string workingDirectory, string dockerFile, string dockerContext, string tag);
         Task<string> DockerCreate(IExecutionContext context, ContainerInfo container);
         Task<int> DockerRun(IExecutionContext context, ContainerInfo container, EventHandler<ProcessDataReceivedEventArgs> stdoutDataReceived, EventHandler<ProcessDataReceivedEventArgs> stderrDataReceived);
         Task<int> DockerStart(IExecutionContext context, string containerId);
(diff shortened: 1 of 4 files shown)


src/Runner.Worker/Container/DockerCommandManager.cs
```diff
@@ -87,9 +87,9 @@ public async Task<int> DockerPull(IExecutionContext context, string image)
             return await ExecuteDockerCommandAsync(context, "pull", image, context.CancellationToken);
         }
 
-        public async Task<int> DockerBuild(IExecutionContext context, string workingDirectory, string dockerFile, string tag)
+        public async Task<int> DockerBuild(IExecutionContext context, string workingDirectory, string dockerFile, string dockerContext, string tag)
         {
-            return await ExecuteDockerCommandAsync(context, "build", $"-t {tag} \"{dockerFile}\"", workingDirectory, context.CancellationToken);
+            return await ExecuteDockerCommandAsync(context, "build", $"-t {tag} -f \"{dockerFile}\" {dockerContext}", workingDirectory, context.CancellationToken);
```
TingluoHuang, line 92: do we need to quote `{dockerContext}`?

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -52,7 +52,11 @@ public async Task RunAsync(ActionRunStage stage)
                 ExecutionContext.Output($"Dockerfile for action: '{dockerFile}'.");
 
                 var imageName = $"{dockerManger.DockerInstanceLabel}:{ExecutionContext.Id.ToString("N")}";
-                var buildExitCode = await dockerManger.DockerBuild(ExecutionContext, ExecutionContext.GetGitHubContext("workspace"), Directory.GetParent(dockerFile).FullName, imageName);
+                var buildExitCode = await dockerManger.DockerBuild(
+                    ExecutionContext,
+                    ExecutionContext.GetGitHubContext("workspace"),
+                    dockerFile,
+                    Directory.GetParent(dockerFile).FullName, imageName);
```
TingluoHuang, line 59: `imageName` on the newline. 😄

src/Runner.Worker/Container/DockerCommandManager.cs
dakale: Probably. It works either way with my samples but I think if theres a whitespace or maybe other weird characters in the path it could get split / not work without quotes so I added them

src/Runner.Worker/Handlers/ContainerActionHandler.cs
dakale: 👍

PR #472, opened 2020-05-11
add help info for '--labels'

TingluoHuang:
Fix: https://github.com/actions/runner/issues/262

```
Config Options:
 --unattended     Disable interactive prompts for missing arguments. Defaults will be used for missing options
 --url string     Repository to add the runner to. Required if unattended
 --token string   Registration token. Required if unattended
 --name string    Name of the runner to configure (default htl-mac)
 --labels string  Extra labels in addition to the default: 'self-hosted,OSX,X64'
 --work string    Relative runner work directory (default _work)
 --replace        Replace any existing runner with the same name (default false)

Examples:
 Configure a runner non-interactively:
  ./config.sh --unattended --url <url> --token <token>
 Configure a runner non-interactively, replacing any existing runner with the same name:
  ./config.sh --unattended --url <url> --token <token> --replace [--name <name>]
 Configure a runner non-interactively with three extra labels:
  ./config.sh --unattended --url <url> --token <token> --labels L1,L2,L3
```

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -466,6 +466,7 @@ Config Options:
  --url string     Repository to add the runner to. Required if unattended
  --token string   Registration token. Required if unattended
  --name string    Name of the runner to configure (default {Environment.MachineName ?? "myrunner"})
+ --labels string  Extra labels in addition to the default: 'self-hosted,{Constants.Runner.Platform},{Constants.Runner.PlatformArchitecture}'
  --work string    Relative runner work directory (default {Constants.Path.WorkDirectory})
  --replace        Replace any existing runner with the same name (default false)");
 #if OS_WINDOWS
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -466,6 +466,7 @@ private void PrintUsage(CommandSettings command)
  --url string     Repository to add the runner to. Required if unattended
  --token string   Registration token. Required if unattended
  --name string    Name of the runner to configure (default {Environment.MachineName ?? "myrunner"})
+ --labels string  Additional labels of the runner (default 'self-hosted', '{Constants.Runner.Platform}', '{Constants.Runner.PlatformArchitecture}')
```
ericsciple, line 469: should this be `default 'self-hosted,{Constants.Runner.Platform},{Constants.Runner.PlatformArchitecture}'`

ericsciple: nm.... this is *additional*

ericsciple: `Extra labels in addition to the default: 'self-hosted,{Constants.Runner.Platform},{Constants.Runner.PlatformArchitecture}'`

PR #473, opened 2020-05-11
Run codeql on all platforms.

TingluoHuang:


dangervslash:
-  @[not a resource no good](url) try the answer no. 
- mark as wrong subject today no use

Issue #370, opened 2020-03-13
Install as service 


Emtec:
**Describe the bug**
Cannot install runner as service, when runner name contains spaces. 

**To Reproduce**
1. Configure new runner with a name that contains spaces
2. Run sudo ./svc.sh install
3. See error

**Expected behavior**
SVC bash should just install runner as service.

## Runner Version and Platform
OS of the machine running the runner? Centos 8

## What's not working?
Creating launch runner in /etc/systemd/system/actions.runner.{Owner}-{RepositoryName}.{Runner name: Development Server}.service
Run as user: centos
Run as uid: 1000
gid: 1000
chmod: cannot access '/etc/systemd/system/actions.runner.{Owner}-{RepositoryName}.{Runner name: Development Server}': No such file or directory
chmod: cannot access 'Server.service': No such file or directory
Failed: failed to set permissions on /etc/systemd/system/actions.runner.{Owner}-{RepositoryName}.{Runner name: Development Server}.service

## Possible solution
In file svc.sh before UNIT_PATH add:
``SVC_NAME=${SVC_NAME// /_}``

TingluoHuang:
@Emtec thanks for reporting, do you want to create a PR to fix this? 😄

PR #474, opened 2020-05-11
Fix configure as service with runner name has space.

TingluoHuang:
fix: #370


src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 SVC_NAME="{{SvcNameVar}}"
+SVC_NAME=${SVC_NAME// /_}
```
ericsciple, line 4: is this safe for existing runners. I would guess so, since template

PR #476, opened 2020-05-12
Validate inputs only for repo action, no warning for small delay.

TingluoHuang:
https://github.com/github/c2c-actions-service/issues/1051

### DIFF
--- a/src/Runner.Worker/ActionRunner.cs
+++ b/src/Runner.Worker/ActionRunner.cs
@@ -170,11 +170,15 @@ namespace GitHub.Runner.Worker
                 }
             }
 
-            foreach (var input in userInputs)
+            // Validate inputs only for actions with action.yml
+            if (Action.Reference.Type == Pipelines.ActionSourceType.Repository)
             {
-                if (!validInputs.Contains(input))
+                foreach (var input in userInputs)
                 {
-                    ExecutionContext.Warning($"Unexpected input '{input}', valid inputs are ['{string.Join("', '", validInputs)}']");
+                    if (!validInputs.Contains(input))
+                    {
+                        ExecutionContext.Warning($"Unexpected input '{input}', valid inputs are ['{string.Join("', '", validInputs)}']");
+                    }
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/ActionRunner.cs
```diff
@@ -170,11 +170,15 @@ public async Task RunAsync()
                 }
             }
 
-            foreach (var input in userInputs)
+            // Validate inputs only for actions with action.yml
+            if (Action.Reference.Type == Pipelines.ActionSourceType.Repository)
```
TingluoHuang, line 174: The real diff is:
```
if (Action.Reference.Type == Pipelines.ActionSourceType.Repository)
{
   existing code
}
```

TingluoHuang: the current cause unwanted warning for `run` step.
ex: https://github.com/actions/canary/actions/runs/102763474
image

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -335,7 +336,7 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
             }
 
             // report total delay caused by server throttling.
-            if (_totalThrottlingDelayInMilliseconds > 0)
+            if (_totalThrottlingDelayInMilliseconds > _throttlingDelayReportThreshold)
```
ericsciple, line 339: Discussed offline, this is related to log lines and timeline record updates. Not general perf tarpitting.

Upload artifact, etc would be reported as 429 during that experience (depends on whether they log).

ericsciple: Existing code so better to just leave it, consider Timespan in future so name doesn't need to convey units :)

src/Runner.Worker/ActionRunner.cs
ericsciple: cool, basically anyhting with action.yml

TingluoHuang:
https://github.com/actions/runner/issues/477

PR #482, opened 2020-05-16
switch GITHUB_URL to GITHUB_SERVER_URL

ericsciple:


src/Runner.Worker/GitHubContext.cs
```diff
@@ -22,8 +22,8 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "repository_owner",
             "run_id",
             "run_number",
+            "server_url",
             "sha",
-            "url",
```
TingluoHuang, line 26: can we safely remove GITHUB_URL?

TingluoHuang: what about checkout action?

TingluoHuang: or anything already using GITHUB_URL

ericsciple: I added back compat code to the [checkout pr](https://github.com/actions/checkout/pull/248) (fallback order: `$GITHUB_SERVER_URL || $GITHUB_URL || 'https://github.com'`

GHES Alpha is pinned to the 2.169.0 runner, but uses checkout from dotcom. So only checkout needs to hold compat (can delete after alpha support is over)

Nothing else uses GITHUB_URL

ericsciple: Feature flag only made it out to Ring 1

Issue #489, opened 2020-05-19
Post step title prefixed with "Post Pre"


hashtagchris:
**Describe the bug**
When an action has both `pre` and `post` stages defined, the `post` step title is prefixed with "Post Pre" instead of just "Post".

**To Reproduce**
1. Create a workflow that executes an action with both `pre` and `post` stages. [Example](https://github.com/h10s/test-workflows/blob/master/.github/workflows/test-simple-actions.yml#L28-L30) 
1. Execute that workflow.
1. Inspect the name of the action's post step.

**Expected behavior**
The `post` step title is prefixed with just "Post".

## Runner Version and Platform
Version of your runner? 2.262.1

OS of the machine running the runner? macOS 10.14.6

## What's not working?
Please include error messages and screenshots.

From https://github.com/h10s/test-workflows/runs/690167853?check_suite_focus=true:

![image](https://user-images.githubusercontent.com/7145717/82368525-95519880-99e3-11ea-813d-3a47d60328b9.png)

TingluoHuang:
fixed.

PR #490, opened 2020-05-19
Fix inputs validation warning, fix post step display name, fix worker crash due to error in step.env

TingluoHuang:
https://github.com/actions/runner/issues/489


src/Runner.Worker/StepsRunner.cs
```diff
@@ -98,124 +98,139 @@ public async Task RunAsync(IExecutionContext jobContext)
...
                     if (step is IActionRunner actionStep)
                     {
                         // Set GITHUB_ACTION
                         step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
 
-                        // Evaluate and merge action's env block to env context
-                        var templateEvaluator = step.ExecutionContext.ToPipelineTemplateEvaluator();
-                        var actionEnvironment = templateEvaluator.EvaluateStepEnvironment(actionStep.Action.Environment, step.ExecutionContext.ExpressionValues, step.ExecutionContext.ExpressionFunctions, VarUtil.EnvironmentVariableKeyComparer);
-                        foreach (var env in actionEnvironment)
+                        try
```
TingluoHuang, line 107: the diff is not good, the real change is adding `try..catch` around `EvaluateStepEnvironment` and fail the step when there is an exception.

Step runner should never blow up with exception back to JobRunner.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -98,124 +98,139 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                            // fail the step since there is an evaluate error.
+                            Trace.Info("Caught exception from expression for step.env");
+                            evaluateStepEnvFailed = true;
+                            step.ExecutionContext.Error(ex);
+                            CompleteStep(step, nextStep, TaskResult.Failed);
                         }
                     }
 
-                    try
+                    if (!evaluateStepEnvFailed)
```
TingluoHuang, line 127: the diff is bad, the real change is to skip execute the step when `EvaluateStepEnvironment` failed, and merge step result with job result.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -98,124 +98,139 @@ public async Task RunAsync(IExecutionContext jobContext)
...
-                            jobCancelRegister?.Dispose();
-                            jobCancelRegister = null;
+                            if (jobCancelRegister != null)
+                            {
+                                jobCancelRegister?.Dispose();
+                                jobCancelRegister = null;
+                            }
                         }
                     }
                 }
```
TingluoHuang, line 236: merge step result and job result happen down belove.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -364,7 +364,11 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
                 }
             }
 
-            _cancellationTokenSource?.Dispose();
+            if (Root != this)
+            {
+                // only dispose TokenSource for step level ExecutionContext
```
TingluoHuang, line 369: when `steprunner` crash with an exception, we complete the job ExecutionContext which causes the token to get disposed, and `Job Finalize` will cause worker crash since it tries to create a new token source from this token.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -94,6 +94,13 @@ public async Task RunAsync()
             if (handlerData.HasPost && (Stage == ActionRunStage.Pre || Stage == ActionRunStage.Main))
             {
                 string postDisplayName = $"Post {this.DisplayName}";
+                if (Stage == ActionRunStage.Pre &&
+                    this.DisplayName.StartsWith("Pre ", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Trim the leading `Pre ` from the display name.
+                    // Otherwise, we will get `Post Pre xxx` as DisplayName for the Post step.
+                    postDisplayName = $"Post {this.DisplayName.Substring("Pre ".Length)}";
+                }
```
ericsciple, line 103: 👍

src/Runner.Worker/ActionRunner.cs
```diff
@@ -155,6 +162,13 @@ public async Task RunAsync()
             }
 
             var validInputs = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            if (handlerData.ExecutionType == ActionExecutionType.Container)
+            {
+                // container action always accept 'entryPoint' and 'args' as inputs
+                // https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswithargs
+                validInputs.Add("entryPoint");
+                validInputs.Add("args");
+            }
```
ericsciple, line 171: 👍

src/Runner.Worker/StepsRunner.cs
ericsciple: did you consider merging this with current local vars for tracking whether error when evaluating `if` condition?

ericsciple: i dont know whether possible (something to consider whether it would simplify)

ericsciple: looking further... ignore my comment probably doesnt make sense

src/Runner.Worker/ExecutionContext.cs
ericsciple: This seems fine. First inclination was a change in JobExtension.FinalizeJob but actually this probably makes more sense (dont dispose since child nodes still should link to get ctrl+c)

PR #492, opened 2020-05-19
Max attempts while adding runner

lokesh755:
Limiting the number of attempts while adding the runner to avoid infinite loop of requests.

Related to https://github.com/github/c2c-actions-runtime/issues/278

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -160,8 +160,11 @@ namespace GitHub.Runner.Listener.Configuration
             }
 
             TaskAgent agent;
+            int attemptCount = 5;
+
             while (true)
             {
+                attemptCount--;
                 runnerSettings.AgentName = command.GetRunnerName();
 
                 _term.WriteLine();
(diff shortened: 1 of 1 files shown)


TingluoHuang:
Do we need to have this kind of limit for all inputs loop?

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -192,10 +195,10 @@ public async Task ConfigureAsync(CommandSettings command)
                             _term.WriteError("Failed to replace the runner.  Try again or ctrl-c to quit");
                         }
                     }
-                    else if (command.Unattended)
+                    else if (command.Unattended || attemptCount <= 0)
                     {
                         // if not replace and it is unattended config.
-                        throw new TaskAgentExistsException($"Pool {runnerSettings.PoolId} already contains a runner with name {runnerSettings.AgentName}.");
+                        throw new TaskAgentExistsException($"A runner already exists with the name {runnerSettings.AgentName}.");
```
TingluoHuang, line 201: do you want to print different message when it’s run out of attempt?

lokesh755:
Abandoning this PR, since this may not really help in decrease in number of requests based on the user pattern described in the github/c2c-actions-runtime#278

PR #500, opened 2020-05-22
add debug option to disable SSL verification

larsxschneider:
Customers playing with the hosted Actions runner preview might not have
a proper SSL certificate setup. Give them an option to disable SSL
verification for Node and Git commands in their actions (e.g. for the
checkout action).

This debug mode is enabled by creating a file `.debug-runner-disable-ssl-verify`
in the directory of the `runsvc.sh` file.

---

I know this is an ugly hack but it might help in a dev setup 😄

### DIFF
--- a/src/Misc/layoutbin/runsvc.sh
+++ b/src/Misc/layoutbin/runsvc.sh
@@ -10,6 +10,13 @@ if [ -f ".path" ]; then
     echo ".path=${PATH}"
 fi
 
+if [ -f ".debug-runner-disable-ssl-verify" ]; then
+    export GIT_SSL_NO_VERIFY=true
+    export NODE_TLS_REJECT_UNAUTHORIZED=0
+    echo "SSL verification for Git and Node disabled"
+    echo "   *** DO NOT RUN THIS IN PRODUCTION ***"
+fi
+
 # insert anything to setup env when running as a service
 
 # run the host process which keep the listener alive


TingluoHuang:
How do you deal with the SSL error that the runner will hit?

larsxschneider:
@TingluoHuang you mean when the runner connects to GH? That SSL verification is indeed not disabled with this switch. But maybe it could be added by passing some env variable to the runner? 

I was just debugging a hosted runner for a customer and this ugly hack fixed *exactly* the problem they were facing. I don't know what was wrong with their SSL setup and therefore I just disabled SSL verification for the problematic parts all together. If you think this might be useful for others, then I am happy to refactor it in whatever way you think is right. If you don't think that is useful, then I am not offended if you close the PR 👍 😄 .

TingluoHuang:
@larsxschneider you can create a `.env` file on the runner root dir and add that 2 env in the file as
```
GIT_SSL_NO_VERIFY=true
NODE_TLS_REJECT_UNAUTHORIZED=0
```
then restart the runner service.

TingluoHuang:
We need to make changes to the runner itself to disable SSL validation in the Dotnet core as well. Close this for now.

LoopsGod:
> @larsxschneider you can create a `.env` file on the runner root dir and add that 2 env in the file as
> 
> ```
> GIT_SSL_NO_VERIFY=true
> NODE_TLS_REJECT_UNAUTHORIZED=0
> ```
> 
> then restart the runner service.

@TingluoHuang How do I restart the GitHub Runner service without rebooting? Tried a ton of commands but can't find the service that runs the GitHub runner.

PR #508, opened 2020-05-29
prepare to switch to service resolves archive download info

ericsciple:
Refactoring repository archive download logic to prepare for server changes. In a future PR, will switch to a REST API for the server to resolve archive link and other download info (token).

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -73,6 +73,11 @@ namespace GitHub.Runner.Worker
             // Clear the cache (for self-hosted runners)
             IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
 
+            // todo: Remove when feature flag DistributedTask.NewActionMetadata is removed
+            var newActionMetadata = executionContext.Variables.GetBoolean("DistributedTask.NewActionMetadata") ?? false;
+
+            var repositoryActions = new List<Pipelines.ActionStep>();
+
             foreach (var action in actions)
             {
                 if (action.Reference.Type == Pipelines.ActionSourceType.ContainerRegistry)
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -73,6 +73,13 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
             // Clear the cache (for self-hosted runners)
             IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
 
+            // todo: Remove when feature flag DistributedTask.NewActionMetadata is removed
```
ericsciple, line 76: Everything compiles. Haven't tested yet.

src/Runner.Worker/ActionManager.cs
```diff
@@ -124,6 +132,81 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+
+            if (repositoryActions.Count > 0)
+            {
+                // Get the download info
+                var downloadInfos = await GetDownloadInfoAsync(executionContext, repositoryActions);
+
+                // Download each action
+                foreach (var action in repositoryActions)
+                {
+                    var lookupKey = GetDownloadInfoLookupKey(action);
```
pjquirk, line 165: Minor: `GetDownloadInfoAsyc` could just return a list of download info objects (i.e. the values of the dictionary it builds), so the for loop doesn't need to retrieve them by key. It should just loop over what was returned and download each one.

src/Runner.Worker/ActionManager.cs
```diff
@@ -464,6 +547,121 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
             }
         }
 
+        // todo: Remove when feature flag DistributedTask.NewActionMetadata is removed
```
pjquirk, line 550: This will be removed when the REST call to the service is implemented, right? It's only called from the temporary implementation below.

src/Runner.Worker/ActionManager.cs
```diff
@@ -124,6 +132,81 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+
+                    if (!downloadInfos.TryGetValue(lookupKey, out var downloadInfo))
+                    {
+                        throw new Exception($"Missing download info for {lookupKey}");
+                    }
+
+                    await DownloadRepositoryActionAsync(executionContext, downloadInfo);
+                }
+
+                // More preparation based on content in the repository (action.yml)
```
pjquirk, line 179: Copy-pasted from above? Can this be pulled into a method?

src/Runner.Worker/ActionManager.cs
```diff
@@ -549,6 +747,32 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             }
         }
 
+        private async Task DownloadRepositoryActionAsync(IExecutionContext executionContext, ActionDownloadInfo downloadInfo)
+        {
+            Trace.Entering();
+            ArgUtil.NotNull(executionContext, nameof(executionContext));
+            ArgUtil.NotNullOrEmpty(downloadInfo.NameWithOwner, nameof(downloadInfo.NameWithOwner));
```
pjquirk, line 754: Missing `ArgUtil.NotNull(downloadInfo)`

src/Runner.Worker/ActionManager.cs
ericsciple: Need to preserve the original list because of local actions (not downloaded). Oh wait local actions do need to be downloaded too.... crap let me think more about this.

src/Runner.Worker/ActionManager.cs
ericsciple: i dont want to reuse, want to cleanly delete the old stuff after the FF rolls out. Reuse will complicate the delete.

src/Runner.Worker/ActionManager.cs
ericsciple: Yea I went through and left comments on all the places where "legacy" code needs to be removed when we delete the feature-flag.

ericsciple: Oops yes i need to fix the comment:

```
// This implementation is temporary and will be replaced with a REST API call to the service to resolve
```

src/Runner.Worker/ActionManager.cs
ericsciple: OK actually local actions are not downloaded. Ting reminded me, they come from checkout.

src/Runner.Worker/ActionManager.cs
```diff
@@ -124,6 +132,81 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+                {
+                    var lookupKey = GetDownloadInfoLookupKey(action);
+                    if (string.IsNullOrEmpty(lookupKey))
+                    {
+                        continue;
+                    }
+
+                    if (!downloadInfos.TryGetValue(lookupKey, out var downloadInfo))
+                    {
+                        throw new Exception($"Missing download info for {lookupKey}");
```
TingluoHuang, line 173: double-check whether we have something to catch the exception and fail the `prepare job` step instead of crash the worker.

src/Runner.Worker/ActionManager.cs
```diff
@@ -872,6 +1108,48 @@ private ActionContainer PrepareRepositoryActionAsync(IExecutionContext execution
...
+                throw new NotSupportedException(repositoryReference.RepositoryType);
+            }
+
+            ArgUtil.NotNullOrEmpty(repositoryReference.Name, nameof(repositoryReference.Name));
+            ArgUtil.NotNullOrEmpty(repositoryReference.Ref, nameof(repositoryReference.Ref));
+            return $"{repositoryReference.Name}@{repositoryReference.Ref}";
+        }
+
+        private static AuthenticationHeaderValue CreateAuthHeader(string authorization)
+        {
```
TingluoHuang, line 1137: do we need to register a secret mask for the authorization?

src/Runner.Worker/ActionManager.cs
```diff
@@ -73,6 +73,13 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
             // Clear the cache (for self-hosted runners)
             IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
 
+            // todo: Remove when feature flag DistributedTask.NewActionMetadata is removed
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
```
TingluoHuang, line 77: do we need ?
```
            var runnerSettings = configurationStore.GetSettings();
```

TingluoHuang:
@ericsciple there are some L0 tests for ActionsManager, see whether you can make them run against your new code flow to double-check for regression.

src/Runner.Worker/ActionManager.cs
ericsciple: will fix

src/Runner.Worker/ActionManager.cs
ericsciple: was going to add later. will go ahead and add now to make sure we dont forget later :)

ericsciple: later = when we switch to the REST API

src/Runner.Worker/ActionManager.cs
ericsciple: Tested to confirm caller handles gracefully. Also follows the pattern of the rest of the file (exceptions bubble)

PR #516, opened 2020-06-04
Fix null ref exception in SecretMasker caused by hashfiles timeout.

TingluoHuang:
Fix:
```
GitHub actions runner worker process crashed with error: Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.DistributedTask.Logging.SecretMasker.MaskSecrets(String input)
   at GitHub.DistributedTask.Expressions2.Sdk.EvaluationTraceWriter.Info(String message)
   at GitHub.Runner.Worker.Expressions.HashFilesFunction.HashFilesTrace.Info(String message)
   at GitHub.Runner.Sdk.ProcessInvoker.SendSignal(Signals signal, TimeSpan timeout)
   at GitHub.Runner.Sdk.ProcessInvoker.SendSIGINT(TimeSpan timeout)
   at GitHub.Runner.Sdk.ProcessInvoker.CancelAndKillProcessTree(Boolean killProcessOnCancel)
   at GitHub.Runner.Sdk.ProcessInvoker.<>c__DisplayClass29_0.<<ExecuteAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
   at System.Threading.QueueUserWorkItemCallbackDefaultContext.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

`Expressions2.Sdk` make a copy of `secret masker` for each evaluation, and call `Dispose()` on the secret masker in the `finally{}` block.


src/Runner.Worker/Expressions/HashFilesFunction.cs
```diff
@@ -89,19 +91,29 @@ public sealed class HashFilesFunction : Function
...
 
-            int exitCode = p.ExecuteAsync(workingDirectory: githubWorkspace,
-                                          fileName: node,
-                                          arguments: $"\"{hashFilesScript.Replace("\"", "\\\"")}\"",
-                                          environment: env,
-                                          requireExitCodeZero: false,
-                                          cancellationToken: new CancellationTokenSource(TimeSpan.FromSeconds(120)).Token).GetAwaiter().GetResult();
-
-            if (exitCode != 0)
+            using (var tokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(_hashFileTimeoutSeconds)))
```
ericsciple, line 94: have you considered passing the step cancellation token through the context and creating a linked token?

src/Runner.Worker/Expressions/HashFilesFunction.cs
```diff
@@ -89,19 +91,29 @@ public sealed class HashFilesFunction : Function
...
+                                                  requireExitCodeZero: false,
+                                                  cancellationToken: tokenSource.Token).GetAwaiter().GetResult();
 
-            return hashResult;
+                    if (exitCode != 0)
+                    {
+                        throw new InvalidOperationException($"hashFiles('{ExpressionUtility.StringEscape(string.Join(", ", patterns))}') failed. Fail to hash files under directory '{githubWorkspace}'");
+                    }
+                }
+                catch (OperationCanceledException) when (tokenSource.IsCancellationRequested)
```
ericsciple, line 110: how does this fix the exception? Also why is secret masker hitting null ref? because it's disposed already?

TingluoHuang: change in process invoker will fix this.

src/Runner.Sdk/ProcessInvoker.cs
```diff
@@ -344,6 +349,13 @@ public ProcessInvoker(ITraceWriter trace)
                 Trace.Info($"Finished process {_proc.Id} with exit code {_proc.ExitCode}, and elapsed time {_stopWatch.Elapsed}.");
             }
 
+            if (cancellationToken.IsCancellationRequested)
```
ericsciple, line 352: this should go inside the using block (otherwise race condition)

TingluoHuang: Do you mean `registration` get disposed?

PR #524, opened 2020-06-08
Update documentation to recommend VS code 2019

ethanchewy:
VS code versions > 2019 doesn't yet support important libraries that are needed to debug/modify runner code effectively such as .NET, etc. 

Ex: https://github.com/OmniSharp/omnisharp-vscode/issues/2507

### DIFF
--- a/docs/contribute.md
+++ b/docs/contribute.md
@@ -54,8 +54,9 @@ cd ./src
 ### Editors
 
 [Using Visual Studio Code](https://code.visualstudio.com/)
-[Using Visual Studio](https://code.visualstudio.com/docs)  
 
+[Using Visual Studio](https://code.visualstudio.com/docs)  
+We recommend using [Visual Studio 2019](https://visualstudio.microsoft.com/vs/) since it has built in support for .NET and other libraries that are helpful for contributing and running this repo while newer versions of Visual Studio does not have support for these libraries as of yet. 
 ### Styling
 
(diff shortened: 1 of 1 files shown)


docs/contribute.md
```diff
@@ -54,8 +54,9 @@ cd ./src
 ### Editors
 
 [Using Visual Studio Code](https://code.visualstudio.com/)
-[Using Visual Studio](https://code.visualstudio.com/docs)  
 
+[Using Visual Studio](https://code.visualstudio.com/docs)  
+We recommend using [Visual Studio 2019](https://visualstudio.microsoft.com/vs/) since it has built in support for .NET and other libraries that are helpful for contributing and running this repo while newer versions of Visual Studio does not have support for these libraries as of yet.
```
TingluoHuang, line 59: I would expect the older VS version might not work with dotnet core, not the newer version.

What error do you get?

I would not say `We recommend using [Visual Studio 2019]` since VsCode is a better IDE for this repo since the runner is for Windows/Linux/macOS, but you can only use VS in Windows.

ethanchewy: I had a lot of issues with omnisharp like: https://github.com/OmniSharp/omnisharp-vscode/issues/2507. But when I reinstall Visual Studio Code, the error disappeared and I wasn't able to replicate it so I'm not sure what's going on. I'll close this PR for now :P

PR #549, opened 2020-06-15
Composite Actions Support for Multiple Run Steps

ethanchewy:
In this PR, we support executing multiple run steps with custom inputs in a composite action. You can use this feature by writing `using: "composite"` in the action.yml file. Our goal for our first version is to be able to run simple, multiple steps in an action.yml file. This first step allows for users to run a list of steps in the `action.yml` file rather than invoking a javascript entry point. This will eventually allow us to support nesting of actions within an action, etc. 

Here is an in-depth example:

`ethanchewy/test-composite/action.yml`:
```
name: "Composite Action Test v2"
description: "Run a simple composite action"
inputs:
  your_name:
    description: 'lol'
    default: 'Ethan'
runs:
    using: "composite"
    steps: 
        - run: echo hello ${{inputs.your_name}}
        - run: echo sha ${{github.sha}}
        - run: echo os ${{runner.os}}
        - run: |
            echo job index ${{ strategy.job-index }}
```
`workflow.yml`

```
name: Composite Action
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
jobs:
  build:
    runs-on: self-hosted
    - uses: actions/checkout@v2
    - uses: ethanchewy/test-composite@v5
      with:
        your_name: "Octocat"
    - name: step 1
      run: echo Workflow step 1
    - name: step 2
      run: echo Workflow step 2
```
Github UI Action View: https://github.com/ethanchewy/testing-actions/runs/778225962?check_suite_focus=true

What this doesn't support yet (for future PRs):

- Input, Output, Env flow (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables)
- Nesting other actions in composite actions (supporting in a step the terminology "using: __")
- etc.

**How this was implemented at a high level:**
We have one queue that keeps track of the composite actions transformed into a job step (in the process, it’s added to the Root node) => then we requeue in a way where these composite actions are executed first. 
None of the steps are actually evaluated by the CompositeActionsHandler but they are pushed to the front of the job step queue and handled by other handlers depending on its type (ex: Script handler).

**This is an experimental feature that can only be turned on via turning the feature flag on:**
`export TESTING_COMPOSITE_ACTIONS_ALPHA=1`
`./run.sh`

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -395,7 +395,7 @@ namespace GitHub.Runner.Worker
                             Trace.Info($"Action cleanup plugin: {plugin.PluginTypeName}.");
                         }
                     }
-                    else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                    else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                     {
                         var compositeAction = definition.Data.Execution as CompositeActionExecutionData;
                         Trace.Info($"Action steps: {compositeAction.Steps}.");
@@ -1106,7 +1106,7 @@ namespace GitHub.Runner.Worker
                     Trace.Info($"Action plugin: {(actionDefinitionData.Execution as PluginActionExecutionData).Plugin}, no more preparation.");
                     return null;
                 }
-                else if (actionDefinitionData.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                else if (actionDefinitionData.Execution.ExecutionType == ActionExecutionType.Composite && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                 {
                     Trace.Info($"Action composite: {(actionDefinitionData.Execution as CompositeActionExecutionData).Steps}, no more preparation.");
                     return null;
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -1221,7 +1221,7 @@ namespace GitHub.Runner.Worker
         NodeJS,
         Plugin,
         Script,
-        Composite
+        Composite,
     }
 
     public sealed class ContainerActionExecutionData : ActionExecutionData
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -397,7 +397,8 @@ namespace GitHub.Runner.Worker
                     }
                     else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                     {
-                        // Don't do anything for now
+                        var compositeAction = definition.Data.Execution as CompositeActionExecutionData;
+                        Trace.Info($"Action steps: {compositeAction.Steps}.");
                     }
                     else
                     {
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -395,6 +395,10 @@ namespace GitHub.Runner.Worker
                             Trace.Info($"Action cleanup plugin: {plugin.PluginTypeName}.");
                         }
                     }
+                    else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                    {
+                        // Don't do anything for now
+                    }
                     else
                     {
                         throw new NotSupportedException(definition.Data.Execution.ExecutionType.ToString());
@@ -1101,6 +1105,11 @@ namespace GitHub.Runner.Worker
                     Trace.Info($"Action plugin: {(actionDefinitionData.Execution as PluginActionExecutionData).Plugin}, no more preparation.");
                     return null;
                 }
+                else if (actionDefinitionData.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                {
+                    Trace.Info($"Action composite: {(actionDefinitionData.Execution as CompositeActionExecutionData).Steps}, no more preparation.");
+                    return null;
+                }
                 else
                 {
                     throw new NotSupportedException(actionDefinitionData.Execution.ExecutionType.ToString());
@@ -1211,6 +1220,7 @@ namespace GitHub.Runner.Worker
         NodeJS,
         Plugin,
         Script,
+        Composite
     }
 
     public sealed class ContainerActionExecutionData : ActionExecutionData
@@ -1265,6 +1275,14 @@ namespace GitHub.Runner.Worker
         public override ActionExecutionType ExecutionType => ActionExecutionType.Script;
         public override bool HasPre => false;
         public override bool HasPost => false;
+    }
+
+    public sealed class CompositeActionExecutionData : ActionExecutionData
+    {
+        public override ActionExecutionType ExecutionType => ActionExecutionType.Composite;
+        public override bool HasPre => false;
+        public override bool HasPost => false;
+        public List<Pipelines.ActionStep> Steps { get; set; }
     }
 
     public abstract class ActionExecutionData
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -398,7 +398,8 @@ namespace GitHub.Runner.Worker
                     else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                     {
                         var compositeAction = definition.Data.Execution as CompositeActionExecutionData;
-                        Trace.Info($"Action steps: {compositeAction.Steps}.");
+                        Trace.Info($"Load {compositeAction.Steps.Count} action steps.");
+                        Trace.Verbose($"Details: {StringUtil.ConvertToJson(compositeAction.Steps)}");
                     }
                     else
                     {
--- a/src/Runner.Worker/ActionManifestManager.cs
+++ b/src/Runner.Worker/ActionManifestManager.cs
@@ -14,6 +14,7 @@ using YamlDotNet.Core;
 using YamlDotNet.Core.Events;
 using System.Globalization;
 using System.Linq;
+using Pipelines = GitHub.DistributedTask.Pipelines;
 
 namespace GitHub.Runner.Worker
 {
@@ -92,7 +93,7 @@ namespace GitHub.Runner.Worker
                             break;
 
                         case "runs":
-                            actionDefinition.Execution = ConvertRuns(context, actionPair.Value);
+                            actionDefinition.Execution = ConvertRuns(executionContext, context, actionPair.Value);
                             break;
                         default:
                             Trace.Info($"Ignore action property {propertyName}.");
@@ -284,7 +285,7 @@ namespace GitHub.Runner.Worker
             // Add the file table
             if (_fileTable?.Count > 0)
             {
-                for (var i = 0 ; i < _fileTable.Count ; i++)
+                for (var i = 0; i < _fileTable.Count; i++)
                 {
                     result.GetFileId(_fileTable[i]);
                 }
@@ -294,6 +295,7 @@ namespace GitHub.Runner.Worker
         }
 
         private ActionExecutionData ConvertRuns(
+            IExecutionContext executionContext,
             TemplateContext context,
             TemplateToken inputsToken)
         {
@@ -311,6 +313,8 @@ namespace GitHub.Runner.Worker
             var postToken = default(StringToken);
             var postEntrypointToken = default(StringToken);
             var postIfToken = default(StringToken);
+            var stepsLoaded = default(List<Pipelines.ActionStep>);
+
             foreach (var run in runsMapping)
             {
                 var runsKey = run.Key.AssertString("runs key").Value;
@@ -355,6 +359,15 @@ namespace GitHub.Runner.Worker
                     case "pre-if":
                         preIfToken = run.Value.AssertString("pre-if");
                         break;
+                    case "steps":
+                        if (!String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                        {
+                            var steps = run.Value.AssertSequence("steps");
+                            var evaluator = executionContext.ToPipelineTemplateEvaluator();
+                            stepsLoaded = evaluator.LoadCompositeSteps(steps);
+                            break;
+                        }
+                        throw new Exception("You aren't supposed to be using Composite Actions yet!");
                     default:
                         Trace.Info($"Ignore run property {runsKey}.");
                         break;
@@ -399,6 +412,20 @@ namespace GitHub.Runner.Worker
                             InitCondition = preIfToken?.Value ?? "always()",
                             Post = postToken?.Value,
                             CleanupCondition = postIfToken?.Value ?? "always()"
+                        };
+                    }
+                }
+                else if (string.Equals(usingToken.Value, "composite", StringComparison.OrdinalIgnoreCase) && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                {
+                    if (stepsLoaded == null)
+                    {
+                        throw new ArgumentNullException($"No steps provided.");
+                    }
+                    else
+                    {
+                        return new CompositeActionExecutionData()
+                        {
+                            Steps = stepsLoaded,
                         };
                     }
                 }
--- a/src/Runner.Worker/ActionManifestManager.cs
+++ b/src/Runner.Worker/ActionManifestManager.cs
@@ -419,6 +419,7 @@ namespace GitHub.Runner.Worker
                 {
                     if (stepsLoaded == null)
                     {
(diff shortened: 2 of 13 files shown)


src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -355,6 +360,15 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
                     case "pre-if":
                         preIfToken = run.Value.AssertString("pre-if");
                         break;
+                    case "steps":
+                        if (!String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                        {
+                            var steps = run.Value.AssertSequence("steps");
+                            var evaluator = executionContext.ToPipelineTemplateEvaluator();
+                            stepsLoaded = evaluator.LoadCompositeSteps(steps, null, null);
```
ericsciple, line 368: lets modify the signature of LoadCompositeSteps so it just takes the steps.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -432,7 +437,10 @@ private bool InitializeScope(IStep step, Dictionary<string, PipelineContextData>
             // Setup expression values
             var scopeName = executionContext.ScopeName;
             executionContext.ExpressionValues["steps"] = stepsContext.GetScope(scopeName);
-            executionContext.ExpressionValues["inputs"] = string.IsNullOrEmpty(scopeName) ? null : scopeInputs[scopeName];
+            // Temp fix to prevent the StepsRunner from overwriting inputs for CompositeActions
+            if (!executionContext.ExpressionValues.ContainsKey("inputs")) {
```
ethanchewy, line 441: Format doc.

src/Runner.Worker/ActionManager.cs
```diff
@@ -395,6 +395,10 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                             Trace.Info($"Action cleanup plugin: {plugin.PluginTypeName}.");
                         }
                     }
+                    else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                    {
+                        // Don't do anything for now
```
TingluoHuang, line 400: check what we do for nodejs, line 373, just dump all steps into log might be good.

TingluoHuang: we can just dump the step as JSON in log.

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -264,5 +263,351 @@ internal static class PipelineTemplateConverter
 
             return result;
         }
+
```
TingluoHuang, line 266: I assume these changes are copied from the existing server code?

src/Runner.Worker/ActionManager.cs
ethanchewy: good idea - thanks. 

(Note to self: do the same for when I add env flow to composite actions)

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
ethanchewy: Yes

src/Runner.Worker/ActionManager.cs
```diff
@@ -395,6 +395,11 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                             Trace.Info($"Action cleanup plugin: {plugin.PluginTypeName}.");
                         }
                     }
+                    else if (definition.Data.Execution.ExecutionType == ActionExecutionType.Composite && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
```
ericsciple, line 398: `string`

src/Runner.Worker/ActionManager.cs
```diff
@@ -395,6 +395,11 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
...
+                    {
+                        var compositeAction = definition.Data.Execution as CompositeActionExecutionData;
+                        Trace.Info($"Action steps: {compositeAction.Steps}.");
```
ericsciple, line 401: I think this will print `Action steps: Object`

ericsciple: If the json representation is desired (too noisy? esp with large inline scripts?) then see StringUtil or IOUtil (has function to convert to/from json)

ethanchewy: @TingluoHuang thoughts?

src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -402,6 +415,20 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
                         };
                     }
                 }
+                else if (string.Equals(usingToken.Value, "composite", StringComparison.OrdinalIgnoreCase) && !String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
+                {
+                    if (stepsLoaded == null)
+                    {
+                        throw new ArgumentNullException($"No steps provided.");
```
ericsciple, line 422: does this error give the user enough context to know where the problem is? Like does it convey that an error happened when loading the path-to-action.yml

ericsciple: i guess we would have the same problem with the other place too (in this file)

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -105,6 +105,8 @@ public interface IExecutionContext : IRunnerService
         // others
         void ForceTaskComplete();
         void RegisterPostJobStep(IStep step);
+        IStep RegisterCompositeStep(IStep step, DictionaryContextData inputsData);
```
ericsciple, line 108: This is a bit of a misnomer. The "composite step" is the outer thing. Here we're an individual inner step.

ericsciple: "InsertNestedStep"?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -105,6 +105,8 @@ public interface IExecutionContext : IRunnerService
...
+        void EnqueueAllCompositeSteps(Queue<IStep> steps);
```
ericsciple, line 109: Then we can eliminate this function? (switch `JobSteps` to List instead of Queue)

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -266,6 +267,64 @@ public void RegisterPostJobStep(IStep step)
             Root.PostJobSteps.Push(step);
         }
 
+        /*
```
ericsciple, line 270: Use:

```
/// <summary>
/// RegisterCompositeStep is a helper function used in CompositeActionHandler::RunAsync to
/// add a child node, aka a step, to the current job to the front of the queue for processing. 
/// </summary>
```

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -266,6 +267,64 @@ public void RegisterPostJobStep(IStep step)
...
+            RegisterCompositeStep is a helper function used in CompositeActionHandler::RunAsync to 
+            add a child node, aka a step, to the current job to the front of the queue for processing. 
+        */
+        public IStep RegisterCompositeStep(IStep step, DictionaryContextData inputsData)
+        {
+            // ~Brute Force Method~
```
ericsciple, line 276: consider removing these if not needed anymore (looks like may have been intended as a review comment)

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -66,6 +66,14 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
                 handler = HostContext.CreateService<IRunnerPluginHandler>();
                 (handler as IRunnerPluginHandler).Data = data as PluginActionExecutionData;
             }
+            else if (data.ExecutionType == ActionExecutionType.Composite)
+            {
+                // TODO
```
ericsciple, line 71: todo what? :)

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -66,6 +66,14 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                // Runner plugin
+                handler = HostContext.CreateService<ICompositeActionHandler>();
+                // handler = CompositeHandler;
```
ericsciple, line 74: dead code

src/Runner.Worker/StepsRunner.cs
```diff
@@ -67,6 +67,11 @@ public async Task RunAsync(IExecutionContext jobContext)
 
                 var step = jobContext.JobSteps.Dequeue();
                 var nextStep = jobContext.JobSteps.Count > 0 ? jobContext.JobSteps.Peek() : null;
+                // TODO: Fix this temporary workaround for Composite Actions
+                if (!String.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
```
ericsciple: is this required anymore? I thought the scope changes worked-around the issue

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -266,6 +267,64 @@ public void RegisterPostJobStep(IStep step)
...
+            //     Root.JobSteps.Enqueue(s);
+
+            // ~Optimized Method~
+            // Alterative solution: We add to another temp Queue
+            // After we add all the transformed composite steps to this temp queue, we requeue the whole JobSteps accordingly in EnqueueAllCompositeSteps()
+            // where the queued composite steps are at the front of the JobSteps Queue and the rest of the jobs maintain its order and are 
+            // placed after the queued composite steps
+            // This will take only O(n+m) time which would be just as efficient if we refactored the code of JobSteps to a PriorityQueue
+            // This temp Queue is created in the CompositeActionHandler. 
+            var newGuid = Guid.NewGuid();
```
ericsciple, line 294: I'm worried a little bit because we're not adding the contexts from the job message, like we normally do when creating steps.

I know this is what Post does, but wouldnt surprise me if there is a bug there too.

It may be safe, since we're copying the root node (maybe job message contexts copied there first so safe?)

Let's add a TODO comment in the code like "// TODO: confirm whether not copying message contexts is safe".

I can help look into this more. It's tedious because values are delay expanded sometimes (sometimes evaluated early, like DisplayName). I wish the code were designed simpler... hmm thinking about alternatives.

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConstants.cs
```diff
@@ -65,6 +65,8 @@ public sealed class PipelineTemplateConstants
         public const String StepEnv = "step-env";
         public const String StepIfResult = "step-if-result";
         public const String Steps = "steps";
+
```
ericsciple, line 68: extra newline

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
```diff
@@ -159,6 +159,32 @@ public class PipelineTemplateEvaluator
             return result;
         }
 
+        public List<ActionStep> LoadCompositeSteps(
```
ericsciple, line 162: lets make sure to get these changes into the server at the same time so we dont forget and stomp over important changes later.

ethanchewy:
Notes: maybe add unit tests as well

src/Runner.Worker/ActionManifestManager.cs
ethanchewy: I think this error makes sense and is appropriate. 

At the top of `ConvertRuns`: 
`var stepsLoaded = default(List<Pipelines.ActionStep>);`

Then, we check in the `action.yml` file to see if there is any `steps` token:
```
case "steps":
                        if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                        {
                            var steps = run.Value.AssertSequence("steps");
                            var evaluator = executionContext.ToPipelineTemplateEvaluator();
                            stepsLoaded = evaluator.LoadCompositeSteps(steps);
                            break;
                        }
                        throw new Exception("You aren't supposed to be using Composite Actions yet!");
```

If there are any issues with loading the steps, there would be an error reported in the `LoadSteps()` or `LoadStep()` function in the pipeline template evaulator. 

Thus, if `stepsLoaded` is null, it must be because there was no `steps` attribute written in the `action.yml` by the user.

Thus, I believe throwing this message is appropriate.

src/Runner.Worker/ActionManager.cs
ethanchewy: Also, `compositeAction.Steps` is of type `List<Pipelines.ActionStep>`. If we were to output this, what would be the best way to output the values in a json string format?

Right now, this Trace.info statement returns the object string representation: `GitHub.DistributedTask.Pipelines.ActionStep`

If we were to print out the steps, it might make more sense to print them out in the `LoadCompositeSteps` function

src/Runner.Worker/StepsRunner.cs
ethanchewy: Tested: This is not needed.

src/Runner.Worker/ExecutionContext.cs
ethanchewy: Added TODO. Leaving open for visibility.

ethanchewy: After more testing for the env flow, I realized that the display name does not evaluate the env variables. This has to do with not passing the context. See: https://github.com/actions/runner/pull/557#issuecomment-647539455

Building in support will occur in the env flow PR: https://github.com/actions/runner/pull/557

ethanchewy: There are probably more reasons that we need context but that is just one reason ^

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
ethanchewy: Leaving open for visibility before merging.

src/Runner.Worker/ExecutionContext.cs
ethanchewy: In the env flow PR, I handle the contexts for each correctly: https://github.com/actions/runner/pull/557/files#diff-e3d54b1d0cbdce4d358272df29857cf2R225 by setting the step.EnvironmentVariables

Demo: https://github.com/ethanchewy/testing-actions/actions/runs/143774560

ethanchewy: Also, since we add each of our composite steps to JobSteps, eventually each of the steps will be run via the ScriptHandler. 

In the ScriptHandler, it exposes the context to the step/job: https://github.com/actions/runner/blob/master/src/Runner.Worker/Handlers/ScriptHandler.cs#L248-L258

src/Runner.Worker/ActionManager.cs
ethanchewy: From @TingluoHuang (thanks Ting!!) : 

> Verbose will shows up only in dev mode (you build the runner from source code.)
> Trace.Info($"Load {compositeAction.Steps.Count} action steps.");
> Trace.Verbose($"Details: {StringUtil.ConvertToJson(compositeAction.Steps}");

So, I'll do that.

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
ethanchewy: @ericsciple Wait, I should probably change `LoadCompositeSteps` to `LoadNestedSteps` since that is clearer?

src/Runner.Worker/ActionManifestManager.cs
ethanchewy: (Maybe add a more helpful error message + including file name, etc.  to show user that it's because of their yaml file)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -0,0 +1,97 @@
+using System.IO;
```
ericsciple, line 1: sort

ericsciple: system first

ericsciple: named at the bottom

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -266,6 +266,20 @@ public void RegisterPostJobStep(IStep step)
...
+        /// add a child node, aka a step, to the current job to the Root.JobSteps based on the location. 
+        /// </summary>
+        public void RegisterNestedStep(IStep step, DictionaryContextData inputsData, int location)
+        {
+            // TODO: For UI purposes, look at figuring out how to condense steps in one node => maybe use the same previous GUID
+            var newGuid = Guid.NewGuid();
+            step.ExecutionContext = Root.CreateChild(newGuid, step.DisplayName, newGuid.ToString("N"), null, null);
+            step.ExecutionContext.ExpressionValues["inputs"] = inputsData;
+            // TODO: confirm whether not copying message contexts is safe
+            Root.JobSteps.Insert(0, step);
```
ethanchewy, line 280: TODO: Replace 0 with location so that the steps are ordered.

src/Runner.Worker/action_yaml.json
```diff
@@ -83,6 +84,32 @@
...
+                "needs",
+                "strategy",
+                "matrix",
+                "secrets",
+                "steps",
+                "inputs",
+                "job",
+                "runner",
+                "env",
+                "hashFiles(1,255)"
```
ericsciple, line 107:
```
        "always(0,0)",
        "failure(0,0)",
        "cancelled(0,0)",
        "success(0,0)",
        "hashFiles(1,255)"
```

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
```diff
@@ -400,7 +426,8 @@ public class PipelineTemplateEvaluator
         private TemplateContext CreateContext(
             DictionaryContextData contextData,
             IList<IFunctionInfo> expressionFunctions,
-            IEnumerable<KeyValuePair<String, Object>> expressionState = null)
+            IEnumerable<KeyValuePair<String, Object>> expressionState = null,
+            bool setMissingContext = true)
```
ethanchewy, line 430: Maybe switch to false when going to server (and also switch the logic in the function to reflect this)

PR #552, opened 2020-06-16
Add Replace to WellKnownFunctions 

Evilweed:
This PR adds Function that allows replacing characters in string.

It is common practice to build and then run Docker Images using branch name in Docker Tag. The problem is that branch name can contain characters (for example: "/") that are not allowed in Docker Tags. 

Other CI providers provide ENV variable that contains normalized branch name (example: `BRANCH=feat/ch-1234/some-feature` is transformed to `BRANCH_SLUG=feat-ch-1234-some-feature`). And this normalized branch is later used in container name:

```
some-job:
  container: some-registry.com/rails:${{ env.BRANCH_SLUG }}-${{ github.sha }}
```

### DIFF
--- a/src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Replace.cs
+++ b/src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Replace.cs
@@ -1,0 +1,34 @@
+using System;
+
+namespace GitHub.DistributedTask.Expressions2.Sdk.Functions
+{
+    internal sealed class Replace : Function
+    {
+        protected sealed override Boolean TraceFullyRealized => false;
+
+        protected sealed override Object EvaluateCore(
+            EvaluationContext context,
+            out ResultMemory resultMemory)
+        {
+            resultMemory = null;
+            var value = Parameters[0].Evaluate(context);
+            var characterToReplace = Parameters[1].Evaluate(context);
+            var characterToReplaceWith = Parameters[2].Evaluate(context);
+
+            if (value.IsPrimitive)
+            {
+                value = value.ConvertToString();
+            }
+            if (characterToReplace.IsPrimitive)
+            {
+                characterToReplace = characterToReplace.ConvertToString();
+            }
+            if (characterToReplaceWith.IsPrimitive)
+            {
+                characterToReplaceWith = characterToReplaceWith.ConvertToString();
+            }
+
+            return String.IsNullOrEmpty(value) ? String.Empty : value.Replace(characterToReplace, characterToReplaceWith);
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


TingluoHuang:
@Evilweed thanks for your contribution, can you also create an issue for the feature request. we need to add this to the service side as well so you can also use the function on expression evaluated on sever-side, ex: Job name, etc.

@chrispat @ericsciple for this Feature request

Evilweed:
@TingluoHuang , sure! Just added :)

chrispat:
This seems great to me.  Not sure how we integrate these changes back to the service.

ericsciple:
sorry missed this before, just saw today...

Will look more tomorrow. Opens up an attack vector so will need to use the expression sdk to track memory (the format function does this). The unit tests are on the server, so it may make sense to add to that repo first (private) then copy here.

src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Replace.cs
```diff
@@ -0,0 +1,34 @@
...
+            out ResultMemory resultMemory)
+        {
+            resultMemory = null;
+            var value = Parameters[0].Evaluate(context);
+            var characterToReplace = Parameters[1].Evaluate(context);
+            var characterToReplaceWith = Parameters[2].Evaluate(context);
+
+            if (value.IsPrimitive)
+            {
+                value = value.ConvertToString();
```
ericsciple, line 20: converttostring is safe to always call (even if it's already a string) more sense to change to `var value = Parameters[0].Evaluate(context).ConvertToString()` above;

src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Replace.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (characterToReplace.IsPrimitive)
+            {
+                characterToReplace = characterToReplace.ConvertToString();
+            }
+            if (characterToReplaceWith.IsPrimitive)
+            {
+                characterToReplaceWith = characterToReplaceWith.ConvertToString();
+            }
+
+            return String.IsNullOrEmpty(value) ? String.Empty : value.Replace(characterToReplace, characterToReplaceWith);
```
ericsciple, line 31: we should follow the precedent of the equality operators and do ordinalignorecase

src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Replace.cs
ericsciple: or i guess if any of the values are not a string then could detect and either throw or convert to empty string... not sure which is better.

Issue #438, opened 2020-04-21
Compose GitHub actions as a single GitHub action


danpintara:
What I like from GitHub action is we can ensure that our action is up-to-date with available fixes in the upstream. E.g.: we can use `actions/checkout@v2` and the CI will still run with the latest version of `checkout@v2`.

Unfortunately, this facility can't be used while we're creating a GitHub action. I think it's better if we can compose multiple GitHub actions into a single GitHub action, or maybe just use one or two GitHub actions for the trivial tasks and continue the other in Javascript or Docker container.

This is an imaginary `action.yml` to serve as an example, maybe it's cool if we can do this:
```
name: 'imaginary-extension'
description: 'Imaginary extension about composing actions'
runs:
  using: composite
  steps:
    - name: Python 3
      uses: actions/setup-python@v1
      with:
        python-version: 3.8
    - name: Requirements
      run: pip install -r requirements.txt
```

TingluoHuang:
It feels like a template reuse scenario, so you can share a list of reusable steps with others.

@chrispat @ericsciple for an idea.

ericsciple:
Yep it's something we want to do. Currently behind other priorities.

danpintara:
Please tell me about the progress later. I'm interested with this feature.

hross:
We are picking up work here so keep an eye out for activity. We will update things as we go.

danpintara:
Really cool. Thank you!

pelotom:
This is our team's number 1 pain point with GH actions, would love to see something like this implemented, or alternatively support for YAML anchors (how we used to accomplish DRY with Travis).

vanppo:
> This is our team's number 1 pain point with GH actions, would love to see something like this implemented, or alternatively support for YAML anchors (how we used to accomplish DRY with Travis).

YAML anchors may not be good enough. For the action below, we can not share common steps between jobs by YAML anchors.
```yaml
test-job:
  steps:
    - checkout
    - cache
    - dependencies
    - test

build-job:
  steps:
    - checkout
    - cache
    - dependencies
    - build
```
Syntax below may be the resolution.
```yaml
defaults:
  steps:
    - checkout
    - cache
    - dependencies
jobs:
  test-job:
    - test
  build-job:
    - build
```

Enrico2:
A single step must be able to reference outputs of previous steps, and so in order to define a step in a centralized way, there must be a way to provide inputs to that step. 
[...]

PR #554, opened 2020-06-17
Composite Run Steps ADR

ethanchewy:
In this ADR, I propose building the first version of Composite Actions which will allow users to run multiple run steps with the support of inputs, outputs, environment, and context for use in any steps as well as the if, timeout-minutes, and the continue-on-error attributes for each Composite Action step. 

https://github.com/actions/runner/blob/users/ethanchewy/compositeADR/docs/adrs/0549-composite-run-steps.md

This feature will eventually lead us on our way to develop a full functioning Composite Steps feature that addresses https://github.com/actions/runner/issues/438 (i.e. allowing actions to compose other actions).

### DIFF
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -46,8 +46,7 @@ echo hello world 2
 echo hello world 3
 echo hello world 4
 ```
-We add a token called "composite" which allows our Runner code to process composite actions. By invoking "using: composite", our Runner code then processes the "steps" attribute, converts this template code to a list of steps, and finally runs each run step 
-sequentially. 
+We add a token called "composite" which allows our Runner code to process composite actions. By invoking "using: composite", our Runner code then processes the "steps" attribute, converts this template code to a list of steps, and finally runs each run step sequentially. If any step fails and there are no `if` conditions defined, the whole composite action job fails. 
 
 ### Inputs
 Example `user/test/composite-action.yml`:
@@ -221,13 +220,15 @@ If the time taken for any of the steps in combination or individually exceed the
 
 For reference, in the example above, if the composite step `foo1` takes 11 minutes to run, that step will fail but the rest of the steps, `foo1` and `foo2`, will proceed as long as their total runtime with the previous failed `foo1` action is less than the composite action's `timeout-minutes` (50 minutes). If the composite step `foo2` takes 51 minutes to run, it will cause the whole composite action job to fail. I
 
-The rationale behind this is that users can configure their steps with the `needs` attribute or `if` condition to conditionally set how steps rely on each other. Due to the additional capabilities that are offered with combining '=`timeout-minutes`, `needs`, and/or `if`, we wanted the `timeout-minutes` condition to be as dumb as possible and not effect other steps. 
+The rationale behind this is that users can configure their steps with the `if` condition to conditionally set how steps rely on each other. Due to the additional capabilities that are offered with combining `timeout-minutes` and/or `if`, we wanted the `timeout-minutes` condition to be as dumb as possible and not effect other steps. 
 
-### Needs
+[Usage limits still apply](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions?query=if%28%29#usage-limits)
+
 
 ### Continue-on-error
 
 **TODO: This continue-on-error condition implementation is up to discussion.** 
+For now, if `continue-on-error` is set to `true` for any of the composite action steps, the composite action job proceeds with the next step and ignores that failure. 
 
 ### Visualizing Composite Action in the GitHub Actions UI
 We want all the composite action's steps to be condensed into the original composite action node. 
@@ -244,3 +245,4 @@ Here is a visual represenation of the [first example](#Steps)
 
 
 ## Conclusion
+This ADR lays the framework for eventually supporting nested Composite Actions within Composite Actions. This ADR allows for users to run multiple run steps within a GitHub Composite Action with the support of inputs, outputs, environment, and context for use in any steps as well as the if, timeout-minutes, and the continue-on-error attributes for each Composite Action step. 
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -18,6 +18,8 @@ An important step towards meeting this goal is to build in functionality for act
 
 We don't want the workflow author to need to know how the internal workings of the action work. Users shouldn't know the internal workings of the composite action (for example, `default.shell` and `default.workingDir` should not be inherited from the workflow file to the action file). When deciding how to design certain parts of composite run steps, we want to think one logical step from the consumer.
 
+A composite action is treated as **one** individual job step. 
+
 
 ## Decision
 
@@ -114,12 +116,12 @@ Example `user/composite/action.yml`:
 
 ```yaml
 outputs:
-  random-number: 
-    description: "random number"
+  random-number: ${{ steps.random-number-generator.outputs.random-id }}
 runs:
   using: "composite"
   steps: 
-    - run: echo "::set-output name=random-number::$(echo $RANDOM)"
+    - id: random-number-generator
+      run: echo "::set-output name=random-id::$(echo $RANDOM)"
 ```
 
 Example Output:
@@ -129,9 +131,9 @@ Example Output:
 random-number 43243
 ```
 
-Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its parent using dot notation (ex `steps.foo.outputs.bar`).
+Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its parent using dot notation (ex `steps.foo.outputs.random-number`).
 
-Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `my-output`). For example, in the `workflow.yml` file, you can't run `foo.steps.my-step.my-output`.
+Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `random-id`). The reason why we are doing this is because we don't want to require the workflow author to know the internal workings of the composite action.
 
 ## Context
 
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -121,6 +121,8 @@ Each of the output variables from the composite action is viewable from the work
 
 Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `my-output`). For example, in the `workflow.yml` file, you can't run `foo.steps.my-step.my-output`.
 
+Our philosophy is we don't want the workflow author to need to know how the internal workings of the action work.
+
 ### Context
 Similar to the workflow file, the composite action has access to the [same context objects](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contexts) (ex: `github`, `env`, `strategy`). 
 
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -14,7 +14,7 @@ Customers want to be able to compose actions from actions (ex: https://github.co
 An important step towards meeting this goal is to build in functionality for actions where users can simply execute any number of steps. 
 
 ## Decision
-In this ADR, we support running multiple steps in an Action. In doing so, we build in support for mapping and flowing the inputs, outputs, and env variables (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables).
+**In this ADR, we only support running multiple steps in an Action.** In doing so, we build in support for mapping and flowing the inputs, outputs, and env variables (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables).
 
 ### Steps
 Example `user/composite/action.yml`
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -111,7 +111,7 @@ Example Output:
 hello Octocat
 oh hello Octocat
 ```
-Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its immediete parent.
+Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its parent using dot notation (ex `steps.foo.outputs.bar`).
 
 Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `my-output`). For example, in the `workflow.yml` file, you can't run `foo.steps.my-step.my-output`.
 
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -291,6 +291,13 @@ For now, if `continue-on-error` is set to `true` for any of the composite action
 
 Note, that since the composite action is not a workflow, it does not have jobs and thus it is not within scope at the moment to support something like `strategy` with `continue-on-error` as seen in this [example](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error).
 
+## Defaults
+TODO
+
+Shell & WorkingDir
+
+Initial thoughts: the composite action takes the workingDir from the workflow file and the shell for the composite action is set by default only by the composite action itself. 
+
 ## Visualizing Composite Action in the GitHub Actions UI
 We want all the composite action's steps to be condensed into the original composite action node. 
 
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -14,7 +14,7 @@ Customers want to be able to compose actions from actions (ex: https://github.co
 An important step towards meeting this goal is to build in functionality for actions where users can simply execute any number of steps. 
 
 ## Decision
-**In this ADR, we only support running multiple steps in an Action.** In doing so, we build in support for mapping and flowing the inputs, outputs, and env variables (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables).
+**In this ADR, we only support running multiple run steps in an Action.** In doing so, we build in support for mapping and flowing the inputs, outputs, and env variables (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables).
 
 ### Steps
 Example `user/composite/action.yml`
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -114,7 +114,9 @@ Example `user/composite/action.yml`:
 
 ```yaml
 outputs:
-  random-number: ${{ steps.random-number-generator.outputs.random-id }}
+  random-number: 
+    description: "Random number"
+    value: ${{ steps.random-number-generator.outputs.random-id }}
 runs:
   using: "composite"
   steps: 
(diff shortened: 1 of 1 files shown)


ethanchewy:
TODO: Add more relevant/exciting examples

SalenaHolmes:
Add example

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,258 @@
+TODO: Change file name to represent the correct PR number (PR is not created yet for this ADR)
+
+# ADR 0549: Composite Run Steps
```
SalenaHolmes, line 3: My traffic stream hope I enjoy the program

ethanchewy:
Thank you for your suggestion for the name `subflow`, I'll discuss it with the rest of the Runner devs to see if it's more appropriate than composite! We are very flexible with the name and we can change it fairly easily so if anyone else who is reading this thread wants to suggest names, feel free to do so. 

Also, yes we are open to community responses. 

The scope of this feature is simply to allow for **running multiple run steps in an action** which already has a requires some technical complexity and significant changes to the runner codebase (here are our beginning PRs for reference: https://github.com/actions/runner/issues/438, https://github.com/actions/runner/pull/557)

Eventually, we will want to create another feature which will allow us to call actions within composite actions (aka supporting nesting) like in https://github.com/actions/runner/issues/438 with the support of `using` etc. We want to take this small step first to ensure we don't introduce a huge number of code changes that will introduce unintended bugs for all actions.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,258 @@
...
+  - run: Server: ${{ env.SERVER }} 
+```
+
+**TODO: This if condition implementation is up to discussion.** 
+
+See the paragraph below for a rudimentary approach:
+
+The immediate if condition holds the most importance and only effects the current if condition. If there is no if condition for a step, it defaults to its parent's if condition (if that parent's if condition is not defined, it takes the grandparent's if condition, and so on)
+
+In the example above, the `always()` condition in the `foo` step applies to the composite action steps `foo2`, `foo3`, and `foo4`. But since `foo3` has an if condition defined, it overrides its parent condition so `foo3` will have an if condition `success()`. Let's say that the `foo2` step fails, then `foo3` will not run since a past step failed and its if condition is `succcess()`, but `foo4` will run because its if condition is inherited from `foo` which is `always()`.
```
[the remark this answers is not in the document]
ethanchewy: Thanks for the suggestion - I appreciate the example and reasoning behind your concern. 

I'll think about this and insert a more thorough response in this thread soon. 

Thanks again!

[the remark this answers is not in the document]
ethanchewy: @cybojenix I agree with most of your suggestions. I'm a little bit confused on the later part of your example. 

> ...
> # user/composite/action.yml
> steps:
>   - run: echo "just succeeding"
>   - run: echo "I will not run, as my current scope is succeeding"
>     if: success()
>   - run: exit 1
>   - run: echo "I will not run, as my current scope is now failing"
> ```

Why would
 ```
- run: echo "I will not run, as my current scope is succeeding
   if: success()
```
 not run in this situation? The previous composite action step does not error. 

Did you mean something like:
 ```
- run: echo "I will not run, as my current scope is succeeding
   if: failure()
```
?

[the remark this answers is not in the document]
cybojenix: Sorry, I accidentally added the "not". I've updated the example. 
It's there to demonstrate that within the composite action, the current status is "success", even though the parent's status is failing.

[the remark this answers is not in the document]
ethanchewy: @cybojenx 

I just incorporated your suggestions in the ADR and cited you: https://github.com/actions/runner/blob/users/ethanchewy/compositeADR/docs/adrs/0549-composite-run-steps.md#if-condition

Thank you again for taking time to make this suggestion. Really appreciate it!

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+jobs:
+  build:
+    runs-on: self-hosted
+    steps:
+    - id: job1
+      uses: actions/setup-python@v1
+    - id: job2
+      uses: actions/setup-node@v2
+    - uses: actions/checkout@v2
+      needs: [job1, job2]
```
thboop, line 40: I don't think we support a step level needs: https://github.com/actions/runner/blob/master/src/Sdk/DTPipelines/workflow-v1.0.json

Is this a new addition as a part of this adr?

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+
+Similar to the above logic, the environment variables will flow from the parent node to its children node. More concretely, whatever workflow/action calls a composite action, that composite action has access to whatever environment variables its caller workflow/action has. Note that the composite action can append its own environment variables or overwrite its parent's environment variables. 
+
+
+### If Condition
+
+Example `user/composite/action.yml`:
+```
+steps:
```
thboop, line 170: is this an `action.yml` or a `workflow.yml`?

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Continue-on-error
+
+**TODO: This continue-on-error condition implementation is up to discussion.** 
+For now, if `continue-on-error` is set to `true` for any of the composite action steps, the composite action job proceeds to the next step and ignores that failure. 
+
+Note, that since the composite action is not a workflow, it does not have jobs and thus it is not within scope at the moment to support something like `strategy` with `continue-on-error` as seen in this [example](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error).
+
+### Visualizing Composite Action in the GitHub Actions UI
+We want all the composite action's steps to be condensed into the original composite action node.
```
thboop, line 259: 🏅

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+
+For reference, in the example above, if the composite step `foo1` takes 11 minutes to run, that step will fail but the rest of the steps, `foo1` and `foo2`, will proceed as long as their total runtime with the previous failed `foo1` action is less than the composite action's `timeout-minutes` (50 minutes). If the composite step `foo2` takes 51 minutes to run, it will cause the whole composite action job to fail. I
+
+The rationale behind this is that users can configure their steps with the `if` condition to conditionally set how steps rely on each other. Due to the additional capabilities that are offered with combining `timeout-minutes` and/or `if`, we wanted the `timeout-minutes` condition to be as dumb as possible and not effect other steps. 
+
+[Usage limits still apply](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions?query=if%28%29#usage-limits)
+
+
+### Continue-on-error
```
thboop, line 251: There are a number of fields that can be populated for a step. Some of these are shared between `uses` and `runs`, and some are specific to one or the other (`shell` for `runs`, `with` for `uses`). Are we going to support overriding all of these by passing them into the composite action? `shell` and `working-directory`, can also be found in the job/workflow defaults defaults

It might be good to have a section on priority. As I understand it the order of precedence is:
- The value defined for an individual step in the action.yaml
- The value defined for the composite action in the workflow.yaml
- The value defined for the job defaults in the workflow.yaml
- The value defined for the workflow level defaults in the workflow.yaml

For example, can I do this?

Example `action.yml`
```
using: 'composite' 
steps: 
  - id: foo1
    run: env #only runs in powershell
    
```

Example `workflow.yml`:
```
steps: 
  - id: foo
    uses: user/composite@v1
    shell: powershell
```

We can talk about if, continue-on-error, timeout-minutes, and the like in that section as examples, rather then providing separate sections for them. Some of these values may also apply to the actual composite action as a whole as well (like timeout-minutes, if), so we can mention those also apply to the composite action rather then being just pass thru.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Example Output:
+```
+hello Octocat
+oh hello Octocat
+```
+Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its parent using dot notation (ex `steps.foo.outputs.bar`).
+
+Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `my-output`). For example, in the `workflow.yml` file, you can't run `foo.steps.my-step.my-output`.
```
thboop, line 122: Seems like a clean solution 👍 to have all the outputs concentrated on the composite action

ethanchewy:
Summary from meeting w/ @ericsciple @TingluoHuang @thboop :

- Additions to ADR
  - `workingDirectory` => workflow file has the highest precedence for workingDirectory which will allow users set their own workingDirectory when they run a composite action.
  - `Defaults` => TBD
  - `shell` => Should we allow users to overwrite the shell that is used in the action file? For now, we say that the composite action has the highest precedence for the shell.
  - `if` condition => Maybe shelve for later and get more community feedback first. Potential issue with original approach is the fact that users would have to attach always() in post steps in the Composite Action.

ethanchewy:
**For the community. If you are reading this thread and are interested in formulating how composite actions work, feel free to leave comments below!**

We are currently thinking about how the `if` conditions will apply to Composite Action Steps as well as how the `workingDirectory`, `Defaults`, and `shell` will work. See notes:  https://github.com/actions/runner/pull/554#issuecomment-648371856

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+
+Example `workflow.yml`:
+```
+...
+steps: 
+  - id: foo
+    uses: user/composite@v1
+    with:
+      your_name: "Octocat"
+  - run: echo hello ${{ steps.foo.inputs.your_name }} 2
```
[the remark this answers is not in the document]
ericsciple: +1 This should suffice for the workflow.yml content:

```yaml
steps: 
  - uses: user/composite@v1
    with:
      your_name: "Octocat"
```

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,275 @@
...
+Our philosophy is we don't want the workflow author to need to know how the internal workings of the action work.
+
+### Context
+Similar to the workflow file, the composite action has access to the [same context objects](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#contexts) (ex: `github`, `env`, `strategy`). 
+
+### Environment
+Example `user/composite/action.yml`:
+```
+using: 'composite' 
+env:
```
ericsciple, line 133: let's consider starting without `env` at this level in the action.yml (extra complexity)

also we should not allow the secrets context inside action.yml. Should pass secrets through.

ericsciple: also should we consider not allowing `env` _context_ inside action.yml

e.g. can't do the following inside action.yml:

```yaml
steps:
  - run: echo ${{ env.foo }}
```

...even though explicily set `env` flow to the processes.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,275 @@
...
+
+Similar to the above logic, the environment variables will flow from the parent node to its children node. More concretely, whatever workflow/action calls a composite action, that composite action has access to whatever environment variables its caller workflow/action has. Note that the composite action can append its own environment variables or overwrite its parent's environment variables. 
+
+
+### If Condition
+
+Example `user/composite/action.yml`:
```
ericsciple, line 170: we may want to explicitly punt on this for now and move this to a separate ADR or appendix

ericsciple: same for `continue-on-error`, `timeout-minutes`

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,275 @@
...
+Example Output:
+```
+hello Octocat
+oh hello Octocat
+```
+Each of the output variables from the composite action is viewable from the workflow file that uses the composite action. In other words, every child action output(s) is viewable only by its parent using dot notation (ex `steps.foo.outputs.bar`).
+
+Moreover, the output ids are only accessible within the scope where it was defined. Note that in the example above, in our `workflow.yml` file, it should not have access to output id (i.e. `my-output`). For example, in the `workflow.yml` file, you can't run `foo.steps.my-step.my-output`.
+
+Our philosophy is we don't want the workflow author to need to know how the internal workings of the action work.
```
ericsciple, line 124: consider moving this up to the summary, a section: `Guiding principles` a bullet list if that makes sense

abstraction, users shouldn't have to know the internal details (so for example `defaults.shell` and `defaults.working-directory` probably shouldn't be inherited)

inputs and outputs contract

ericsciple: Also one logical step from the perspective of the consumer.... so that influences how we think about `if` condition.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,273 @@
...
+## Decision
+In this ADR, we support running multiple steps in an Action. In doing so, we build in support for mapping and flowing the inputs, outputs, and env variables (ex: All nested steps should have access to its parents' input variables and nested steps can overwrite the input variables).
+
+### Steps
+Example `user/composite/action.yml`
+```
+...
+using: 'composite' 
+steps:
+  - run: pip install -r requirements.txt
```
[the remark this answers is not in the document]
ethanchewy: @thboop I like this idea!

I think for the scope of this ADR (only supporting multiple run steps), I can't think of a way to refactor the starter workflows since we aren't supporting `uses` in this ADR. 

I think for the next ADR where we support `uses`, that would be a good example. 

Any ideas on a better example for multiple steps in a composite action?

ethanchewy:
> Couple questions (please feel free to say that they're outside the scope of this ADR)
> 
> ### 1. Can a user publish composite actions to the marketplace?
> ### 2. Will there be any editor support for writing composite actions?
> Currently, we have a workflow editor which helps with autocomplete and writing workflows. With composite actions, we're now allowing a portion of a workflow to live in the `action.yml` file, so a similar editing experience would be worthwhile.

Those are great questions @joshmgross ! I haven't thought much about those areas. I think this is beyond the scope of this ADR. The plan for this ADR is to support running only multiple run steps in the composite action. This will lay the framework in the runner codebase to add additional features which will allow us to create the "fully functioning" Composite Action feature where users will be able to call/nest other actions within actions. 

I think after I finish implementing this ADR and writing up the corresponding documentation, I'll talk to the teams responsible for marketplace and the editor to build in support for composite run steps.

docs/adrs/0549-composite-run-steps.md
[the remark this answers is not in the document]
thboop: We can keep the original example then!

Nearly all of our starter workflows run `actions/checkout`, we could take one like this: https://github.com/actions/starter-workflows/blob/master/ci/c-cpp.yml and just have it run all the steps after checkout in the composite action, but I don't think we need to at the moment.

docs/adrs/0549-composite-run-steps.md
thboop: We spoke about this offline.

we aren't planning on passing through any of the values that are passed into the composite action directly to each step inside the composite action. As a design principle, we are hoping workflow authors do not need to understand the implementation of the composite action, but can use it as a black box to accomplish what they want. If there needs to be input from the workflow author inside a composite action, they can use the `inputs` feature

docs/adrs/0549-composite-run-steps.md
thboop: Agreed, I'd like the idea of starting with a more barebones implementation, and letting user feedback and scenarios drive how we can improve it going forward!

chrispat:
> ### 1. Can a user publish composite actions to the marketplace?

This should work today with no modifications.  The metadata that the marketplace looks at is present in the actions.yml.

> ### 2. Will there be any editor support for writing composite actions?

It would be good to do this eventually but I would say it is out of scope for this effort.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,321 @@
...
+        echo Server $SERVER 
+      env:
+        NAME2: test3
+```
+
+Example Output:
+
+```
+NAME2 test3
+Server development
```
cybojenix, line 176: According to this logic, workflow authors won't be able to override environment variables.
On the one hand, I can see how it would be useful for an action author to want full control over their scope, to avoid collisions (I can't think of an example off-hand). 
On the other, I can see it as useful for the action author to set a default, with the intention of it being overwritten (all our deploy scripts use the environment variable `STAGE` to control which environment to manage).

A good compromise here might be to allow the use of `input` within the top level `env`.

cybojenix: For additional clarity in the example, can we also explicitly say what happens when you specify `env` on the step itself. 
My assumption is that it will behave the same as above, though it could be a bit of a surprise to workflow authors when explicitly setting an env doesn't change what the action has set

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,321 @@
...
+  - id: foo
+    uses: user/test@v1
+  - run: echo Server $SERVER
+```
+
+Example `user/composite/action.yml`:
+
+```yaml
+using: 'composite' 
+env:
```
cybojenix, line 158: Will `env` be top level, or nested under `runs`, as it is when writing a docker action?

Does the behaviour also align with how `env` works for docker actions? (where the workflow can't override `runs.env`)

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,321 @@
...
+
+Similar to the above logic, the environment variables will flow from the parent node to its children node. More concretely, whatever workflow/action calls a composite action, that composite action has access to whatever environment variables its caller workflow/action has. Note that the composite action can append its own environment variables or overwrite its parent's environment variables. 
+
```
TingluoHuang, line 183: what about all the commands we have, ex: `::set-env::` will it only scope to the composite action, or still affect steps within the job?

ethanchewy: I think following our logic that the workflow author shouldn't know the internal workings of the composite action, we should say that `::set-env::` will only scope to the composite action.

docs/adrs/0549-composite-run-steps.md
```diff
@@ -0,0 +1,321 @@
...
+    - id: foo3
+      run: echo test 3
+      timeout-minutes: 10
+```
+
+**TODO: This timeout-minutes condition implementation is up to discussion.** 
+
+A composite action in its entirety is a job. You can set both timeout-minutes for the whole composite action or its steps as long as the the sum of the `timeout-minutes` for each composite action step that has the attribute `timeout-minutes` is less than or equals to `timeout-minutes` for the composite action. There is no default timeout-minutes for each composite action step. 
+
+If the time taken for any of the steps in combination or individually exceed the whole composite action `timeout-minutes` attribute, the whole job will fail. If an individual step exceeds its own `timeout-minutes` attribute but the total time that has been used including this step is below the overall composite action `timeout-minutes`, the individual step will fail but the rest of the steps will run.
```
TingluoHuang, line 282: ` the individual step will fail but the rest of the steps will run.` -> the rest of the steps will run base on condition.

docs/adrs/0549-composite-run-steps.md
cybojenix: Given that actions can set environment variables for the job, it feel like `set-env` should also work the same way for composite actions.

An usecase would be if someone wants to configure AWS credentials as part of their boilerplate for preparing a deploy.
This involves setting up environment variables.

The workaround here would be for the action author to output everything using `outputs`, and have the workflow author do `set-env` themselves.. which is a fair amount of boilerplate, especially that to do it properly you need to handle escaping.

mgudmund:
A question on this PR that’s important from an Enterprise perspective. Not sure if it’s the right PR though. 
We need composite actions like this to be internal. Will this ADR address this need? 
What we need is for us to share composite workflows across the entire organization, without being exposed publicly. And without PATs. I guess it requires a GITHUB_TOKEN with org level credentials. 

Is this anything that will be part of this work?

PR #568, opened 2020-06-25
Composite Run Steps Outputs

ethanchewy:
**In this PR, I add support for handling Outputs as described in https://github.com/actions/runner/blob/users/ethanchewy/compositeADR/docs/adrs/0549-composite-run-steps.md#outputs** Throughout this process, we remove unnecessary old code relating to the scope.
 
Demo: https://github.com/ethanchewy/testing-actions/runs/855647900?check_suite_focus=true

action.yml
```yaml
outputs:
  random_number: 
    description: "Random number"
    value: ${{ steps.random_number_generator.outputs.random_id }}
  food: 
    description: "Food"
    value: ${{ steps.food.outputs.test }}
runs:
    using: "composite"
    steps: 
      - id: random_number_generator
        run: echo "::set-output name=random_id::$(echo $RANDOM)"
      - id: food
        run: echo "::set-output name=test::strawberry"
```

workflow.yml

```yaml
jobs:
  build:
    runs-on: self-hosted
    steps:
    - uses: actions/checkout@v2
    - id: composite1
      uses: ethanchewy/test-composite@outputs
      with:
        your_name: "Octocat"
      env:
        NAME3: testasdfasdfads
    - name: test outputs
      run: | 
        echo test outputs random: ${{ steps.composite1.outputs.random_number}} 
        echo test outputs food: ${{ steps.composite1.outputs.food}} 
```

Goal: Handle Composite Run steps. 

Pretty large in scope.

To achieve this, we need to do:

- [x] Move Initialize Scope to Composite Action Handler
- [x] Generate Scope name for Composite Action Steps. Right now, by default there is no scope name passed to any instance of ExecutionContext in actions since we didn't have to handle `set-output` in steps since there were no steps. We need scope name for creating the right scopes for setting output variables.
- [x] Refactor code so that if the scope is null (aka if the step doesn't have an id), we generate a unique ID for it (in our case we do, "__GUIDSTRING"
- [x] Handle `outputs` token in action yaml file
- [x] Handle `::set-output::`
- [x] Create Composite Actions Outputs Handler to handle outputs.
  - [x] Create Composite actions output execution type
  - [x] Add Reference type to the action step that is used for cleaning up the outputs.
  - [x] Append a new step after adding all the composite action steps to handle outputs in 1 step 
      - [x] Add all necessary attributes for the step to trigger the Composite Action Handler. 

TODO 7/6/20: 
- [x] Clean up code
- [x] Refactor how we trigger the composite action output handler so that we don't use reference type... 

**Refactoring Info for those who are curious:** 
We return the same outputs execution data type as the Composite Action Output Handler
* But this time, we add an attribute to the ExecutionContext to trigger the right handler in ActionHandler
* Moreover, we share the same data, so the data will be passed along correctly 

In the Composite Action Output Handler, we refactor so that we evaluate the scope outputs in the same context as the StepsContext. Then, we go up a scope and set the outputs in the whole scope of the whole Composite Action.


ethanchewy:
TODO: For Handling the Outputs in the CompositeActionHandler, we need to make sure that we set the steps.<step_id>.outputs in the StepsContext for each step is configured correctly.

See debug for reference:
```
[2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Evaluating: steps.random-number-generator.outputs.random-id
4459 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Evaluating Index:
   1 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]..Evaluating Index:
   2 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]....Evaluating Index:
   3 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]......Evaluating steps:
   4 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]......=> null
   5 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]....=> null
   6 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]..=> null
   7 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]=> null
   8 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Result: null
   9 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Evaluating: steps.food.outputs.test
  10 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Evaluating Index:
  11 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]..Evaluating Index:
  12 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]....Evaluating Index:
  13 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]......Evaluating steps:
  14 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]......=> null
  15 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]....=> null
  16 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]..=> null
  17 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]=> null
  18 [2020-07-01 20:58:23Z VERB JobServerQueue] Enqueue web console line queue: ##[debug]Result: null
```

ethanchewy:
Basic outputs support done: https://github.com/ethanchewy/testing-actions/runs/843234904?check_suite_focus=true

Will need to refactor how we trigger our outputs handler. Also, will need to clean up outputs and consolidate scope code.

ethanchewy:
Finished. 

I also tested Errror Handling, and error messages seem to be helpful since it points the user to the correct file and line number: https://github.com/ethanchewy/testing-actions/actions/runs/160835569

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -79,6 +79,30 @@
       "string": {}
     },
 
+    "composite-outputs": {
```
ethanchewy, line 82: We also have to modify workflow.json because when we EvaluateCompositeOutputs in the outputs handler, the PipelineEvaluator uses workflow_1.0.json for evaluating tokens.

src/Runner.Worker/action_yaml.json
```diff
@@ -120,6 +128,24 @@
...
+                "secrets",
+                "steps",
+                "inputs",
+                "job",
+                "runner",
+                "env"
+            ],
+            "one-of": [
+                "string",
+                "mapping"
```
ethanchewy, line 146: we add mapping as well as to support something like this which follows the same format as container's outputs:
```
outputs:
 time: # id of output
   description: 'The time we did the greeting'
```

ethanchewy:
Fixed unit tests issue - ready for review. (See comments for more info about design decisions)

ericsciple:
Since existing schema....

Let's do the below for now, we can follow-up with product later. I dont want to break existing customers ([docs](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action#creating-an-action-metadata-file)). We can iterate and change before we release Composite Actions.

```yaml
outputs:
  time:
    description: string
    value: string (expression allowed)
```

if not composite action... dont allow "value".

src/Runner.Worker/Handlers/CompositeActionOutputHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        public CompositeActionExecutionData Data { get; set; }
+        public Task RunAsync(ActionRunStage stage)
+        {
+            // Evaluate the mapped outputs value
+            if (Data.Outputs != null)
+            {
+                // Evaluate the outputs in the steps context to easily retrieve the values
+                var evaluator = ExecutionContext.ToPipelineTemplateEvaluator();
+
+                DictionaryContextData actionOutputs = evaluator.EvaluateCompositeOutputs(Data.Outputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
```
ericsciple, line 33: Because of the back compat thing... Be careful to handle the structure, and only add the "value"

```
time:
  description: asdf
  value: the value
```

src/Sdk/DTPipelines/workflow-v1.0.json
ethanchewy: In the CompositeOutputHandler, 

we do:
```
var evaluator = ExecutionContext.ToPipelineTemplateEvaluator();
                DictionaryContextData actionOutputs = evaluator.EvaluateCompositeOutputs(Data.Outputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
```

This ExecutionContext will generate a template evaluator using the workflow yaml since in PipelineTemplateSchemaFactory, GetSchema will always return workflow yaml:
```
public static TemplateSchema GetSchema()
        {
            if (s_schema == null)
            {
                var assembly = Assembly.GetExecutingAssembly();
                var json = default(String);
                using (var stream = assembly.GetManifestResourceStream("GitHub.DistributedTask.Pipelines.ObjectTemplating.workflow-v1.0.json"))
                using (var streamReader = new StreamReader(stream))
                {
                    json = streamReader.ReadToEnd();
                }

                var objectReader = new JsonObjectReader(null, json);
                var schema = TemplateSchema.Load(objectReader);
                Interlocked.CompareExchange(ref s_schema, schema, null);
            }

            return s_schema;
        }
```

ethanchewy: I tried moving the code for GetSchema over to the CompositeActionOutputHandler but Assembly.GetExecutingAssembly() won't get the right directory for loading "GitHub.Runner.Worker.action_yaml.json"

Another possibility is to store the _actionManifestSchema (from ActionManifestManager) as a static variable in the ExecutionContext but since the ExecutionContext in the output handler is IExecutionContext, you can't access it unfortunately. 

I think it makes the most amount of sense to refactor this during abstraction.

ethanchewy: or we could just store it in our ExecutionContext via LoadAction

ethanchewy:
Will refactor some of the evaluator code so that we don't have to copy over action_yaml.json to composite_action.yaml

ethanchewy:
Pushed another commit to prevent duplication of action_yaml outputs attributes into workflow outputs. See https://github.com/actions/runner/pull/568/files#r452379009 for reference. 

We'll be able to clean this up more when we start working on abstraction of ExecutionContext.

src/Runner.Worker/ActionManager.cs
```diff
@@ -400,6 +400,8 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                         var compositeAction = definition.Data.Execution as CompositeActionExecutionData;
                         Trace.Info($"Load {compositeAction.Steps.Count} action steps.");
                         Trace.Verbose($"Details: {StringUtil.ConvertToJson(compositeAction.Steps)}");
+                        Trace.Info($"Load: {compositeAction.Outputs} outputs");
```
ericsciple, line 403: does this print anything useful, or does it print "object"

src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -56,6 +56,9 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
             var templateContext = CreateContext(executionContext);
             ActionDefinitionData actionDefinition = new ActionDefinitionData();
 
+            // Set ActionManifestSchema so that we can use it for post processing in Composite Actions
+            executionContext.ActionManifestSchema = templateContext.Schema;
```
ericsciple, line 60: instead of adding to execution context, lets expose from IActionManifestManager

Consumer can access like:

```csharp
var actionManifestManager = HostContext.GetService<IActionCommandManager>().ActionManifestSchema;
actionManifestManager.ActionManifestSchema;
```

src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -108,8 +122,9 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
                             break;
 
                         case "runs":
-                            actionDefinition.Execution = ConvertRuns(executionContext, templateContext, actionPair.Value);
+                            actionDefinition.Execution = ConvertRuns(executionContext, templateContext, actionPair.Value, actionOutputs);
```
ericsciple, line 125: Let's store `runs` (i.e. `actionPair.Value`) and then call `ConvertRuns` outside of the for loop. Otherwise this adds a constraint that the order matters. Order shouldn't matter for json (for yaml :shrug: but lets make it not matter).

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -70,6 +69,10 @@ public interface IExecutionContext : IRunnerService
 
         bool EchoOnActionCommand { get; set; }
 
+        IExecutionContext FinalizeContext { get; set; }
```
ericsciple, line 72: we'll be able to clean this up when the handler performs the steps-runner-ish things

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -270,17 +277,37 @@ public void RegisterPostJobStep(IStep step)
         /// Helper function used in CompositeActionHandler::RunAsync to
         /// add a child node, aka a step, to the current job to the Root.JobSteps based on the location. 
         /// </summary>
-        public void RegisterNestedStep(IStep step, DictionaryContextData inputsData, int location, Dictionary<string, string> envData)
+        public IStep RegisterNestedStep(
+            IActionRunner step, DictionaryContextData inputsData, int location,
```
ericsciple, line 281: nit: lets format all on one liner or all separate

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -105,7 +108,7 @@ public interface IExecutionContext : IRunnerService
         // others
         void ForceTaskComplete();
         void RegisterPostJobStep(IStep step);
-        void RegisterNestedStep(IStep step, DictionaryContextData inputsData, int location, Dictionary<string, string> envData);
+        IStep RegisterNestedStep(IActionRunner step, DictionaryContextData inputsData, int location, Dictionary<string, string> envData, Boolean cleanUp = false);
```
ericsciple, line 111: nit: `bool` not `Boolean`

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -270,17 +277,37 @@ public void RegisterPostJobStep(IStep step)
...
+            Dictionary<string, string> envData,
+            Boolean cleanUp = false)
         {
             // TODO: For UI purposes, look at figuring out how to condense steps in one node => maybe use the same previous GUID
             var newGuid = Guid.NewGuid();
-            step.ExecutionContext = Root.CreateChild(newGuid, step.DisplayName, newGuid.ToString("N"), null, null);
+
+            // Set Scope Name. Note, for our design, we consider each step in a composite action to have the same scope
+            // This makes it much simpler to handle their outputs at the end of the Composite Action
+            var childScopeName = !string.IsNullOrEmpty(this.ScopeName) ? $"{this.ScopeName}.{this.ContextName}" : this.ContextName;
```
ericsciple, line 290: i think there's a bug here for the future (nested composite actions)

Should generate `__<GUID>` only when context name is empty.

ericsciple:
```
var safeContextName = !string.IsNullOrEmpty(ContextName) ? ContextName : $"__{newGuid}";
var childScopeName = !string.IsNullOrEmpty(this.ScopeName) ? $"{this.ScopeName}.{safeContextName}" : safeContextName;
```

is this ^ right? i'm getting confused now because childContextName is passed to Root.CreateChild and i dont know what that's for....


OK i just realized i'm way confused b/c i was getting confused whether in the for loop yet or not... i'll leave these notes here and will think through again later.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -466,7 +496,10 @@ public void SetOutput(string name, string value, out string reference)
         {
             ArgUtil.NotNullOrEmpty(name, nameof(name));
 
-            if (String.IsNullOrEmpty(ContextName))
+            // Checks if scope name is null or 
+            // if the ScopeName follows the __GUID format which is set as the default value for ScopeNames if null for Composite Actions. 
+            bool scopeNameCondition = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")) && !String.IsNullOrEmpty(ScopeName) && ScopeName.Length >= 36 && String.Equals(ScopeName.Substring(0, 2), "__") && Guid.TryParse(ScopeName.Substring(2, 36), out Guid test);
```
ericsciple, line 501: consider static regex precompiled

src/Runner.Worker/ActionManager.cs
TingluoHuang: `compositeAction.Outputs.Count`

src/Runner.Worker/ExecutionContext.cs
ericsciple:
```
using System.Text.RegularExpressions;
[...]
private static readonly s_generatedContextNamePattern = new Regex("^__[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$", RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
```

Then you can do "_generatedContextNamePattern.IsMatch( )"

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -633,16 +666,6 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             // Steps context (StepsRunner manages adding the scoped steps context)
             StepsContext = new StepsContext();
 
-            // Scopes
```
ericsciple, line 636: i love seeing dead code deleted :)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -47,6 +54,12 @@ public Task RunAsync(ActionRunStage stage)
 
             // Add each composite action step to the front of the queue
             int location = 0;
+
+            Dictionary<string, string> scopesAndContexts = new Dictionary<string, string>();
```
ericsciple, line 58: this local variable isnt used

src/Runner.Worker/Handlers/CompositeActionOutputHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+                // }
+                foreach (var pair in actionOutputs)
+                {
+                    var outputsName = pair.Key;
+                    var outputsAttributes = pair.Value as DictionaryContextData;
+                    outputsAttributes.TryGetValue("value", out var val);
+                    var outputsValue = val as StringContextData;
+
+                    // Set output in the whole composite scope. 
+                    if (!String.IsNullOrEmpty(outputsName) && !String.IsNullOrEmpty(outputsValue))
```
ericsciple, line 67: nit: `string` not `String`

src/Runner.Worker/StepsRunner.cs
```diff
@@ -83,171 +82,170 @@ public async Task RunAsync(IExecutionContext jobContext)
                 step.ExecutionContext.ExpressionFunctions.Add(new FunctionInfo<SuccessFunction>(PipelineTemplateConstants.Success, 0, 0));
                 step.ExecutionContext.ExpressionFunctions.Add(new FunctionInfo<HashFilesFunction>(PipelineTemplateConstants.HashFiles, 1, byte.MaxValue));
 
-                // Initialize scope
-                if (InitializeScope(step, scopeInputs))
-                {
-                    // Populate env context for each step
-                    Trace.Info("Initialize Env context for step");
+                step.ExecutionContext.ExpressionValues["steps"] = step.ExecutionContext.StepsContext.GetScope(step.ExecutionContext.ScopeName);
```
ericsciple, line 85: What's different about this file?

`if (InitializeScope[...]` was removed and the different indent-level is causing too much red/green. I can't tell what changed.

I see the InitializeScope method was deleted. Anything else?

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
```diff
@@ -183,6 +183,33 @@ public class PipelineTemplateEvaluator
             return result;
         }
 
+        public DictionaryContextData EvaluateCompositeOutputs(
```
ericsciple, line 186: This should be in ActionManifestManager.cs instead. For example see `EvaluateDefaultInput`

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConstants.cs
```diff
@@ -15,6 +16,7 @@ public sealed class PipelineTemplateConstants
         public const String Clean = "clean";
         public const String Container = "container";
         public const String ContinueOnError = "continue-on-error";
+        public const String CompositeOutputs = "composite-outputs";
```
ericsciple, line 19: revert this file (see comment futher below)

src/Runner.Worker/ActionManager.cs
ethanchewy: Good catch - thanks!

src/Runner.Worker/StepsRunner.cs
ethanchewy: @ericsciple  InitializeScope was completely removed.

Also, most of the CompleteStep was removed so that it looks like. All of the code removed related to the "Scopes" attribute of the ExecutionContext which is no longer used afaik:
```
private void CompleteStep(IStep step, TaskResult? result = null, string resultCode = null)
{
  var executionContext = step.ExecutionContext;
}
``

see: https://github.com/actions/runner/pull/568/files#diff-4fa5184d693636c90a417707cf39d422L468-L522

File: src/Runner.Worker/Handlers/CompositeActionOutputHandler.cs [hunk 18]
```diff
@@ -0,0 +1,77 @@
...
+            {
+                // We have to have this function because ExecutionContext.ToPipelineTemplateEvaluator calls
+                // PipelineTemplateSchemaFactory.GetSchema() which always uses workflow_v1.0.json 
+                // So, we have to manually create the evaluator
+                var evaluator = new Pipelines.ObjectTemplating.PipelineTemplateEvaluator(ExecutionContext.ToTemplateTraceWriter(), ExecutionContext.ActionManifestSchema, ExecutionContext.FinalizeContext.FileTable);
+
+                // Evaluate the outputs in the steps context to easily retrieve the values
+                DictionaryContextData actionOutputs = evaluator.EvaluateCompositeOutputs(Data.Outputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
+
+                // Each pair is structured like this
```
ericsciple, line 41: Since comment intended for conceptual understanding of the structure, and not how the object is serialized, consider plain json dictionaries to describe the structure instead

```
{
  "the-output-name": {
    "description": "",
    "value": "the value"
  },
  ...
}
```

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConstants.cs
TingluoHuang: +1, otherwise you will need to port this back to server code

src/Runner.Worker/action_yaml.json
```diff
@@ -1,4 +1,6 @@
 ﻿{
+    "version": "action_v1.0",
```
TingluoHuang, line 2: do we need to version this file for now?

src/Runner.Worker/Handlers/CompositeActionOutputHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+            // Evaluate the mapped outputs value
+            if (Data.Outputs != null)
+            {
+                // We have to have this function because ExecutionContext.ToPipelineTemplateEvaluator calls
+                // PipelineTemplateSchemaFactory.GetSchema() which always uses workflow_v1.0.json 
+                // So, we have to manually create the evaluator
+                var evaluator = new Pipelines.ObjectTemplating.PipelineTemplateEvaluator(ExecutionContext.ToTemplateTraceWriter(), ExecutionContext.ActionManifestSchema, ExecutionContext.FinalizeContext.FileTable);
+
+                // Evaluate the outputs in the steps context to easily retrieve the values
+                DictionaryContextData actionOutputs = evaluator.EvaluateCompositeOutputs(Data.Outputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
```
ethanchewy, line 39: Consider moving this to ActionManifestManager

src/Runner.Worker/action_yaml.json
```diff
@@ -120,6 +137,21 @@
             ],
             "string": {}
         },
+        "composite-output-value": {
+            "context": [
```
TingluoHuang, line 141: we might want to double-check what is allowed.
Ex: as a composite action author, how do i use `needs`, `secrets` and `steps`?

src/Runner.Worker/Handlers/CompositeActionOutputHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    var outputsName = pair.Key;
+                    var outputsAttributes = pair.Value as DictionaryContextData;
+                    outputsAttributes.TryGetValue("value", out var val);
+                    var outputsValue = val as StringContextData;
+
+                    // Set output in the whole composite scope. 
+                    if (!String.IsNullOrEmpty(outputsName) && !String.IsNullOrEmpty(outputsValue))
+                    {
+                        ExecutionContext.FinalizeContext.SetOutput(outputsName, outputsValue, out string test);
```
ericsciple, line 69: use discards: `ExecutionContext.FinalizeContext.SetOutput(outputsName, outputsValue, out _);`

src/Runner.Worker/action_yaml.json
```diff
@@ -7,7 +9,8 @@
                     "name": "string",
                     "description": "string",
                     "inputs": "inputs",
-                    "runs": "runs"
+                    "runs": "runs",
+                    "outputs": "composite-outputs"
```
TingluoHuang, line 13: is `composite-outputs` or just `outputs` even it only matter for composite actions for now?

src/Runner.Worker/ActionManager.cs
ericsciple: beware null ref

TingluoHuang: `compositeAction.Outputs?.Count ?? 0`

src/Runner.Worker/action_yaml.json
TingluoHuang: also, remove `secrets`, `needs`, all the condition func from `composite-steps`

ethanchewy: tl;dr clean up secrets + needs from composite-steps and compomsite-output-value.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -23,6 +23,13 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
     {
         public CompositeActionExecutionData Data { get; set; }
 
+        private void InitializeScope(IStep step)
```
TingluoHuang, line 26: move `private` method down after `public` method

PR #584, opened 2020-07-10
(In progress) Refactor Codebase to Support Modular Composite Actions 

ethanchewy:
In this PR, we plan to support **recognizing any** type of steps in a composite action. 

We are going to accomplish this by doing:

- [ ] Refactoring StepsRunner to ensure that it's only used for Job Level Steps
- [ ] Create a separate StepsRunner for only Composite Actions
- [ ] Modify existing code to parse through every type of step
  - [ ] Modify LoadCompositeSteps in PipelineTemplate Evaluator to support any step
  - [ ] Change for loop in composite actions handler to handle every type of step and create the "correct" child node with the right context attributes to trigger the correct handler!

- [ ] Refactor Execution Context
  - [ ] Global Execution Context
  - [ ] Job Steps Execution Context
  - [ ] Composite Execution Context
** Let's first start with seperating the composite stuff from the execution context and packaging that into the composite execution context. 

Additional changes that we might do in this PR or another PR depending on complexity:

1. Refactor ExecutionContext => Global => Job, Action => Composite Action

High level view: https://docs.google.com/document/d/1euUp55hp0CcAuaoATDieVvJMUTFbHc6zveNaTEKIY8A/edit#

### DIFF
--- a/src/Runner.Worker/ActionManifestManager.cs
+++ b/src/Runner.Worker/ActionManifestManager.cs
@@ -376,6 +376,7 @@ namespace GitHub.Runner.Worker
                         {
                             var steps = run.Value.AssertSequence("steps");
                             var evaluator = executionContext.ToPipelineTemplateEvaluator();
+                            // TODO: Change this so that we process each type of step
                             stepsLoaded = evaluator.LoadCompositeSteps(steps);
                             break;
                         }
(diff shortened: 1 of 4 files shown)


ethanchewy:
Actually, we would need to make another ExecutionContext for our Composite Actions StepsRunner to make sense as well. Hmmmmm... I'll try to do both refactorings at the same time and see how it goes. 

If it becomes to large, I'll try to seperate the logic.

ethanchewy:
Yup, gonna create another PR based on the master branch for the Global Execution Context refactoring. That makes the most sense.

ethanchewy:
Closing. Will create another branch since we elected to choose a different route for refactoring.

PR #585, opened 2020-07-12
Add an environment variable to indicate which repository the currently running Action came from.

chrisgavin:
For certain types of Actions, it can be useful for the Action to know what repository it came from. For example the [github/codeql-action](https://github.com/github/codeql-action/) needs to access binary files stored in GitHub Releases artifacts on its own repository.

Currently if this Action is mirrored on GitHub Enterprise Server under a different name it will be unable to locate its release artifacts as it always expects them to be in the `github/codeql-action` repository.

This change adds a `github.action_repository` context variable (and corresponding `GITHUB_ACTION_REPOSITORY` environment variable) that is set to the NWO of the repository which the Action came from. It is not set for script or Container Registry Actions. It is set to the empty string if the Action reference is relative to the checkout repository.

This allows Actions to be aware of their own source repository and therefore reach back out to them to do things like downloading extra artifacts.

Apologies if I've missed an existing way of getting this information, but I couldn't find one other than looking at the path which the Action is checked out to; I wasn't sure that this was a stable interface to rely on.


src/Runner.Worker/StepsRunner.cs
```diff
@@ -120,6 +120,11 @@ public async Task RunAsync(IExecutionContext jobContext)
                     {
                         // Set GITHUB_ACTION
                         step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
+                        // Set GITHUB_ACTION_REPOSITORY if this Action is from a repository
+                        if (actionStep.Action.Reference is Pipelines.RepositoryPathReference repositoryReference)
+                        {
+                            step.ExecutionContext.SetGitHubContext("action_repository", repositoryReference.Name);
```
robertbrignull, line 126: I notice the `RepositoryPathReference` also has a `Ref` field. Would it be ok to also make that available, perhaps as a `action_ref` var? That would be very useful for code scanning as we've been wanting to include that data in our status reports, but couldn't work out how to get it. Doesn't have to be in this PR if the extra change would slow it down.

chrisgavin: Yes, sure. I've added that field as well.

chrisgavin:
@joshhale Do you know what the best way is to get this reviewed?

joshhale:
@TingluoHuang do you have time to review this PR?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -119,6 +119,12 @@ public async Task RunAsync(IExecutionContext jobContext)
                 {
                     // Set GITHUB_ACTION
                     step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
+                    // Set GITHUB_ACTION_REPOSITORY if this Action is from a repository
+                    if (actionStep.Action.Reference is Pipelines.RepositoryPathReference repositoryReference)
+                    {
+                        step.ExecutionContext.SetGitHubContext("action_repository", repositoryReference.Name);
+                        step.ExecutionContext.SetGitHubContext("action_ref", repositoryReference.Ref);
+                    }
```
TingluoHuang, line 127: I would add these into `src/Runner.Worker/ActionRunner.cs` after line 137
```C#
            // adds the GitHub event path/file if the event exists
            if (gitHubEvent != null)
            {
                var workflowFile = Path.Combine(workflowDirectory, "event.json");
                Trace.Info($"Write event payload to {workflowFile}");
                File.WriteAllText(workflowFile, gitHubEvent, new UTF8Encoding(false));
                ExecutionContext.SetGitHubContext("event_path", workflowFile);
            }

            // Set GITHUB_ACTION_REPOSITORY if this Action is from a repository
                    if (Action.Reference is Pipelines.RepositoryPathReference repositoryReference)
                    {
                        ExecutionContext.SetGitHubContext("action_repository", repositoryReference.Name);
                        ExecutionContext.SetGitHubContext("action_ref", repositoryReference.Ref);
                    }
```

src/Runner.Worker/GitHubContext.cs
```diff
@@ -9,6 +9,8 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
         private readonly HashSet<string> _contextEnvWhitelist = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
         {
             "action",
+            "action_ref",
```
TingluoHuang, line 12: @chrispat any objection to adding these 2 new environment variables?

chrispat (approved):

src/Runner.Worker/StepsRunner.cs
TingluoHuang: @joshhale do you have a chance to address this comment?

src/Runner.Worker/ActionRunner.cs
```diff
@@ -135,6 +135,13 @@ public async Task RunAsync()
                 ExecutionContext.SetGitHubContext("event_path", workflowFile);
             }
 
+            // Set GITHUB_ACTION_REPOSITORY if this Action is from a repository
+            if (Action.Reference is Pipelines.RepositoryPathReference repoPathReferenceAction)
```
TingluoHuang, line 139: what's the value would be if i use the action from the checkout repo.

ex:
```
- use: ./localaction
  with:
      foo: bar
```

chrisgavin: It looks like currently they are both set to the empty string. Do you think that's reasonable or should they just not be set at all in that case?

chrisgavin:
It looks like retargeting from `master` to `main` has brought in the commit a711bd94 (it's on `master` but not on `main`). Would you like me to rebase to take this commit out?

src/Runner.Worker/ActionRunner.cs
TingluoHuang: maybe just not set.
```
if (Action.Reference is Pipelines.RepositoryPathReference repoPathReferenceAction && 
    !string.Equals(repoPathReferenceAction.RepositoryType, Pipelines.PipelineConstants.SelfAlias, StringComparison.OrdinalIgnoreCase))
{
 // set action_repo/action_ref
}
```

TingluoHuang:
@chrisgavin 

> It looks like retargeting from `master` to `main` has brought in the commit [a711bd9](https://github.com/actions/runner/commit/a711bd9494e2a3e00fcac3f8a685cf1f30572026) (it's on `master` but not on `main`). Would you like me to rebase to take this commit out?

it fine to include that commit. :)

TingluoHuang (approved):

PR #591, opened 2020-07-13
Composite Run Steps Refactoring

ethanchewy:
**In this PR, we consolidate the Composite Output Handler and the StepsRunner code all into the composite action handler as well as remove as much code as possible for ExecutionContext!**

Previously, we built another steps runner for processing composite actions but that made us realized we actually don't need one and we can fit all the StepsRunner logic in the handler! 

This provides a variety of benefits. Namely, 
- Easy status resolution for step progress
- Separated logic for processing composite steps vs. regular action steps. 
- Easy to build step specific logic for an individual composite step logic like continue-on-error 

**Our main goal is to 1) refactor codebase so that Composite Actions has its own StepsRunner 2) Make sure this works with our previous built in functionalities for composite run steps.**

In another PR, we'll build off of this PR to add on additional support for continue-on-error and timeout-minutes. After that, we'll start building support for nested actions of _all_ types. We can also explore abstracting Execution Context for Composite Actions.

Based off of exploration here: https://github.com/actions/runner/pull/584

Previous composite changes for reference: https://github.com/actions/runner/commits?author=ethanchewy

One tangible benefit you can directly see from this PR, is that the UI shows the nested steps proceeding in order in an intuitive fashion now: https://drive.google.com/file/d/1JzAWi0qOHQSOK-fBMp8Q_AFwPkkRqvDl/view?usp=sharing

Old ScreenRecording for Reference (We don't update the timeline name for every step in the newest PR version):
Screen-Recording-2020-07-14-at-5


ethanchewy:
Current Progress on how UI Shows Composite Steps (each step is delayed for 5 secs):
https://drive.google.com/file/d/1-NrI8M1SoMKwS0y7LAl3JgmafzeVkapU/view?usp=sharing

ethanchewy:
experimenting to see if I can get my new commits to show in this PR...

src/Runner.Worker/StepsRunner.cs
```diff
@@ -92,26 +92,11 @@ public async Task RunAsync(IExecutionContext jobContext)
                 var envContext = new CaseSensitiveDictionaryContextData();
 #endif
 
-                // Global env
                 foreach (var pair in step.ExecutionContext.EnvironmentVariables)
```
ethanchewy, line 95: These changes are to revert the workaround we used to flow Env to composite actions. This is now handled in the Composite Steps Runner.

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                    foreach (var pair in dict)
+                    {
+                        envContext[pair.Key] = pair.Value;
+                    }
+                }
+
+                step.ExecutionContext.ExpressionValues["env"] = envContext;
+
+
+
```
ericsciple, line 81: remove extra line breaks

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                step.ExecutionContext.Error(ex);
+            }
+            if (timeoutMinutes > 0)
+            {
+                var timeout = TimeSpan.FromMinutes(timeoutMinutes);
+                step.ExecutionContext.SetTimeout(timeout);
+            }
+
+#if OS_WINDOWS
+            try
```
ericsciple, line 151: Do we need to do this? Wouldn't steps-runner have already attempted this?

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                    step.ExecutionContext.Outcome = step.ExecutionContext.Result;
+                    step.ExecutionContext.Result = TaskResult.Succeeded;
+                    Trace.Info($"Updated step result (continue on error)");
+                }
+            }
+            Trace.Info($"Step result: {step.ExecutionContext.Result}");
+
+            // Complete the step context.
+            step.ExecutionContext.Debug($"Finishing: {step.DisplayName}");
+
```
ericsciple, line 246: extra newline

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        Task RunAsync(IExecutionContext Context);
+    }
+
+
+    public sealed class CompositeStepsRunner : RunnerService, ICompositeStepsRunner
+    {
+        public async Task RunAsync(IExecutionContext actionContext)
+        {
+            // Another approach we can explore, is moving all this logic to the CompositeActionHandler if it's small enough.
```
ericsciple, line 32: +1 to moving into CompositeActionHandler since that approach would eliminate the need to add a `CompositeSteps` property to ExecutionContext

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                }
+
+                step.ExecutionContext.ExpressionValues["env"] = envContext;
+
+
+
+                if (step is IActionRunner actionStep)
+                {
+                    // Set GITHUB_ACTION
+                    step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
```
ericsciple, line 85: We need to figure out what the correct value to set here is. I think this should be the fully qualified context name - i.e. "{scopeName}.{contextName}"

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -72,6 +72,9 @@ public interface IExecutionContext : IRunnerService
 
         IExecutionContext FinalizeContext { get; set; }
 
+        // Only Composite Level ExecutionContext has CompositeSteps (public setter for handler)
+        List<IStep> CompositeSteps { get; set; }
```
TingluoHuang, line 76: Any reason we have to store `CompositeSteps` as part of `ExecutionContext`?
The reason I store the JobStep is for register `post` steps.
what about `CompositeStepsRunner. RunAsync(ExecituonContext context, IStep compositeSteps)`, since you know what's going to run up front (for now. 😄 ).

src/Runner.Worker/CompositeStepsRunner.cs
ericsciple: And we need to fix this to be "{contextName}" for regular steps. I am working on refactoring the context name generation code so we can kill legacy ref name.

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                            envContext[env.Key] = new StringContextData(env.Value ?? string.Empty);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        // fail the step since there is an evaluate error.
+                        Trace.Info("Caught exception in Composite Steps Runner from expression for step.env");
+                        // evaluateStepEnvFailed = true;
+                        step.ExecutionContext.Error(ex);
+                        // CompleteStep(step, TaskResult.Failed);
```
ericsciple, line 103: Curious why this is commented out. I think this should fail the step. Thoughts?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -48,6 +48,12 @@ public Task RunAsync(ActionRunStage stage)
             // Add each composite action step to the front of the queue
             int location = 0;
```
TingluoHuang, line 49: do you need `location` anymore?
also, you don't need `tempDirectory`

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+            }
+#endif
+
+            try
+            {
+                await step.RunAsync();
+            }
+            catch (OperationCanceledException ex)
+            {
+                if (step.ExecutionContext.CancellationToken.IsCancellationRequested)
```
ericsciple, line 191: do we need to distinguish between this step timing-out vs the outer step timing-out?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -102,9 +111,22 @@ public Task RunAsync(ActionRunStage stage)
             actionRunner2.Action = cleanOutputsStep;
             actionRunner2.Stage = ActionRunStage.Main;
             actionRunner2.Condition = "always()";
-            ExecutionContext.RegisterNestedStep(actionRunner2, inputsData, location, Environment, true);
+            var cleanUpStep = ExecutionContext.RegisterNestedStep(actionRunner2, inputsData, location, Environment, true);
```
TingluoHuang, line 114: You don't need a step for setting outputs, inline the logic here.
the handler might look like:
- gather steps from composite action.
- run steps with compositeStepsRunner.
- set outputs.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -102,9 +111,22 @@ public Task RunAsync(ActionRunStage stage)
...
+                await compositeStepsRunner.RunAsync(ExecutionContext);
+            }
+            catch (Exception ex)
+            {
+                // Composite StepRunner should never throw exception out.
+                Trace.Error($"Caught exception from composite steps {nameof(CompositeStepsRunner)}: {ex}");
+                ExecutionContext.Error(ex);
+                ExecutionContext.Result = TaskResult.Failed;
+            }
+            ExecutionContext.Result = TaskResult.Succeeded;
```
TingluoHuang, line 129: you don't need to set `ExecutionContext.Result = TaskResult.Succeeded;`, the result defaults to succeed.

src/Runner.Worker/CompositeStepsRunner.cs
ericsciple: Otherwise let's move into the Handlers namespace (only used by CompositeActionHandler) and pass the composite steps instead of adding to ExecutionContext

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -23,7 +23,7 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
     {
         public CompositeActionExecutionData Data { get; set; }
 
-        public Task RunAsync(ActionRunStage stage)
+        public async Task RunAsync(ActionRunStage stage)
```
ericsciple, line 26: Can the "RegisterNestedSteps" method be deleted now from ExecutionContext?

And can we delete the CompositeActionOutputHandler too?

src/Runner.Worker/CompositeStepsRunner.cs
ericsciple: I guess technically a step could change it? Is that why we run it before each step? @TingluoHuang do you happen to remember?

ericsciple: If we need to keep it here, it would be good to move into IOUtil.cs so we don't duplicate this lengthy code.

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        public async Task RunAsync(IExecutionContext actionContext)
+        {
+            // Another approach we can explore, is moving all this logic to the CompositeActionHandler if it's small enough. 
+
+            ArgUtil.NotNull(actionContext, nameof(actionContext));
+            ArgUtil.NotNull(actionContext.CompositeSteps, nameof(actionContext.CompositeSteps));
+
+            // The parent StepsRunner of the whole Composite Action Step handles the cancellation stuff already. 
+            while (actionContext.CompositeSteps.Count > 0)
```
TingluoHuang, line 38: I think you can just foreach composite steps since the list won't change while steps are running.
the `while` loop was for handling `post` steps that are registered while steps are running.

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+                    {
+                        envContext[pair.Key] = pair.Value;
+                    }
+                }
+
+                step.ExecutionContext.ExpressionValues["env"] = envContext;
+
+
+
+                if (step is IActionRunner actionStep)
```
TingluoHuang, line 82: you don't need the check `if (step is IActionRunner actionStep)` since it must be an ActionRunner for composite action.

src/Runner.Worker/CompositeStepsRunner.cs
```diff
@@ -0,0 +1,249 @@
...
+        }
+
+        private async Task RunStepAsync(IStep step)
+        {
+            // Check to see if we can expand the display name
+            if (step is IActionRunner actionRunner &&
+                actionRunner.Stage == ActionRunStage.Main &&
+                actionRunner.TryEvaluateDisplayName(step.ExecutionContext.ExpressionValues, step.ExecutionContext))
+            {
+                step.ExecutionContext.UpdateTimelineRecordDisplayName(actionRunner.DisplayName);
```
TingluoHuang, line 123: you don't want to update the displayname for the record.

```
if (step is IActionRunner actionRunner && actionRunner.Stage == ActionRunStage.Main)
{
 actionRunner.TryEvaluateDisplayName(step.ExecutionContext.ExpressionValues, step.ExecutionContext);
}
```

src/Runner.Worker/CompositeStepsRunner.cs
TingluoHuang: +1 we don't really need this.

src/Runner.Worker/CompositeStepsRunner.cs
TingluoHuang: i don't think timeout will work properly right now, since the `step.ExecutionContext.SetTimeout(timeout);` is called on the StepRunner for the whole composite action

src/Runner.Worker/CompositeStepsRunner.cs
ethanchewy: (tl;dr from discussion, we should move this if step into IOUtil.cs)

ethanchewy: @ericsciple 

I don't actually can't abstract this away into IOUtil.cs because, 

`var p = HostContext.CreateService<IProcessInvoker>()` uses `RunnerService.HostContext` which I can't import into IOUtil.cs

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: good point - thanks!

src/Runner.Worker/CompositeStepsRunner.cs
ethanchewy: fixed - thanks for pointing out.

src/Runner.Worker/CompositeStepsRunner.cs
ethanchewy: I'll add a TODO so that I remember to do this for `timeout-minutes`

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: We do `Root.CreateChild(_record.Id, step.DisplayName,...` in the RegisterNestedStep. `Root` is private to the ExecutionContext so we can't just move that logic into the handler.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: Good idea - thanks!

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +83,64 @@ public Task RunAsync(ActionRunStage stage)
...
 
-            // Create a step that handles all the composite action steps' outputs
-            Pipelines.ActionStep cleanOutputsStep = new Pipelines.ActionStep();
-            cleanOutputsStep.ContextName = ExecutionContext.ContextName;
-            // Use the same reference type as our composite steps.
-            cleanOutputsStep.Reference = Action;
+            // All steps have the same initial ExecutionContext/
+            // We store one of these ExecutionContexts in this variable 
+            // so that we can process the outputs correctly.
+            var stepExecutionContext = compositeSteps?[0].ExecutionContext;
```
ericsciple: have you considered using `ExecutionContext` instead of `compositeSteps?[0].ExecutionContext`?

src/Runner.Common/Util/EncodingUtil.cs
```diff
@@ -0,0 +1,50 @@
+using GitHub.Runner.Sdk;
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Common.Util
+{
+    using System;
+
+    public static class EncodingUtil
+    {
+        public static async void EnsureUtfEncoding(IHostContext hostContext, Tracing trace, System.Threading.CancellationToken cancellationToken)
```
ericsciple, line 10: `SetEncoding` might be a more accurate name. EnsureUtfEncoding might be a slight misnomer. On Linux i think dotnet will follow whatever the `LANG` or related env vars. So on Linux this method doesnt need to do anything.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: `ExecutionContext` is different than `compositeSteps?[0].ExecutionContext`

`ExecutionContext` contains the whole composite action's execution context. 

The `compositeSteps?[0].ExecutionContext` is the direct result of Root.CreateChild(..) in `RegisterNestedStep` and has the pointer to the correct `outputs` attribute in StepsContext that we need to use when we evaluate the outputs:
`DictionaryContextData actionOutputs = actionManifestManager.EvaluateCompositeOutputs(stepExecutionContext, Data.Outputs, evaluateContext);`

**tl;dr We can't use `ExecutionContext` since it doesn't contain the pointer to`ExpressionValues["outputs"]` while `compositeSteps?[0].ExecutionContext` has that pointer.**

src/Runner.Common/Util/EncodingUtil.cs
ethanchewy: Noted - will fix. Thanks!

src/Runner.Common/Util/EncodingUtil.cs
```diff
@@ -0,0 +1,50 @@
+using GitHub.Runner.Sdk;
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Common.Util
+{
+    using System;
+
+    public static class EncodingUtil
+    {
+        public static async void SetEncoding(IHostContext hostContext, Tracing trace, System.Threading.CancellationToken cancellationToken)
```
ericsciple, line 10: typically `async Task SetEncoding` technically void works but history... intention was for GUI back compat.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,63 @@ public Task RunAsync(ActionRunStage stage)
...
 
-            var actionRunner2 = HostContext.CreateService<IActionRunner>();
-            actionRunner2.Action = cleanOutputsStep;
-            actionRunner2.Stage = ActionRunStage.Main;
-            actionRunner2.Condition = "always()";
-            ExecutionContext.RegisterNestedStep(actionRunner2, inputsData, location, Environment, true);
+                // This is where we set the outputs.
+                // All steps have the same scope for their ExecutionContext/
+                // So we can use one of them so that we can get the right "outputs" attribute.
+                HandleOutput(compositeSteps?[0].ExecutionContext);
```
ericsciple, line 103: compositeSteps is never null

ericsciple: should we check if compositeSteps.Count \> 0 (guard against index out of range)?

ericsciple: Should handle-output get it's own child node instead? I wonder what unintentional side effects this will have by re-using the context for a previous step that already ran.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,63 @@ public Task RunAsync(ActionRunStage stage)
...
+                foreach (var pair in stepExecutionContext.ExpressionValues)
+                {
+                    evaluateContext[pair.Key] = pair.Value;
+                }
+
+                // Get the evluated composite outputs' values mapped to the outputs named
+                DictionaryContextData actionOutputs = actionManifestManager.EvaluateCompositeOutputs(stepExecutionContext, Data.Outputs, evaluateContext);
+
+                // Set the outputs for the outputs object in the whole composite action
+                actionManifestManager.SetAllCompositeOutputs(ExecutionContext, actionOutputs);
```
ericsciple, line 142: feels like the responsibility for setting the outputs belongs in this handler (above evaluate function on manifest manager makes sense)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +149,188 @@ private void InitializeScope(IStep step)
...
+                        envContext[pair.Key] = pair.Value;
+                    }
+                }
+
+                step.ExecutionContext.ExpressionValues["env"] = envContext;
+
+                var actionStep = step as IActionRunner;
+
+                // Set GITHUB_ACTION
+                if (!String.IsNullOrEmpty(step.ExecutionContext.ScopeName))
```
ericsciple, line 200: add a todo to fix this. i have a related pr

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +149,188 @@ private void InitializeScope(IStep step)
             var scopeName = step.ExecutionContext.ScopeName;
             step.ExecutionContext.ExpressionValues["steps"] = stepsContext.GetScope(scopeName);
         }
+
+        private async Task RunStepsAsync(List<IStep> compositeSteps)
```
ericsciple, line 153: To track the overall step-result, you need to do something like stepsrunner does to track the overall job-result:

Look at this code in StepsRunner since it's a similar problem:

```
                // Update the job result.
                if (step.ExecutionContext.Result == TaskResult.Failed)
                {
                    Trace.Info($"Update job result with current step result '{step.ExecutionContext.Result}'.");
                    jobContext.Result = TaskResultUtil.MergeTaskResults(jobContext.Result, step.ExecutionContext.Result.Value);
                    jobContext.JobContext.Status = jobContext.Result?.ToActionResult();
                }
                else
                {
                    Trace.Info($"No need for updating job result with current step result '{step.ExecutionContext.Result}'.");
                }
```

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +149,188 @@ private void InitializeScope(IStep step)
...
+            if (step is IActionRunner actionRunner && actionRunner.Stage == ActionRunStage.Main)
+            {
+                actionRunner.TryEvaluateDisplayName(step.ExecutionContext.ExpressionValues, step.ExecutionContext);
+            }
+
+            // Start the step.
+            Trace.Info("Starting the step.");
+            step.ExecutionContext.Debug($"Starting: {step.DisplayName}");
+
+            // Set the timeout
```
ericsciple, line 249: Should we yank this code for timeout?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +149,188 @@ private void InitializeScope(IStep step)
...
+                if (continueOnError)
+                {
+                    step.ExecutionContext.Outcome = step.ExecutionContext.Result;
+                    step.ExecutionContext.Result = TaskResult.Succeeded;
+                    Trace.Info($"Updated step result (continue on error)");
+                }
+            }
+            Trace.Info($"Step result: {step.ExecutionContext.Result}");
+
+            // Complete the step context.
```
ericsciple, line 332: if the result is not succeeded, we need to short circuit and return the result (basically it is the outer step result)

ericsciple: (still need to compute "outputs")

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: (as talked in our meeting, we decided to follow a short-circuiting logic: https://github.com/actions/runner/pull/591#discussion_r456004813)

src/Runner.Worker/CompositeStepsRunner.cs
ethanchewy: @TingluoHuang 

If I remove that, the UI now looks like: https://drive.google.com/file/d/1JzAWi0qOHQSOK-fBMp8Q_AFwPkkRqvDl/view?usp=sharing

But having the old code actually creates a more intuitive UI based on my conversations with Chris: https://drive.google.com/file/d/1-NrI8M1SoMKwS0y7LAl3JgmafzeVkapU/view?usp=sharing

ethanchewy: For now, I'll keep it as is

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,67 @@ public Task RunAsync(ActionRunStage stage)
...
+                // This is where we run each step.
+                await RunStepsAsync(compositeSteps, ExecutionContext);
+
+                // This is where we set the outputs.
+                // All steps have the same scope for their ExecutionContext/
+                // So we can use one of them so that we can get the right "outputs" attribute.
+                if (compositeSteps.Count > 0)
+                {
+                    // TODO: Find better solution. 
+                    HandleOutput(compositeSteps[0].ExecutionContext);
```
ericsciple, line 106: Just had an idea...

Let's use the `ExecutionContext`. Technically speaking, calculating the outputs is part of the outer step (this handler) and therefore (arguably) makes sense to use to this handler's execution context.

Before evaluating the outputs, you will need to set `ExecutionContext.ExpressionValues["inputs"] = inputsData;` and `ExecutionContext.ExpressionValues["steps"] = StepsContext.GetScope(fullyQualifiedContextName);`

ethanchewy: Genius

src/Runner.Common/Util/EncodingUtil.cs
```diff
@@ -0,0 +1,52 @@
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Runner.Sdk;
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Common.Util
+{
+    using System;
```
TingluoHuang, line 8: move the `using` to top.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,67 @@ public Task RunAsync(ActionRunStage stage)
                 actionRunner.Stage = stage;
                 actionRunner.Condition = aStep.Condition;
 
-                var step = ExecutionContext.RegisterNestedStep(actionRunner, inputsData, location, Environment);
-
+                var step = ExecutionContext.RegisterNestedStep(actionRunner, inputsData, Environment);
```
TingluoHuang, line 87: should we rename `RegisterNestedStep` to something like `CreateCompositeStep`?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,67 @@ public Task RunAsync(ActionRunStage stage)
...
                 InitializeScope(step);
 
-                location++;
+                // Add all steps to the Composite StepsRunner
+                // Follows similar logic to how JobRunner invokes the StepsRunner for job steps!
```
TingluoHuang, line 91: the comment seems out-dated. :)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
TingluoHuang: +1

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +153,209 @@ private void InitializeScope(IStep step)
...
+        }
+
+        private async Task RunStepsAsync(List<IStep> compositeSteps, IExecutionContext executionContext)
+        {
+            ArgUtil.NotNull(compositeSteps, nameof(compositeSteps));
+
+            // The parent StepsRunner of the whole Composite Action Step handles the cancellation stuff already. 
+            foreach (IStep step in compositeSteps)
+            {
+                // This is used for testing UI appearance.
```
TingluoHuang, line 192: remove these.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,67 @@ public Task RunAsync(ActionRunStage stage)
...
+                ExecutionContext.Error(ex);
+                ExecutionContext.Result = TaskResult.Failed;
+            }
+        }
+
+        private void CompleteStep(IStep step, TaskResult? result = null, string resultCode = null)
+        {
+            var executionContext = step.ExecutionContext;
+
+            executionContext.Complete(result, resultCode: resultCode);
```
TingluoHuang, line 122: not sure whether should ever call `ExecutionContext.Complete()` for steps inside a composite action. `ExecutionContext.Complete()` will update the timeline record.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,5 +153,209 @@ private void InitializeScope(IStep step)
...
+                var outputsAttributes = pair.Value as DictionaryContextData;
+                outputsAttributes.TryGetValue("value", out var val);
+                var outputsValue = val as StringContextData;
+
+                // Set output in the whole composite scope. 
+                if (!String.IsNullOrEmpty(outputsName) && !String.IsNullOrEmpty(outputsValue))
+                {
+                    parentExecutionContext.SetOutput(outputsName, outputsValue, out _);
+                }
+            }
```
TingluoHuang, line 182: just inline these into `HandleOutput` since the `SetAllCompositeOutputs()` only get used once.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -85,26 +84,67 @@ public Task RunAsync(ActionRunStage stage)
...
+        }
+
+        private void CompleteStep(IStep step, TaskResult? result = null, string resultCode = null)
+        {
+            var executionContext = step.ExecutionContext;
+
+            executionContext.Complete(result, resultCode: resultCode);
+        }
+
+        private void HandleOutput(IExecutionContext stepExecutionContext)
```
TingluoHuang, line 125: maybe rename `HandleOutput` to `ProcessCompositeActionOutputs`?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: Changed to:
```
                // Get the pointer of the correct "steps" object and pass it to the ExecutionContext so that we can process the outputs correctly
                // This will always be the same for every step so we can pull this from the first step if it exists
                var stepExecutionContext = compositeSteps.Count > 0 ? compositeSteps[0].ExecutionContext : null;
                ExecutionContext.ExpressionValues["inputs"] = inputsData;
                ExecutionContext.ExpressionValues["steps"] = stepExecutionContext.StepsContext.GetScope(stepExecutionContext.ScopeName);

                HandleOutput();
```

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: Sure

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: Yeah, now that I think about it, we don't really need to call Complete() in each step's execution context. 

It doesn't really do anything since we'll eventually call Complete() in the StepsRunner for the whole composite action step in the job. 

I'll take it out.

ethanchewy: Yeah, this is confusing, I'll remove it. I was just using that for failing a step. I'll consolidate the code and just put the parts we need for failing a step.

PR #593, opened 2020-07-16
Switch GITHUB_ACTION to ContextName

ericsciple:
For composite actions, we need a distinct value GITHUB_ACTION for each nested action.

Today GITHUB_ACTION is always set to the refname (which is an otherwise deprecated property). The server generates the refname.

For composite actions, we need the runner to generate a value for each nested step. And that value cannot collide with outer steps.

The solution proposed by this PR, is to set GITHUB_ACTION to the step ContextName instead. The ContextName is denoted in the workflow YAML by the step `id` property.

For nested steps, $"{ScopeName}.{ContextName}" will be used to ensure no collisions.

When a step `id` is not supplied by the customer, a unique ContextName will be generated. Generated names will start with `__`. Step context (result/outputs/etc) is not tracked for generated context names.


src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -454,9 +454,9 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
                     case "steps":
                         if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TESTING_COMPOSITE_ACTIONS_ALPHA")))
                         {
-                            var steps = run.Value.AssertSequence("steps");
-                            var evaluator = executionContext.ToPipelineTemplateEvaluator();
-                            stepsLoaded = evaluator.LoadCompositeSteps(steps);
+                            var stepsToken = run.Value.AssertSequence("steps");
+                            steps = PipelineTemplateConverter.ConvertToSteps(templateContext, stepsToken);
```
ericsciple, line 458: I changes this to use PipelineTemplateConverter instead of PipelineTemplateEvaluator

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -14,8 +14,66 @@
...
+                if (step != null) // step = null means we are hitting error during step conversion, there should be an error in context.errors
+                {
+                    if (step.Enabled)
+                    {
+                        result.Add(step);
+                    }
+                }
+            }
+
+            // Generate context name if empty
```
ericsciple, line 40: this foreach block is new

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -14,8 +14,66 @@
 
 namespace GitHub.DistributedTask.Pipelines.ObjectTemplating
 {
-    internal static class PipelineTemplateConverter
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class PipelineTemplateConverter
     {
+        public static List<Step> ConvertToSteps(
```
ericsciple, line 20: most of this method is existing, i just moved it up because it's public now

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/ReferenceNameBuilder.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
```
ericsciple, line 1: this is an existing class from the server

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
```diff
@@ -425,8 +400,7 @@ public class PipelineTemplateEvaluator
         private TemplateContext CreateContext(
             DictionaryContextData contextData,
             IList<IFunctionInfo> expressionFunctions,
-            IEnumerable<KeyValuePair<String, Object>> expressionState = null,
-            bool setMissingContext = true)
+            IEnumerable<KeyValuePair<String, Object>> expressionState = null)
```
ericsciple, line 403: i reverted changes from this file back to their original state before composite-actions-changes

switched to PipelineTemplateConverter instead to convert steps token into `Pipelines.Steps`

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -337,9 +373,12 @@ internal static class PipelineTemplateConverter
 
                     case PipelineTemplateConstants.Id:
                         id = stepProperty.Value.AssertString($"{PipelineTemplateConstants.Steps} item {PipelineTemplateConstants.Id}");
-                        if (!NameValidation.IsValid(id.Value, true))
+                        if (!String.IsNullOrEmpty(id.Value))
```
ericsciple, line 376: TryAddKnownName performs name validation

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -400,14 +435,12 @@ internal static class PipelineTemplateConverter
             }
 
             // Fixup the if-condition
-            var isDefaultScope = String.IsNullOrEmpty(scope?.Value);
-            ifCondition = ConvertToIfCondition(context, ifToken, false, isDefaultScope);
+            ifCondition = ConvertToIfCondition(context, ifToken, false);
```
ericsciple, line 438: The rest of the changes in this file are because legacy template-prototype code was removed (PR already approved on server, just got copied here)

src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -57,7 +57,7 @@ public override void Initialize(IHostContext hostContext)
 
         public ActionDefinitionData Load(IExecutionContext executionContext, string manifestFile)
         {
-            var templateContext = CreateContext(executionContext);
+            var templateContext = CreateTemplateContext(executionContext);
```
ericsciple, line 60: sorry about the diff for this file, i renamed `context` local variable everywhere to `templateContext` for future readability (distinguish from executionContext)

src/Sdk/DTPipelines/Pipelines/ActionStepDefinitionReference.cs
```diff
@@ -150,4 +153,36 @@ public override ActionStepDefinitionReference Clone()
             return new ScriptReference(this);
         }
     }
+
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
ericsciple, line 157: copied this class from the server

ericsciple: actually.... i don't think we need this anymore. Is this because originally we had a `checkout` keyword?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -118,7 +118,7 @@ public async Task RunAsync(IExecutionContext jobContext)
                 if (step is IActionRunner actionStep)
                 {
                     // Set GITHUB_ACTION
-                    step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
+                    step.ExecutionContext.SetGitHubContext("action", step.ExecutionContext.ContextName);
```
ericsciple, line 121: i need to add a FF

ericsciple: And when scopename not empty, need to set to `$"{step.ExecutionContext.ScopeName}.{step.ExecutionContext.ContextName}"`..... i might do that part after Ethan's PR lands for the CompositeActionOutputHandler changes

src/Sdk/DTPipelines/Pipelines/ActionStepDefinitionReference.cs
```diff
@@ -16,7 +16,10 @@ public enum ActionSourceType
         ContainerRegistry = 2,
 
         [DataMember]
-        Script = 3
+        Script = 3,
```
ericsciple, line 19: todo: copy server changes where scopename was removed from the step and react

src/Sdk/DTPipelines/Pipelines/ActionStep.cs
```diff
@@ -24,7 +24,6 @@ private ActionStep(ActionStep actionToClone)
             Environment = actionToClone.Environment?.Clone();
             Inputs = actionToClone.Inputs?.Clone();
             ContextName = actionToClone?.ContextName;
-            ScopeName = actionToClone?.ScopeName;
```
ericsciple, line 27: Message contract changes copied from server (already merged on server side)

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -39,7 +39,6 @@ internal AgentJobRequestMessage()
             DictionaryContextData contextData,
             WorkspaceOptions workspaceOptions,
             IEnumerable<JobStep> steps,
-            IEnumerable<ContextScope> scopes,
```
ericsciple, line 42: Message contract changes copied from server (already merged on server side)

src/Sdk/DTPipelines/Pipelines/ContextScope.cs
```diff
@@ -1,53 +0,0 @@
-﻿using System;
```
ericsciple, line 1: Message contract changes copied from server (already merged on server side)

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -16,116 +16,6 @@
       }
     },
 
-    "steps-template-root": {
```
ericsciple, line 19: copied latest from server

src/Runner.Worker/StepsRunner.cs
```diff
@@ -118,7 +118,19 @@ public async Task RunAsync(IExecutionContext jobContext)
                 if (step is IActionRunner actionStep)
                 {
                     // Set GITHUB_ACTION
-                    step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
+                    if (step.ExecutionContext.Variables.GetBoolean("DistributedTask.UseContextNameForGITHUBACTION") ?? false)
```
ericsciple, line 121: this kills refname in favor of ContextName (dont need both)

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -327,7 +324,9 @@ public IExecutionContext CreateChild(Guid recordId, string displayName, string r
             var child = new ExecutionContext();
             child.Initialize(HostContext);
             child.ScopeName = scopeName;
-            child.ContextName = contextName;
+            // Temporary code to generate a context name. After M271-ish the server will never send an empty context name.
+            // Generated context names start with "__"
+            child.ContextName = !string.IsNullOrEmpty(contextName) ? contextName : $"__{Guid.NewGuid()}";
```
ericsciple, line 329: dont do this

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -434,7 +433,9 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
 
             _logger.End();
 
-            if (!string.IsNullOrEmpty(ContextName))
+            // The job-level context name is always empty.
+            // Generated step-level context names start with "__".
+            if (!string.IsNullOrEmpty(ContextName) && !ContextName.StartsWith("__"))
```
ericsciple, line 438: check ff - dont introduce risk

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -14,8 +14,62 @@
...
+                {
+                    continue;
+                }
+
+                var name = default(string);
+                switch (step.Reference.Type)
+                {
+                    case ActionSourceType.ContainerRegistry:
+                        var containerReference = step.Reference as ContainerRegistryReference;
+                        name = containerReference.Image;
```
TingluoHuang, line 53: what will `namebuilder` do for special characters in docker image, like `:`?

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
TingluoHuang: is this already on server-side code or you need to copy it over?

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
ericsciple: replace with _

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
ericsciple: i need to make the same change on the server side. Waiting on until we have telemetry first to make sure we can use `__` to indicate generated

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -273,23 +270,14 @@ public void RegisterPostJobStep(IStep step)
...
-            // This makes it much simpler to handle their outputs at the end of the Composite Action
-            var childScopeName = !string.IsNullOrEmpty(ScopeName) ? $"{ScopeName}.{safeContextName}" : safeContextName;
-
-            var childContextName = !string.IsNullOrEmpty(step.Action.ContextName) ? step.Action.ContextName : $"__{Guid.NewGuid()}";
-
-            step.ExecutionContext = Root.CreateChild(_record.Id, step.DisplayName, _record.Id.ToString("N"), childScopeName, childContextName, logger: _logger);
-
+            step.ExecutionContext = Root.CreateChild(_record.Id, step.DisplayName, _record.Id.ToString("N"), scopeName, step.Action.ContextName, logger: _logger);
             step.ExecutionContext.ExpressionValues["inputs"] = inputsData;
+            step.ExecutionContext.ExpressionValues["steps"] = StepsContext.GetScope(step.ExecutionContext.GetFullyQualifiedContextName());
```
ericsciple, line 280: @ethanchewy fyi i moved this here where the rest of the contexts are setup. and simplified to the new `GetFullyQualfiedContextName` extension method

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -273,23 +270,14 @@ public void RegisterPostJobStep(IStep step)
         /// add a child node, aka a step, to the current job to the Root.JobSteps based on the location. 
         /// </summary>
         public IStep CreateCompositeStep(
+            string scopeName,
             IActionRunner step,
             DictionaryContextData inputsData,
             Dictionary<string, string> envData)
         {
-            // If the context name is empty and the scope name is empty, we would generate a unique scope name for this child in the following format:
```
ericsciple, line 280: @ethanchewy fyi i fixed a bug here. When ContextName was empty, each child scope name got a new guid (i.e. each nested step had a different scope name). I switched to pass the child scopeName as a parameter instead.

After the server generates default context names (M271-ish), we can eliminate the `scopeName` parameter since the outer step's context name will never be empty.

ethanchewy: ok gotcha - thanks!

PR #594, opened 2020-07-16
move shared ExecutionContext properties under .Global

ericsciple:


ericsciple:
Thoughts?

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -188,7 +188,7 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
                 throw new Exception("Required field 'name' is missing in ##[set-env] command.");
             }
 
-            context.EnvironmentVariables[envName] = command.Data;
+            context.Global.EnvironmentVariables[envName] = command.Data;
```
ericsciple, line 191: The only change is in ExecutionContext.cs and everything else is reaction

src/Runner.Worker/GlobalExecutionContext.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Common.Util;
+using GitHub.Runner.Worker.Container;
+
+namespace GitHub.Runner.Worker
+{
+    public sealed class GlobalExecutionContext
```
ericsciple, line 9: GlobalContext

ericsciple:
reviewed offline w/ @TingluoHuang

PR #599, opened 2020-07-17
Fix Timeout-minutes for Whole Composite Action Step

ethanchewy:
The timeout-minutes token set in the workflow file for an individual composite action is correctly processed with the CancellationToken for the ExecutionContext of the whole composite action step **but for some reason my composite action cannot differentiate between cancellation and timing out** 

In this PR, we fix that. We also remove unnecessary code as well.

Example Cancel: https://github.com/ethanchewy/testing-actions/runs/891636687
Example of 1 min. Timeout: https://github.com/ethanchewy/testing-actions/runs/891641094?check_suite_focus=true


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -273,30 +279,7 @@ private async Task RunStepAsync(IStep step)
                 step.ExecutionContext.Result = Common.Util.TaskResultUtil.MergeTaskResults(step.ExecutionContext.Result, step.ExecutionContext.CommandResult.Value);
             }
 
-            // Fixup the step result if ContinueOnError.
-            if (step.ExecutionContext.Result == TaskResult.Failed)
```
ethanchewy, line 277: We don't need this since we'll explicitly disallow Continue on error for each composite step for now.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -356,7 +358,7 @@ private async Task RunStepAsync(IStep step, CancellationToken jobCancellationTok
                     Trace.Info($"Updated step result (continue on error)");
                 }
             }
-            Trace.Info($"Step result: {step.ExecutionContext.Result}");
+            Trace.Info($"Step result: {StringUtil.ConvertToJson(step.ExecutionContext.Result)}");
```
ethanchewy, line 361: Add formatting

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -273,30 +279,7 @@ private async Task RunStepAsync(IStep step)
...
-
-                if (continueOnError)
-                {
-                    step.ExecutionContext.Outcome = step.ExecutionContext.Result;
-                    step.ExecutionContext.Result = TaskResult.Succeeded;
-                    Trace.Info($"Updated step result (continue on error)");
-                }
-            }
-            Trace.Info($"Step result: {step.ExecutionContext.Result}");
+            Trace.Info($"Step result: {StringUtil.ConvertToJson(step.ExecutionContext.Result)}");
```
ethanchewy, line 282: Added JSON formatting

src/Runner.Worker/StepsRunner.cs
ericsciple: Do we need to do this? `Result` is a simple type (an enum), not a structure.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -60,10 +60,12 @@ public interface IExecutionContext : IRunnerService
 
         bool EchoOnActionCommand { get; set; }
 
+        IExecutionContext JobExecutionContext { get; set; }
```
ericsciple, line 63: Should we use `Root`?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -254,7 +258,7 @@ public void RegisterPostJobStep(IStep step)
             DictionaryContextData inputsData,
             Dictionary<string, string> envData)
         {
-            step.ExecutionContext = Root.CreateChild(_record.Id, step.DisplayName, _record.Id.ToString("N"), scopeName, step.Action.ContextName, logger: _logger);
+            step.ExecutionContext = Root.CreateChild(_record.Id, step.DisplayName, _record.Id.ToString("N"), scopeName, step.Action.ContextName, logger: _logger, cancellationTokenSource: _cancellationTokenSource);
```
ericsciple, line 261: Would it make more sense for the interface to pass `CancellationToken.CreateLinkedTokenSource(_cancellationTokenSource)`

PR #605, opened 2020-07-20
Explicitly define what is allowed for a composite action

ethanchewy:
In this PR, we explicitly define what is allowed for our composite action feature.

For each composite run step, we explicitly remove support for `if`, `timeout-minutes`, and `continue-on-error`. We'll support, `id`, `shell`, `run`, `working-directory`, and `env` all of which are optional except for `run`.

We don't want the defaults to flow to the composite action. Here is proof of them not flowing: https://github.com/ethanchewy/testing-actions/runs/895343022?check_suite_focus=true

Here is an example of the step level `working-directory` and `env` working: https://github.com/ethanchewy/testing-actions/runs/896147997?check_suite_focus=true

Shell requirement: https://github.com/ethanchewy/testing-actions/actions/runs/184525896

Example outputs demo working with explicit parameters: https://github.com/ethanchewy/testing-actions/runs/915495023?check_suite_focus=true

ID Context Expressions Not Allowed: https://github.com/ethanchewy/testing-actions/actions/runs/184536224, https://github.com/ethanchewy/testing-actions/actions/runs/184534064

Shell expression not allowed: https://github.com/ethanchewy/testing-actions/runs/915600890


ethanchewy:
Tested if the defaults shell and working-dir values override the composite action steps' shell and working-dir. They do not override the steps inside of a composite action's shell or working-dir. This is want we want because we want the  workflow to have no influence over the composite action in terms of defaults.

Why does this already work?

We don't check the `ExecutionContext.Global.JobDefaults` in `ScriptHandler.cs` unless the `ExecutionContext.ScopeName` is null for both the `shell` and `working-dir`: 

- working-dir: https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ScriptHandler.cs#L152-L162
- shell: https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ScriptHandler.cs#L168-L179

cc @ericsciple

ethanchewy:
Originally had planned to just null the ExectionContext.Global.JobDefaults for each composite action step ]'s ExecutionContext but this would have nullified all .Global.JobDefaults since we share the same pointer to Global for every instance of ExecutionContext. 

So, I think my `isComposite` solution is better suited: https://github.com/actions/runner/pull/605/commits/aeae15de7b621fcc500f44442c09f395d2aa39ac

ethanchewy:
Reverting changes - if we change the run defaults logic in the future, we'll change it then.

ethanchewy:
example of if not being allowed: https://github.com/ethanchewy/testing-actions/runs/895611092?check_suite_focus=true

 (all others should not work too like "needs", etc.)

ethanchewy:
Step level `shell` and `working-directory` not working in a composite action since the `shell` and `working-directory` are mapped into the ActionStep.Inputs when we convert the step in the pipeline template converter. 

The ActionRunner tries to retrieve these values from the ActionDefinitionData Inputs and then sets the handler's `Inputs` to this inputs dictionary. 

Then, the ScriptHandler tries to get this `working-directory` and `shell` values from this handlerData.Inputs.

To fix this, I'm thinking about explicitly adding the "Inputs" attribute to pass the `working-directory` and `shell` values.

ethanchewy:
> Step level `shell` and `working-directory` not working in a composite action since the `shell` and `working-directory` are supposed to be mapped into the ActionStep.Inputs when we convert the step in the pipeline template converter. But for some reason it's not mapped when we process it in the Composite Action Handler.
> 
> The ActionRunner tries to retrieve these values from the ActionDefinitionData Inputs and then sets the handler's `Inputs` to this inputs dictionary.
> 
> Then, the ScriptHandler tries to get this `working-directory` and `shell` values from this handlerData.Inputs.
> 
> To fix this, I'm trying to figure out how to add the `working-directory` and `shell` values to the definition.Inputs dictionary. Currently debugging...

This issue stems from the ConvertToSteps losing specificity. It originally returned a list of `Step`s which doesn't include the `.Inputs` attribute which is a dictionary that will contain `working-directory` and `shell`. Since we return it as a list of Steps and cast it back to a list of ActionSteps, we "lose" this `.Inputs` attribute.

ethanchewy:
Step level `working-dir` and `shell` works now!

ethanchewy:
Thought there was a bug with setting the working-dir and shell but it works as intended (it was a bug with my test :/). Reverting back to original code.

ethanchewy:
Ready for review.

src/Runner.Worker/action_yaml.json
```diff
@@ -120,7 +120,22 @@
...
+                "properties": {
+                    "name": "string-steps-context",
+                    "id": "non-empty-string",
+                    "run": {
+                        "type": "string-steps-context",
+                        "required": true
+                    },
+                    "env": "step-env",
+                    "working-directory": "string-steps-context",
+                    "shell": "string-steps-context"
```
TingluoHuang, line 137: Is shell required?

src/Runner.Worker/action_yaml.json
```diff
@@ -157,6 +172,36 @@
...
+                "runner",
+                "env",
+                "hashFiles(1,255)"
+            ],
+            "string": {}
+        },
+        "step-env": {
+            "context": [
+                "github",
+                "needs",
```
TingluoHuang, line 192: Remove needs?

src/Runner.Worker/action_yaml.json
ethanchewy: @ericsciple and I talked about it and we decided that it's best to leave it optional. If the user doesn't put anything there, in ScriptHandler, it will default to the runner's OS default shell.

ethanchewy: Also, this follows the same consistency and logic as how e do it in the workflow so I think it makes sense.

src/Runner.Worker/action_yaml.json
ethanchewy: shoot - didn't see. Thanks - will remove.

ethanchewy:
(leaving this unmerged till @ericsciple has the chance to look at it since this is a really important PR)

src/Runner.Worker/action_yaml.json
```diff
@@ -120,7 +120,22 @@
                 "hashFiles(1,255)"
             ],
             "sequence": {
-                "item-type": "any"
+                "item-type": "composite-step"
```
ericsciple, line 123: We need to remove `context` from `composite-steps`. Expressions aren't allowed anywhere under `composite-steps`; only certain places.

ericsciple, line 123: do we have validation on the server that at least one step is required?

src/Runner.Worker/action_yaml.json
ericsciple: Thinking about this a bit more... let's make it require (and `non-empty-string`).

I think during the original ADR review with product we said required. Rationale because it's different than adding a `run` step to your own workflow where you control the OS. In this case, action-author doesn't control the OS. Should be explicit to avoid unexpected behavior.

src/Runner.Worker/action_yaml.json
ericsciple: Make sure to test and verify an expression can't be used for `id`

src/Runner.Worker/action_yaml.json
ericsciple, line 137: `shell` in `workflow-v1.0.json` is `non-empty-string` (expressions not allowed). Let's match that value. Currently you have `string-steps-context` instead. Good to test after making the change and verify expression not allowed for `shell` value.

src/Runner.Worker/action_yaml.json
```diff
@@ -157,6 +172,35 @@
             "string": {
                 "require-non-empty": true
             }
+        },
+        "string-steps-context": {
+            "context": [
+                "github",
```
ericsciple, line 178: Add `inputs`

src/Runner.Worker/action_yaml.json
```diff
@@ -157,6 +172,35 @@
...
+                "job",
+                "runner",
+                "env",
+                "hashFiles(1,255)"
+            ],
+            "string": {}
+        },
+        "step-env": {
+            "context": [
+                "github",
```
ericsciple, line 191: Add `inputs`

src/Runner.Worker/action_yaml.json
ethanchewy: Yup, that makes sense. 

(Reminder to self: will need to update all documentation to have required shell attribute for each run step)

src/Runner.Worker/action_yaml.json
ethanchewy: For server side, if there are no "steps" token used, it handles it: https://github.com/ethanchewy/testing-actions/runs/898870893?check_suite_focus=true. If there are no steps within the "steps" token, the user will get an unexpected value from the server side: https://github.com/ethanchewy/testing-actions/runs/898862578?check_suite_focus=true

ethanchewy:
Sanity checks:

ID Context Expressions Not Allowed: https://github.com/ethanchewy/testing-actions/actions/runs/184536224, https://github.com/ethanchewy/testing-actions/actions/runs/184534064

Shell Context Expression Not Allowed: https://github.com/ethanchewy/testing-actions/actions/runs/184538648

src/Runner.Worker/action_yaml.json
ethanchewy: shell is evaluated unfortunately due to the if condition in ScriptHandler: https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ScriptHandler.cs#L255-L271. Demo: https://github.com/ethanchewy/testing-actions/runs/915535061?check_suite_focus=true. 

Will put in fix soon.

src/Runner.Worker/action_yaml.json
```diff
@@ -108,19 +108,26 @@
             }
         },
         "composite-steps": {
-            "context": [
```
ethanchewy, line 111: if the attribute needs a context, they'll have access (see "string-steps-context")

ethanchewy:
Ready for final review.

PR #607, opened 2020-07-21
Clarify details for defaults, shell, and working-dir

ethanchewy:
In this commit, I clarify some details for our implementation of shell and working-dir (cc: @ericsciple )

### DIFF
--- a/docs/adrs/0549-composite-run-steps.md
+++ b/docs/adrs/0549-composite-run-steps.md
@@ -254,7 +254,7 @@ For the composite action steps, it follows the same logic as above. In this exam
 We will not support "defaults" in a composite action. 
 
 ### Shell and Working-dir
-For each run step in a composite action, the action author can set the `shell` and `working-dir` attributes for that step. These attributes are optional for each run step - by default, the `shell` is set to whatever default value is associated with the runner os (ex: bash =\> Mac). Moreover, the composite action author can map in values from the `inputs` for it's `shell` and `working-dir` attributes at the step level for an action. 
+For each run step in a composite action, the action author can set the `shell` and `working-directory` attributes for that step. These attributes are optional for each run step - by default, the `shell` is set to whatever default value is associated with the runner os (ex: bash =\> Mac). Moreover, the composite action author can map in values from the `inputs` for it's `shell` and `working-directory` attributes at the step level for an action. 
 
 For example,
 
@@ -271,7 +271,7 @@ steps:
     shell: ${{ inputs.shell_1 }}
 ```
 
-Note, the workflow file and action file are treated as separate entities. **So, the workflow `defaults` will never change the `shell` and `working-dir` value in the run steps in a composite action.** Note, `defaults` in a workflow only apply to run steps not "uses" steps (steps that use an action).
+Note, the workflow file and action file are treated as separate entities. **So, the workflow `defaults` will never change the `shell` and `working-directory` value in the run steps in a composite action.** Note, `defaults` in a workflow only apply to run steps not "uses" steps (steps that use an action).
 
 ### Visualizing Composite Action in the GitHub Actions UI
 We want all the composite action's steps to be condensed into the original composite action node. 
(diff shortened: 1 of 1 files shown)


docs/adrs/0549-composite-run-steps.md
```diff
@@ -250,9 +250,21 @@ If any of the steps fail in the composite action and the `continue-on-error` is
 
 For the composite action steps, it follows the same logic as above. In this example, `"Hello World 2"` will be outputted because the previous step has `continue-on-error` set to `true` although that previous step errored. 
 
-### Defaults
+### Defaults, Shell, and Working-dir
+We will not support "defaults" in a composite action. 
 
-The composite action author will be required to set the `shell` and `workingDir` of the composite action. Moreover, the composite action author will be able to explicitly set the shell for each composite run step. The workflow author will not have the ability to change these attributes. 
+For each run step in a composite action, the action author can set the `shell` and `working-dir` attributes for that step. These attributes are optional for each run step - by default, the `shell` is set to whatever default value is associated with the runner os (ex: bash => Mac). Moreover, the composite action author can map in values from the `inputs` for it's `shell` and `working-dir` attributes at the step level for an action. For example,
```
ericsciple, line 256: > (ex: bash => Mac)

markdown allows html so you should escape html-special characters:

```
(ex: bash =\> Mac)
```

docs/adrs/0549-composite-run-steps.md
```diff
@@ -250,9 +250,21 @@ If any of the steps fail in the composite action and the `continue-on-error` is
...
+inputs:
+  shell_1:
+    description: 'Your name'
+    default: 'pwsh'
+steps:
+  - run: echo 1
+    shell: ${{ inputs.shell_1 }}
+```
+
+Note, the workflow file and action file are treated as seperate entities. **So, the workflow `defaults` will never change the `shell` and `working-dir` value in the run steps in a composite action.** Note, `defaults` in a workflow only apply to run steps not "uses" steps (steps that use an action).
```
ericsciple, line 267: sp: separate

docs/adrs/0549-composite-run-steps.md
```diff
@@ -250,9 +250,21 @@ If any of the steps fail in the composite action and the `continue-on-error` is
 
 For the composite action steps, it follows the same logic as above. In this example, `"Hello World 2"` will be outputted because the previous step has `continue-on-error` set to `true` although that previous step errored. 
 
-### Defaults
+### Defaults, Shell, and Working-dir
+We will not support "defaults" in a composite action.
```
ericsciple, line 254: may need to escape `"` as `\"`

ethanchewy: Just double checked, since it's a .md, I believe you don't need to escape. I double checked in the preview and it looks as intended

docs/adrs/0549-composite-run-steps.md
```diff
@@ -251,8 +251,27 @@ If any of the steps fail in the composite action and the `continue-on-error` is
 For the composite action steps, it follows the same logic as above. In this example, `"Hello World 2"` will be outputted because the previous step has `continue-on-error` set to `true` although that previous step errored. 
 
 ### Defaults
+We will not support "defaults" in a composite action. 
 
-The composite action author will be required to set the `shell` and `workingDir` of the composite action. Moreover, the composite action author will be able to explicitly set the shell for each composite run step. The workflow author will not have the ability to change these attributes. 
+### Shell and Working-directory
+For each run step in a composite action, the action author can set the `shell` and `working-directory` attributes for that step. These attributes are optional for each run step - by default, the `shell` is set to whatever default value is associated with the runner os (ex: bash =\> Mac). Moreover, the composite action author can map in values from the `inputs` for it's `shell` and `working-directory` attributes at the step level for an action.
```
roryabraham, line 257: > These attributes are optional for each run step - by default, the `shell` is set to whatever default value is associated with the runner os

This [doesn't appear to be true](https://github.com/Andrew-Test-Org/Public-Test-Repo/runs/7140706556?check_suite_focus=true). Removing the `shell` attribute from any `run` step of a composite action results in:

> The template is not valid ... Required property is missing: shell

PR #609, opened 2020-07-22
Improve Debugging Messages for Empty Tokens

ethanchewy:
As part of the process for polishing Composite Run Steps, I've revisited my old TODOs. 

There’s only one TODO related directly to composite run steps (the other TODOs have to do with nested steps).

Namely:
`// TODO: Add a more helpful error message + including file name, etc. to show user that it's because of their yaml file`

In this PR, I add the filename associated with this error as well as add more clear language in the error message. I also change similar error messages that could use some improvement.

**Basically, I change the error message for null tokens to say, "You are using \<TypeOfAction\> but the \<TokenName\> is empty in \<FileRelativePath\>".**

Example: https://github.com/ethanchewy/testing-actions/runs/898870893?check_suite_focus=true

### DIFF
--- a/src/Test/L0/Worker/ActionManagerL0.cs
+++ b/src/Test/L0/Worker/ActionManagerL0.cs
@@ -846,7 +846,7 @@ namespace GitHub.Runner.Common.Tests.Worker
                 {
                     var traceFile = Path.GetTempFileName();
                     File.Copy(_hc.TraceFileName, traceFile, true);
-                    Assert.Contains("Entry javascript file is not provided.", File.ReadAllText(traceFile));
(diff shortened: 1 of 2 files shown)


src/Test/L0/Worker/ActionManagerL0.cs
```diff
@@ -846,7 +846,7 @@ public async void PrepareActions_RepositoryActionWithInvalidWrapperActionfile_No
                 {
                     var traceFile = Path.GetTempFileName();
                     File.Copy(_hc.TraceFileName, traceFile, true);
-                    Assert.Contains("Entry javascript file is not provided.", File.ReadAllText(traceFile));
+                    Assert.Contains("You are using a JavaScript Action but there is not an entry JavaScript file provided in", File.ReadAllText(traceFile));
```
ericsciple, line 849: > file provided in

in what?

PR #610, opened 2020-07-22
Clean Up Composite UI

ethanchewy:
In this PR, I remove all of the "extra" information for each run step inside of the composite action. 

It's easier to see through an example:

- Before UI: https://github.com/ethanchewy/testing-actions/runs/895584439?check_suite_focus=true (debug off)
- New UI: https://github.com/ethanchewy/testing-actions/runs/899580810 (debugging turned on)

As you can see, our existing code has the equivalent information written to the debugging console as the "extra" information node.

Before Changes:
Screen Shot 2020-07-22 at 1 52 57 PM

After Changes:
Screen Shot 2020-07-22 at 1 54 17 PM

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -52,7 +52,7 @@ namespace GitHub.Runner.Worker.Handlers
                     firstLine = firstLine.Substring(0, firstNewLine);
                 }
 
-                writeDetails($"##[group]Run {firstLine}");
+                writeDetails(ExecutionContext.InsideComposite ? $"Run {firstLine}" : $"##[group]Run {firstLine}");
             }
             else
             {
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -23,109 +23,113 @@ public sealed class ScriptHandler : Handler, IScriptHandler
 
         public override void PrintActionDetails(ActionRunStage stage)
         {
-            if (stage == ActionRunStage.Post)
+            // We don't want to display the internal workings if composite (similar/equivalent information can be found in debug)
+            if (!ExecutionContext.InsideComposite)
```
ethanchewy, line 27: All I did was wrap all the existing code in this if statement

TingluoHuang: add the `if` check to the caller of `PrintActionDetails` to reduce diff?

TingluoHuang: or should we print all these info using `ExecutionContext.Debug()` instead of `ExecutionContext.Output()`?

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -23,109 +23,113 @@ public sealed class ScriptHandler : Handler, IScriptHandler
 
         public override void PrintActionDetails(ActionRunStage stage)
         {
-            if (stage == ActionRunStage.Post)
+            // We don't want to display the internal workings if composite (similar/equivalent information can be found in debug)
```
ericsciple, line 26: var writeDetails = ExecutionContext.InsideComposite ? ExecutionContext.Debug : ExecutionContext.Output as Action<string>;

...

writeDetails($"##[group]Run {firstLine}");

ethanchewy: I tried using delegate stuff but couldn't figure it out. 

I'm just going to do this: 
```csharp
void writeDetails(string message)
            {
                if (ExecutionContext.InsideComposite)
                {
                    ExecutionContext.Debug(message);
                }
                else 
                {
                    ExecutionContext.Output(message);
                }
            }
```

PR #613, opened 2020-07-23
Add ability to register a runner to the non-default self-hosted runne…

thechrisjohnson:
This adds the ability for a runner to be registered against a non-default self-hosted runner group if one exists.

### DIFF
--- a/src/Runner.Listener/CommandSettings.cs
+++ b/src/Runner.Listener/CommandSettings.cs
@@ -42,7 +42,7 @@ namespace GitHub.Runner.Listener
             Constants.Runner.CommandLine.Args.Labels,
             Constants.Runner.CommandLine.Args.MonitorSocketAddress,
             Constants.Runner.CommandLine.Args.Name,
-            Constants.Runner.CommandLine.Args.Pool,
+            Constants.Runner.CommandLine.Args.RunnerGroup,
             Constants.Runner.CommandLine.Args.StartupType,
             Constants.Runner.CommandLine.Args.Token,
             Constants.Runner.CommandLine.Args.Url,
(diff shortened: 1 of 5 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -159,17 +159,34 @@ public async Task ConfigureAsync(CommandSettings command)
...
+                agentPool = agentPools.Where(x => string.Equals(poolName, x.Name, StringComparison.CurrentCultureIgnoreCase) && !x.IsHosted).FirstOrDefault();
+            }
+            else
+            {
+                agentPool = defaultPool;
+            }
+
+            if (agentPool == null && poolName == null)
+            {
+                throw new TaskAgentPoolNotFoundException($"Could not find any self-hosted runner group. Contact support.");
```
joshmgross, line 181:
```suggestion
                throw new TaskAgentPoolNotFoundException($"Could not find any self-hosted runner groups. Contact support.");
```

src/Runner.Listener/CommandSettings.cs
```diff
@@ -169,6 +169,15 @@ public string GetRunnerName()
                 validator: Validators.NonEmptyValidator);
         }
 
+        public string GetRunnerGroupName(string defaultPoolName = null)
+        {
+            return GetArgOrPrompt(
+                name: Constants.Runner.CommandLine.Args.RunnerGroup,
+                description: "Enter the name of the runner group to add this runner to:",
```
joshmgross, line 176: Would it be worth listing the available runner groups that the user can choose?

thechrisjohnson: So we didn't in the past. I think the problem is around how many runner groups there are in the tenant.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -159,17 +159,34 @@ public async Task ConfigureAsync(CommandSettings command)
...
-            TaskAgentPool agentPool = agentPools?.Where(x => x.IsHosted == false).FirstOrDefault();
+            TaskAgentPool defaultPool = agentPools?.Where(x => x.IsInternal).FirstOrDefault();
 
-            if (agentPool == null)
+            if (agentPools?.Where(x => !x.IsHosted).Count() > 1)
             {
-                throw new TaskAgentPoolNotFoundException($"Could not find any private pool. Contact support.");
+                poolName = command.GetRunnerGroupName(defaultPool?.Name);
+                _term.WriteLine();
+                agentPool = agentPools.Where(x => string.Equals(poolName, x.Name, StringComparison.CurrentCultureIgnoreCase) && !x.IsHosted).FirstOrDefault();
```
TingluoHuang, line 172: Why current culture?

thechrisjohnson: Whoops, good catch!

PR #615, opened 2020-07-24
GITHUB_ACTION_PATH + GITHUB_ACTION so that we can run scripts for Composite Run Steps

ethanchewy:
In this PR, we add an environment variable GITHUB_ACTION_PATH, which contains the path in the runner that contains the action code. The user will only be able to access this through `github.action_path` and `$GITHUB_ACTION_PATH`

Also in this PR, we will change GITHUB_ACTION so all nested run steps match the parent.

Demo: 
https://github.com/ethanchewy/testing-actions/runs/915248358?check_suite_focus=true

`action.yml`

```
runs:
  using: "composite"
  steps: 
    - run: chmod +x ${{ github.action_path }}/test/script2.sh
       shell: bash
    - run: chmod +x $GITHUB_ACTION_PATH/script.sh
       shell: bash
    - run: ${{ github.action_path }}/test/script2.sh
       shell: bash
    - run: $GITHUB_ACTION_PATH/script.sh
       shell: bash
```

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -98,7 +98,7 @@ namespace GitHub.Runner.Worker
         // others
         void ForceTaskComplete();
         void RegisterPostJobStep(IStep step);
-        IStep CreateCompositeStep(string scopeName, IActionRunner step, DictionaryContextData inputsData, Dictionary<string, string> envData);
+        IStep CreateCompositeStep(string scopeName, IActionRunner step, DictionaryContextData inputsData, Dictionary<string, string> envData, string actionDirectory);
     }
 
     public sealed class ExecutionContext : RunnerService, IExecutionContext
@@ -258,7 +258,8 @@ namespace GitHub.Runner.Worker
             string scopeName,
             IActionRunner step,
             DictionaryContextData inputsData,
-            Dictionary<string, string> envData)
+            Dictionary<string, string> envData,
+            string actionDirectory)
         {
             step.ExecutionContext = Root.CreateChild(_record.Id, step.DisplayName, _record.Id.ToString("N"), scopeName, step.Action.ContextName, logger: _logger, insideComposite: true, cancellationTokenSource: CancellationTokenSource.CreateLinkedTokenSource(_cancellationTokenSource.Token));
             step.ExecutionContext.ExpressionValues["inputs"] = inputsData;
@@ -275,6 +276,9 @@ namespace GitHub.Runner.Worker
                 envContext[pair.Key] = new StringContextData(pair.Value ?? string.Empty);
             }
             step.ExecutionContext.ExpressionValues["env"] = envContext;
+
+            // Set GITHUB_ACTION_PATH for each step
+            step.ExecutionContext.SetGitHubContext("action_path", actionDirectory);
 
             return step;
         }
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -98,7 +98,7 @@ namespace GitHub.Runner.Worker
         // others
         void ForceTaskComplete();
         void RegisterPostJobStep(IStep step);
-        IStep CreateCompositeStep(string scopeName, IActionRunner step, DictionaryContextData inputsData, Dictionary<string, string> envData, string actionDirectory);
+        IStep CreateCompositeStep(string scopeName, IActionRunner step, DictionaryContextData inputsData, Dictionary<string, string> envData);
     }
 
     public sealed class ExecutionContext : RunnerService, IExecutionContext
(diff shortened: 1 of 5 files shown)


ethanchewy:
oopsies we need to move this to the github context.

ethanchewy:
New issue is that the env context is empty and doesn't contain the "GITHUB_" attributes. Currently investigating...

ethanchewy:
Done. Ready for review.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -56,6 +56,9 @@ public async Task RunAsync(ActionRunStage stage)
                 childScopeName = $"__{Guid.NewGuid()}";
             }
 
+            // Set GITHUB_ACTION_PATH
+            ExecutionContext.SetGitHubContext("action_path", ActionDirectory);
```
ericsciple, line 60: Since this set's the value globally in the `github` context. We need to make sure to clear it afterward

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -177,7 +180,8 @@ private async Task RunStepsAsync(List<IStep> compositeSteps)
                 var actionStep = step as IActionRunner;
 
                 // Set GITHUB_ACTION
-                step.ExecutionContext.SetGitHubContext("action", step.ExecutionContext.GetFullyQualifiedContextName());
+                // For composite + their nested steps, we want the all of them to have the same GITHUB_ACTION
+                step.ExecutionContext.SetGitHubContext("action", ExecutionContext.GetGitHubContext("action"));
```
ericsciple, line 184: can we delete this instead?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: oh shoot totally forgot. Will do.

TingluoHuang: does this work when using job container?

ethanchewy: (I'll just set it for the step level github context)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: Yes, we can just delete this. I double checked and printed the `step.ExecutionContext.ExpressionValues["github"]` at this location and it has the same github.action value as expected.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -276,6 +277,9 @@ public void RegisterPostJobStep(IStep step)
             }
             step.ExecutionContext.ExpressionValues["env"] = envContext;
 
+            // Set GITHUB_ACTION_PATH for each step
+            step.ExecutionContext.SetGitHubContext("action_path", actionDirectory);
```
ethanchewy, line 281: we don't have to remove it now since we just set it at the run step level for each step.

ethanchewy: (this will stomp over each step [we will need to revert and then pop off the action_path]).

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: @TingluoHuang Sorry just realized my PR title was misleading. 

I was planning to support scripts for just Composite Run Steps as of now so I didn't explore if it would work for using job containers.

I think I'll explore that when we start building functionality for supporting "uses".

Right now, I haven't explored the "uses" for a container action. I'll write a note down and remember to explore this when I build functionality for that.

Thanks!

ethanchewy: (Will test after my Linux VM finishes)

ethanchewy:
Unfortunately, I'm running into segmentation faults issues, etc. while setting up runner on my ubuntu vm. 

After I test this PR on my linux version runner and make sure everything works, I'll merge this.

ethanchewy:
Weird bugs with not being able to run the runner without seg faulting. 

Trying to use a script: https://github.com/ericsciple/setup-runner to get locally make a runner but the runner still doesn't run the right build.

Testing for linux still put on hold.

ethanchewy:
works for job containers: https://github.com/ethanchewy/testing-actions/runs/924499009?check_suite_focus=true

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ethanchewy: works for job containers: https://github.com/ethanchewy/testing-actions/runs/924499009?check_suite_focus=true

Issue #629, opened 2020-07-31
ARM-based runners throw ICU error on AWS Graviton/Graviton2 instances


bdruth:
**Describe the bug**
When attempting to run `./config.sh` (or anything else) on freshly baked AWS Graviton/Graviton2 instances running Amazon Linux 2, the .NET Core runtime throws an error indicating that it `Couldn't find a valid ICU package installed on the system`. However, `libicu` is already installed, and installing the only other `icu` related packages (`icu`, `libicu-devel`) doesn't affect change in this error. The only "fix" for this appears to be setting the `configProperties` to include `"System.Globalization.Invariant": true`.

**To Reproduce**
1. Launch an AWS Graviton/Graviton2 (a1/*6g*) instance with Amazon Linux 2
2. Follow the steps to install an ARM64 action-runner from your GitHub repo's _Actions_ -> _Add Runner_

**Expected behavior**
When running `./config.sh` command, the runner configuration should complete successfully.

## Runner Version and Platform
`2.267.1`

Linux/ARM64

## What's not working?
`./config.sh` crashes.
```
Process terminated. Couldn't find a valid ICU package installed on the system. Set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support.
   at System.Environment.FailFast(System.String)
   at System.Globalization.GlobalizationMode.GetGlobalizationInvariantMode()
   at System.Globalization.GlobalizationMode..cctor()
   at System.Globalization.CultureData.CreateCultureWithInvariantData()
   at System.Globalization.CultureData.get_Invariant()
   at System.Globalization.CultureInfo..cctor()
   at System.StringComparer..cctor()
   at System.StringComparer.get_OrdinalIgnoreCase()
   at GitHub.Runner.Sdk.RunnerWebProxy..ctor()
   at GitHub.Runner.Common.HostContext..ctor(System.String, System.String)
   at GitHub.Runner.Listener.Program.Main(System.String[])
./config.sh: line 79:  1689 Aborted                 ./bin/Runner.Listener configure "$@"
```

### Workaround

Applying the following patch sets the `System.Globalization.Invariant` property in 3 `configProperties` locations.

```
patch -p1 <<ICU_PATCH
diff -Naur a/bin/Runner.Listener.runtimeconfig.json b/bin/Runner.Listener.runtimeconfig.json
--- a/bin/Runner.Listener.runtimeconfig.json	2020-07-01 02:21:09.000000000 +0000
[...]

PR #630, opened 2020-07-31
Add configuration to ignore ICU requirement.

bdruth:
See https://github.com/actions/runner/issues/629

* Three runtimes are present in the ARM actions-runner: [Listener, PluginHost, Worker]
* Use `runtimeconfig.template.json` to define the `System.Globalization.Invariant` value to true.

**Caveat**: I am no .NET Core dev, never used it, just following the docs I found [here](https://natemcmaster.com/blog/2019/01/09/netcore-primitives-3/). There's likely a better way to do this - just throwing this up as a follow-up/contribution to #629.

### DIFF
--- a/src/Runner.Listener/runtimeconfig.template.json
+++ b/src/Runner.Listener/runtimeconfig.template.json
@@ -1,0 +1,6 @@
+{
+ "$schema": "https://gist.githubusercontent.com/natemcmaster/0bdee16450f8ec1823f2c11af880ceeb/raw/runtimeconfig.template.schema.json",
+ "configProperties": {
+  "System.Globalization.Invariant": true
+ }
+}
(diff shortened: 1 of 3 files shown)


TingluoHuang:
@bdruth 
I don't think we should disable this in the config file since it's going to disable it on all platforms.
base on the doc https://docs.microsoft.com/en-us/dotnet/core/run-time-config/globalization#invariant-mode 
I think you can set an env `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1` before running `./config.sh` to run dotnet in invariant mode.

bdruth:
@TingluoHuang - thanks! I will give that a try. I agree, it's not a great solution - I'll also see if I can try to get icu support on the Graviton2, as was referenced [here](https://github.com/actions/runner/issues/629#issuecomment-667313409).

PR #640, opened 2020-08-04
fix endgroup maker

timoschilling:


TingluoHuang:
@timoschilling thanks for fixing this.

PR #641, opened 2020-08-04
Free up memory from step level outputs in composite action

ethanchewy:
Our goal is to save memory.

In this PR, we free up memory related to the composite action step, so we just remove the dictionary attribute in StepsContext which is associated with the whole composite action step.

Demo: https://github.com/ethanchewy/testing-actions/runs/946829801


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -100,6 +102,26 @@ public async Task RunAsync(ActionRunStage stage)
             }
         }
 
+        private void ClearScopes(List<IStep> compositeSteps)
```
ethanchewy, line 105: simplify + move to stepscontext

ethanchewy: done.

ethanchewy:
TODO: stomp over whole scope in one instance. 

See:
contextData:
```
 "":
   "foo1":
     "outcome": "success"
     "conclusion": "success"
     "outputs":
       "some-output": "asdf"
   "foo2": ...
 "foo1":
   "bar1":
     "outcome": ...
     "conclusion": ...
     "outputs": ...
   "bar2":
     "outcome": ...
     "conclusion": ...
     "outputs": ...

```

PR #645, opened 2020-08-05
Fix DisplayName Changing in middle of composite action run

ethanchewy:
Display Name for Composite Action changes to "run" in the middle of when the action runs. See: https://drive.google.com/file/d/15loBtzgIIOZjnLgbZtSz7tp6_H3bolCB/view?usp=sharing

In this PR, we fix it so that the DisplayName of the whole step maintains the same as the name defined in the workflow step.

The issue stems in the `ExecutionContext.CreateChild()` function which calls `ExecutionContext.InitializeTimelineRecord` which modifies the parent node display name **since they share the same record id**.

So, in our solution, we only run `QueueTimelineRecordUpdate` if the scope name is null (aka the step is a job step running in the workflow).

ex: https://github.com/ethanchewy/testing-actions/runs/950260297?check_suite_focus=true


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -856,7 +856,11 @@ private void InitializeTimelineRecord(Guid timelineId, Guid timelineRecordId, Gu
             var configuration = HostContext.GetService<IConfigurationStore>();
             _record.WorkerName = configuration.GetSettings().AgentName;
 
-            _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+            // If we are inside an action, we don't want the steps inside to update the displayname of the parent so we pass the same parent displayname.
```
TingluoHuang, line 859: should we not call `InitializeTimelineRecord` when inside an action?

ethanchewy: @TingluoHuang we need to call it so that the step info is shown while the step is running. 

If you don't call `InitializeTimelineRecord`, the composite action node will not expand with the steps' results while it's running.

ethanchewy: (resolved in zoom call)

Issue #510, opened 2020-05-30
Ephemeral (single use) runner registrations


MatisseHack:
**Describe the bug**
When starting a self hosted runner with `./run.cmd --once`, the runner sometimes accepts a second job before shutting down, which causes that second job to fail with the message:

```
The runner: [runner-name] lost communication with the server. Verify the machine is running and has a healthy network connection.
```

This looks like the same issue recently fixed here: microsoft/azure-pipelines-agent#2728

**To Reproduce**
1. Create a repo, enable GitHub Actions, and add a new workflow
1. Configure a new runner on your machine
1. Run the runner with `./run.cmd --once`
1. Queue two runs of your workflow
1. The first job will run and the runner will go offline
1. (Optionally) configure and start a second runner
1. The second job will time out after several minutes with the message:

    ```
    The runner: [runner-name] lost communication with the server. Verify the machine is running and has a healthy network connection.
    ```

    (where `[runner-name]` is the name of the first runner)

1. Also: trying to remove the first runner with the command `./config.cmd remove --token [token]` will result in the following error until the second job times out:

    ```
    Failed: Removing runner from the server
    Runner "[runner-name]" is running a job for pool "Default"
    ```

**Expected behavior**
The second job should run on (and wait for) any new runner that comes online rather than try to run as a second job on the, now offline, original runner.

## Runner Version and Platform
2.262.1 on Windows

## Runner and Worker's Diagnostic Logs
[_diag.zip](https://github.com/actions/runner/files/4704906/_diag.zip)

TingluoHuang:
@MatisseHack 
How did you find the flag? 😆 
the run once flag `--once` is not in the official actions doc and better not be used in production, there is a known issue that we haven't fix. 😄

bryanmacfarlane:
Yeah, this is why it's not part of the documentation or command line help.  Because it doesn't work and has races so we didn't ship it.  

We have work on the backlog to register a runner with the service as ephemeral.  It has to be a service side feature so the service is cooperative and doesn't assign it another job in the window when the runner or outside / in automation decides to tear it down.

MatisseHack:
I've used the `--once` flag extensively with Azure Pipelines to create an automatically scaling pool of single use agents. I'm trying to convert that system to work with GitHub Actions. I was pleased to find that the flag still worked (at least partially)!

rclmenezes:
Please support `--once` when you have a chance!

[...]

PR #660, opened 2020-08-14
Support configure runner as ephemeral.

TingluoHuang:
The service will make sure to only ever send one job to this runner.
The service will remove the runner registration from service after the job finish.

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -125,9 +125,10 @@ namespace GitHub.Runner.Common
                 {
                     public static readonly string Check = "check";
                     public static readonly string Commit = "commit";
+                    public static readonly string Ephemeral = "ephemeral";
                     public static readonly string Help = "help";
                     public static readonly string Replace = "replace";
-                    public static readonly string Once = "once";
+                    public static readonly string Once = "once"; // TODO: Remove in 10/2021
                     public static readonly string RunAsService = "runasservice";
                     public static readonly string Unattended = "unattended";
                     public static readonly string Version = "version";
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -194,6 +194,7 @@ namespace GitHub.Runner.Listener.Configuration
             TaskAgent agent;
             while (true)
             {
+                runnerSettings.Ephemeral = command.Ephemeral;
                 runnerSettings.AgentName = command.GetRunnerName();
 
                 _term.WriteLine();
@@ -210,7 +211,7 @@ namespace GitHub.Runner.Listener.Configuration
                     if (command.GetReplace())
                     {
                         // Update existing agent with new PublicKey, agent version.
-                        agent = UpdateExistingAgent(agent, publicKey, userLabels);
+                        agent = UpdateExistingAgent(agent, publicKey, userLabels, runnerSettings.Ephemeral);
 
                         try
                         {
@@ -233,7 +234,7 @@ namespace GitHub.Runner.Listener.Configuration
                 else
                 {
                     // Create a new agent.
-                    agent = CreateNewAgent(runnerSettings.AgentName, publicKey, userLabels);
+                    agent = CreateNewAgent(runnerSettings.AgentName, publicKey, userLabels, runnerSettings.Ephemeral);
 
                     try
                     {
@@ -458,7 +459,7 @@ namespace GitHub.Runner.Listener.Configuration
         }
 
 
-        private TaskAgent UpdateExistingAgent(TaskAgent agent, RSAParameters publicKey, ISet<string> userLabels)
+        private TaskAgent UpdateExistingAgent(TaskAgent agent, RSAParameters publicKey, ISet<string> userLabels, bool ephemeral)
         {
             ArgUtil.NotNull(agent, nameof(agent));
             agent.Authorization = new TaskAgentAuthorization
@@ -469,6 +470,8 @@ namespace GitHub.Runner.Listener.Configuration
             // update should replace the existing labels
             agent.Version = BuildConstants.RunnerPackage.Version;
             agent.OSDescription = RuntimeInformation.OSDescription;
+            agent.Ephemeral = ephemeral;
+            agent.MaxParallelism = 1;
 
             agent.Labels.Clear();
 
@@ -484,7 +487,7 @@ namespace GitHub.Runner.Listener.Configuration
             return agent;
         }
 
-        private TaskAgent CreateNewAgent(string agentName, RSAParameters publicKey, ISet<string> userLabels)
+        private TaskAgent CreateNewAgent(string agentName, RSAParameters publicKey, ISet<string> userLabels, bool ephemeral)
         {
             TaskAgent agent = new TaskAgent(agentName)
             {
@@ -495,6 +498,7 @@ namespace GitHub.Runner.Listener.Configuration
                 MaxParallelism = 1,
                 Version = BuildConstants.RunnerPackage.Version,
                 OSDescription = RuntimeInformation.OSDescription,
+                Ephemeral = ephemeral,
             };
 
             agent.Labels.Add(new AgentLabel("self-hosted", LabelType.System));
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -234,7 +234,7 @@ namespace GitHub.Runner.Listener
                     HostContext.StartupType = startType;
 
                     // Run the runner interactively or as service
-                    return await RunAsync(settings, command.RunOnce);
+                    return await RunAsync(settings, command.RunOnce || settings.Ephemeral);  // TODO: Remove RunOnce later.
                 }
                 else
                 {
@@ -512,7 +512,9 @@ Config Options:
  --labels string        Extra labels in addition to the default: 'self-hosted,{Constants.Runner.Platform},{Constants.Runner.PlatformArchitecture}'
  --work string          Relative runner work directory (default {Constants.Path.WorkDirectory})
  --replace              Replace any existing runner with the same name (default false)
- --pat                  GitHub personal access token used for checking network connectivity when executing `.{separator}run.{ext} --check`");
+ --pat                  GitHub personal access token used for checking network connectivity when executing `.{separator}run.{ext} --check`
+ --ephemeral            Configure the runner to only take one job and then let the service un-configure the runner after the job finishes (default false)");
+
 #if OS_WINDOWS
     _term.WriteLine($@" --runasservice   Run the runner as a service");
     _term.WriteLine($@" --windowslogonaccount string   Account to run the service as. Requires runasservice");
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -466,8 +466,16 @@ namespace GitHub.Runner.Listener
                         await jobDispatcher.ShutdownAsync();
                     }
 
-                    //TODO: make sure we don't mask more important exception
-                    await _listener.DeleteSessionAsync();
+                    try
+                    {
+                        await _listener.DeleteSessionAsync();
+                    }
+                    catch (Exception ex) when (runOnce)
+                    {
+                        // ignore exception during delete session for ephemeral runner since the runner might already be deleted from the server side
+                        // and the delete session call will ends up with 401.
+                        Trace.Info($"Ignore any exception during DeleteSession for an ephemeral runner. {ex}");
+                    }
 
                     messageQueueLoopTokenSource.Dispose();
                 }
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -34,6 +34,9 @@ namespace GitHub.Runner.Common
         public string PoolName { get; set; }
 
         [DataMember(EmitDefaultValue = false)]
+        public bool Ephemeral { get; set; }
+
+        [DataMember(EmitDefaultValue = false)]
         public string ServerUrl { get; set; }
 
         [DataMember(EmitDefaultValue = false)]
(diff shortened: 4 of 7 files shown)


TingluoHuang:
https://github.com/actions/runner/issues/510

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -468,6 +469,8 @@ private TaskAgent UpdateExistingAgent(TaskAgent agent, RSAParameters publicKey,
             // update should replace the existing labels
             agent.Version = BuildConstants.RunnerPackage.Version;
             agent.OSDescription = RuntimeInformation.OSDescription;
+            agent.Ephemeral = ephemeral;
```
ericsciple, line 472: Curious when the service will fully support the behavior - i.e. has the service code rolled out yet?

Also curious whether it will work on GHES? Did the server-side code make it into GHES. If not, curious if we can detect/error during runner config.

Also are we ready to document and forever support the new behavior?

ericsciple (dismissed):

src/Runner.Listener/Configuration/ConfigurationManager.cs
ericsciple: I confirmed the server changes did not make into 2.22

lokesh755 (dismissed):

src/Runner.Listener/Configuration/ConfigurationManager.cs
hross: note the service behavior will go into GA @ericsciple

hross:
If someone uses this on a beta version of GHES how will we handle error messages?

Shegox:
Hi, this looks like a very great change and really looking forward to that.
I would actually have a question and hope it is fine to directly ask it here. 
We're running currently the GHES 2.22.0 beta and also looking into providing runners on a large scale for the enterprise account. 
If I understand ephemeral runners correctly, they will be registered just like normal runners, but the big difference being that they only receive a single job. Meaning that we could execute untrusted code inside the self-hosted runner and the untrusted code wouldn't be able to extract credentials to get another job and potentially steal their GitHub secrets?

Would perhaps also be good to extend the [docs/design/auth.md](https://github.com/actions/runner/blob/main/docs/design/auth.md) with such information. 

Already many thanks in advance and if you need someone testing that on GHES, I would be happy to help.

Temtaime:
It is a very limited solution for creating fresh environments.
It doesn't provide a way for creating VM with specific configuration.
More elegant solution is create custom executor like how it is done in gitlab.
https://docs.gitlab.com/runner/executors/custom.html

dakale:
@TingluoHuang I tried this out, and one thing I found is that the process doesnt seem to exit if the runner was auto updated prior to running its one job. Is that something you are aware of?

shwuhk:
May I know how to use ephemeral with run.sh/runsvc.sh now?

Dids:
What's the status on this? I'm assuming this is still waiting for server-side changes, if so, is that publicly being tracked anywhere?

I've been working around the "single use" self-hosted runner issues by creating an orchestrator of sorts, which keeps N amount of runners running (all running inside a Docker container) with the `--once` flag, then destroys and (de-)registers them when the jobs are done.

This has been fairly unreliably for several reasons:
- If the amount of queued jobs exceeds the amount of runners, jobs will end up timing out because they were assigned to a non-existent runner (sometimes takes 24 hours for them to timeout)
- Having a `--once` runner running for long periods of time, and especially when there are connection issues during that time, will leave the runner in some kind of inconsistent state, where it is unable to accept new jobs and sometimes even shows up as offline on GH's side

The upside is that this has provided a very nice way to provide semi-isolated environments for runners/jobs, as each runner would run in a fresh Docker container, but with the downside of additional action containers running on the same host.

FloThinksPi:
Also interested in an answer to @Shegox question.

Given one is running unknown code on a runner. To safely run unknown code we`d like to reset a VM to a Snapshot after every run. To do so one can run on an ephemeral runner. As the unknown code has root permissions there (to access docker or install packages) the assumption is this code could also alter the runner process. 

Is it guaranteed that with the access token of the ephemeral runner, malicious code on this runner can not initiate another workflow to be scheduled on his instance ? E.g. malicious workflows could extract api tokens from the runner and start a second runner process to pick up another workflow and be able to extract secrets from that workflow.

As of https://github.com/actions/runner/blob/main/docs/design/auth.md i would expect the runner can pick up an additional workflow as long as its token is valid ? In above case this would mean as long as the first workflow is running and did not finish, the ephemeral runners token is valid and malicious code would be able to use that token to fetch more Workflow jobs and extract secrets from them ?

sethvargo:
@bryanmacfarlane @TingluoHuang 👋 could you please provide an update on whether this will be merged (or general support for ephemeral runners in general)?

lokesh755:
@sethvargo We're actively working on this. We should merge this sometime this month probably sooner :)

thinkafterbefore:
@lokesh755 @TingluoHuang Any updates to when this will hit production?

src/Runner.Listener/Runner.cs
```diff
@@ -469,7 +469,8 @@ private void PrintUsage(CommandSettings command)
  --runnergroup string   Name of the runner group to add this runner to (defaults to the default runner group)
  --labels string        Extra labels in addition to the default: 'self-hosted,{Constants.Runner.Platform},{Constants.Runner.PlatformArchitecture}'
  --work string          Relative runner work directory (default {Constants.Path.WorkDirectory})
- --replace              Replace any existing runner with the same name (default false)");
+ --replace              Replace any existing runner with the same name (default false)
+ --ephemeral            Configure the runner only take one job and let service un-configured the runner after the job finish (default false)");
```
martin389, line 473: 👋 Small suggestion here:

```suggestion
 --ephemeral            Configure the runner to only take one job and then let the service un-configure the runner after the job finishes (default false)");
```

pje: Thanks, @martin389!

src/Runner.Common/Constants.cs
```diff
@@ -125,9 +125,10 @@ public static class Flags
                 {
                     public static readonly string Check = "check";
                     public static readonly string Commit = "commit";
+                    public static readonly string Ephemeral = "ephemeral";
                     public static readonly string Help = "help";
                     public static readonly string Replace = "replace";
-                    public static readonly string Once = "once";
+                    public static readonly string Once = "once"; // TODO: Remove in 10/2020
```
paride, line 131: That comment was a bit optimistic :-) Bump it to 10/2021?

seemethere (approved):

seemethere (dismissed):

chaychoong:
@TingluoHuang @hross on this PR and this comment here: https://github.com/actions-runner-controller/actions-runner-controller/issues/470#issuecomment-841428853

Do you guys have a rough estimate on when this feature is expected to be merged? I understand that there's some delays to be expected; an updated timeline would help a lot for us enterprise folks who are trying to iron out the kinks when autoscaling runners on k8s.

seemethere (approved):

pje:
> @TingluoHuang @hross on this PR and this comment here: [actions-runner-controller/actions-runner-controller#470 (comment)](https://github.com/actions-runner-controller/actions-runner-controller/issues/470#issuecomment-841428853)
> 
> Do you guys have a rough estimate on when this feature is expected to be merged? I understand that there's some delays to be expected; an updated timeline would mean a lot for us enterprise folks who are trying to iron out the kinks when autoscaling runners on k8s.

We expect to see all the pieces launch in September!

seemethere (approved):

pje (approved):

joeyparrish:
This doesn't work for me.  I tried the v2.282.0 release, and I found that:

1. The runner accepted multiple jobs instead of just one
2. The runner did not deregister itself after it became idle

Is this feature not ready yet?  Please don't remove `--once` until this is resolved, because `--ephemeral` is not yet usable as far as I can tell.

thboop:
@joeyparrish are you using `github.com` or a GHES/GHAE instance?

Could you provide more details on the repo you tried this? This feature should be available and working on `github.com`. Please follow up with a bug if that is not the case

joeyparrish:
Actually, let me double-check my usage.  I'm suspicious that my issues are my fault.  Makes more sense.

joeyparrish:
Is this flag meant for config.sh or run.sh?

Shegox:
It is a configuration option for `config.sh` (as I assume the server then as well only provides it with one-job credentials): 
```
$ ./config.sh --help
...
Config Options:
...
 --ephemeral            Configure the runner to only take one job and then let the service un-configure the runner after the job finishes (default false)
 ```

**Comment by thboop** [comment]
> Is this flag meant for config.sh or run.sh?

The flag is meant to be run during `config.sh` on mac/linux. We should have official docs for these on github.com soon, which hopefully will make this easier, but feel free to use this feature now!

**Comment by joeyparrish** [comment]
Thank you!  I was adding it to `run.sh` as I had with `--once`.

**Comment by tbando** [comment]
Can I already use `--ephemeral` feature on self-hosted GHES?  Or need to wait for newer release of GHES?

**Comment by joeyparrish** [comment]
This shipped in the latest release, v2.282.0.

**Comment by tbando** [comment]
Oh, I would like to know if I need to update server-side GHES as I got `Internal Error` when I issued `.config.sh --ephemeral`.

**Comment by MichaelJJ** [comment]
With the new --ephemeral flag, is there a way to have the config.sh wait until the runner has de-registered?  As an example, if I make a docker image that runs a shell script on startup to register an ephemeral runner, what is the best way to have the script wait until the runner is done so the container doesn't exit?

**Comment by joeyparrish** [comment]
I use a shell script as a docker entrypoint, which calls `config.sh --ephemeral` followed by `run.sh`, then terminates.  I wrap that in a systemd service that removes and restarts the docker container automatically.

An older version of this (based on `--once`) is currently available at https://github.com/myoung34/docker-github-actions-runner#ephemeral-mode

I'm working on a PR to update that to use `--ephemeral` instead.

**Comment by zetaab** [comment]
This PR currently breaks github actions in GHES. --once does not work anymore and --ephemeral not supported great

**Comment by TingluoHuang** [comment]
@zetaab I don't think we changed any behavior for `--once` in this PR, what exactly error/issue did you are run into on GHES?

**Comment by rofafor** [comment]
`once` was removed from valid flags resulting `Unrecognized command-line input arguments: 'once'. `

**Comment by TingluoHuang** [comment]
I think that should only give you a warning but not actually fail anything.

**Comment by rofafor** [comment]
I'm getting `An Internal Error Occurred. Activity Id: ...` errors when enabling `--ephemeral` against our GHES 3.0 / 3.1.

**Comment by TingluoHuang** [comment]
`--ephemeral ` does not support GHES

**Comment by aidan-mundy** [comment]
@TingluoHuang As far as I can tell, --once is no longer usable with this update. As @rofafor said, it is no longer accepted as a valid flag. See https://github.com/actions/runner/pull/660/files#diff-b1f59ae3d34d9d3811ce43ed0214576cb4d9f3373a6734adf1318b5ab7e535eeL35

**Comment by zetaab** [comment]
Like @rofafor said: I think your idea was to deprecate flag, but you actually removed it also. So now the problem is that --once does not work anymore. When using GHES --ephemeral does not work either.

**Comment by thboop** [comment]
@aidan-mundy , @zetaab can you confirm that you are unable to use the `--once` flag when configuring the newest runner? You may see an error saying the flag is not available (which is intended, we want people to eventually move off of it), but the flag still works.

If it doesn't work, please file an issue and provide your runner version and os.

**Comment by TingluoHuang** [comment]
Here is what i just tried.
```
ting@htl-mac _layout % ./run.sh --once
Unrecognized command-line input arguments: 'once'. For usage refer to: .\config.cmd --help or ./config.sh --help

√ Connected to GitHub

2021-09-21 13:19:26Z: Listening for Jobs
2021-09-21 13:20:15Z: Running job: build
2021-09-21 13:20:18Z: Job build completed with result: Succeeded
ting@htl-mac _layout % $?
zsh: command not found: 0
```

We do print out an error but the flag is no longer recognized, but the runner is still able to connect to the server and run a single job, and exit.

Do I miss something here?

rofafor:
The `--once` seems to be working despite the warning message. According to code comments, you've scheduled to remove the once in 10/2021 - what happens to GHES after that? No more ephemeral runners?

TingluoHuang:
@rofafor 
We have not decided when to really remove `--once`, given the fact that so many customers depend on it today and have various reasons can't leverage `--ephemeral`, ex: GHES.

We will keep the `--once` around for a long time until everyone is good to move off.

TingluoHuang:
I created a PR to update the comment to make it less confusing. https://github.com/actions/runner/pull/1360

@thboop ☝️

hross:
I want to add a summary here so it's obvious if you land here wondering about `--once`:

* We don't plan to deprecate this command any time soon and will give notification before we do so. We realize customers still use it (you will receive the warning, though)
* We recommend that you stop using `--once` and start using `--ephemeral` (except on current versions of GHES). The reason is that this is a server side change to ensure there are not race conditions with job assignment. `--once` was not "officially supported" and is client side, which exposes you to the risk of multiple job assignment.
* `--ephemeral` will ship in the next version of GHES (but as I said above it requires server changes to fix the race condition with client side only assignment)

If you have any issues with ephemeral/once please feel free to reach out (this issue works but you can also use the [community support forms](https://github.community/c/code-to-cloud/github-actions/41) which might have better support for customer questions and let us file support tickets to help you).

Shegox:
Disclaimer: not a GitHub employee

GitHub normally releases feature only in minor (3.x) releases and not in patch releases (3.2.x). So I wouldn't actually expect it before 3.3.x (and maybe even later, but thats up to GitHub to confirm). 

[The GitHub roadmap](https://github.com/github/roadmap/projects/1?card_filter_query=label%3Aserver) currently doesn't specify any concrete date for it.

aidan-mundy:
For those of you that are enterprise server users and are waiting for this functionality, 3.3.0.rc1 is now available for preview. It includes the `--ephemeral` flag and a number of other neat features/changes.

Manouchehri:
Is there an easy way to run a command after `--ephemeral` has finished one job?

sethvargo:
@Manouchehri I've achieved this by running ephemeral under systemd and then using a `ExecStop` or `ExecStopPost`.

PR #684, opened 2020-08-27
Add Runner File Commands

thboop:


src/Runner.Worker/ActionRunner.cs
```diff
@@ -145,6 +145,12 @@ public async Task RunAsync()
                 stepHost = containerStepHost;
             }
 
+            // Setup File Command Manager
+            var fileCommandManager = HostContext.CreateService<IFileCommandManager>();
+            // Container Action Handler will handle the conversion for Container Actions
+            var container = handlerData.ExecutionType == ActionExecutionType.Container ? null : ExecutionContext.Global.Container;
```
ericsciple, line 151: let's be consistent with existing bugs and add this as the host-path always.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -161,16 +161,21 @@ public async Task RunAsync(ActionRunStage stage)
...
 
             container.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
             container.MountVolumes.Add(new MountVolume(tempHomeDirectory, "/github/home"));
             container.MountVolumes.Add(new MountVolume(tempWorkflowDirectory, "/github/workflow"));
+            container.MountVolumes.Add(new MountVolume(tempFileCommandDirectory, "/github/file_commands"));
             container.MountVolumes.Add(new MountVolume(defaultWorkingDirectory, "/github/workspace"));
 
             container.AddPathTranslateMapping(tempHomeDirectory, "/github/home");
             container.AddPathTranslateMapping(tempWorkflowDirectory, "/github/workflow");
+            container.AddPathTranslateMapping(tempFileCommandDirectory, "/github/file_commands");
```
ericsciple, line 178: nit `file-commands`

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -161,16 +161,21 @@ public async Task RunAsync(ActionRunStage stage)
             Directory.CreateDirectory(tempHomeDirectory);
             this.Environment["HOME"] = tempHomeDirectory;
 
+            var tempFileCommandDirectory = Path.Combine(tempDirectory, "_runner_file_commands");
```
ericsciple, line 164: nit: `_runner-file-commands`?

src/Runner.Worker/Handlers/ContainerActionHandler.cs
ericsciple: doesnt matter really, temp file anyway

src/Runner.Worker/GitHubContext.cs
```diff
@@ -13,11 +13,13 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "actor",
             "api_url",
             "base_ref",
+            "env",
```
ericsciple, line 16: can we change to `allow` terminology (existing code above)

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -0,0 +1,140 @@
...
+                return false;
+            }
+            return true;
+        }
+    }
+
+    public interface IFileCommandExtension : IExtension
+    {
+        string ContextName { get; }
+        string FileName { get; }
```
ericsciple, line 95: consider `FilePrefix`

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        public void TryProcessFiles(IExecutionContext context, ContainerInfo container)
+        {
+            foreach (var fileCommand in _commandExtensions)
+            {
+                fileCommand.ProcessCommand(context, Path.Combine(_fileCommandDirectory, fileCommand.FileName + _fileSuffix),container);
+            }
+        }
+
+        private bool TryDeleteFile(string path)
```
ericsciple, line 73: `IOUtil`

src/Runner.Worker/ActionRunner.cs
```diff
@@ -239,6 +245,8 @@ public async Task RunAsync()
 
             // Run the task.
             await handler.RunAsync(Stage);
+            fileCommandManager.TryProcessFiles(ExecutionContext, ExecutionContext.Global.Container);
```
ericsciple, line 248: What if the step fail? Do the file-commands still get processed?

For example, esp consider continue-on-error (script exits 1).

src/Runner.Worker/ActionRunner.cs
```diff
@@ -145,6 +145,12 @@ public async Task RunAsync()
                 stepHost = containerStepHost;
             }
 
+            // Setup File Command Manager
```
ericsciple, line 148: check whether job extension step (initailize job) uses action runner

ericsciple: may not matter if it doesnt set env/path

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    TryDeleteFile(oldPath);
+                }
+
+                var newPath = Path.Combine(_fileCommandDirectory, fileCommand.FileName + _fileSuffix);
+                TryDeleteFile(newPath);
+                File.Create(newPath).Dispose();
+
+                var pathToSet = container != null ? container.TranslateToContainerPath(newPath) : newPath; 
+                context.SetGitHubContext(fileCommand.ContextName, pathToSet);
```
ericsciple, line 61: nit: would be rad if more explicitly set in ActionRunner. Hard to mentally keep track where the global context is changed. Today it's only StepsRunner/CompositeActionRunner. May be drawback of plugin abstraction.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+
+        public void TryProcessFiles(IExecutionContext context, ContainerInfo container)
```
ericsciple, line 65: nit: Do, do not try :)

src/Runner.Worker/ActionRunner.cs
ericsciple: May need try/finally. If we move into finally, need to make sure doesn't clobber original exception

ericsciple: if it fails, set task result failed

src/Runner.Common/ExtensionManager.cs
```diff
@@ -56,6 +56,10 @@ public sealed class ExtensionManager : RunnerService, IExtensionManager
                     Add<T>(extensions, "GitHub.Runner.Worker.EndGroupCommandExtension, Runner.Worker");
                     Add<T>(extensions, "GitHub.Runner.Worker.EchoCommandExtension, Runner.Worker");
                     break;
+                case "GitHub.Runner.Worker.IFileCommandExtension":
```
ericsciple, line 59: nit: personally don't like extension :P

nit: technically not "commands". Wonder if different naming would be more intuitive (less overloading)

PR #694, opened 2020-09-03
Allow registry credentials for job/service containers

dakale:
This lets you specify credentials for job and service containers, with the following example syntax:

```yaml
jobs:
  private-container:
    runs-on: [ self-hosted ]
    env:
      USER: username
    services:
      myservice: 
        image: ghcr.io/OWNER/NAME
        credentials:
          username: ${{ env.USER }}
          password: ${{ secrets.GHCR_TOKEN }}
    container:
      image: ghcr.io/OWNER/NAME
      credentials:
        username: ${{ github.actor }}
        password: ${{ secrets.GHCR_TOKEN }}
    steps:
    - run: echo hi
```

Where values can come from `github`,`env`, or `secrets` contexts.

Mechanically, it just runs `docker login` before running `docker pull`, creating a temporary, container-specific docker config file, which is deleted after the pull

### DIFF
--- a/src/Runner.Worker/Container/ContainerInfo.cs
+++ b/src/Runner.Worker/Container/ContainerInfo.cs
@@ -34,6 +34,9 @@ namespace GitHub.Runner.Worker.Container
             _environmentVariables = container.Environment;
             this.IsJobContainer = isJobContainer;
             this.ContainerNetworkAlias = networkAlias;
+            this.RegistryAuthUsername = container.Credentials?.Username;
+            this.RegistryAuthPassword = container.Credentials?.Password;
+            this.RegistryServer = ParseRegistryServer(this.ContainerImage);
 
 #if OS_WINDOWS
             _pathMappings.Add(new PathMapping(hostContext.GetDirectory(WellKnownDirectory.Work), "C:\\__w"));
@@ -79,6 +82,9 @@ namespace GitHub.Runner.Worker.Container
         public string ContainerWorkDirectory { get; set; }
         public string ContainerCreateOptions { get; private set; }
         public string ContainerRuntimePath { get; set; }
+        public string RegistryServer { get; set; }
+        public string RegistryAuthUsername { get; set; }
+        public string RegistryAuthPassword { get; set; }
         public bool IsJobContainer { get; set; }
 
         public IDictionary<string, string> ContainerEnvironmentVariables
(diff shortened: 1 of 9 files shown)


src/Runner.Worker/Container/DockerCommandManager.cs
```diff
@@ -346,6 +358,31 @@ public async Task<List<PortMapping>> DockerPort(IExecutionContext context, strin
...
+        {
+            string args = $"--config {configFileDirectory} login {registry} -u {username} --password-stdin";
+            context.Command($"{DockerPath} {args}");
+
+            var input = Channel.CreateBounded<string>(new BoundedChannelOptions(1) { SingleReader = true, SingleWriter = true });
+            input.Writer.TryWrite(password);
+
+            var processInvoker = HostContext.CreateService<IProcessInvoker>();
+
+            var task = processInvoker.ExecuteAsync(
```
TingluoHuang, line 371: you can just `return processInvoker.ExecuteAsync()` and remove `async` from the method name.

src/Runner.Worker/Container/DockerCommandManager.cs
```diff
@@ -82,9 +85,18 @@ public async Task<DockerVersion> DockerVersion(IExecutionContext context)
...
+            return DockerPull(context, image, null);
+        }
+
+        public async Task<int> DockerPull(IExecutionContext context, string image, string configFileDirectory)
+        {
+            if (string.IsNullOrEmpty(configFileDirectory))
+            {
+                return await ExecuteDockerCommandAsync(context, $"pull", image, context.CancellationToken);
+            }
+            return await ExecuteDockerCommandAsync(context, $"--config {configFileDirectory} pull", image, context.CancellationToken);
```
TingluoHuang, line 99: do we need to merge the config file with the default in $HOME to get any system-wide config?

dakale: I dont think so. For self-hosted where there might be a config, you can just not specify credentials in the workflow and it will continue working the exact same, ie the pull will use global/default config

src/Runner.Worker/Container/ContainerInfo.cs
```diff
@@ -226,6 +232,16 @@ public void AddPathTranslateMapping(string hostCommonPath, string containerCommo
             _pathMappings.Insert(0, new PathMapping(hostCommonPath, containerCommonPath));
         }
 
+        private string ParseRegistryServer(string image)
```
dakale, line 235: This doesnt actually work. I need to update this and add tests

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -198,12 +198,18 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
                 }
             }
 
+            // TODO: Add at a later date. This currently no local package registry to test with
+            // UpdateRegistryAuthForGitHubToken(executionContext, container);
```
TingluoHuang, line 202: finish todo?

TingluoHuang (approved):

src/Runner.Worker/ContainerOperationProvider.cs
dakale: This would either be added in the near future, or Id remove the TODO if we decide to not do the feature. Basically I couldnt figure out how to verify the change because I dont know how to run a local/development package registry. Since we (at least I do) test certain changes with a development runner against prod actions, I was going to test and finally add this feature back in once actions service changes were deployed 

A quick search looked like we have some TODOs in the runner code, so I didnt feel like it was breaking any established pattern

Ill open a small PR resolving this particular TODO soon

PR #696, opened 2020-09-08
log runner group name

hfaulds:
This PR logs the runner group name in the Set up job step.

This variable isn't being sent yet but this PR should work fine without it (it'll have no affect) and then once it starts getting sent it'll start being logged.

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
@@ -74,6 +74,10 @@ namespace GitHub.Runner.Worker
                         {
                             // print out HostName for self-hosted runner
                             context.Output($"Runner name: '{setting.AgentName}'");
+                            if (message.Variables.TryGetValue("system.runnerGroupName", out VariableValue runnerGroupName))
+                            {
+                              context.Output($"Runner group name: '{runnerGroupName}'");
+                            }
                             context.Output($"Machine name: '{Environment.MachineName}'");
                         }
                     }
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -76,7 +76,7 @@ namespace GitHub.Runner.Worker
                             context.Output($"Runner name: '{setting.AgentName}'");
                             if (message.Variables.TryGetValue("system.runnerGroupName", out VariableValue runnerGroupName))
                             {
-                              context.Output($"Runner group name: '{runnerGroupName}'");
+                                context.Output($"Runner group name: '{runnerGroupName.Value}'");
                             }
                             context.Output($"Machine name: '{Environment.MachineName}'");
                         }


src/Runner.Worker/JobExtension.cs
```diff
@@ -74,6 +74,11 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         {
                             // print out HostName for self-hosted runner
                             context.Output($"Runner name: '{setting.AgentName}'");
+                            VariableValue runnerGroupName = "";
+                            if (message.Variables.TryGetValue("RunnerGroupName", out runnerGroupName))
```
juliobbv, line 78:
```
if (message.Variables.TryGetValue("RunnerGroupName", out VariableValue runnerGroupName))
```

TingluoHuang: `if (message.Variables.TryGetValue("RunnerGroupName", out var runnerGroupName))`

TingluoHuang: we normally namespace these kinds of variables, like `system.runnergroupname`

src/Runner.Worker/JobExtension.cs
```diff
@@ -74,6 +74,10 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         {
                             // print out HostName for self-hosted runner
                             context.Output($"Runner name: '{setting.AgentName}'");
+                            if (message.Variables.TryGetValue("system.runnerGroupName", out VariableValue runnerGroupName))
+                            {
+                              context.Output($"Runner group name: '{runnerGroupName}'");
```
ericsciple, line 79: nit: indent using 4 spaces :)

src/Runner.Worker/JobExtension.cs
```diff
@@ -74,6 +74,10 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         {
                             // print out HostName for self-hosted runner
                             context.Output($"Runner name: '{setting.AgentName}'");
+                            if (message.Variables.TryGetValue("system.runnerGroupName", out VariableValue runnerGroupName))
```
ericsciple, line 77: do you need to check `!string.IsNullOrEmpty(runnerGroupName.Value)`

src/Runner.Worker/JobExtension.cs
ericsciple, line 79: i think you might need to do:

```csharp
context.Output($"Runner group name: '{runnerGroupName.Value}'");
```

otherwise i think it will always output `Runner group name: 'Object'` since VariableValue does not override ToString()

ericsciple (approved):

src/Runner.Worker/JobExtension.cs
hfaulds: We outright don't set this variable on the payload so I think this should be sufficient. @lokesh755 is going to test so should be able to say

src/Runner.Worker/JobExtension.cs
hfaulds: Good spot!

hfaulds:
I tested this. For organizations and enterprises it works great. For repository runners it always logs `Default` as the runner group name. I confirmed this is acceptable so am merging.

thboop (approved):
LGTM

PR #714, opened 2020-09-17
Expose retention days in env for toolkit/artifacts package

yacaovsnc:
We want to enable custom retention days for artifacts.  Artifacts are backed by the `toolkit/artifacts` package so we need to expose this value as an environment variable, and then the `artifacts` package will be able to consume it.

### DIFF
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -23,6 +23,7 @@ namespace GitHub.Runner.Worker
             "ref",
             "repository",
             "repository_owner",
+            "retention_days",
             "run_id",
             "run_number",
             "server_url",


ericsciple (approved):

src/Runner.Worker/GitHubContext.cs
```diff
@@ -23,6 +23,7 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "ref",
             "repository",
             "repository_owner",
+            "retention_days",
```
TingluoHuang, line 26: This means a environment variable GITHUB_RETENTION_DAYS will shows up, what retention is this for? Workflow run? Log? Artifacts?

yacaovsnc: This is "Artifacts & Logs Retention" on UI.  We will store this value on each run, so it's effectively the run retention.  This value will be consumed by the artifact toolkit to verify user do not enter a larger value.  This value is also checked on service layer.

PR #722, opened 2020-09-28
Make it possible to set the runner's root dir through GITHUB_ACTIONS_ROOT_DIR env var

rickynils:
I've successfully used this patch for using the runner installed in a non-writeable location. Specifically, this allows the runner to be packaged and installed with the Nix (https://nixos.org/) package manager.


navagjonaj188-ops (commented):
#3724

PR #731, opened 2020-10-01
Notify on unsecure commands

thboop:
The purpose of this PR is to add telemetry and notify PR's when they are using the [unsecure commands in the runner](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/). Eventually, we will need to disable these commands, but getting people to migrate off of them is the first step. This pr will create annotations on runs where these commands are used. For example:
image

### Testing
- [x] MacOS
- [x] Ubuntu
- [x] Ubuntu Job container
- [x] Ubuntu container action
- [x] GHES (In progress)

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -142,7 +142,7 @@ namespace GitHub.Runner.Common
             public static readonly string WorkerCrash = "WORKER_CRASH";
             public static readonly string UnsupportedCommand = "UNSUPPORTED_COMMAND";
             public static readonly string UnsupportedCommandMessage = "The `{0}` command is depreciated and will soon be disabled. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
-            public static readonly string UnsupportedCommandMessageGHES = "The `{0}` command is disabled. Please upgrade to using Environment Files or to opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable on the Runner to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
+            public static readonly string UnsupportedCommandMessageDisabled = "The `{0}` command is disabled. Please upgrade to using Environment Files or to opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable on the Runner to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
         }
 
         public static class RunnerEvent
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -141,7 +141,7 @@ namespace GitHub.Runner.Common
             public static readonly string InternalTelemetryIssueDataKey = "_internal_telemetry";
             public static readonly string WorkerCrash = "WORKER_CRASH";
             public static readonly string UnsupportedCommand = "UNSUPPORTED_COMMAND";
-            public static readonly string UnsupportedCommandMessage = "The `{0}` command is disabled. Please upgrade to using Environment Files to set the env, or to opt into unsecure command execution by setting `ACTIONS_ALLOW_UNSECURE_COMMANDS` to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
+            public static readonly string UnsupportedCommandMessage = "The `{0}` command is depreciated and will soon be disabled. Please upgrade to using Environment Files to set the env. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
             public static readonly string UnsupportedCommandMessageGHES = "The `{0}` command is disabled. Please upgrade to using Environment Files to set the env, or to opt into unsecure command execution by setting `ACTIONS_ALLOW_UNSECURE_COMMANDS` on the Runner to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
         }
 
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -141,8 +141,8 @@ namespace GitHub.Runner.Common
             public static readonly string InternalTelemetryIssueDataKey = "_internal_telemetry";
             public static readonly string WorkerCrash = "WORKER_CRASH";
             public static readonly string UnsupportedCommand = "UNSUPPORTED_COMMAND";
-            public static readonly string UnsupportedCommandMessage = "The `{0}` command is depreciated and will soon be disabled. Please upgrade to using Environment Files to set the env. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
-            public static readonly string UnsupportedCommandMessageGHES = "The `{0}` command is disabled. Please upgrade to using Environment Files to set the env, or to opt into unsecure command execution by setting `ACTIONS_ALLOW_UNSECURE_COMMANDS` on the Runner to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
+            public static readonly string UnsupportedCommandMessage = "The `{0}` command is depreciated and will soon be disabled. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
+            public static readonly string UnsupportedCommandMessageGHES = "The `{0}` command is disabled. Please upgrade to using Environment Files or to opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable on the Runner to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
         }
 
         public static class RunnerEvent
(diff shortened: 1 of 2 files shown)


src/Runner.Common/Constants.cs
```diff
@@ -140,6 +140,9 @@ public static class ReturnCode
 
             public static readonly string InternalTelemetryIssueDataKey = "_internal_telemetry";
             public static readonly string WorkerCrash = "WORKER_CRASH";
+            public static readonly string UnsupportedCommand = "UNSUPPORTED_COMMAND";
+            public static readonly string UnsupportedCommandMessage = "The `{0}` command is depreciated and will soon be disabled. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/";
```
juliobbv, line 144: depreciated -> deprecated

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,41 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
...
+            var serverUrl = context.GetGitHubContext("server_url");
+            var isGHES = false;
+            if(!String.IsNullOrWhiteSpace(serverUrl) && serverUrl != "https://github.com")
+            {
+                isGHES = true;
+            }
+            
+            var allowUnsecureCommands = false;
+            if (!String.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands)) 
+                && Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands).ToUpper() == "TRUE")
```
juliobbv, line 196: The usual pattern we use in the runner is to use the `TryParse()` family of methods to convert from the string to the object type of choice. For example: https://github.com/actions/runner/blob/2709cbc0eae592e6004b016c52c04382f19c7683/src/Runner.Sdk/Util/IOUtil.cs#L398

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,41 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
 
         public void ProcessCommand(IExecutionContext context, string line, ActionCommand command, ContainerInfo container)
         {
+
+            var serverUrl = context.GetGitHubContext("server_url");
+            var isGHES = false;
+            if(!String.IsNullOrWhiteSpace(serverUrl) && serverUrl != "https://github.com")
```
juliobbv, line 189: Nit:

```
            var isGhes = false;

            if (!String.IsNullOrWhiteSpace(serverUrl) && serverUrl != "https://github.com")
```

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,41 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
...
+                isGHES = true;
+            }
+            
+            var allowUnsecureCommands = false;
+            if (!String.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands)) 
+                && Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands).ToUpper() == "TRUE")
+                {
+                    allowUnsecureCommands = true;
+                }
+            // TODO REMOVE DEBUG
```
juliobbv, line 200: Do you need to remove this? :)

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -282,6 +317,40 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
...
```
juliobbv, line 333: Same here.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -282,6 +317,40 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
...
+                // Log Telemetry and let user know they shouldn't do this
+                var issue = new Issue() 
+                { 
+                    Type = IssueType.Warning, 
+                    Message = String.Format(Constants.Runner.UnsupportedCommandMessage, this.Command)
+                };
+                issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.UnsupportedCommand;
+                context.AddIssue(issue);
+            }
+
```
juliobbv, line 353: The same comments from above would also apply here.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,41 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
 
         public void ProcessCommand(IExecutionContext context, string line, ActionCommand command, ContainerInfo container)
         {
+
+            var serverUrl = context.GetGitHubContext("server_url");
+            var isGHES = false;
```
juliobbv, line 188: Now that I think of it, I'd rename this variable to `isNotGithub`, or something like that. Our internal environments are not github.com, but also not GHES, so currently what the current is saying is not what the code is actually doing for those environments.

juliobbv (commented):
Left you some comments. I'll take another look once they're addressed 👍

src/Runner.Worker/ActionCommandManager.cs
thboop: It is removed now! Was still testing GHES!

thboop:
Should have responded to all your feedback Julio!

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,33 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
...
         {
+            var serverUrl = context.GetGitHubContext("server_url");
+            var isNotGithub = false;
+            if(!String.IsNullOrWhiteSpace(serverUrl) && serverUrl != "https://github.com")
+            {
+                isNotGithub = true;
+            }
+            
+            var allowUnsecureCommands = false;
+            bool.TryParse(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands), out allowUnsecureCommands);
```
thboop, line 194: @ericsciple this pulls from the `.env` file in the layout folder of the runner. Users won't be able to set it in their [workflow job env.](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenv)

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -282,6 +309,33 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
...
+
+            // TODO: Eventually remove isNotGithub and apply this to dotcom customers as well
+            if (isNotGithub && !allowUnsecureCommands)
+            {
+                throw new Exception(String.Format(Constants.Runner.UnsupportedCommandMessageDisabled, this.Command));
+            }
+            else if(!allowUnsecureCommands)
```
ericsciple, line 327: nit: space between `if (`

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -282,6 +309,33 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
 
         public void ProcessCommand(IExecutionContext context, string line, ActionCommand command, ContainerInfo container)
         {
+            var serverUrl = context.GetGitHubContext("server_url");
```
ericsciple, line 312:
```csharp
            var isHostedServer = configurationStore.GetSettings().IsHostedServer;
```

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -282,6 +309,33 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
...
+            var isNotGithub = false;
+            if(!String.IsNullOrWhiteSpace(serverUrl) && serverUrl != "https://github.com")
+            {
+                isNotGithub = true;
+            }
+            
+            var allowUnsecureCommands = false;
+            bool.TryParse(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands), out allowUnsecureCommands);
```
ericsciple, line 320: i would suggest checking both the environment variable as well as `env` from the workflow.

Folks who have many definitions may find it easier to set on the machine. Folks who don't have admin access to the machine find it easier to set on the workflow.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -183,6 +183,33 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
...
+
+            // TODO: Eventually remove isNotGithub and apply this to dotcom customers as well
+            if (isNotGithub && !allowUnsecureCommands)
+            {
+                throw new Exception(String.Format(Constants.Runner.UnsupportedCommandMessageDisabled, this.Command));
+            }
+            else if(!allowUnsecureCommands)
```
ericsciple, line 201: nit: `if (`

src/Runner.Worker/ActionCommandManager.cs
ericsciple, line 194: Be careful, i think bool.TryParse can null ref. You might want to coalesce with empty string.

```
bool.TryParse(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowUnsupportedCommands) ?? string.Empty, out allowUnsecureCommands);
```

src/Runner.Worker/ActionCommandManager.cs
ericsciple: ^ better to do it that way for consistency with the other places in the runner

src/Runner.Worker/ActionCommandManager.cs
thboop: It doesn't, I tested manually and from the docs

```
        //     When this method returns, if the conversion succeeded, contains true if value
        //     is equal to System.Boolean.TrueString or false if value is equal to System.Boolean.FalseString.
        //     If the conversion failed, contains false. The conversion fails if value is null
        //     or is not equal to the value of either the System.Boolean.TrueString or System.Boolean.FalseString
        //     field.
```

src/Runner.Worker/ActionCommandManager.cs
ericsciple, line 327: kind of weird that `allow` is false, and then we warn but allow anyway

wondering whether the default value of `allow` should be based on whether is hosted server. And then if the env var is specified, respect it.

ericsciple (approved):

juliobbv (approved):

Issue #743, opened 2020-10-06
Error when auto-upgrading runner service on MacOS - service fails to restart after upgrade


JohnLBergqvist:
**Describe the bug**
When attempting to auto-upgrade the runner service on MacOS, the runner service repeatedly fails to re-launch, and the machine has to be rebooted in order for it to launch succesfully.

**To Reproduce**
1. Install the runner (but don't start it)
2. Install the service (but don't start it) via `./svc.sh install`
2. Ensure the service runs automatically upon boot without needing to log in (as per https://github.com/actions/runner/issues/349): 
`
sudo cp /Users/dbadmin/Library/LaunchAgents/actions.runner.diffblue-cover.ci-mac-04.plist /Library/LaunchDaemons/actions.runner.diffblue-cover.ci-mac-04.plist
`
3. Reboot machine and check (via actions page) that github can see the runner is online
4. Wait for an upgrade - the service will then go offline and not restart unless you reboot the machine manually.

**Expected behavior**

The service should upgrade and restart correctly (which I was able to observe happening fine when upgrading from 2.273.3 to 2.273.4):

```
2020-09-23 14:42:59Z: Listening for Jobs
Runner update in progress, do not shutdown runner.
Downloading 2.273.4 runner
Waiting for current job finish running.
Generate and execute update script.
Runner will exit shortly for update, should back online within 10 seconds.
Runner listener exited with error code 3
Runner listener exit because of updating, re-launch runner in 5 seconds.
Starting Runner listener with startup type: service
Started listener process
√ Connected to GitHub
2020-09-23 14:46:49Z: Listening for Jobs
```

## Runner Version and Platform
Upgrading 2.273.4 to 2.273.5 on Mac OS 10.15.5

## Job Log Output
If applicable, include the relevant part of the job / step log output here.  All sensitive information should already be masked out, but please double-check before pasting here.

## Runner and Worker's Diagnostic Logs

`/Users/dbadmin/Library/Logs/actions.runner.diffblue-cover.ci-mac-04/stderr.og` is empty

Output of `/Users/dbadmin/Library/Logs/actions.runner.diffblue-cover.ci-mac-04/stdout.log`

```
Runner update in progress, do not shutdown runner.
Downloading 2.273.5 runner
Waiting for current job finish running.
Generate and execute update script.
Runner will exit shortly for update, should back online within 10 seconds.
Runner listener exited with error code 3
Runner listener exit because of updating, re-launch runner in 5 seconds.
[...]

PR #746, opened 2020-10-07
Move restart-on-update reponsability to launchd/systemd

kevindumanoir:
After an update, RunnerService.js should be relaunched, as its
interpretor might be deleted from hard drive (leads to security issues
on macOS). To do so, restart can be managed by system service manager.

Resolve issue #743


kevindumanoir:
There might be an issue with these changes. In fact, these changes will not apply on configured systemd/launchd systems. Worse, it could lead to situations where runner never restarts after an upgrade, because current systemd/launchd config never restart process.

To preserve a consistent state after upgrade, this change requires user to reinstall service.

kevindumanoir:
In fact the second commit is enough to solve our issue, since it re-invoke the whole service chain.

src/Misc/layoutbin/update.sh.template
```diff
@@ -125,6 +125,14 @@ date "+[%F %T-%4N] Update succeed" >> "$logfile"
 date "+[%F %T-%4N] Rename $logfile to be $logfile.succeed" >> "$logfile" 2>&1
 mv -fv "$logfile" "$logfile.succeed" >> "$logfile" 2>&1
 
+# reinstall service if needed (update systemd/launchd template)
+if [[ -e "$rootfolder/.service" ]]
+then
+    "$rootfolder/svc.sh" uninstall
```
TingluoHuang, line 131: this will be a problem for Linux runner since service in Linux is installed via [sudo ./svc.sh install](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service#installing-the-service) during runner config time, the update script is running under runner execution context, might not have permission to `sudo`.

kevindumanoir: The other "OS-agnostic" solution I can see here is to remove this part, change service settings to always restart watchdog and make watchdog exits on update. However this service update will still require user intervention on Linux.

We can also run the service reinstall exclusively on Darwin, but it will be, for now, the only UNIX OS to handle service template updates.

vincentbrison:
@TingluoHuang any chance to take a second look? It breaks macOS runners at each auto update 😔. Or is there an option to disable auto-update feature? I am pretty sure @kevindumanoir could help on the implementation, if we have some feedback on his work 🙏. Thanks!

PR #760, opened 2020-10-15
Fixes #759 doesn't change proxy environment variables

### LINKED ISSUE #759: Proxies with port 80 are stripped of their port
**Describe the bug**
Proxies are stripped of their port if it is 80. This causes anything that uses libcurl to fail because it's default proxy port is 1080.

**To Reproduce**
1. Setup self-hosted runner
2. Add https_proxy with port 80 in .env file (ex. `https_proxy=http://proxy:80`)
3. Start runner
4. Configure and run workflow to use something that uses libcurl (ex. checkout plugin)
5. See operation timeout error with proxy port set to 1080
6. You can also print the environment variables and see that port 80 has been stripped from https_proxy

Related Issue: [actions/checkout#216](https://github.com/actions/checkout/issues/216)

**Expected behavior**
Keep port 80 as a part of the proxy environment variable

Version of your runner?
2.273.5
OSX and Linux
## What's not working?
Environment variable is being stripped of proxy port
Expected
```
http_proxy=http://proxy:80
HTTP_PROXY=http://proxy:80
https_proxy=http://proxy:80
HTTPS_PROXY=http://proxy:80
```
Actual
```
http_proxy=http://proxy/
HTTP_PROXY=http://proxy/
https_proxy=http://proxy/
HTTPS_PROXY=http://proxy/
```
## Job Log Output
If applicable, include the relevant part of the job / step log output here.  All sensitive information should already be masked out, but please double-check before pasting here.
```
fatal: unable to access 'https://github.com/actions/runner.git/': Failed to connect to proxy port 1080: Operation timed out
The process '/usr/bin/git' failed with exit code 128
```

## Code causing problem
[RunnerWebProxy.cs:72-130](https://github.com/actions/runner/blob/2709cbc0eae592e6004b016c52c04382f19c7683/src/Runner.Sdk/RunnerWebProxy.cs#L72)

JasonLaquaTR:
Fixes #759 

Sets the environment variables for proxies based on what the user set them as and not based on .net's Uri which strips default ports off

### DIFF
--- a/src/Runner.Sdk/RunnerWebProxy.cs
+++ b/src/Runner.Sdk/RunnerWebProxy.cs
@@ -74,8 +74,8 @@ namespace GitHub.Runner.Sdk
                 _httpProxyAddress = proxyHttpUri.AbsoluteUri;
 
                 // Set both environment variables since there are tools support both casing (curl, wget) and tools support only one casing (docker)
-                Environment.SetEnvironmentVariable("HTTP_PROXY", _httpProxyAddress);
-                Environment.SetEnvironmentVariable("http_proxy", _httpProxyAddress);
+                Environment.SetEnvironmentVariable("HTTP_PROXY", httpProxyAddress);
+                Environment.SetEnvironmentVariable("http_proxy", httpProxyAddress);
 
                 // the proxy url looks like http://[user:pass@]127.0.0.1:8888
                 var userInfo = Uri.UnescapeDataString(proxyHttpUri.UserInfo).Split(':', 2, StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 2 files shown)


src/Runner.Sdk/RunnerWebProxy.cs
```diff
@@ -74,8 +74,8 @@ public RunnerWebProxy()
                 _httpProxyAddress = proxyHttpUri.AbsoluteUri;
```
TingluoHuang, line 74: can you delete `_httpProxyAddress` since it's no longer used?

src/Runner.Sdk/RunnerWebProxy.cs
```diff
@@ -104,8 +104,8 @@ public RunnerWebProxy()
                 _httpsProxyAddress = proxyHttpsUri.AbsoluteUri;
```
TingluoHuang, line 104: can you delete `_httpProxyAddress` since it's no longer used?

TingluoHuang (approved):

TingluoHuang:
@JasonLaquaTR Thanks for filing and fixing this. 🎉 🙇‍♂️

src/Runner.Sdk/RunnerWebProxy.cs
JasonLaquaTR: That still gets used via the `HttpProxyAddress` property in a couple places. Would it make more sense to set `_httpProxyAddress = httpProxyAddress;` instead of setting just the environment variables to it?

src/Runner.Sdk/RunnerWebProxy.cs
JasonLaquaTR: That still gets used via the `HttpsProxyAddress` property in a couple places. Would it make more sense to set `_httpsProxyAddress = httpsProxyAddress;` instead of setting just the environment variables to it?

src/Runner.Sdk/RunnerWebProxy.cs
TingluoHuang: feel like `_httpProxyAddress = proxyHttpUri.OriginalString;` is better, i pushed it to your branch. 😄

JasonLaquaTR: oh yeah that's much better. I must've missed that when I was looking at the Uri class

Issue #351, opened 2020-03-02
Powershell does not return the correct non-zero exit code


johnstevenson:
**Describe the bug**
Powershell does not return the correct exit code if it is non-zero and not 1. 

**To Reproduce**
1. Create an action that outputs an exit code that is not 0 or 1.
```
run: php -r "exit(42);"
shell: pwsh
```
2. See error, as in this example from https://github.com/johnstevenson/runner-actions-test/actions/runs/48177580 which shows the correct behaviour for `cmd` and `bash` and the error from `pwsh`.

![action-error-exit](https://user-images.githubusercontent.com/881777/75681599-d0b5b400-5c8b-11ea-9980-1839776dff9c.png)

**Expected behavior**
For powershell to return the exit code returned by the script.

## Runner Version and Platform
Current runner version: 2.165.2
Tested on ubuntu-latest, macos-latest and windows-latest

See https://github.com/johnstevenson/runner-actions-test/actions/runs/48202446 for tests that use `.bat`, `.sh` and `.ps1` scripts rather than using the php code.

## What's not working?
The code that is appended to the powershell script in `FixUpScriptContents`:

https://github.com/actions/runner/blob/d80ab095a55324be2777de87a3c70791ae917885/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs#L59

This can be fixed by either changing the line to:
```cs
var append = @"if (Test-Path -LiteralPath variable:\LASTEXITCODE) { $Host.SetShouldExit($LASTEXITCODE) }"; 
```
or by executing the script as a file, rather than a command which always returns the correct exit code:

https://github.com/actions/runner/blob/d80ab095a55324be2777de87a3c70791ae917885/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs#L12-L13

```
        ["pwsh"] = "-file \"'{0}'\""
        ["powershell"] = "-file \"'{0}'\"",
```
Incidentally, the documentation at https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell shows the powershell internal commands as `pwsh -command "& '{0}'"` rather than `pwsh -command ". '{0}'"`, but I'm not sure where to report this.

bryanmacfarlane:
@thboop - this is a dupe right?

thboop:
> @thboop - this is a dupe right?

I do not believe so.

cprosser-clear:
Is this being worked on?  This is causing a lot of problems for us.

[...]

PR #773, opened 2020-10-28
Output correct error codes #351

anthonysessa:
Output correct error codes #351


anthonysessa:
@ericsciple @TingluoHuang -- Hey! anyway to get a review sometime soon?  

We are continuing to run into the issue in multiple parts of our workflows.

`yarn, linting, building executables, jest / testing, other testing frameworks like spectron, etc.` 

We have some workarounds, but not for all of the scenarios we need to cover.

Thanks!

PR #779, opened 2020-10-30
Disable set-env and add-pathcommands

thboop:
### Context
A moderate security vulnerability has been identified in the GitHub Actions runner that can allow environment variable and path injection in workflows that log untrusted data to stdout. This can result in environment variables being introduced or modified without the intention of the workflow author. To address this issue we have introduced a new set of files to manage environment and path updates in workflows. This PR disables the old commands.

[Blog Post](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

### Mitigation
If you are using these old commands, the steps that use them in your workflow will fail. You will want to move towards using [Environment Files](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files).

You may also opt into unsecure command execution as well, at a job level or for all jobs on your self hosted runner. We recommend you do not choose to do this, and instead update to the new Environment Files. 

- To opt in at a [job level](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenv), set `ACTIONS_ALLOW_UNSECURE_COMMANDS` to  `true`
```
jobs:
  test:
    runs-on: self-hosted
    env:
      ACTIONS_ALLOW_UNSECURE_COMMANDS: true
```
- To opt in for all jobs run on a self hosted runner, set `ACTIONS_ALLOW_UNSECURE_COMMANDS=true` in the `.env` file found at the root of the runner, much like you would set an [http_proxy](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners#using-a-env-file-to-set-the-proxy-configuration)


src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -339,10 +324,7 @@ public sealed class AddPathCommandExtension : RunnerService, IActionCommandExten
         public Type ExtensionType => typeof(IActionCommandExtension);
 
         public void ProcessCommand(IExecutionContext context, string line, ActionCommand command, ContainerInfo container)
-        {
-            var configurationStore = HostContext.GetService<IConfigurationStore>();
-            var isHostedServer = configurationStore.GetSettings().IsHostedServer;
-
+        {
```
TingluoHuang, line 327: should this method only has one line?
```
throw new Exception(String.Format(Constants.Runner.UnsupportedCommandMessageDisabled, this.Command));
```

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -184,9 +184,6 @@ public sealed class SetEnvCommandExtension : RunnerService, IActionCommandExtens
 
         public void ProcessCommand(IExecutionContext context, string line, ActionCommand command, ContainerInfo container)
         {
```
TingluoHuang, line 186: should this method only has one line?
```
throw new Exception(String.Format(Constants.Runner.UnsupportedCommandMessageDisabled, this.Command));
```

thboop: No because we want users to be able to set `ACTIONS_ALLOW_UNSECURE_COMMANDS` to use the old commands if needed, some actions may not have updated already and giving users a workaround is helpful until they do get updated!

src/Runner.Worker/ActionCommandManager.cs
thboop: See above :)

juliobbv (approved):
Changes LGTM 👍

src/Runner.Worker/ActionCommandManager.cs
TingluoHuang: confirmed with @chrispat 
APPROVED.

TingluoHuang (approved):

PR #784, opened 2020-11-02
raise error for set-env, block set node_options.

TingluoHuang:
- create error instead of warning when the customer uses deprecated commands.
- block set NODE_OPTIONS using `::set-env::` 

image

### DIFF
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -232,7 +232,7 @@ namespace GitHub.Runner.Worker
                     var issue = new Issue()
                     {
                         Type = IssueType.Error,
-                        Message = "Can't update NODE_OPTIONS environment variable using ::set-env:: command."
+                        Message = $"Can't update {blocked} environment variable using ::set-env:: command."
                     };
                     issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = $"{Constants.Runner.UnsupportedCommand}_{envName}";
                     context.AddIssue(issue);
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -186,7 +186,7 @@ namespace GitHub.Runner.Worker
         {
             var configurationStore = HostContext.GetService<IConfigurationStore>();
             var isHostedServer = configurationStore.GetSettings().IsHostedServer;
-            
+
             var allowUnsecureCommands = false;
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -223,6 +223,24 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
...
+
+            foreach (var blocked in _setEnvBlockList)
+            {
+                if (string.Equals(blocked, envName, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Log Telemetry and let user know they shouldn't do this
+                    var issue = new Issue()
+                    {
+                        Type = IssueType.Error,
+                        Message = "Can't update NODE_OPTIONS environment variable using ::set-env:: command."
```
hross, line 235: can we make this message part of a hash or something since the actual list of env's and blocks could be generic? otherwise we should just hard code the variable too

hross (approved):
Approving but I think you should fix the error message based on the block

src/Runner.Worker/ActionCommandManager.cs
TingluoHuang: fixed.

juliobbv (commented):
Changes LGTM.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -223,6 +223,24 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
...
+                    // Log Telemetry and let user know they shouldn't do this
+                    var issue = new Issue()
+                    {
+                        Type = IssueType.Error,
+                        Message = $"Can't update {blocked} environment variable using ::set-env:: command."
+                    };
+                    issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = $"{Constants.Runner.UnsupportedCommand}_{envName}";
+                    context.AddIssue(issue);
+
+                    return;
```
juliobbv, line 240: Minor: this `return;` statement was surprising to me to read at first.

I'm guessing that if we ever add more items to `_setEnvBlockList`, we want the runner to report an error on the first offending env var, instead of all of them, @TingluoHuang?

TingluoHuang: I was just want to make sure we don't run 
```
context.Global.EnvironmentVariables[envName] = command.Data;
context.SetEnvContext(envName, command.Data);
context.Debug($"{envName}='{command.Data}'");
```

PR #785, opened 2020-11-03
Add initial support for Windows arm64

dennisameling:
I wanted to test if it's possible to cross-compile GitHub Actions for Windows arm64, turns out it works! 🚀 

Here's an example release with the arm64 binaries: https://github.com/dennisameling/runner/releases/tag/v2.273.6-win-arm64

**UPDATE December 28 2020**: rebased against the recent .NET 5 upgrade. New release: https://github.com/dennisameling/runner/releases/tag/v2.275.1-win-arm64 - confirmed to work on Surface Pro X including the RunnerService 👍🏼 

I was even able to add a self-hosted runner on my Surface Pro X with the arm64 binaries:

image

Running as a Windows service also works:

image

image

I know I haven't discussed this in a Feature Request here before to ask for the feature, but would appreciate if the team wants to consider this PR. It would bring Windows arm64 to a pre-release state, just like Linux arm64 and arm.

Thanks in advance!

### DIFF
--- a/releaseNote.md
+++ b/releaseNote.md
@@ -29,6 +29,20 @@ Add-Type -AssemblyName System.IO.Compression.FileSystem ;
 [System.IO.Compression.ZipFile]::ExtractToDirectory("$PWD\actions-runner-win-x64-<RUNNER_VERSION>.zip", "$PWD")
 ```
 
+## Windows arm64 (Pre-release)
+We recommend configuring the runner in a root folder of the Windows drive (e.g. "C:\actions-runner"). This will help avoid issues related to service identity folder permissions and long file path restrictions on Windows.
+
+The following snipped needs to be run on `powershell`:
+``` powershell
+# Create a folder under the drive root
+mkdir \actions-runner ; cd \actions-runner
+# Download the latest runner package
+Invoke-WebRequest -Uri https://github.com/actions/runner/releases/download/v<RUNNER_VERSION>/actions-runner-win-arm64-<RUNNER_VERSION>.zip -OutFile actions-runner-win-arm64-<RUNNER_VERSION>.zip
+# Extract the installer
+Add-Type -AssemblyName System.IO.Compression.FileSystem ; 
+[System.IO.Compression.ZipFile]::ExtractToDirectory("$PWD\actions-runner-win-arm64-<RUNNER_VERSION>.zip", "$PWD")
+```
+
 ## OSX
 
 ``` bash
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -134,6 +134,16 @@ if [[ "$PACKAGERUNTIME" == "win-x64" || "$PACKAGERUNTIME" == "win-x86" ]]; then
     fi
 fi
 
+if [[ "$PACKAGERUNTIME" == "win-arm64" ]]; then
+    acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/$PACKAGERUNTIME/node.exe" node12/bin
+    acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/$PACKAGERUNTIME/node.lib" node12/bin
+    acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.exe" node16/bin
+    acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.lib" node16/bin
+    if [[ "$PRECACHE" != "" ]]; then
+        acquireExternalTool "https://github.com/microsoft/vswhere/releases/download/2.6.7/vswhere.exe" vswhere
+    fi
+fi
+
 # Download the external tools only for OSX.
 if [[ "$PACKAGERUNTIME" == "osx-x64" ]]; then
     acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/node-v${NODE12_VERSION}-darwin-x64.tar.gz" node12 fix_nested_dir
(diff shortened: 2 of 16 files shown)


src/Misc/externals.sh
```diff
@@ -3,6 +3,7 @@ PACKAGERUNTIME=$1
 PRECACHE=$2
 
 NODE_URL=https://nodejs.org/dist
+NODE_UNOFFICIAL_URL=https://unofficial-builds.nodejs.org/download/release
```
TingluoHuang, line 6: what is unoffical-builds?  is that trusted?

TingluoHuang:
@chrispat @thejoebourneidentity for this.

src/Misc/externals.sh
dennisameling: The NodeJS can't currently publish official builds for Windows arm64 yet due to missing Hosted ARM64 runners (tracked in https://github.com/nodejs/build/issues/2450#issuecomment-705853342). Because of this, the NodeJS Build team is currently providing unofficial builds at that URL.

Compilation support for ARM64 [is at Tier 2](https://github.com/nodejs/node/blob/master/BUILDING.md#platform-list), meaning that they won't release new versions if NodeJS can't be built on arm64.

releaseNote.md
```diff
@@ -22,6 +22,20 @@ Add-Type -AssemblyName System.IO.Compression.FileSystem ;
 [System.IO.Compression.ZipFile]::ExtractToDirectory("$PWD\actions-runner-win-x64-<RUNNER_VERSION>.zip", "$PWD")
 ```
 
+## Windows arm64 (Pre-release)
```
dscho, line 25: Shouldn't this be upper-case `ARM`, i.e. `ARM64` instead of `arm64`?

Biswa96:
Any progress on this?

jeremyd2019:
This has worked pretty well for me for making a Raspberry Pi Windows ARM64 runner.  Fast it ain't, but it's better than the nothing that's available in hosted runners.  I just had to rebase and rebuild (CI) because a new version was released and auto-update failure (not finding a windows-arm64 version) was apparently fatal.  Upgrading from a 3B+ to a 4B 8GB model with a USB3 flash drive has really helped with speed.

dennisameling:
Yep, had the same issue where the auto-update process got in a loop. Just rebased and rebuilt it as well, and it's once again running really well on my Surface Pro X 🚀

Biswa96:
Why is this not merged yet? What does bother GitHub to add Windows ARM official runners?

jeremyd2019:
Time for another rebase, my runner quit on me again 😖 

One minor issue I see is that the CI job "Package Release" step results in "Packaging actions-runner-win-x64-2.280.2" on arm64, and the zip file inside the artifact zip says x64

dennisameling:
@jeremyd2019 When building on an x64 host the following works for me:

```
./dev layout Release win-arm64
./dev package Release win-arm64
```

It simply creates `actions-runner-win-arm64-2.280.2.zip` with the correct binaries in it. Note that building **on** arm64 might have unintended side-effects since a native arm64 SDK is available only in .NET 5 and higher (the runner runs on .NET Core 3 at the moment of writing).

jeremyd2019:
> @jeremyd2019 When building on an x64 host the following works for me:
> 
> ```
> ./dev layout Release win-arm64
> ./dev package Release win-arm64
> ```

I see. 
https://github.com/actions/runner/blob/04679b56a9d4901f051bd18ab8e532abc56707b2/.github/workflows/build.yml#L64

Needs `${{ matrix.runtime }}` arg like release.yml has

https://github.com/actions/runner/blob/04679b56a9d4901f051bd18ab8e532abc56707b2/.github/workflows/release.yml#L99

ntindle:
Hello,

is this scheduled for review by the actions team?

jeremyd2019:
rebasing onto v2.285.0 finally got a conflict, because they added node16 support.  And it looks like the unofficial build for 12.22.7 for windows-arm64 is not there ☹️

dennisameling:
Have just rebased. The code is now ready for the situation where the NodeJS team will be publishing official ARM64 builds, which hopefully will be happening soon: https://github.com/nodejs/build/issues/2540

jeremyd2019:
As a workaround for the time being, it seems the server is not trying to force my runner to update if it is registered as ephemeral (despite #1396 not being done yet).  And I've read on other issues that you can just lie in your runner's version as something greater than the released version, and that will make the auto-update STFU.  But I'd much rather actually be able to keep my ARM64 runner up-to-date.

src/Misc/externals.sh
bryanmacfarlane: Would the existing node we package work on the windows arm64 through an emulation layer?

dennisameling: @bryanmacfarlane it would indeed, however it would result in a loss of performance as `process.arch` would also return `x64` in that case. We should [actually be very close to official arm64 builds](https://github.com/nodejs/build/issues/2918#issuecomment-1156497518), so may I suggest that we wait 2-3 more weeks? Otherwise I'm happy to update this PR to use `x64` Node until the native version becomes available.

jeremyd2019:
It looks like this has finally been done in https://github.com/actions/runner/releases/tag/v2.297.0 (#2200).  I had been rebasing this against new releases of the runner to keep my runner going, but now I can try out the "official" unofficial build! 🎉

thboop:
Closing this out in favor of the official unofficial builds, but thanks for all teh work you've done here @jeremyd2019 . Let us know if you see any issues in the "official" release

PR #799, opened 2020-11-11
Update SDK to .NET 5

am11:
* Update dotnet-install scripts to match the latest revision.
* TFM from netcoreapp3.1 to net5.0.
* Fix warning about rethrow exception about corrupting stack (`throw ex` -> `throw`).
* `HttpRequestMessage.Properties` is deprecated, use the recommended `HttpRequestMessage.Options`.
* Fix tests where cases are using the same instance of `HttpResponseMessage` which get disposed and then in async context same method is called multiple times by test executor and we get `ObjectDisposedException`.
  * The fix was to transform `ReturnAsync(new HttpResponseMessage(..))` to always return a fresh instance; `ReturnAsync(() => new HttpResponseMessage(..))`.


TingluoHuang (approved):
LGTM, we need to do E2E test before merging.

PR #800, opened 2020-11-11
Only pull if not exist

### LINKED ISSUE #791: Option to not do docker pull?
Thank you 🙇‍♀ for wanting to create a feature in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions or issues with the service, writing workflows or actions, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

**Describe the enhancement**
If I have a local runner and a locally maintained docker image i.e. not stored on a repository, I would like to avoid the "docker pull" step, which then fails.

**Code Snippet**
```
name: Self Hosted Tests

on: [push]

jobs:
  integration_tests:
      runs-on: self-hosted
      container:
          image: python3.6:latest
      steps:
      - name: Checkout
        uses: actions/checkout@v2
      - name: What java version do we have
        run: java -version
```

**Additional information**
The image python3.6:latest is in my self-hosted runner's docker images.  It doesn't exist in DockerHub or any other repository, and I don't want it to, because it should only ever be run on my self-hosted runner.

rowleya:
First check if a docker image exists locally before trying to pull it.  This resolves #791.  Note that most of the change shown below is whitespace (from indent), so adding the "w=1" option to the URL shows the actual change, which is to do a "docker inspect" before the "docker pull".

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -198,41 +198,51 @@ namespace GitHub.Runner.Worker
                 }
             }
 
-            // TODO: Add at a later date. This currently no local package registry to test with
-            // UpdateRegistryAuthForGitHubToken(executionContext, container);
-
-            // Before pulling, generate client authentication if required
-            var configLocation = await ContainerRegistryLogin(executionContext, container);
-
-            // Pull down docker image with retry up to 3 times
-            int retryCount = 0;
-            int pullExitCode = 0;
-            while (retryCount < 3)
+            try
             {
-                pullExitCode = await _dockerManger.DockerPull(executionContext, container.ContainerImage, configLocation);
-                if (pullExitCode == 0)
-                {
-                    break;
-                }
-                else
-                {
-                    retryCount++;
-                    if (retryCount < 3)
-                    {
-                        var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10));
-                        executionContext.Warning($"Docker pull failed with exit code {pullExitCode}, back off {backOff.TotalSeconds} seconds before retry.");
-                        await Task.Delay(backOff);
-                    }
-                }
+
+                // See if the container already exists locally
+                await _dockerManager.DockerInspect(executionContext, container.ContainerImage, "")
             }
-
-            // Remove credentials after pulling
-            ContainerRegistryLogout(configLocation);
-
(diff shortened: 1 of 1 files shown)


dakale:
Could we check the actual output of `docker inspect` instead of catching an exception for control flow? There could be other exceptions that happen when running that command that we would want to fail on or handle differently. I understand its sort of what we are already doing since we arent catching exceptions here anyways, but it might be nicer for long term

rowleya:
Happy to update this if it helps make it get in to the main branch and a future release.  Do you know:
1. What exceptions can come out of this call and what might be done differently on each exception?
1. What outputs can come out of this call and what might be done differently on each output?

If there is nothing different to do at this point, it may be better doing this when there is; I don't think anything here would prevent a future PR with functionality that does different things if needed at a later date.

baparham:
Would it be more versatile to instead add an option to `jobs.[job-id].container` to skip the pull altogether rather than pull if it doesn't exist?

something like 
```
jobs:
  my_job:
    container:
      image: node:14.16
      # could be simply true/false for now, and extended later to include 'never/false', 'if-not-present', or 'always/true'
      pull: false
```

and of course, if the `await _dockerManger.DockerInspect(executionContext, container.ContainerImage, "");` fails, then the job fails, as expected.

I think it's a valid use case to have basically the equivalent of GitLab's `pull-policy: never` for Action container support, and this may be a quicker way than fully implementing all the possible pull policy options that GitLab runners support.

SamsRM (approved):

rowleya:
Thanks @SamsRM for approving this, but note that #1086 does more than this by adding an option rather than making it required... Still happy to use this version if preferred though (anything to make it work)!

rowleya:
Master also now merged in here

SamsRM (approved):

SamsRM (approved):

SamsRM (approved):

SamsRM (approved):

SamsRM (approved):

xihajun:
> Master also now merged in here

has this been merged ready?

rowleya:
Nope; I don't think anyone is looking at these PRs!

PR #806, opened 2020-11-13
Crypto cleanup and enable usage of FIPS compliant crypto when required

dakale:
- Updates RSA api usage to match recommendations in https://docs.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography#rsa, and use APIs that support fips requirements like Pss crossplatform

- Adds a runner setting to be sent by the server when upgrading to fips complaint algorithms is required

- Cleans up some dead crypto code that came in from sdk port to simplify auditing for compliant usage

### DIFF
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -45,9 +45,6 @@ namespace GitHub.Runner.Common
         [DataMember(EmitDefaultValue = false)]
         public string MonitorSocketAddress { get; set; }
 
-        [DataMember(EmitDefaultValue = false)]
-        public bool? _requireFipsCryptography { get; set; }
-
         [IgnoreDataMember]
         public bool IsHostedServer
         {
@@ -60,21 +57,6 @@ namespace GitHub.Runner.Common
             set
             {
                 _isHostedServer = value;
-            }
-        }
-
-        // This should be temporary. Used for compatibility while both are supported
-        [IgnoreDataMember]
-        public bool RequireFipsCryptography
-        {
-            get
-            {
-                return _requireFipsCryptography ?? false;
-            }
-
-            set
-            {
-                _requireFipsCryptography = value;
             }
         }
 
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -45,6 +45,9 @@ namespace GitHub.Runner.Common
         [DataMember(EmitDefaultValue = false)]
         public string MonitorSocketAddress { get; set; }
 
+        [DataMember(EmitDefaultValue = false)]
+        public bool? _requireFipsCryptography { get; set; }
+
         [IgnoreDataMember]
         public bool IsHostedServer
         {
@@ -57,6 +60,21 @@ namespace GitHub.Runner.Common
             set
             {
                 _isHostedServer = value;
+            }
+        }
+
+        // This should be temporary. Used for compatibility while both are supported
+        [IgnoreDataMember]
+        public bool RequireFipsCryptography
+        {
+            get
+            {
+                return _requireFipsCryptography ?? false;
+            }
+
+            set
+            {
+                _requireFipsCryptography = value;
             }
         }
 
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -276,19 +276,6 @@ namespace GitHub.Runner.Listener.Configuration
                 throw new NotSupportedException("Message queue listen OAuth token.");
             }
 
-            _term.WriteSection("Runner settings");
-
-            // We will Combine() what's stored with root.  Defaults to string a relative path
-            runnerSettings.WorkFolder = command.GetWork();
-
-            runnerSettings.MonitorSocketAddress = command.GetMonitorSocketAddress();
-
-            _store.SaveSettings(runnerSettings);
-
-            _term.WriteLine();
-            _term.WriteSuccessMessage("Settings Saved.");
-            _term.WriteLine();
-
             // Testing agent connection, detect any potential connection issue, like local clock skew that cause OAuth token expired.
             var credMgr = HostContext.GetService<ICredentialManager>();
             VssCredentials credential = credMgr.LoadCredentials();
@@ -309,6 +296,19 @@ namespace GitHub.Runner.Listener.Configuration
                 Trace.Error(ex);
                 throw new Exception("The local machine's clock may be out of sync with the server time by more than five minutes. Please sync your clock with your domain or internet time and try again.");
             }
+
+            _term.WriteSection("Runner settings");
+
+            // We will Combine() what's stored with root.  Defaults to string a relative path
+            runnerSettings.WorkFolder = command.GetWork();
+
+            runnerSettings.MonitorSocketAddress = command.GetMonitorSocketAddress();
+
+            _store.SaveSettings(runnerSettings);
+
+            _term.WriteLine();
+            _term.WriteSuccessMessage("Settings Saved.");
+            _term.WriteLine();
 
 #if OS_WINDOWS
             // config windows service
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -275,6 +275,23 @@ namespace GitHub.Runner.Listener.Configuration
                 throw new NotSupportedException("Message queue listen OAuth token.");
             }
 
+            // Configure to use fips-compliant encryption schemes for communication if required by server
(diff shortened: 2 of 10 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -275,6 +275,23 @@ public async Task ConfigureAsync(CommandSettings command)
                 throw new NotSupportedException("Message queue listen OAuth token.");
             }
 
+            // Configure to use fips-compliant encryption schemes for communication if required by server
```
dakale, line 278: Moved up so that the crypto settings are available before doing the connection test

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -275,6 +275,23 @@ public async Task ConfigureAsync(CommandSettings command)
...
+            var requireFipsCryptography = agent.Properties.GetValueOrDefault("RequireFipsCryptography") as bool?;
+            runnerSettings.RequireFipsCryptography = requireFipsCryptography ?? false;
```
TingluoHuang, line 280: should we store these as part of `credentialData` instead of `runnerSettings`?

src/Runner.Listener/MessageListener.cs
```diff
@@ -319,7 +319,8 @@ private TaskAgentMessage DecryptMessage(TaskAgentMessage message)
                 var keyManager = HostContext.GetService<IRSAKeyManager>();
                 using (var rsa = keyManager.GetKey())
                 {
-                    return aes.CreateDecryptor(rsa.Decrypt(_session.EncryptionKey.Value, RSAEncryptionPadding.OaepSHA1), message.IV);
+                    var padding = _settings.RequireFipsCryptography ? RSAEncryptionPadding.OaepSHA256 : RSAEncryptionPadding.OaepSHA1;
```
TingluoHuang, line 322: this is the tricky part that makes you do a runner setting, i think.

src/Runner.Listener/Configuration/ConfigurationManager.cs
TingluoHuang: I assume you don't need most of these change you if store it into `credentialData`?

src/Runner.Listener/MessageListener.cs
dakale: Yea it was find a single best place for this. Putting it in `credentialData` makes sense too, I will try that

src/Runner.Listener/Configuration/ConfigurationManager.cs
dakale: It doesnt seem like there would be much benefit to changing it. We still need to read the setting from the server response (agent properties) since the server will dictate what mode to use

So instead of saving it in the runner settings, it could be saved in the credentialData, but then in the messagelistener to get that data back out, I think we need to get the credentialData out of the settings store anyways

src/Runner.Common/ConfigurationStore.cs
```diff
@@ -45,6 +45,9 @@ public sealed class RunnerSettings
         [DataMember(EmitDefaultValue = false)]
         public string MonitorSocketAddress { get; set; }
 
+        [DataMember(EmitDefaultValue = false)]
```
TingluoHuang, line 48: `DataMember(Name= "RequireFipsCryptography", EmitDefaultValue = false)`

src/Runner.Listener/Configuration/ConfigurationManager.cs
TingluoHuang: I am not sure whether we should change the command line prompt during config, I still prefer to store it into the `credentialData.Data` dictionary in line 265.

you can get the data out in MessageListener by 
```
var configStore = HostContext.GetService<IConfigurationStore>();
var credData = configStore.GetCredentials();
var requireFipsCryptography = StringUtil.ConvertToBoolean(credData.Data.GetValueOrDefault("requireFipsCryptography"), false);                    
```

So we don't need to change the runner setting at all to minimize the overall code changes.

src/Runner.Listener/MessageListener.cs
TingluoHuang, line 322: how does this work with Hosted Runner? i want to make sure we don't kill all hosted run when the runner rolls out.

src/Runner.Listener/Configuration/OAuthCredential.cs
```diff
@@ -36,7 +36,8 @@ public override VssCredentials GetVssCredentials(IHostContext context)
             // We expect the key to be in the machine store at this point. Configuration should have set all of
             // this up correctly so we can use the key to generate access tokens.
             var keyManager = context.GetService<IRSAKeyManager>();
-            var signingCredentials = VssSigningCredentials.Create(() => keyManager.GetKey());
+            var settings = context.GetService<IConfigurationStore>().GetSettings();
```
TingluoHuang, line 39: if we store in the `credentialdata.data`, you can get it by
```
var requireFipsCryptography = StringUtil.ConvertToBoolean(this.CredentialData.Data.GetValueOrDefault("requireFipsCryptography"), false);
```
in line 32

dakale:
Theres actually two distinct settings now:

1. Padding used in signing tokens. This is sent once in the runner properties when its created and is used for all future token signing. Server can support both backwards compatibly by checking (pseudo code)`Verify(token, Pss) || Verify(token, Pkcs1)`. This setting is related to the signing credentials so its saved in the credential data

2. Padding used for decrypting job message encryption key. I added this to the TaskAgentSession because it relates to decrypting messages in a session. It can change session to session depending on server side feature flags, so it is sent down with each session

In both cases, a runner will continue to work even if server side feature flag changes during the lifetime of a single runner. Because server side token signing verification is backwards compatible we can upgrade to signing with Pss, or downgrade back to Pcks1 and they will both pass verification. The type of encryption used for a session key is now stored in the session so it will always know the correct way to decrypt a session key

src/Runner.Listener/Configuration/OAuthCredential.cs
dakale: 👍 This update has been made

src/Runner.Listener/MessageListener.cs
dakale: See my comment at the bottom of this PR. Its changed a bit since you last looked, but the gist is that we can safely upgrade forward here, as long as the server changes are deployed to support both styles with back compat, so I wont merge until Im sure server changes are rolled out everywhere, to prevent bad runner deploy

src/Runner.Common/ConfigurationStore.cs
dakale: Outdated

src/Runner.Listener/MessageListener.cs
dakale: I opted to add an extra property to the session since thats where the session key is, and the way that key is encrypted wont change for the life of a session (and its key)

src/Runner.Listener/Configuration/ConfigurationManager.cs
dakale: 👍  I updated it to store that setting in credentialData.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -275,6 +276,19 @@ public async Task ConfigureAsync(CommandSettings command)
                 throw new NotSupportedException("Message queue listen OAuth token.");
             }
 
+            _term.WriteSection("Runner settings");
```
TingluoHuang, line 279: do you still need to move these codes up? since we no longer store in runner settings. 😄

src/Runner.Listener/Configuration/OAuthCredential.cs
TingluoHuang: you can now remove `var settings = context.GetService<IConfigurationStore>().GetSettings();` i think.

src/Runner.Listener/Configuration/ConfigurationManager.cs
dakale: Nope, good catch. I reverted this change

dakale:
@TingluoHuang Do you have a couple minutes to finish looking over this? It should be safe to merge or even deploy with or without server side changes since it defaults to compatible behavior

thboop (approved):
LGTM

mhart:
This looks cool – is there some documentation around this?

mhart:
@dakale question here – if you're changing the JWT to use PSS if `requireFipsCryptography` is set, won't it be an invalid `RS256` JWT signature?

dakale:
@mhart Thats technically correct. This is just part of the plan to upgrade to use PS256 by default. On the server, we know what method to expect based on feature flags regardless of the JWT algorithm sent in headers. Its a bit messy that way but we really werent checking the header anyways, implementation wise. Since that will be removed anyway and we will default to and assume PS256 (for jwt) and RSA SHA256 padding for the session encryption, I chose to leave it out. Basically, its a remnant of something that wont need to be supported going forward (and clients never actually had an option to begin with)

Issue #570, opened 2020-06-26
Using proxy with self-hosted runner's container jobs


jwarumze:
Hi there,

I am trying to get self-hosted runner container jobs to work inside the corporate network.
According to the documentation I need to do two things:

1. Set the proxy for the runner using .env file

https://help.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners#using-a-env-file-to-set-the-proxy-configuration

And that works, runner picks up the proxy just fine.

2. Set the proxy for Docker containers

https://help.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners#setting-proxy-configuration-for-docker-containers
https://docs.docker.com/network/proxy/

I did follow the documentation - docker proxy has been set for both root and runner_user. If I try to launch the container directly on the machine hosting the runner, everything is right - proxy settings are fetched.
Now when I launch a test action using containers on that machine - proxy settings **are not** fetched.
Same results for setting the proxy as environmental variables.

The only way I can get it working is setting 'env http_proxy' in the Dockerfile, but I'd rather not do that. Any way around that?

klassiker:
As a workaround: you can set environment variables in the workflow file with `jobs.<job-id>.env`

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenv

or for each individual step using `jobs.<job-id>.steps.env`

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsenv

PR #840, opened 2020-12-01
Add proxy support for container actions

klassiker:
Currently the proxy environment variables used by the runner are not passed on to container actions.

Manually starting the container and accessing the network works, but not when it is started by the runner.

This fixes https://github.com/actions/runner/issues/570

Not sure if this will work correctly on Windows.


klassiker:
This should probably be in `Runner.Worker/Handlers/ContainerActionHandler.cs` to reduce possible side-effects and since it is specific to that action, steps without a container work fine.

Let me know what you think.

dakale (approved):
I think this looks fine as is.

PR #847, opened 2020-12-05
Makes the user keychains available to the service

### LINKED ISSUE #350: Switching default keychain
**Describe the bug**

Ultimately, I'm trying to use match/gym to build apps through a self-hosted runner.

I've tried many combinations of the following:
```
bundle exec fastlane run create_keychain
bundle exec fastlane run unlock_keychain
```

However, the result of `security default-keychain` is always `/Library/Keychain/System`.  I can only change the system default with `sudo`

Has anyone been able to work though a similar issue?  I've found a lot of _similar_ or outdated issues, but they all point to things like "use setup_ci" or unlock the keychain, but I think the crux of the problem here is the inability to change the default keychain.

Alternatively, I'm able to see valid certs with 
```
security find-identity -v -p codesigning mykeychain
```
but unable to see them with
```
security find-identity -v -p codesigning
```

Is there a way for me to get `xcodebuild` to use the specific keychain?

**To Reproduce**
See above

**Expected behavior**
```
$ security default-keychain mykeychain
$ security default-keychain
/path/to/mykeychain
```

OSX - tar xzf ./actions-runner-linux-x64-2.165.2.tar.gz

raphaelcruzeiro:
Without creating a session, the service is not able to access the keychains for the user specified under `UserName`. This causes any workflow that deals with code signing to fail as the only keychain loaded with be the system one. This should fix #350

### DIFF
--- a/src/Misc/layoutbin/actions.runner.plist.template
+++ b/src/Misc/layoutbin/actions.runner.plist.template
@@ -25,5 +25,7 @@
     </dict>
     <key>ProcessType</key>
     <string>Interactive</string>
+    <key>SessionCreate</key>
+    <true/>
   </dict>
 </plist>


bmbferreira:
I just had this issue today and this fixed it for me. When will this be merged?

src/Misc/layoutbin/actions.runner.plist.template
```diff
@@ -25,5 +25,7 @@
     </dict>
     <key>ProcessType</key>
     <string>Interactive</string>
+    <key>SessionCreate</key>
```
TingluoHuang, line 28: This should be fine since we are running as `LaunchAgents`

TingluoHuang (approved):

Issue #769, opened 2020-10-27
Bad tokens being passed to runners, resulting in 401 errors


mhart:
**Describe the bug**
We see 401 errors in the setup portions of our actions, especially when trying to download action tarball URLs like `https://api.github.com/repos/actions/checkout/tarball/a81bbbf8298c0fa03ea29cdc473d45769f953675`, and especially in workflows with a number of jobs (~20). It doesn't happen every build, but happens frequently enough to be annoying.

Examples can also be seen around the web of this happening to others:

- https://github.com/tailscale/tailscale/actions/runs/92926574
- https://github.community/t/failure-during-job-set-up/16142
- https://github.com/fastai/fastai2/actions/runs/25226353

## Runner Version and Platform
Version of your runner? 2.273.5

OS of the machine running the runner? Linux

## Job Log Output

```
Current runner version: '2.273.5'
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v2'
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/a81bbbf8298c0fa03ea29cdc473d45769f953675'. Error: Response status code does not indicate success: 401 (Unauthorized).
Warning: Back off 18.746 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/a81bbbf8298c0fa03ea29cdc473d45769f953675'. Error: Response status code does not indicate success: 401 (Unauthorized).
Warning: Back off 18.006 seconds before retry.
Error: Response status code does not indicate success: 401 (Unauthorized).
```

## Suggestions

EDIT: a previous version of this description had presumed it was due to rate-limiting, but subsequent debugging has shown it's actually bad tokens being passed to the runners and the 401s are actually `Bad credentials` errors.

So I'm not sure what to suggest here beyond:

- Don't pass bad/expired tokens to runners (still unclear why this is happening – why aren't the tokens valid?)
- Provide better debugging of errors (include response bodies)
- Provide a way to debug tokens, and why api.github.com would be returning `Bad credentials` for them.

mhart:
I've modified the source code to add some logging for these errors. The cause in most cases seems *not* to be rate limits, but rather a cryptic `Bad credentials` error:

```json
{"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
```

[...]

PR #851, opened 2020-12-07
Count actions resolve failures as infra failures

yacaovsnc:
Closes https://github.com/github/c2c-actions-runtime/issues/986

During job run we may fail to resolve actions download info, and this stack is fully controlled by GitHub actions so it should be counted as infrastructure failure instead of user failure.


mhart:
Hi @yacaovsnc – that repo isn't public so we can't see the details of that issue. Is it related to https://github.com/actions/runner/issues/769 at all?

yacaovsnc:
> Hi @yacaovsnc – that repo isn't public so we can't see the details of that issue. Is it related to #769 at all?

No, this doesn't change functionality in anyway, it only affects how we report the error in our statistics.

thboop (commented):
LGTM lets hold off on merging until server changes roll out

thboop (approved):

PR #864, opened 2020-12-17
Introduce function count() for arrays & strings

sshymko:
Usage example with [dynamic jobs](https://github.com/cschleiden/actions-examples):
```yaml
on:
  workflow_dispatch:
    inputs:
      versions:   
jobs:
  deploy:
    name: Deploy ${{ count(fromJson(github.event.inputs.versions)) }} versions
    runs-on: ubuntu-latest
    strategy:
      matrix:
        version: ${{ fromJson(github.event.inputs.versions) }}
    name: Deploy ${{ matrix.version }} version
    steps:
      - run: echo "hello world"
```


sshymko:
I'd prefer à la JavaScript property syntax `array.length` over the `count()` function, but have no idea how to implement that.

Desirable syntax:
```yaml
${{ fromJson(github.event.inputs.versions).length }}
${{ fromJson(steps.example.outputs.json).items.*.length }}
```

sshymko:
See related #865 introducing `split()` function to parse arrays from strings instead of JSON.

PR #865, opened 2020-12-17
Introduce function split() for strings

sshymko:
Introduce counterpart to [join()](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#join) function for simple arrays where JSON format would be an overkill.

Usage example with [dynamic jobs](https://github.com/cschleiden/actions-examples/blob/main/.github/workflows/dynamic-jobs.yml):
```yaml
name: Dynamic jobs
on:
  workflow_dispatch:
    inputs:
      versions:   
jobs:
  deploy:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        version: ${{ split(github.event.inputs.versions, ',') }}
    name: Deploy version ${{ matrix.version }}
    steps:
      - run: echo "hello world"
```

Sample input:
```json
{
  "versions": "1,2,3"
}
```

### DIFF
--- a/src/Sdk/DTExpressions2/Expressions2/ExpressionConstants.cs
+++ b/src/Sdk/DTExpressions2/Expressions2/ExpressionConstants.cs
@@ -12,6 +12,7 @@ namespace GitHub.DistributedTask.Expressions2
             AddFunction<Contains>("contains", 2, 2);
             AddFunction<EndsWith>("endsWith", 2, 2);
             AddFunction<Format>("format", 1, Byte.MaxValue);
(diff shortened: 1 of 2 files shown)


sshymko:
Would appreciate if someone could help to fix the compilation error:
> The type or namespace name 'FilteredArray' could not be found

Or provide an example of how to return array from expression functions.

76creates:
This is good fn addition, would be nice if we could get this ball rolling ⚡

PR #873, opened 2020-12-19
add warning when running out of disk.

TingluoHuang:
Base on telemetry, lots of workflow run failed as the runner machine run out of disk.

When the machine runs out of disk, the runner may crash due to can't write to the log files, the OS may also freeze up.

The workflow run will look like hanging there do nothing and eventually gets abandoned by the service.

The PR tries to add a warning annotation to the job when there is less than 100MB left on the disk, to provide a hint to customers about why their job failed.

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -1,10 +1,11 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Runtime.Serialization;
+using System.Threading;
 using System.Threading.Tasks;
 using GitHub.DistributedTask.Expressions2;
 using GitHub.DistributedTask.ObjectTemplating.Tokens;
@@ -41,6 +42,8 @@ namespace GitHub.Runner.Worker
         private readonly HashSet<string> _existingProcesses = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         private bool _processCleanup;
         private string _processLookupId = $"github_{Guid.NewGuid()}";
+        private CancellationTokenSource _diskSpaceCheckToken = new CancellationTokenSource();
+        private Task _diskSpaceCheckTask = null;
 
         // Download all required actions.
         // Make sure all condition inputs are valid.
@@ -325,6 +328,12 @@ namespace GitHub.Runner.Worker
                         }
                     }
 
+                    jobContext.Global.EnvironmentVariables.TryGetValue(Constants.Runner.Features.DiskSpaceWarning, out var enableWarning);
+                    if (StringUtil.ConvertToBoolean(enableWarning, defaultValue: true))
+                    {
+                        _diskSpaceCheckTask = CheckDiskSpaceAsync(context, _diskSpaceCheckToken.Token);
+                    }
+
                     return steps;
                 }
                 catch (OperationCanceledException ex) when (jobContext.CancellationToken.IsCancellationRequested)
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -325,6 +328,11 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    if (jobContext.Global.Variables.GetBoolean("__DISK.CHECK") ?? true)
```
hross, line 331: Is this convention? Do we have other variables like this? 

I think we should save this off as a constant somewhere and explain where it comes from in the code.

hross (commented):
How many customers have this issue, do we think? I'm wondering if there are runs that will be impacted and get this notation all the time and it will annoy those customers (maybe not).

Do we already have some telemetry in place that can tell us how frequently this happens so when we release this we can stop it from rolling further if it has unintended noise consequences?

src/Runner.Worker/JobExtension.cs
```diff
@@ -544,6 +557,32 @@ public void FinalizeJob(IExecutionContext jobContext, Pipelines.AgentJobRequestM
...
+        private async Task CheckDiskSpaceAsync(IExecutionContext context, CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                var workDirRoot = Directory.GetDirectoryRoot(HostContext.GetDirectory(WellKnownDirectory.Work));
+                var driveInfo = new DriveInfo(workDirRoot);
+                var freeSpaceInMB = driveInfo.AvailableFreeSpace / 1024 / 1024;
+                if (freeSpaceInMB < 100) // Add warning when disk is lower than 100 MB
+                {
+                    var issue = new Issue() { Type = IssueType.Warning, Message = $"You are running out of disk space, the runner will stop working when machine run out of disk space, free space left: {freeSpaceInMB} MB" };
```
hross, line 569:
```suggestion
                    var issue = new Issue() { Type = IssueType.Warning, Message = $"You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: {freeSpaceInMB} MB" };
```

src/Runner.Worker/JobExtension.cs
```diff
@@ -544,6 +557,32 @@ public void FinalizeJob(IExecutionContext jobContext, Pipelines.AgentJobRequestM
...
         }
 
+        private async Task CheckDiskSpaceAsync(IExecutionContext context, CancellationToken token)
+        {
+            while (!token.IsCancellationRequested)
+            {
+                var workDirRoot = Directory.GetDirectoryRoot(HostContext.GetDirectory(WellKnownDirectory.Work));
+                var driveInfo = new DriveInfo(workDirRoot);
+                var freeSpaceInMB = driveInfo.AvailableFreeSpace / 1024 / 1024;
+                if (freeSpaceInMB < 100) // Add warning when disk is lower than 100 MB
```
hross, line 567: should we make this configurable? Maybe instead of a boolean free disk check we have it be an int and then if the int exists and is non-zero we check for that number of MB/KB?

hross (approved):
I approved this but probably should merge/update some based on my suggestions.

src/Runner.Worker/JobExtension.cs
bryanmacfarlane: Is there any way we can add telemetry so we can discern what portion of abandons are due to this?

TingluoHuang: `issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.LowDiskSpace;` will make this issue shows in Kusto

PR #889, opened 2021-01-04
Ignore certain scenarios so they are not counted as infra failures

yacaovsnc:
Closes https://github.com/github/c2c-actions-runtime/issues/1050

There are some cases that should not be counted as infrastructure failures:
1. Repo is rate-limited
2. Action lives in a non-public repo
3. Job is canceled due to fail fast strategy

This PR will ignore all `422` response code (Server maps `UnresolvableActionDownloadInfoException` to status code `422`), those scenarios should be considered user error.

We should also stop retry if the job is canceled.

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -594,14 +594,8 @@ namespace GitHub.Runner.Worker
                     actionDownloadInfos = await jobServer.ResolveActionDownloadInfoAsync(executionContext.Global.Plan.ScopeIdentifier, executionContext.Global.Plan.PlanType, executionContext.Global.Plan.PlanId, new WebApi.ActionReferenceList { Actions = actionReferences }, executionContext.CancellationToken);
                     break;
                 }
-                catch (Exception ex)
+                catch (Exception ex) when (!executionContext.CancellationToken.IsCancellationRequested) // Do not retry if the run is canceled.
                 {
-                    // Action download was canceled, due to fail fast strategy. There is no point in retrying anymore as the run is canceled.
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -596,6 +596,12 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
                 }
                 catch (Exception ex)
                 {
+                    // Action download was canceled, due to fail fast strategy. There is no point in retrying anymore as the run is canceled.
+                    if (ex.Message != null && string.Equals(ex.Message, "A task was canceled.", StringComparison.OrdinalIgnoreCase))
```
TingluoHuang, line 600: you can do something like 
```
catch (Exception ex) when (!executionContext.CancellationToken.IsCancelled)
```

So you don't have to check the exception message.

src/Sdk/DTWebApi/WebApi/Exceptions.cs
```diff
@@ -2459,6 +2459,25 @@ private InvalidContinuationTokenException(SerializationInfo info, StreamingConte
         }
     }
 
+    [Serializable]
+    public class UnresolvableActionDownloadInfoException : DistributedTaskException
```
hashtagchris, line 2463: I don't see any `new UnresolvableActionDownloadInfoException` calls being added in this PR. Does this exception type get used automatically somehow?

hross: Is this auto-resolved in the task client library and this is just a missed port?

yacaovsnc: Yes, from my local testing this is auto resolved once I ported the exception type.

TingluoHuang (approved):

PR #892, opened 2021-01-05
Add new ANDROID_SDK_ROOT environment variable

sapuglha:
`ANDROID_SDK_ROOT` is the current way of specifying the location of the SDK.
`ANDROID_HOME` is deprecated, but not everyone updated to the new format.

More info about the environment variables: https://developer.android.com/studio/command-line/variables

This change will publish the new and old environment variables. The reason is because some actions are [starting to use the new environment variable and getting broken because of that](https://github.com/ReactiveCircus/android-emulator-runner/issues/116).

### DIFF
--- a/src/Misc/layoutroot/env.sh
+++ b/src/Misc/layoutroot/env.sh
@@ -6,6 +6,7 @@ varCheckList=(
     'ANT_HOME' 
     'M2_HOME' 
     'ANDROID_HOME' 
+    'ANDROID_SDK_ROOT'
     'GRADLE_HOME' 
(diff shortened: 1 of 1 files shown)


TingluoHuang (approved):

TingluoHuang:
@sapuglha just want to make sure, you still need to have `ANDROID_SDK_ROOT` set before running the `env.sh` to get the env to populate into the `.env` file

sapuglha:
Correct, the shell environment must have that variable configured, just like it used to be with ANDROID_HOME.

Issue #808, opened 2020-11-17
[docker] the input device is not a TTY


liukangcc:
**Expected behavior**
```shell
i want to use pytest-sugar plugin, but it doesn't work in github actions.
```

## Runner Version and Platform
```shell
Current runner version: '2.274.1'
```

OS of the machine running the runner? OSX/Windows/Linux/...
```
Linux
```
## What's not working?
```shell
the input device is not a TTY
```
## Job Log Output
```shell
Run cd /home/runner/work/sdk-index/sdk-index   
  cd /home/runner/work/sdk-index/sdk-index   
  docker pull realthread/sdk_index:latest   
  docker run -v "`pwd`:/rt-thread/sdk-index" -it realthread/sdk_index:latest bash -c ". /etc/profile && python3 /rt-thread/sdk- 
  index/tools/sdk_check/sdk_check.py"   
  cd /home/runner/work/sdk-index/sdk-index/tools/sdk_check && python3 check_report.py   
  shell: /bin/bash --noprofile --norc -e -o pipefail {0}  
  env:   
    pythonLocation: /opt/hostedtoolcache/Python/3.8.6/x64   
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.6/x64/lib   
latest: Pulling from realthread/sdk_index   
bb79b6b2107f: Pulling fs layer   
35e30c3f3e2b: Pulling fs layer   
eb0efc438b37: Pulling fs layer   
6398f6ad5ea6: Pulling fs layer   
2dc106129d9e: Pulling fs layer   
89b93dc5af9e: Pulling fs layer   
c11bf1a86a45: Pulling fs layer   
22f73f17c8fc: Pulling fs layer   
6398f6ad5ea6: Waiting   
2dc106129d9e: Waiting   
89b93dc5af9e: Waiting   
c11bf1a86a45: Waiting   
22f73f17c8fc: Waiting   
35e30c3f3e2b: Verifying Checksum   
35e30c3f3e2b: Download complete   
6398f6ad5ea6: Verifying Checksum  
6398f6ad5ea6: Download complete   
eb0efc438b37: Verifying Checksum   
eb0efc438b37: Download complete   
bb79b6b2107f: Download complete   
89b93dc5af9e: Verifying Checksum
89b93dc5af9e: Download complete
2dc106129d9e: Verifying Checksum
2dc106129d9e: Download complete
22f73f17c8fc: Verifying Checksum
22f73f17c8fc: Download complete
[...]

PR #916, opened 2021-01-19
Enable tty output from Docker Actions

dicko2:
Thoughts?

Discussion in #808

### DIFF
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -197,7 +197,8 @@ namespace GitHub.Runner.Worker.Container
 
             dockerOptions.Add($"--workdir {container.ContainerWorkDirectory}");
             dockerOptions.Add($"--rm");
-
+            dockerOptions.Add($"-t");
+            
             foreach (var env in container.ContainerEnvironmentVariables)
             {
                 // e.g. -e MY_SECRET maps the value into the exec'ed process without exposing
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/Container/DockerCommandManager.cs
```diff
@@ -197,7 +197,8 @@ public async Task<int> DockerRun(IExecutionContext context, ContainerInfo contai
 
             dockerOptions.Add($"--workdir {container.ContainerWorkDirectory}");
             dockerOptions.Add($"--rm");
-
+            dockerOptions.Add($"-t");
```
TingluoHuang, line 200: you might want to add the `-t` to where the runner run `docker exec` as well.

dicko2: in the public method here?

https://github.com/actions/runner/blob/6e2fa11b267e79fb553b1b759fbe041eb4db945f/src/Runner.Worker/Container/DockerCommandManager.cs#L301

TingluoHuang: I added it for you. 😄

dicko2: Should we use "-it" in both places?

dicko2:
Good to merge @TingluoHuang ?

TingluoHuang (approved):

everops-alex:
This was reverted in #977 (`v2.277.1`) -- the issue is, `teletype` was reverted, as well as the `-t` docker `tags` command.

Please add it back! 🙏 see details in #977 thread.

cc @chrispat

src/Runner.Worker/Container/DockerCommandManager.cs
mary-doc: Thanks all

PR #920, opened 2021-01-19
 support authenticated package download

ericsciple:
Enables on-prem air-gap scenarios where runner package is downloaded from the appliance


ericsciple:
reviewed offline w/ @TingluoHuang

PR #939, opened 2021-01-25
Update AgentPlatform for job timeline record

lokesh755:
Updates AgentPlatform for job timeline record.

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -5,7 +5,6 @@ using System.Collections.Specialized;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


hross (approved):

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -5,6 +5,7 @@
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
```
TingluoHuang, line 8: do we need this?

lokesh755: Removed it.

TingluoHuang (approved):

PR #948, opened 2021-01-27
Generate SHA's for released packages and include them in package notes

thboop:
This PR updates the release process to include sha256 hashes of the runner downloads. Users can use these to verify that they have downloaded the correct runner and it has not been tampered with.

Example (fake) runner release is found [here](https://github.com/thboop/runner/releases/tag/v2.276.2)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,7 @@ jobs:
               repo: '${{ github.event.repository.name }}',
               tag: 'v' + runnerVersion
             })
+            core.setFailed('Release with same tag already created: ' + release.data.html_url)
           } catch (e) {
             // We are good to create the release if release with same tag doesn't exists
             if (e.status != 404) {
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -101,7 +107,19 @@ jobs:
...
         path: _package
-
+    # compute shas and set as job outputs to use in release notes
+    - run: brew install coreutils #needed for shasum util
+      if: ${{ matrix.os == 'macOS-latest' }}
+      name: Install Dependencies for SHA Calculation (osx)
+    - run: |
+        file=$(ls)
+        sha=$(sha256sum $file | awk '{ print $1 }')
+        echo "Computed sha256: $sha for $file"
```
thboop, line 117: helpful for debugging and lets us ensure the sha gets into the build log.

TingluoHuang (approved):

.github/workflows/release.yml
```diff
@@ -45,6 +45,12 @@ jobs:
   
   build:
     needs: check
+    outputs:
+      linux-x64-sha: ${{ steps.sha.outputs.linux-x64-sha256 }}
```
ericsciple, line 49: curious whether `${{ matrix.runtime }}-sha: ${{ steps.compute-sha.outputs.sha }}` would work?

ericsciple: ok to be explicit too... looks like there is another place below where we have to repeat the runtimes, so whatevs :)

ericsciple: other way around repeating platforms in multiple places might be to upload as part of artifact

PR #949, opened 2021-01-27
Add http POST to --check.

TingluoHuang:
As part of investigation for https://github.com/github/c2c-actions/issues/2231
We learned that Nginx may change the HTTP request header and cause an issue with a runner HTTP request to the Actions server written in ASP.netcore.

Add an HTTP POST test to `--check` and update the documentation.

### DIFF
--- a/src/Runner.Listener/Checks/ActionsCheck.cs
+++ b/src/Runner.Listener/Checks/ActionsCheck.cs
@@ -39,6 +39,7 @@ namespace GitHub.Runner.Listener.Check
             string githubApiUrl = null;
             string actionsTokenServiceUrl = null;
             string actionsPipelinesServiceUrl = null;
+            string actionsPipelinesServicePostUrl = null;
             var urlBuilder = new UriBuilder(url);
             if (UrlUtil.IsHostedServer(urlBuilder))
             {
@@ -47,6 +48,7 @@ namespace GitHub.Runner.Listener.Check
                 githubApiUrl = urlBuilder.Uri.AbsoluteUri;
                 actionsTokenServiceUrl = "https://vstoken.actions.githubusercontent.com/_apis/health";
                 actionsPipelinesServiceUrl = "https://pipelines.actions.githubusercontent.com/_apis/health";
+                actionsPipelinesServicePostUrl = "https://pipelines.actions.githubusercontent.com/_apis/distributedtask/pools/1/agents";
             }
             else
             {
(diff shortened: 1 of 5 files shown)


src/Runner.Listener/Checks/ActionsCheck.cs
```diff
@@ -47,6 +48,7 @@ public async Task<bool> RunCheck(string url, string pat)
                 githubApiUrl = urlBuilder.Uri.AbsoluteUri;
                 actionsTokenServiceUrl = "https://vstoken.actions.githubusercontent.com/_apis/health";
                 actionsPipelinesServiceUrl = "https://pipelines.actions.githubusercontent.com/_apis/health";
+                actionsPipelinesServicePostUrl = "https://pipelines.actions.githubusercontent.com/_apis/distributedtask/pools/1/agents";
```
ericsciple, line 51: is this an unauthenticated response? and we're just checking response headers?

src/Runner.Listener/Checks/CheckUtil.cs
```diff
@@ -189,6 +189,67 @@ public static async Task<CheckResult> CheckHttpsRequests(this IHostContext hostC
...
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} Http request 'POST' to {url} has expected HTTP response header");
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} ***************************************************************************************************************");
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} ");
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} ");
+                    }
+                    else
+                    {
+                        result.Pass = false;
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} ***************************************************************************************************************");
+                        result.Logs.Add($"{DateTime.UtcNow.ToString("O")} Http request 'POST' to {url} doesn't have expected HTTP Header.");
```
ericsciple, line 233: should we clarify that it doesn't have the expected HTTP _response_ header

ericsciple (approved):

src/Runner.Listener/Checks/ActionsCheck.cs
TingluoHuang: we just check the response header to make sure it was from the Actions service, the request will 401 or 404.

PR #963, opened 2021-02-03
Update run.sh

dennybaa:
Fix runner update script, in fact the behavior is:

```
Runner update in progress, do not shutdown runner.
Downloading 2.276.1 runner
Waiting for current job finish running.
Generate and execute update script.
Runner will exit shortly for update, should back online within 10 seconds.
+ returnCode=3
+ [[ 3 == 4 ]]
+ exit 3
```

### DIFF
--- a/src/Misc/layoutroot/run.sh
+++ b/src/Misc/layoutroot/run.sh
@@ -26,10 +26,10 @@ if [[ "$1" == "localRun" ]]; then
 else
     "$DIR"/bin/Runner.Listener run $*
 
-# Return code 4 means the run once runner received an update message.
-# Sleep 5 seconds to wait for the update process finish and run the runner again.
+# Return code 3 means the run once runner received an update message.
+# Sleep 5 seconds to wait for the update process finish
     returnCode=$?
-    if [[ $returnCode == 4 ]]; then
+    if [[ $returnCode == 3 ]]; then
         if [ ! -x "$(command -v sleep)" ]; then
             if [ ! -x "$(command -v ping)" ]; then
                 COUNT="0"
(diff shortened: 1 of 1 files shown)


src/Misc/layoutroot/run.sh
```diff
@@ -27,9 +27,9 @@ else
     "$DIR"/bin/Runner.Listener run $*
 
 # Return code 4 means the run once runner received an update message.
```
TingluoHuang, line 29: update the comment as well?
` Return code 3 means the run once runner received an update message.`

TingluoHuang (approved):

src/Misc/layoutroot/run.sh
```diff
@@ -43,8 +43,6 @@ else
         else
             sleep 5 >nul
         fi
-        
-        "$DIR"/bin/Runner.Listener run $*
```
TingluoHuang, line 47: why we don't start the runner again after update?

dennybaa: Well, what's interesting the runner is restarted forked without requiring this, moreover the script exits!
As stated `Runner will exit shortly for update, should back online within 10 seconds.`

Pay attention to the last line where I'm already back in terminal the parent has exited as well as the run.sh
```
2021-02-04 13:07:56Z: Listening for Jobs
Runner update in progress, do not shutdown runner.
Downloading 2.276.1 runner
Waiting for current job finish running.
Generate and execute update script.
Runner will exit shortly for update, should back online within 10 seconds.
+ returnCode=3
+ [[ 3 == 4 ]]
+ exit 3
root@github-actions-deployer-pool-pod-44wd6:/home/runner# 
√ Connected to GitHub
```

dennybaa: This is hot automatic update performed by the runner itself, we need only to wait for the update process finish, so that we don't break unfinished upgrade if the container which spawned run.sh is terminated immediately.

TingluoHuang (approved):

PR #966, opened 2021-02-04
Display GITHUB_TOKEN permissions

yacaovsnc:
Right now it's not clear what permission the GITHUB_TOKEN has.  With this change we will display the permissions in one single group in the job log.

Example run:
image

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -123,16 +123,24 @@ namespace GitHub.Runner.Worker
                         }
                     }
 
-                    var tokenPermissions = jobContext.Global.Variables.Get("system.github.token.permissions") ?? "";
-                    if (!string.IsNullOrEmpty(tokenPermissions))
+                    try 
                     {
-                        context.Output($"##[group]GITHUB_TOKEN Permissions");
-                        var permissions = JsonConvert.DeserializeObject<Dictionary<string, string>>(tokenPermissions);
-                        foreach(KeyValuePair<string, string> entry in permissions)
+                        var tokenPermissions = jobContext.Global.Variables.Get("system.github.token.permissions") ?? "";
+                        if (!string.IsNullOrEmpty(tokenPermissions))
                         {
-                            context.Output($"{entry.Key}: {entry.Value}");
+                            context.Output($"##[group]GITHUB_TOKEN Permissions");
+                            var permissions = JsonConvert.DeserializeObject<Dictionary<string, string>>(tokenPermissions);
+                            foreach(KeyValuePair<string, string> entry in permissions)
+                            {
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -122,6 +123,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    var tokenPermissions = jobContext.Global.Variables.Get("system.github.token.permissions") ?? "";
+                    if (!string.IsNullOrEmpty(tokenPermissions))
+                    {
+                        context.Output($"##[group]GITHUB_TOKEN Permissions");
+                        var permissions = JsonConvert.DeserializeObject<Dictionary<string, string>>(tokenPermissions);
```
thboop, line 130: consider catching exceptions here `JsonException` and logging to runner logs (Trace), may be good to not fail the job if we can't output the Token permissions. It likely won't happen, as we control the api side of this as well.

src/Runner.Worker/JobExtension.cs
```diff
@@ -122,6 +123,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    var tokenPermissions = jobContext.Global.Variables.Get("system.github.token.permissions") ?? "";
+                    if (!string.IsNullOrEmpty(tokenPermissions))
+                    {
+                        context.Output($"##[group]GITHUB_TOKEN Permissions");
```
thboop, line 129: Love the usage of group

thboop (dismissed):

src/Runner.Worker/JobExtension.cs
yacaovsnc: Updated to catch all exceptions, display permissions is best effort.

src/Runner.Worker/JobExtension.cs
```diff
@@ -122,6 +123,26 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    try 
+                    {
+                        var tokenPermissions = jobContext.Global.Variables.Get("system.github.token.permissions") ?? "";
+                        if (!string.IsNullOrEmpty(tokenPermissions))
+                        {
+                            context.Output($"##[group]GITHUB_TOKEN Permissions");
+                            var permissions = JsonConvert.DeserializeObject<Dictionary<string, string>>(tokenPermissions);
```
TingluoHuang, line 132: Use Stringutil instead of jsonconvert

yacaovsnc: Cool, removed `newtonsoft` dependency.

TingluoHuang (approved):

PR #967, opened 2021-02-04
Check Runner Zip Hash on Upgrade

thboop:
This PR updates the Self-Updater to validate the downloaded Runner against the sha256 of the file if the sha256 is provided by the backend service.


src/Runner.Listener/SelfUpdater.cs
```diff
@@ -220,12 +222,25 @@ private async Task DownloadLatestRunner(CancellationToken token)
...
                                 using (Stream result = await httpClient.GetStreamAsync(_targetPackage.DownloadUrl))
                                 {
                                     //81920 is the default used by System.IO.Stream.CopyTo and is under the large object heap threshold (85k). 
                                     await result.CopyToAsync(fs, 81920, downloadCts.Token);
                                     await fs.FlushAsync(downloadCts.Token);
+
+                                    // Validate Hash Matches if it is provided
+                                    if (!String.IsNullOrEmpty(_targetPackage.HashValue))
+                                    {
+                                        fs.Seek(0, SeekOrigin.Begin);
```
TingluoHuang, line 235: I would suggest don't seek the stream back to 0, just read the after the download finish to make the code less indented with `using...using..using..if...if...`

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -220,12 +222,25 @@ private async Task DownloadLatestRunner(CancellationToken token)
...
+                                        var hash = GetSHA256(fs);
+                                        if (hash != _targetPackage.HashValue)
+                                        {
+                                            // Hash did not match, we can't recover from this, just throw
+                                            throw new Exception($"Computed runner hash {hash} did not match expected Runner Hash {_targetPackage.HashValue} for {_targetPackage.Filename}");
```
TingluoHuang, line 240: What will happen after this throw? I would bet the runner will be stuck in an upgrade loop...

thboop: It tries to download three times then aborts, I can try to short circuit after one if thats preferable.

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -256,6 +258,24 @@ private async Task DownloadLatestRunner(CancellationToken token)
...
+                    if (!String.IsNullOrEmpty(_targetPackage.HashValue))
+                    {
+                        using (SHA256 sha256 = SHA256.Create())
+                        {
+                            byte[] srcHashBytes = await sha256.ComputeHashAsync(stream);
+                            var hash = PrimitiveExtensions.ConvertToHexString(srcHashBytes);
+                            if (hash != _targetPackage.HashValue)
+                            {
+                                // Hash did not match, we can't recover from this, just throw
+                                throw new Exception($"Computed runner hash {hash} did not match expected Runner Hash {_targetPackage.HashValue} for {_targetPackage.Filename}");
```
TingluoHuang, line 273: Will we start doing this on github.com ?
What will the runner do after this hash match failed? stuck in an upgrade loop?
We better setup canary runs in Ring 0 to verify every runner release has the right checksum.

TingluoHuang (approved):

src/Runner.Listener/SelfUpdater.cs
thboop: We spoke offline!

PR #977, opened 2021-02-09
Release 2.277.1 runner

thboop:
[Enabling TTY](https://github.com/actions/runner/pull/916) appeared to break container scenarios, lets revert this PR for now to not block this release.


mhart:
Sure would be great to have some more comprehensive integration tests – been a fair bit of churn over the last few releases

lokesh755 (approved):

everops-alex:
This breaks any self-hosted runner that uses an older `cloud-build-local` version.

everops-alex:
This breaks any self-hosted runner that uses an older `cloud-build-local` version, because older cloud-build-local doesn't yet support --all-tags.

The problem here is, neither does GCB. So if we build with our runners for some projects, and build in GCB for other projects, we're forced to have 2 separate `cloudbuild.yaml` templates.

everops-alex:
This breaks any self-hosted runner that uses an older `cloud-build-local` version, because older cloud-build-local doesn't yet support `--all-tags`.

The problem here is, neither does GCB. So if we build with our runners for some projects, and build in GCB for other projects, we're forced to have 2 separate `cloudbuild.yaml` templates.

I've tested and confirmed by downgrading all gcloud components to several old versions, along with docker. Downgrading docker works, however `cloud-build-local` fails (even on the old versions) when it tries to push only `latest` tags.

Our cloudbuild.yaml's in my tests added `commit-sha` and `git-branch` tags to my image, but `cloud-build-local` pushes only `latest`.

This is by design in newer docker versions, but we've explicitly pinned back our `cloud-build-local` and `docker` versisons due to the GCB constraints I mentioned above. When I test the same setup on an ubuntu machine that isn't a github-actions runner, the old `cloud-build-local` works as desired, it pushes my tags and doesn't try to push the image with tags that I didn't create.

Not every docker image push is a `latest`.

cc @chrispat

everops-alex:
I also reviewed the CI and there doesn't seem to be any testing surrounding docker. 

There's 2 things going on here -- `-it` is teletype and that is fine to revert -- but `-t` is `tags` for docker, in `src/Runner.Worker/Container/DockerCommandManager.cs`.

PR #985, opened 2021-02-12
Add support for windows x86

### LINKED ISSUE #423: windows x86 release
please add an x86 windows release

BickelLukas:
This adds support for Windows x86. 

I confirmed it running on our own build server by building the package myself but an official package would be great. 

closes #423 

(I wasn't completely sure about the E2E tests so I added them in a separate commit. If they are not needed the commit can be dropped)


TingluoHuang:
this is not completed.
- need to check with the product manager to decide whether we want to support x85 windows. @thejoebourneidentity 
- package node.js within the runner.
- update docs

BickelLukas:
@TingluoHuang Let me know if there is anything more that I can provide you with. I would love to see x86 support because otherwise we won't be able to use actions to build our legacy application.

rnayabed:
any updates?

PR #990, opened 2021-02-17
Use GITHUB_TOKEN for ghcr.io containers if credentials are not provided

gimenete:
This PR allows us to use the GITHUB_TOKEN for job containers hosted in ghcr.io without providing credentials. Example:

```yaml
name: ghcr.io test
on:
  workflow_dispatch:
jobs:
  push:
    runs-on:
      - self-hosted
      - Linux
      - X64
    
    # Use private GHCR image as build container
    container:
      image: ghcr.io/owner/repo

    steps:
      - uses: actions/checkout@v2

      - name: Run Action
        run: |
          echo 'Hello, World!'
```

I just tested it here https://github.com/sweethaven-village/ghtoken_test/runs/1920493095?check_suite_focus=true Let me know if you want access.

I used some existing code, but removed the repo check because we want to use containers in any repo in the same org and eventually in other orgs too, as long as the GITHUB_TOKEN has access to them of course.

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -493,7 +493,7 @@ namespace GitHub.Runner.Worker
 
         private void UpdateRegistryAuthForGitHubToken(IExecutionContext executionContext, ContainerInfo container)
         {
-            var registryIsTokenCompatible = container.RegistryServer.Equals("ghcr.io", StringComparison.OrdinalIgnoreCase);
+            var registryIsTokenCompatible = container.RegistryServer.Equals("ghcr.io", StringComparison.OrdinalIgnoreCase) || container.RegistryServer.Equals("containers.pkg.github.com", StringComparison.OrdinalIgnoreCase);
             if (!registryIsTokenCompatible)
             {
                 return;
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -198,8 +198,7 @@ namespace GitHub.Runner.Worker
                 }
             }
 
-            // TODO: Add at a later date. This currently no local package registry to test with
-            // UpdateRegistryAuthForGitHubToken(executionContext, container);
+            UpdateRegistryAuthForGitHubToken(executionContext, container);
 
             // Before pulling, generate client authentication if required
             var configLocation = await ContainerRegistryLogin(executionContext, container);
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -494,31 +493,14 @@ private void ContainerRegistryLogout(string configLocation)
...
-            {
-                executionContext.Warning($"Could not identify owner and repo for container image {container.ContainerImage}. Skipping automatic token auth");
-                return;
-            }
-            var owner = imageParts[1];
-            var repo = imageParts[2];
-            var nwo = $"{owner}/{repo}";
-            if (nwo.Equals(executionContext.GetGitHubContext("repository"), StringComparison.OrdinalIgnoreCase))
-            {
-                registryMatchesWorkflow = true;
```
TingluoHuang, line 517: so the `GITHUB_TOKEN` can auth with any image from any repo?

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -494,31 +493,14 @@ private void ContainerRegistryLogout(string configLocation)
 
         private void UpdateRegistryAuthForGitHubToken(IExecutionContext executionContext, ContainerInfo container)
         {
-            var registryIsTokenCompatible = container.RegistryServer.Equals("docker.pkg.github.com", StringComparison.OrdinalIgnoreCase);
+            var registryIsTokenCompatible = container.RegistryServer.Equals("ghcr.io", StringComparison.OrdinalIgnoreCase);
```
TingluoHuang, line 496: any name other than `ghcr.io` we should also check?

src/Runner.Worker/ContainerOperationProvider.cs
gimenete: We want to use containers in any repo in the same org and eventually in other orgs too. So I guess we cannot do any validation here and users will just get an error if the GITHUB_TOKEN doesn't have permissions.

src/Runner.Worker/ContainerOperationProvider.cs
gimenete: I don't think so. I'm asking the packages team to take a look first to this PR just in case.

taywrobel: If we could check against `containers.pkg.github.com` as well that'd be good. It was the development hostname used before we had `ghcr.io` and shouldn't be used anywhere but internally, but they are aliased to one another so would be good to have that corner case covered.

src/Runner.Worker/ContainerOperationProvider.cs
taywrobel: @TingluoHuang We have fine grained permissions allowing/disallowing access to containers scoped to individual repositories, so while this looks more open than previously, we have policies in authzd ensuring that access is restricted as needed when pulling actual images.

src/Runner.Worker/ContainerOperationProvider.cs
taywrobel: We also should leave `docker.pkg.github.com` as an option, since it hasn't been sunset yet and I believe this will break some current workflows that use GITHUB_TOKEN with the legacy registry

taywrobel (changes requested):
Changes overall LGTM, but we should also not break the current behavior for `docker.pkg.github.com`, and should add `containers.pkg.github.com` to the list of token-compatible registry hostnames

src/Runner.Worker/ContainerOperationProvider.cs
TingluoHuang: 👍

src/Runner.Worker/ContainerOperationProvider.cs
TingluoHuang: @gimenete do you mind add those 2 strings to the checklist? i think we are good to merge after this.

gimenete: 👍 I've just pushed a change to do the same if the server is `containers.pkg.github.com`

gimenete:
> but we should also not break the current behavior for docker.pkg.github.com

@taywrobel actually the code that was supposed to work with `docker.pkg.github.com` so it was never used.

> should add containers.pkg.github.com to the list of token-compatible registry hostnames

@taywrobel done here https://github.com/actions/runner/pull/990/commits/021e03c2e7c90234d7e93d036111564063d0eff7 and tested in `sweethaven-village/ghtoken_test`

taywrobel (approved):
LGTM with the extended registry list 👍🏽

TingluoHuang (approved):

PR #992, opened 2021-02-18
Allow auth token env var for pulling internal/org based actions

mike-donlon:
Follow similar logic used for GHES, to allow non GHES to pass an auth token for pulling private/internal/org actions

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -1135,10 +1135,21 @@ namespace GitHub.Runner.Worker
             {
                 return null;
             }
+            var authToken = Environment.GetEnvironmentVariable("_GITHUB_ACTION_TOKEN");
 
-            var base64EncodingToken = Convert.ToBase64String(Encoding.UTF8.GetBytes($"x-access-token:{token}"));
-            HostContext.SecretMasker.AddValue(base64EncodingToken);
-            return new AuthenticationHeaderValue("Basic", base64EncodingToken);
+            if (!string.IsNullOrEmpty(authToken))
+            {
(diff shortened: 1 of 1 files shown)


mike-donlon:
Adds functionality for [1005](https://github.com/actions/runner/issues/1005)

missedone:
wow, it's a long waiting feature, hope it can be approved and merge soon.

tahirmt:
Will this also allow private actions across GitHub enterprise?

mike-donlon:
> Will this also allow private actions across GitHub enterprise?

Yes, but only per runner, and only to the token allowed access assigned to the runner. I think they might be hesitant about this because an assumption is being made that this token will have elevated perms. But I have tested this with a read only token, specifically assigned to allow access to an internal repo(action repo). Basically, it could be dangerous if you are careless about permissions.

hross:
Yes we are definitely very hesitant about this. It may also allow circumventing policy features on the server (like restricting what actions execute) by providing a token that circumvents those permissions. This is a cool idea but it opens a pretty big policy/security hole so it's not something we are considering merging right now.

mike-donlon:
> Yes we are definitely very hesitant about this. It may also allow circumventing policy features on the server (like restricting what actions execute) by providing a token that circumvents those permissions. This is a cool idea but it opens a pretty big policy/security hole so it's not something we are considering merging right now.

For the record though, this code is already present in your codebase, but only available for github enterprise self hosted stuff. This doesn't really circumvent anything, only on the pull action stage it uses this env var rather than the runner's scoped working token. Technically, if this token is setup intelligently, then it would likely have less permissions than the scoped token because it would only have read permissions to your internal actions and nothing more.

PR #1028, opened 2021-03-29
Split out env parser and add tests

hross:
Makes environment variable parsing testable and adds a test


src/Runner.Listener/Program.cs
```diff
@@ -16,7 +16,7 @@ public static class Program
         public static int Main(string[] args)
         {
             // Add environment variables from .env file
-            LoadAndSetEnv();
+            EnvironmentParser.LoadAndSetEnvironment();
```
TingluoHuang, line 19: should we get the return dictionary and set the environment variables there? otherwise, it's a little bit weird that the return value is only used by L0 tests.

PR #1033, opened 2021-04-01
Make FileShare ReadWrite

lupine86:
L2 tests that use runners sometimes fail when attempting to retrieve the runner diagnostic logs with a message like: 
```
01:01.600  Could not read log file Runner_20210309-013553-utc.log, is the runner still running? System.IO.IOException: The process cannot access the file 'C:\l2-test-runners\StreamingLogsTests\_diag\Runner_20210309-013553-utc.log' because it is being used by another process.
             at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
             at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
             at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
             at System.IO.StreamReader.ValidateArgsAndOpenPath(String path, Encoding encoding, Int32 bufferSize)
             at System.IO.StreamReader..ctor(String path, Encoding encoding)
             at System.IO.File.InternalReadAllLines(String path, Encoding encoding)
             at System.IO.File.ReadAllLines(String path)
             at DistributedTask.TestFramework.Client.L2.ActionsTestRunner.LogDiagnosticLogs() in C:\a\1\s\Actions\Runtime\Client\TestFramework\L2\ActionsTestRunner.cs:line 102
```

I'm not able to reproduce this issue locally but I suspect it occurs when other file handles are opened while uploading the logs? Possibly this prevents other processes from accessing unless the same share level is specified on all handles.  I'm not sure why `FileShare.Read` which is the default isn't sufficient, but have seen others in stackoverflow and MSDN run into the same issue: 

https://stackoverflow.com/questions/27771319/cant-read-all-lines-in-file-that-being-used-by-another-process/27771453
https://social.msdn.microsoft.com/Forums/vstudio/en-US/df81389a-68bb-4fed-9f97-285da2f89415/systemiofilereadalllines-file-already-in-use?forum=csharpgeneral

### DIFF
--- a/src/Runner.Common/Logging.cs
+++ b/src/Runner.Common/Logging.cs
@@ -101,7 +101,7 @@ namespace GitHub.Runner.Common
             EndPage();
             _byteCount = 0;
             _dataFileName = Path.Combine(_pagesFolder, $"{_timelineId}_{_timelineRecordId}_{++_pageCount}.log");
-            _pageData = new FileStream(_dataFileName, FileMode.CreateNew, FileAccess.Write, FileShare.ReadWrite);
+            _pageData = new FileStream(_dataFileName, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.ReadWrite);
             _pageWriter = new StreamWriter(_pageData, System.Text.Encoding.UTF8);
         }
 
--- a/src/Runner.Common/Logging.cs
+++ b/src/Runner.Common/Logging.cs
@@ -101,7 +101,7 @@ namespace GitHub.Runner.Common
             EndPage();
             _byteCount = 0;
             _dataFileName = Path.Combine(_pagesFolder, $"{_timelineId}_{_timelineRecordId}_{++_pageCount}.log");
-            _pageData = new FileStream(_dataFileName, FileMode.CreateNew);
+            _pageData = new FileStream(_dataFileName, FileMode.CreateNew, FileAccess.Write, FileShare.ReadWrite);
             _pageWriter = new StreamWriter(_pageData, System.Text.Encoding.UTF8);
         }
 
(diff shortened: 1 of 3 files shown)


src/Runner.Common/Logging.cs
```diff
@@ -101,7 +101,7 @@ private void NewPage()
             EndPage();
             _byteCount = 0;
             _dataFileName = Path.Combine(_pagesFolder, $"{_timelineId}_{_timelineRecordId}_{++_pageCount}.log");
-            _pageData = new FileStream(_dataFileName, FileMode.CreateNew);
+            _pageData = new FileStream(_dataFileName, FileMode.CreateNew, FileAccess.Write, FileShare.ReadWrite);
```
TingluoHuang, line 104: maybe `FileAccess.ReadWrite, FileShare.Read` to only allow other processes to read?

lupine86: From [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor?view=net-5.0#System_IO_FileStream__ctor_System_String_System_IO_FileMode_) `FileShare.Read is the default for those FileStream constructors without a FileShare parameter.`

I'm not clear why FileShare.ReadWrite would be needed, but others have indicated it helped solve the problem.  My only theory is that we set the share level to ReadWrite when opening additional file handles for uploading the file?

You are right though that the access should probably be ReadWrite.  The docs say the default when not specified for this FileMode is ReadWrite.

TingluoHuang (approved):

PR #1038, opened 2021-04-06
do not trucate error message from template evaluation

ericsciple:


src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateEvaluator.cs
```diff
@@ -40,7 +40,7 @@ public class PipelineTemplateEvaluator
         /// <summary>
         /// Gets the maximum error message length before the message will be truncated.
         /// </summary>
-        public Int32 MaxErrorMessageLength => 500;
+        public Int32 MaxErrorMessageLength { get; set; } = 500;
```
TingluoHuang, line 43: any reason to not default this to `int.max`?

ericsciple: safe defaults for the server

ericsciple: (whereas the fixes i'm adding are on the client)

TingluoHuang (approved):

PR #1045, opened 2021-04-12
minor cleanup in composite

ericsciple:


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -254,8 +254,7 @@ public void RegisterPostJobStep(IStep step)
         }
 
         /// <summary>
-        /// Helper function used in CompositeActionHandler::RunAsync to
-        /// add a child node, aka a step, to the current job to the Root.JobSteps based on the location.
+        /// Creates a child step within a composite action
```
ericsciple, line 257: The comment was stale. This method doesn't take a location param anymore.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -153,29 +143,29 @@ private async Task RunStepsAsync(List<IStep> compositeSteps)
...
+                // Merge global env
                 foreach (var pair in ExecutionContext.Global.EnvironmentVariables)
                 {
                     envContext[pair.Key] = new StringContextData(pair.Value ?? string.Empty);
                 }
 
-                // Stomps over with outside step env
-                if (step.ExecutionContext.ExpressionValues.TryGetValue("env", out var envContextData))
+                // Merge composite-step env
+                if (ExecutionContext.ExpressionValues.TryGetValue("env", out var envContextData))
```
ericsciple, line 168: Simplified this a little bit to remove the indirection (i.e. previously `ExecutionContext.CreateCompositeStep` copied the composite step's env to the child step)

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -266,19 +265,6 @@ public void RegisterPostJobStep(IStep step)
             step.ExecutionContext = Root.CreateChild(_record.Id, _record.Name, _record.Id.ToString("N"), scopeName, step.Action.ContextName, logger: _logger, insideComposite: true, cancellationTokenSource: CancellationTokenSource.CreateLinkedTokenSource(_cancellationTokenSource.Token));
             step.ExecutionContext.ExpressionValues["inputs"] = inputsData;
             step.ExecutionContext.ExpressionValues["steps"] = Global.StepsContext.GetScope(step.ExecutionContext.GetFullyQualifiedContextName());
-
```
ericsciple, line 269: Simplified this a little bit. See comment in CompositeActionHandler.cs

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,37 +113,27 @@ private void ProcessCompositeActionOutputs()
...
                 foreach (var pair in actionOutputs)
                 {
-                    var outputsName = pair.Key;
-                    var outputsAttributes = pair.Value as DictionaryContextData;
-                    outputsAttributes.TryGetValue("value", out var val);
-
-                    if (val != null)
+                    var outputName = pair.Key;
+                    var outputDefinition = pair.Value as DictionaryContextData;
+                    if (outputDefinition.TryGetValue("value", out var val))
```
ericsciple, line 133: Simplified this a little bit. Empty value is OK

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,69 +109,59 @@ private void ProcessCompositeActionOutputs()
...
-                        }
+                        var outputValue = val.AssertString("output value");
+                        ExecutionContext.SetOutput(outputName, outputValue.Value, out _);
                     }
                 }
             }
         }
 
-        private async Task RunStepsAsync(List<IStep> compositeSteps)
+        private async Task RunStepsAsync(List<IStep> nestedSteps)
```
ericsciple, line 138: "compositeSteps" was a misnomer. These are the nested steps.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -136,7 +135,7 @@ public async Task RunAsync(IExecutionContext jobContext)
                             // Test the condition again. The job was canceled after the condition was originally evaluated.
                             jobCancelRegister = jobContext.CancellationToken.Register(() =>
                             {
-                                // mark job as cancelled
+                                // Mark job as cancelled
```
ericsciple, line 138: Spent a while skimming this file.... formatted all the comments consistently so it skims easier

src/Runner.Worker/action_yaml.json
```diff
@@ -32,10 +32,10 @@
         "outputs": {
             "mapping": {
                 "loose-key-type": "non-empty-string",
-                "loose-value-type": "outputs-attributes"
+                "loose-value-type": "output-definition"
```
ericsciple, line 35: Avoiding new terms here like "attributes" and instead using "definition" which aligns with the parser classes (technically schema classes)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ericsciple: Checked the schema, `value` is a `string` and will never be null

src/Runner.Sdk/BuildConstants.cs
```diff
@@ -1,19 +1,16 @@
...
 {
-    /***
-     * WARNING: This file is automatically regenerated on layout so the runner can provide version/commit info (do not manually edit it).
-     */
     public static class BuildConstants
     {
         public static class Source
         {
-            public static readonly string CommitHash = "N/A";
+            public static readonly string CommitHash = "83bafa4ab151239450460ee7cc906ddfba409571";
```
ericsciple, line 7: revert

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -274,9 +244,7 @@ private async Task RunStepAsync(IStep step)
             }
 
             Trace.Info($"Step result: {step.ExecutionContext.Result}");
-
-            // Complete the step context.
```
ericsciple, line 278: This comment was copied from stepsrunner but doesn't make sense here. We never call `.Complete()` on the nested steps.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -61,14 +61,15 @@ public interface IExecutionContext : IRunnerService
 
         bool EchoOnActionCommand { get; set; }
 
-        bool InsideComposite { get; }
+        bool IsEmbedded { get; }
```
ericsciple, line 64: Renamed this since ExecutionContext lower level and ideally doesn't know about "Composite Actions"

src/Runner.Worker/ExecutionContext.cs
ericsciple: Setting up the child execution context is now handled by `CreateEmbeddedChild` further below in this class

The logic to setup the expression values is performed by CompositeActionHandler now. That is a better place for it since it has to deal with setting up other expression values. And also is analogous to StepsRunner

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
 
         public async Task RunAsync(ActionRunStage stage)
         {
-            // Validate args.
```
ericsciple, line 29: Use the settings gear above -\> \"Hide whitespace changes\" when reviewing this file

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
...
-
-            foreach (Pipelines.ActionStep actionStep in actionSteps)
+            try
             {
-                var actionRunner = HostContext.CreateService<IActionRunner>();
-                actionRunner.Action = actionStep;
-                actionRunner.Stage = stage;
-                actionRunner.Condition = actionStep.Condition;
-
-                var step = ExecutionContext.CreateCompositeStep(childScopeName, actionRunner, inputsData, Environment);
```
ericsciple, line 63: Previously the CreateCompositeStep took `actionRunner` as a param, and returned the same instance after updating it, which was then assigned to a local variable `steps`. We had two local variables pointing to the same instance.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
...
-
-            // Create Context Data to reuse for each composite action step
-            var inputsData = new DictionaryContextData();
-            foreach (var i in Inputs)
-            {
-                inputsData[i.Key] = new StringContextData(i.Value);
-            }
-
-            // Initialize Composite Steps List of Steps
-            var compositeSteps = new List<IStep>();
```
ericsciple, line 46: Changed this variable name to `nestedSteps`. From a certain perspective, could be viewed as a misnomer before. Changed the name to draw a better distinction.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
...
-            {
-                // This is where we run each step.
-                await RunStepsAsync(compositeSteps);
+                // Run nested steps
+                await RunStepsAsync(nestedSteps);
 
-                // Get the pointer of the correct "steps" object and pass it to the ExecutionContext so that we can process the outputs correctly
+                // Set outputs
                 ExecutionContext.ExpressionValues["inputs"] = inputsData;
-                ExecutionContext.ExpressionValues["steps"] = ExecutionContext.Global.StepsContext.GetScope(ExecutionContext.GetFullyQualifiedContextName());
```
ericsciple, line 84: nothing changed here, just re-using the local variable instead since it's more descriptive (better for readability)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -113,69 +108,57 @@ private void ProcessCompositeActionOutputs()
...
-            ArgUtil.NotNull(compositeSteps, nameof(compositeSteps));
+            ArgUtil.NotNull(nestedSteps, nameof(nestedSteps));
 
-            // The parent StepsRunner of the whole Composite Action Step handles the cancellation stuff already. 
-            foreach (IStep step in compositeSteps)
+            foreach (IStep step in nestedSteps)
             {
-                Trace.Info($"Processing composite step: DisplayName='{step.DisplayName}'");
-
-                step.ExecutionContext.ExpressionValues["steps"] = ExecutionContext.Global.StepsContext.GetScope(step.ExecutionContext.ScopeName);
```
ericsciple, line 161: This line was previously here because of a bug. The wrong steps context was added when the step was initially created. So this line was needed to stomp over it with the correct one :)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -188,13 +171,11 @@ private async Task RunStepsAsync(List<IStep> compositeSteps)
                     }
                 }
 
-                step.ExecutionContext.ExpressionValues["env"] = envContext;
```
ericsciple, line 191: this line just moved up

ericsciple: At this point in the code, we're still in the middle of merging values into the env dictionary. So it makes more sense to move this statement up or down (i moved it up)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
...
+                // context name. Generated context names start with "__"
+                var childScopeName = ExecutionContext.GetFullyQualifiedContextName();
+                if (string.IsNullOrEmpty(childScopeName))
+                {
+                    childScopeName = $"__{Guid.NewGuid()}";
+                }
 
-                compositeSteps.Add(step);
-            }
+                // Create nested steps
```
ericsciple, line 52: per @TingluoHuang feedback, call this "embedded steps" which lines up w/ the execution context change

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -26,65 +26,60 @@ public sealed class CompositeActionHandler : Handler, ICompositeActionHandler
...
+                var nestedSteps = new List<IStep>();
+                foreach (Pipelines.ActionStep stepData in Data.Steps)
+                {
+                    var step = HostContext.CreateService<IActionRunner>();
```
ericsciple, line 56: per @TingluoHuang consider rename `embeddedStep`

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -188,13 +171,11 @@ private async Task RunStepsAsync(List<IStep> compositeSteps)
...
                 }
 
-                step.ExecutionContext.ExpressionValues["env"] = envContext;
-
                 var actionStep = step as IActionRunner;
 
                 try
                 {
-                    // Evaluate and merge action's env block to env context
+                    // Evaluate and merge nested-step env
```
ericsciple, line 178: find everywhere we use "nested" and replace with "embedded"

TingluoHuang (approved):

Issue #795, opened 2020-11-07
Post run actions/checkout@v2 fails if NODE_ICU_DATA is set


eemeli:
**Describe the bug**
Node.js before v14 ships with the small-icu dataset by default. If a CI job needs the full-icu data instead, these commands will set that up:
```
npm install -g full-icu
echo "NODE_ICU_DATA=`node-full-icu-path 2>/dev/null`" >> $GITHUB_ENV
```

In a Node.js v12 environment, this will lead to the "Post run actions/checkout@v2" step failing with this error:
```
/home/runner/runners/2.273.6/externals/node12/bin/node: could not initialize ICU (check NODE_ICU_DATA or --icu-data-dir parameters)
```

This almost certainly isn't a bug in the checkout action, as the failure is happening before its script even gets a chance to execute during the post run step.

**To Reproduce**
1. Go to https://github.com/eemeli/gh-action-test/runs/1367527380?check_suite_focus=true
2. See error

**Expected behavior**
The post run script should not fail.

## Runner Version and Platform
Version of your runner? 2.273.6

OS of the machine running the runner? GitHub Actions

## What's not working?
I suspect that the `/opt/hostedtoolcache/node/12.19.0/x64/lib/node_modules/full-icu/icudt67l.dat` location where the full-icu data is installed is not available during the post run step, even though the env var is.

Manually setting `NODE_ICU_DATA` to an empty value as a last step allows for the post run step to complete successfully.

## Job Log Output
See https://github.com/eemeli/gh-action-test/actions

## Runner and Worker's Diagnostic Logs
See https://github.com/eemeli/gh-action-test/actions

PR #1060, opened 2021-04-23
Unset NODE_ICU_DATA before starting node in NodeHandler.

TingluoHuang:
Fix: https://github.com/actions/runner/issues/795


src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -97,6 +97,9 @@ public async Task RunAsync(ActionRunStage stage)
             Encoding outputEncoding = null;
 #endif
 
+            // Remove environment variable that may conflict the the node within the runner.
```
hross, line 100:
```suggestion
            // Remove environment variable that may cause conflicts with the node within the runner.
```

hross (dismissed):

hross (approved):

PR #1067, opened 2021-04-27
Add Runner group support to sample startup script.

kassemsandarusi:
This seems like something that should have already been in the script, but basically we have a few use cases where we have runners join groups and we want that to happen as early as possible and as simply as possible.

Considering its already an argument on the underlying runner CLI this seems relatively simple to add.


lastthyme (approved):

kassemsandarusi:
Ping!

kassemsandarusi:
Closed because another PR was accepted that did the exact same thing.

PR #1086, opened 2021-05-11
Pull option

### LINKED ISSUE #791: Option to not do docker pull?

rowleya:
Resolves #791, replaces #800

This provides an option to do the pull rather than just testing if the container exists locally.  This has the advantage over the #800 that this can still do a pull even if the container exists locally, and also avoids issues with exceptions mentioned there.


AkhilTThomas:
I would also like to have this feature, Can this be merged?

Issue #772, opened 2020-10-28
Supporting Node.js 14 and 16 LTS as the runtime of JavaScript action


peaceiris:
**Describe the enhancement**

Node.js v14.15.0 has been released and the 14.x release line now moves into Active LTS. Can the GitHub Actions runner support it as the runtime of a JavaScript (TypeScript) action? Is there a roadmap for this?

**Code Snippet**

Today, we can set `node12` to `runs.using`.

```yaml
runs:
  using: 'node12'
  main: 'main.js'
```

Can we set `node14` (or `node16`) as follows?

```yaml
runs:
  using: 'node14'
  main: 'main.js'
```

**Additional information**

https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-javascript-actions

We also need to update the documentation when `node14` is available.

viceice:
Looks like this is not easy 😭 

https://github.com/actions/runner/blob/89a13db2c3b72e87d7c42798f39edb16df822f31/src/Runner.Worker/ActionManifestManager.cs#L454-L471

ad-m:
Is it done via actions/virtual-environments#1953 ?

peaceiris:
No. https://github.com/peaceiris/actions-gh-pages/pull/539

> https://github.com/peaceiris/actions-gh-pages/pull/539/checks?check_run_id=1437553870#step:19:20
> Error: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter ''using: node14' is not supported, use 'docker' or 'node12' instead.')

TingluoHuang:
We need:
- compile alpine node14
- update the ./external.sh script to download the `node14`
- update C# code and yaml schema to support `node14`
- update documentation

littlewall:
Hello,
Is there any update on this or timeline of when to expect this feature? 

Thank you,
Milan Zítka

ianwalter:
@TingluoHuang I started working on this but I don't have any significant experience with C# so any help or direction would be much appreciated.

TingluoHuang:
Should we just wait for node.js 16 to become LTS and then directly goes to node 16?  I don't see too many benefits that go to node 14.

mktcode:
"Active LTS Start: 2021-10-26"
https://nodejs.org/en/about/releases/

👍🏻

bryanmacfarlane:
[...]

PR #1090, opened 2021-05-13
Support for Node.js 14 runtime

ianwalter:
A solution for #772: Supporting Node.js 14 LTS as the runtime of JavaScript action


ethomson:
Thanks for opening this!  Now that node16 is LTS, we can target that instead.  See #1439 for that work.  I'm going to close this issue.

PR #1100, opened 2021-05-20
Add message size in chars and bytes

nodeselector:
Adds `Job Message` size to both Worker and Listener logs for debugging. This would be useful in a situation where a user wants to debug is being malformed while being sent from Listener to Worker.

### DIFF
--- a/src/Runner.Common/ProcessChannel.cs
+++ b/src/Runner.Common/ProcessChannel.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.IO.Pipes;
 using System.Threading;
 using System.Threading.Tasks;
+using GitHub.Runner.Sdk;
 
 namespace GitHub.Runner.Common
 {
@@ -68,6 +69,7 @@ namespace GitHub.Runner.Common
 
         public async Task SendAsync(MessageType messageType, string body, CancellationToken cancellationToken)
         {
+            Trace.Info($"Sending message of length {body.Length}, with hash '{IOUtil.GetSha256Hash(body)}'");
             await _writeStream.WriteInt32Async((int)messageType, cancellationToken);
             await _writeStream.WriteStringAsync(body, cancellationToken);
         }
@@ -77,6 +79,7 @@ namespace GitHub.Runner.Common
             WorkerMessage result = new WorkerMessage(MessageType.NotInitialized, string.Empty);
             result.MessageType = (MessageType)await _readStream.ReadInt32Async(cancellationToken);
             result.Body = await _readStream.ReadStringAsync(cancellationToken);
+            Trace.Info($"Receiving message of length {result.Body.Length}, with hash '{IOUtil.GetSha256Hash(result.Body)}'");
             return result;
         }
 
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -268,7 +268,6 @@ namespace GitHub.Runner.Listener
                 }
 
                 Trace.Info($"Message '{message.MessageId}' received from session '{_session.SessionId}'.");
-                Trace.Info($"Message '{message.MessageId}' char length: '{message.Body.Length}', byte length '{System.Text.ASCIIEncoding.Unicode.GetByteCount(message.Body)}'.");
                 return message;
             }
         }
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -268,6 +268,7 @@ namespace GitHub.Runner.Listener
                 }
 
                 Trace.Info($"Message '{message.MessageId}' received from session '{_session.SessionId}'.");
+                Trace.Info($"Message '{message.MessageId}' char length: '{message.Body.Length}', byte length '{System.Text.ASCIIEncoding.Unicode.GetByteCount(message.Body)}'.");
                 return message;
             }
         }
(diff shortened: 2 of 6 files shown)


src/Runner.Listener/MessageListener.cs
```diff
@@ -268,6 +268,7 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                 }
 
                 Trace.Info($"Message '{message.MessageId}' received from session '{_session.SessionId}'.");
+                Trace.Info($"Message '{message.MessageId}' char length: '{message.Body.Length}', byte length '{System.Text.ASCIIEncoding.Unicode.GetByteCount(message.Body)}'.");
```
ericsciple, line 271: i think we want to do this tracing in src/Runner.Common/ProcessChannel.cs instead in SendAsync and ReceiveAsync

ericsciple: thinking.... instead of tracing byte length can we trace the string length and a hash of the contents

That should tell us whether the string got malformed in the process. For instance if any codepoints didnt roundtrip properly

ericsciple: Actually maybe not bad to go ahead and trace the string length here.

But you definitely want to trace in the processchannel too. The reason is, message.Body isn't what get's sent to the worker process. It gets deserialized and serialized again before being sent via processchannel.

nodeselector: 👋  Thanks for the feedback! That's a good point about tracing in the process channel since that's closer to the failure point in the scenario that this trace would be used for. I've gone ahead and done that and left the other traces in. If you think they're valuable let's keep them, if not let's yank them out.

src/Runner.Worker/Worker.cs
```diff
@@ -59,6 +59,9 @@ public async Task<int> RunAsync(string pipeIn, string pipeOut)
                         channelMessage = await channel.ReceiveAsync(csChannelMessage.Token);
                     }
 
+                    // This is useful for validating that the message wasn't truncated when being sent over the anonymous pipe.        
+                    Trace.Info($"Job Message char length: '{channelMessage.Body.Length}', byte length '{System.Text.Encoding.UTF8.GetByteCount(channelMessage.Body)}'.");
```
anywherepilot, line 63: Even just this would be very useful. Can we merge this?

anywherepilot (dismissed):
This would really help debug https://github.com/github/c2c-actions-support/issues/128
Let's deploy it.

src/Runner.Common/ProcessChannel.cs
```diff
@@ -96,5 +98,21 @@ private void Dispose(bool disposing)
                 _outClient?.Dispose();
             }
         }
+
+        //  Create a deterministic hash for a string to ensure that the message is not malformed on delivery
+        private static string GetMessageHash(string message)
+        {
+            var md5 = System.Security.Cryptography.MD5.Create();
+            byte[] input = System.Text.Encoding.UTF8.GetBytes(message);
+            byte[] hash = md5.ComputeHash(System.Text.Encoding.UTF8.GetBytes(message));
```
ericsciple, line 107: @TingluoHuang do you know whether this will this flag security scans ? do we need to use sha256 or something?

src/Runner.Worker/Worker.cs
ericsciple: nit: do we need this now that ProcessChannel.ReceiveAsync will traces?

src/Runner.Listener/MessageListener.cs
```diff
@@ -268,6 +268,7 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                 }
 
                 Trace.Info($"Message '{message.MessageId}' received from session '{_session.SessionId}'.");
+                Trace.Info($"Message '{message.MessageId}' char length: '{message.Body.Length}', byte length '{System.Text.Encoding.UTF8.GetByteCount(message.Body)}'.");
```
ericsciple, line 271: nit: does this add any value now that ProcessChannel.SendAsync traces?

src/Runner.Common/ProcessChannel.cs
ericsciple: for fips compliance

TingluoHuang: we should avoid using MD5, sha256 is better.

src/Runner.Listener/MessageListener.cs
nodeselector: I don't think it does.

src/Runner.Worker/Worker.cs
nodeselector: Nope!

src/Runner.Common/ProcessChannel.cs
```diff
@@ -96,5 +98,21 @@ private void Dispose(bool disposing)
                 _outClient?.Dispose();
             }
         }
+
+        //  Create a deterministic hash for a string to ensure that the message is not malformed on delivery
+        private static string GetMessageHash(string message)
```
TingluoHuang, line 103: There is an `IOUtil.GetPathHash()` that doing the exact thing, you might want to just reuse that.
Maybe rename that method to `IOUtil.GetSha256Hash()`

TingluoHuang (approved):

Issue #1023, opened 2021-03-25
Make Runner Start up and Compile/Debug Friendlier


hross:
When you fire up the runner for the first time, we lack some basics that we should improve:

* Standardize compile directories for worker and listener so they can be launched from a solution file
* Add a better debug story/breakpoint for the worker process (maybe a flag to break on debug or something similar)
* Fix SDK standardization, if needed (not sure we really need this)
* Does auto-generated compile file need to be autogenerated? Should we just do this on release?
* Make sure this works in Rider, VS Code, Visual Studio

fhammerl:
- Standardize compile directories for worker and listener so they can be launched from a solution file
  - Will not fix: Decided to focus on VS Code developer experience
- Add a better debug story/breakpoint for the worker process (maybe a flag to break on debug or something similar)
  - Done
- Fix SDK standardization, if needed (not sure we really need this)
  - For now, we keep the SDK source as part of the runner repository
- Does auto-generated compile file need to be autogenerated? Should we just do this on release?
  - Follow up issue: [] Convert BuildConstants.cs to BuildConstants.json
- Make sure this works in Rider, VS Code, Visual Studio
  - Will not fix: Decided to focus on VS Code developer experience

PR #1101, opened 2021-05-20
Wait for debugger to attach in Runner.Worker process

fhammerl:
This PR stalls the Runner.Worker process until a debugger is connected if `GITHUB_ACTIONS_RUNNER_ATTACH_DEBUGGER` environment variable is set to true.

Related to https://github.com/actions/runner/issues/1023

### DIFF
--- a/src/Runner.Worker/Program.cs
+++ b/src/Runner.Worker/Program.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Globalization;
 using System.Threading.Tasks;
 using GitHub.Runner.Common;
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/Program.cs
```diff
@@ -64,5 +69,24 @@ public static async Task<int> MainAsync(IHostContext context, string[] args)
...
+        /// <summary>
+        /// Runner.Worker is started by Runner.Listener in a separate process,
+        /// so the two can't be debugged in the same session.
+        /// This method halts the Runner.Worker process until a debugger is attached,
+        /// allowing a developer to debug Runner.Worker from start to finish.
+        /// </summary>
+        private static async Task WaitForDebugger(Tracing trace)
+        {
+            trace.Info($"Waiting for a debugger to be attached. Edit the 'GITHUB_ACTIONS_RUNNER_ATTACH_DEBUGGER' environment variable to toggle this feature.");
+            while (!Debugger.IsAttached)
```
hross, line 84: can we wait for some max time in case you don't attach a debugger

fhammerl: Sure, I'm thinking 30s would probably be appropriate.

src/Runner.Worker/Program.cs
```diff
@@ -19,11 +19,16 @@ public static int Main(string[] args)
 
         public static async Task<int> MainAsync(IHostContext context, string[] args)
         {
+            Tracing trace = context.GetTrace(nameof(GitHub.Runner.Worker));
+            if (bool.TryParse(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_ATTACH_DEBUGGER"), out bool waitForDebugger) && waitForDebugger)
```
TingluoHuang, line 23: use `StringUtil.ConvertToBoolean()` instead of `bool.TryParse()`

TingluoHuang: Do you have your listener side change ready? also the launch.task changes for VSCode?

TingluoHuang (approved):

Issue #1056, opened 2021-04-22
`Load failed: 5: Input/output error` on ./svc.sh run


callum-tait-pbx:
**Describe the bug**
Just deployed 2 new self-hosted runners on Mac Minis, I get the error `Load failed: 5: Input/output error` when I install and then start the runner as a service. Interestingly the service appears to start and the runner appears online on github.com
**Silicon** Intel
**OS Version** Big Sur 11.2.3

**To Reproduce**
1. Install the latest self-hosted runner `./config.sh --url $URL --token $TOKEN`
2. Install the runner as a service `./svc.sh install`
2. Run the service  `./svc.sh run`

**Expected behavior**
No error is posted

## Runner Version and Platform
**Silicon** Intel
**OS Version** Big Sur 11.2.3
**Runner Version** v2.278.0 

## What's not working?
Currently testing will update

## Job Log Output
```
starting actions.runner.org.hostname
Load failed: 5: Input/output error
status actions.runner.$org.$hostname:

/Users/runner/Library/LaunchAgents/actions.runner.org.hostname.plist

Started:
512 0 actions.runner.$org.$hosntame
```

## Runner and Worker's Diagnostic Logs
If you google `load failed 5 input/output error` you will find lots of posts about this error on a wide array of services with lots fo references to Big Slur. Something has changed in relation to Big Sur and services and the plist that gets auto generated by the install script / the values within the plist no longer being valid for some reason.

TingluoHuang:
🤦  we need to test the runner on `Big sur` and verify the behavior.
@hross FYI, since we can't upgrade our MacBook to Big Sur yet... 😢

liya2017:
@callum-tait-pbx Hi,have you solved the problem?

callum-tait-pbx:
I haven't got around to trying to solve the issue personally, functionally they seem to work but it's definitely an issue that needs solving. Based on my research at the time I think it relates to the deprecation of the `load` command. If I get a chance next week I'll messing with the svc.sh script.

brandon-bird-yolabs:
Checking ` /var/log/system.log` I saw what looked like a permission error and not one with the `load` command itself. 
`(com.apple.xpc.launchd.domain.system): Caller not allowed to perform action: launchctl.9739, action = bootout, code = 1: Operation not permitted,` 

A potential temporary workaround that seems functional is to remove the sudo check from `svc.sh` then run the install and start commands as sudo.

brandon-bird-yolabs:
[...]

Issue #947, opened 2021-01-27
Provide an installer for running in a service account on self hosted Mac


ericvergnaud:
Thank you 🙇‍♀ for wanting to create a feature in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions or issues with the service, writing workflows or actions, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

If you have a feature request that is relevant to this repository, the runner, then please include the information below:

**Describe the enhancement**
Currently the Github action runner service can only be installed by an interactive user.
This goes against what one expects from a CI i.e. run as a service independently of any user logged in.

**Code Snippet**
Not relevant

**Additional information**
Whilst it may seem a good idea to have 1 user account open a session automatically at startup, and have multiple runners installed under that user account (by running the Github agent runner many times with different names and tokens), it is not practical. Software such as Maven relies on a single .settings.xml tied to the account, so 4 runners running maven in parallel would compete and potentially corrupt the maven repo. This is just one example. We need n accounts in order to have more than 1 runner per self-hosted Mac. And they need to be service accounts in order for the runner to restart automatically.

Creating a service account on a Mac is quite difficult (there is no adduser) but it's possible (see attached).
You can then su into the service account and install the standard runner (see attached).
However, you'll find that the action runner service fails to run.
There are many causes for that: missing dirs (Library/LaunchAgents...), authorizations...

What would be really great would be the following:
-when initiating creation of a self-hosted runner from the Github UI, propose selection between UI account and service account

 - if the latter, propose a downloadable script that can be run using sudo after a chmod 777
 - the script itself would receives a few parameters (service group name, user name...)
 - it would automate the creation of the service account
 - it would automate the creation of the runner, and install a service that:
    - does not rely on interactive account pre-requisites
    - restarts automatically without requiring the user to login
    

[create-action-runner.txt](https://github.com/actions/runner/files/5882542/create-action-runner.txt)
[create-ci-users.txt](https://github.com/actions/runner/files/5882543/create-ci-users.txt)

[...]

Issue #349, opened 2020-03-01
OSX Runner service fails to start


jl-gogovapps:
**Describe the bug**

When I ssh into my OSX box, and try to start the service, I receive the following error:
`Could not find domain for port (Aqua)`

The command I'm running is `cd actions-runner && ./svc.sh install`

The ./run.sh command works without issue

**To Reproduce**
./svc.sh start

**Expected behavior**
- The service starts
- Rebooting the machine starts the service automatically

## Runner Version and Platform
Version of your runner?
osx-x64-2.165.2

OS of the machine running the runner? OSX/Windows/Linux/...
OSX Catalina

## What's not working?

`Could not find domain for port (Aqua)`

## Job Log Output
```
$ ./svc.sh install
Creating launch runner in /Users/xxx/Library/LaunchAgents/actions.runner.xxx-xxx.xxx-build-osx.plist
Creating /Users/xxx/Library/Logs/actions.runner.xxx-xxx.xxx-build-osx
Creating /Users/xxx/Library/LaunchAgents/actions.runner.xxx-xxx.xxx-build-osx.plist
Creating runsvc.sh
Creating .service
svc install complete
```

## Runner and Worker's Diagnostic Logs
N/A

TingluoHuang:
@jl-gogovapps i did a quick google search on the error, looks like related permission around the SSH user and the service installed user.
what user did you use to run `./svc.sh install` and what user did you use to SSH?

jl-gogovapps:
Hey @TingluoHuang, thanks for checking in on it.   I'm ssh-ing, and then running `./svc.sh install` so it's the same user.  It's also the only primary user on the machine.

jl-gogovapps:
Looks like this did the trick for me!

```
sudo cp {/Users/xxx/Library/LaunchAgents,/Library/LaunchDaemons}/your.plist
```

I was able to reboot my machine without logging in and see the runner active

raphaelcruzeiro:
This is happening on a fresh macOS install.

ivan-gaydamakin:
Solution by @jl-gogovapps not good, because LaunchDaemons doesn't have access to keychain.

Better copy to LaunchAgents Like:
sudo cp {/Users/xxx/Library/LaunchAgents,/Library/LaunchAgents}/your.plist

zackrspv:
[...]

PR #1102, opened 2021-05-20
Add SessionCreate so keychain can be used in macos runners

brandonbirdj:
When using a self-hosted macos runner with big sur I ran into several errors like the following

```
Failure to decode /[REDACTED PROVISIONING PROFILE]. Exit: 1: security: cert import failed: Write permissions error.
```

I found this similar issue effecting jenkins https://github.com/fastlane/fastlane/issues/9182

I have tested this by adding these changes to my plist, restarting the service and no longer have this issue.

Issues: https://github.com/actions/runner/issues/1056 https://github.com/actions/runner/issues/947 https://github.com/actions/runner/issues/349

### DIFF
--- a/src/Misc/layoutbin/actions.runner.plist.template
+++ b/src/Misc/layoutbin/actions.runner.plist.template
@@ -13,7 +13,9 @@
     <key>WorkingDirectory</key>
     <string>{{RunnerRoot}}</string>
     <key>RunAtLoad</key>
-    <true/>    
+    <true/>
+    <key>SessionCreate</key>
+    <true/>
     <key>StandardOutPath</key>
     <string>{{UserHome}}/Library/Logs/{{SvcName}}/stdout.log</string>
     <key>StandardErrorPath</key>


src/Misc/layoutbin/actions.runner.plist.template
```diff
@@ -13,7 +13,9 @@
     <key>WorkingDirectory</key>
     <string>{{RunnerRoot}}</string>
     <key>RunAtLoad</key>
-    <true/>    
+    <true/>
+    <key>SessionCreate</key>
```
TingluoHuang, line 17: not sure we can set this for everyone, it might break codesigning in some cases.

brandon-bird-yolabs: Can I help alleviate that concern? I can list some example tests
The following behaves very differently if running the service via `./run.sh` or `./svc.sh` 
`- run: security list-keychains` 

My changes bring `./svc.sh` inline with `./run.sh`.

Would you recommend somehow conditionally generating the template instead?

sanmai-NL: @TingluoHuang Can you pick up this PR again please?

frederickjansen:
This PR is no longer relevant. `SessionCreate` was added in https://github.com/actions/runner/pull/847.

PR #1104, opened 2021-05-23
Update 0263-proxy-support.md

jsoref:
`ect` looks like a typo for `etc.`, but it's mostly incompatible with `such as`, and thus, it's better to just drop it.


fhammerl (approved):
LGTM

PR #1112, opened 2021-05-23
Regularize notation

jsoref:
The docs https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions clearly indicate `::`.

I understand a previous iteration of the environment may have used `##[...]`, but the current one does not, and messages should reflect how users are supposed to use things today.

### DIFF
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -92,7 +92,7 @@ namespace GitHub.Runner.Worker
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


jsoref:
Fwiw, without this change, I have a workflow which used an action which did: `echo ::set-output::` and it triggered this:
> workflow Unable to process command '::set-output::' successfully.
> workflow Required field 'name' is missing in ##[set-output] command.

Which is pretty clearly wrong.

fhammerl (dismissed):

PR #1124, opened 2021-05-27
Self update fix when running under impersonation

### LINKED ISSUE #1123: Windows runner loses identity after self-update
**Describe the bug**

If the runner starts using impersonation (e.g. a scheduled task run under a logged out user) and the runner self-updates, the resulting process does not keep the full user profile loaded. Workflows using certain Windows cryptographic APIs will fail until the runner process is restarted.

**To Reproduce**

1. Create a local user account (e.g. runneruser)
2. Make sure that user is not logged in (so that their profile is not already loaded)
3. Install a version of the Actions Runner that is at least one version older than the latest
4. Configure the runner with GitHub.com using the `config.cmd` file
5. Launch the runner with impersonation (for example using a Scheduled Task, passing in the user’s credentials)
6. Run a workflow, which will cause the runner to self-update before execution
7. Any workflows run after this point will fail if they need access to certain Windows cryptographic APIs
8. Stop the currently executing runner process and launch it again using a scheduled task
9. Run the same workflow again. Note that the workflow now runs fine.

**Expected behavior**

Workflows should not fail or exhibit different behavior after the runner updates

Windows runner 2.277.1 (updating to 2.278.0)

## What's not working?

The update completes successfully, but workflows using certain Windows cryptographic APIs will fail when run by the new runner process. For example, this line of PowerShell:

```powershell
New-Object System.Security.Cryptography.X509Certificates.X509Certificate2("key.pfx", "")
```

Will fail with:

```
Exception calling ".ctor" with "2" argument(s): "The specified network password is not correct."
```

[This example workflow](https://github.com/MatisseHack/pfx/blob/main/.github/workflows/pfx.yml) is a good demonstration of the issue.

MatisseHack:
Closes #1123 by reusing the existing ephemeral runner update process for static runners. The user profile remains properly loaded since ephemeral runners use the original `run.cmd` script to restart the process rather than starting a new process from the update script.

### DIFF
--- a/src/Misc/layoutroot/run.cmd
+++ b/src/Misc/layoutroot/run.cmd
@@ -24,10 +24,10 @@ if /i "%~1" equ "localRun" (
   rem ********************************************************************************
   "%~dp0bin\Runner.Listener.exe" run %*
 
-  rem Return code 4 means the run once runner received an update message.
+  rem Return codes 3 and 4 mean the runner received an update message.
   rem Sleep 5 seconds to wait for the update process finish and run the runner again.
-  if ERRORLEVEL 4 (
+  if ERRORLEVEL 3 (
     timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
+    "%~dp0run.cmd"
   )
 )
--- a/src/Misc/layoutroot/run.cmd
+++ b/src/Misc/layoutroot/run.cmd
@@ -28,6 +28,6 @@ if /i "%~1" equ "localRun" (
   rem Sleep 5 seconds to wait for the update process finish and run the runner again.
   if ERRORLEVEL 3 (
     timeout /t 5 /nobreak > NUL
-    "%~dp0run.cmd"
+    "%~dp0run.cmd" %*
   )
 )
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -396,7 +396,14 @@ namespace GitHub.Runner.Listener
                                     autoUpdateInProgress = true;
                                     var runnerUpdateMessage = JsonUtility.FromString<AgentRefreshMessage>(message.Body);
                                     var selfUpdater = HostContext.GetService<ISelfUpdater>();
-                                    selfUpdateTask = selfUpdater.SelfUpdate(runnerUpdateMessage, jobDispatcher, !runOnce && HostContext.StartupType != StartupType.Service, HostContext.RunnerShutdownToken);
+
+#if OS_WINDOWS
+                                    var restartInteractiveRunner = false;
+#else
+                                    var restartInteractiveRunner = !runOnce;
+#endif
+
+                                    selfUpdateTask = selfUpdater.SelfUpdate(runnerUpdateMessage, jobDispatcher, restartInteractiveRunner, HostContext.RunnerShutdownToken);
                                     Trace.Info("Refresh message received, kick-off selfupdate background process.");
                                 }
                                 else


src/Runner.Listener/Runner.cs
```diff
@@ -396,7 +396,14 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                                     autoUpdateInProgress = true;
                                     var runnerUpdateMessage = JsonUtility.FromString<AgentRefreshMessage>(message.Body);
                                     var selfUpdater = HostContext.GetService<ISelfUpdater>();
-                                    selfUpdateTask = selfUpdater.SelfUpdate(runnerUpdateMessage, jobDispatcher, !runOnce && HostContext.StartupType != StartupType.Service, HostContext.RunnerShutdownToken);
+
+#if OS_WINDOWS
+                                    var restartInteractiveRunner = false;
```
ericsciple, line 401: is this going to break runners running interactively? also is this going to break runners running as a windows service?

ericsciple: i.e. will they come back online

src/Misc/layoutroot/run.cmd
```diff
@@ -24,10 +24,10 @@ if /i "%~1" equ "localRun" (
...
   "%~dp0bin\Runner.Listener.exe" run %*
 
-  rem Return code 4 means the run once runner received an update message.
+  rem Return codes 3 and 4 mean the runner received an update message.
   rem Sleep 5 seconds to wait for the update process finish and run the runner again.
-  if ERRORLEVEL 4 (
+  if ERRORLEVEL 3 (
     timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
+    "%~dp0run.cmd"
```
ericsciple, line 31: can you explain more about this change. It's not obvious to me what this fixes

src/Runner.Listener/Runner.cs
MatisseHack: It did not affect interactive runners in my tests. This line just disables restarting the runner from the update script:

https://github.com/actions/runner/blob/93ec16e14f6d3bca65df10e137dc0132aed8fe73/src/Misc/layoutbin/update.cmd.template#L128-L133

The runner listener will still return one of these two update related exit codes (4 and 3 respectively):

https://github.com/actions/runner/blob/93ec16e14f6d3bca65df10e137dc0132aed8fe73/src/Runner.Listener/Runner.cs#L352-L359

Both those exit codes are then handled by the run.cmd script. The only difference from a user perspective is that the updated runner reuses the original command prompt window instead of creating a new one.

src/Misc/layoutroot/run.cmd
```diff
@@ -24,10 +24,10 @@ if /i "%~1" equ "localRun" (
   rem ********************************************************************************
   "%~dp0bin\Runner.Listener.exe" run %*
 
-  rem Return code 4 means the run once runner received an update message.
+  rem Return codes 3 and 4 mean the runner received an update message.
   rem Sleep 5 seconds to wait for the update process finish and run the runner again.
-  if ERRORLEVEL 4 (
+  if ERRORLEVEL 3 (
```
ericsciple, line 29: @TingluoHuang can i get your eyes on this. It looks like a bug but all related history goes back to the initial squash so i'm confused why `if ERRORLEVEL4` works (see const `RunnerUpdating`)

ericsciple: @fhammerl and i looked at this together

TingluoHuang: 3 is a regular update, 4 is run once agent update.

ericsciple: @TingluoHuang why sometimes does this script restart the runner, and sometimes the runner is restarted by `start "Actions Runner" cmd.exe /k "_ROOT_FOLDER_\run.cmd"` in `src/Misc/layoutbin/update.cmd.template`?

src/Misc/layoutroot/run.cmd
```diff
@@ -24,10 +24,10 @@ if /i "%~1" equ "localRun" (
...
     timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
+    "%~dp0run.cmd" %*
```
ericsciple, line 31: if in some cases, this script restarts the runner (i.e. when "run once" runner) then it looks like this change is required if the runner were to receive a second update

(although i wonder if a `goto` would make more sense, e.g. to effectively create a loop)

src/Runner.Listener/Runner.cs
TingluoHuang: If you run `whoami` to check the runner run-as user, what is the user before and after the update?

src/Misc/layoutroot/run.cmd
ericsciple: @MatisseHack curious is `start` what is causing the issue you are seeing? if so then based on your proposed fix, it looks like `start` command would never be invoked anymore (if we go that route, we should delete it)

ericsciple: i guess `start` inherits the user profile from the logged in session, not from the process?

src/Runner.Listener/Runner.cs
ericsciple: ok i see now there are two different places where we restart - sometimes the cmd file starts the runner again and sometimes the update script restarts the runner

TingluoHuang: I am not sure whether this is the right fix, 
I think we should check:
- The way we invoke the generated update.cmd script from the runner, we use `Process.Start()`, do we miss anything?
- How the update.cmd restart the runner process, we just `START`, do we miss any parameter?

src/Misc/layoutroot/run.cmd
ericsciple: otherwise if you dont use `goto` then you end up with a bunch of parent cmd.exe processes after a bunch of updates.

src/Misc/layoutroot/run.cmd
TingluoHuang: https://github.com/actions/runner/pull/1124#discussion_r641764098

MatisseHack: Replied above, but unfortunately removing `START` was not the fix. I did narrow down the issue to C#'s method of creating a new process.

src/Misc/layoutroot/run.cmd
MatisseHack: I'm certainly not a CMD guru, so I might well be wrong, but I thought calling another CMD file resulted in that file "taking over" the current process. In my testing I've at least observed that the output from the updated runner shows up in the same command prompt window. Happy to use `goto` if that is preferred though!

src/Runner.Listener/Runner.cs
MatisseHack: Interesting, I just ran another test where I changed `"%~dp0run.cmd" %*` to `start "Actions Runner" cmd.exe /k "%~dp0run.cmd" %*` in run.cmd, which also resulted in a failure. So `START` certainly compounds the issue, but removing it is not enough. Creating a new process from C# is still an issue.

MatisseHack: For completeness, I ran another test where I used `invokeScript.StartInfo.UseShellExecute = true` and removed the `START` call from the update script (I wasn't 100% sure I had tested this scenario before). Unfortunately it still resulted in failure.

src/Misc/layoutroot/run.cmd
ericsciple: ok cool i think you might be right. i forgot the default behavior isn't CALL. instead the script interpreter executes the other script never returns :)

side note, iirc we can't update the cmd file because the cmd.exe interpreter only loads a line at a time and we can't replace the cmd file while it's being executed. So calling the cmd file again does not enable us to use an updated cmd file. Disclaimer, i havent thought about these details in years, so i might be misremembering. @TingluoHuang might remember offhand whether i'm wrong :)

TingluoHuang:
I will ask @fhammerl to take over this PR and test/merge the fix if needed.

src/Runner.Listener/Runner.cs
ericsciple: Thanks @MatisseHack this is awesome data

Would you be interested in reopening this PR against a [feature branch](https://github.com/actions/runner/tree/features/consolidate-restart)? I like the proposed approach in this PR, but i think there is some more follow-up we want to do (@fhammerl has bandwidth to follow-through if that helps). I like that we're converging the restart code into one place. We should also delete the code from the update.template.cmd. And we should change Linux/macOS code paths so behaviors align across all platforms.

Since we have additional changes to make, I would rather do one big test pass at the end rather. So i'm hesitant for this to go into main yet. Runner update high risk. If we mess something up, then all customer's self-hosted runners will fail to update or they may have to manually restart their runners. IIRC sometimes hosted runners also encounter the auto-update path too.

@fhammerl one thing to look out for, is to think-through or verify the runner can successfully update twice :) - we had that bug in the past :(

MatisseHack: No problem, I'll retarget this PR to that branch!

I think keeping Linux and macOS consistent with Windows is a great idea. I'm happy to make those changes if it would help move things along, but happy to let @fhammerl take over as well if you would like.

If there are going to be more involved changes, I also wonder if it might make sense to call (and wait for) the update script from the run script instead of starting it from C# and waiting a set period of time before restarting the runner? Obviously the current approach has been working well so far, but removing the `timeout` would be a small chance to make the process a bit more deterministic.

Also, in regards to removing code from the update script template, I think that will have to wait for a future update. Existing runners expect the update script to restart the runner once it completes. If we remove that code now, runners will never restart after updating because their existing run.cmd script will not have this updated logic yet. Unfortunately we need to get these changes in run.cmd deployed before removing that dead code in the update script.

Issue #1079, opened 2021-05-06
remove-svc.sh: sudo config.sh wipes the RUNNER_ALLOW_RUNASROOT env var


b-sturgeon:
https://github.com/actions/runner/blob/main/scripts/remove-svc.sh#L71-L76

Within the `remove-svc.sh` script, when `./config.sh remove` is run with `sudo` (the case for Linux users), user set env-vars are ignored.

This creates a catch-22 for linux users, `./config.sh remove` is always run as root, but it fails because `RUNNER_ALLOW_RUNASROOT` is therefore unset. The script operates as expected when `${prefix}`is removed from the`./config.sh remove` call. 

I appear unable to use the handy `remove-svc.sh` script as-is on Linux. I apologize if I am blatantly missing something here.

PR #1127, opened 2021-06-01
Run config remove as user instead of root

fhammerl:
Closes https://github.com/actions/runner/issues/1079

This PR ensures that when `config.sh remove` is called as part of uninstalling the runner as a service on a Linux machine, it is run as the current user instead of `root`. Calling it as `root` fails because the `RUNNER_ALLOW_RUNASROOT` environment variable (which we need to set to `1` in order to run `config.sh` as root) is usually set not set for root. The [issue](https://github.com/actions/runner/issues/1079) does a great job of explaining it.

### DIFF
--- a/docs/automate.md
+++ b/docs/automate.md
@@ -11,7 +11,7 @@ export RUNNER_CFG_PAT=yourPAT
 
 ## Create running as a service
 
-**Scenario**: Run on a machine or VM (not container) which automates:
+**Scenario**: Run on a machine or VM ([not container](#why-cant-i-use-a-container)) which automates:
 
  - Resolving latest released runner
  - Download and extract latest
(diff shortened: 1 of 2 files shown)


scripts/remove-svc.sh
```diff
@@ -73,4 +73,4 @@ if [ "${runner_plat}" == "linux" ]; then
 fi 
 ${prefix}./svc.sh stop
 ${prefix}./svc.sh uninstall
-${prefix}./config.sh remove --token $REMOVE_TOKEN
+,/config.sh remove --token $REMOVE_TOKEN
```
TingluoHuang, line 76: you might want to check how we configure the runner in the [config script](https://github.com/actions/runner/blob/main/scripts/create-latest-svc.sh), and make sure we lines up with that.

docs/automate.md
```diff
@@ -26,9 +26,13 @@ Run as a one-liner. NOTE: replace with yourorg/yourrepo (repo level) or just you
 curl -s https://raw.githubusercontent.com/actions/runner/automate/scripts/create-latest-svc.sh | bash -s yourorg/yourrepo
 ```
 
+### Why can't I use a container?
```
TingluoHuang, line 29: why do we need this?

fhammerl: I tested out the `create-svc.sh` and `remove-svc.sh` on a VM using these docs.

I thought we could further discourage trying it on a container by offering an explanation why the runner can't be installed as a service on docker.

scripts/remove-svc.sh
fhammerl: Good point, in the create script we run `sudo -E -u ${svc_user} ./config.sh` to preserve the environment variables so we don't trip on `RUNNER_ALLOW_RUNASROOT`. I agree it should be consistent.

Is there a reason to run `./config.sh` with `sudo` at all? If there isn't, shouldn't both just be `./config.sh`?

TingluoHuang (approved):

Issue #1129, opened 2021-06-03
Find a way to remove BuildConstants.cs changes from git diff


fhammerl:
Running a build using `dir.proj` generates a file called `BuildConstants.cs` with string constants like version, commit hash and os platform.
We cannot just remove and ignore this file, because the project doesn't compile if it doesn't exist. But every time you make a commit and build, this file changes. Currently, we just ignore it when it shows up in the diff.

(non-trivial) Solution 1: Convert it into a config file so the compiler is not reliant on it

(quick win) Solution 2: Generate (at least a dummy version of) `BuildConstants.cs` in `Runner.Sdk.csproj` using `
  <Target [...] BeforeTargets="Build">`. Now the project can be built without the compiler missing the file.

TingluoHuang:
Fixed.

PR #1130, opened 2021-06-03
Remove buildconstants

fhammerl:
This PR implements solution 2 in https://github.com/actions/runner/issues/1129

Running a build using `dir.proj` generates a file called `BuildConstants.cs` with string constants like version, commit hash and os platform.
We cannot just remove and ignore this file, because the project doesn't compile if it doesn't exist. But every time you make a commit and build, this file changes. Currently, we just ignore it when it shows up in the diff.

This PR generates a dummy version of `BuildConstants.cs` in `Runner.Sdk.csproj` using `
  <Target [...] BeforeTargets="Build">`. Now the project can be built without the compiler missing the `BuildConstants.cs`, so we can safely delete and ignore it.

Closes https://github.com/actions/runner/issues/1129


src/Runner.Sdk/BuildConstants.cs
```diff
@@ -1,19 +0,0 @@
-namespace GitHub.Runner.Sdk
```
TingluoHuang, line 1: if we remove this, then after you clone the repo for the first time, you will get some error in the IDE about `BuildConstants` doesn't exist. 😢

fhammerl:
https://github.com/actions/runner/pull/1132 fixes the issue

Issue #646, opened 2020-08-05
Next Steps for Fully Functioning Composite Actions


ethanchewy:
NOTE:

For those who are not aware, I'm a college student who interned at GitHub for Summer 2020. Thank you all for all the thoughtful comments and feedback on this thread but since I'm not working at GitHub currently, I will not be responding to any new comments below in this thread.

================================================

Hi there 👋

We just released a new feature called composite run steps actions(https://docs.github.com/en/actions/creating-actions/about-actions#composite-run-steps-actions).

This issue thread will serve as a point of reference for people who are curious about the next steps after this feature and what we plan to work on next. There are probably quite a few questions about when we are going to support in the future like "when are you going to support 'uses' steps in a composite action"? In this Issue thread, I'll talk about the technical problems that we are planning to solve to accomplish this. Throughout the next week, I'll update it with any additional updates.

**tl;dr in this issue, I'll go over what we currently support for composite run steps and how we plan on developing fully functioning composite actions.**

### What is a composite run steps action?

Many of you have been asking us to build a feature that enables them to nest actions within actions (ex: https://github.com/actions/runner/issues/438).

An important first step for supporting nesting within action is to first start supporting the execution of multiple run steps in an action.

For some developers, they may want to execute a variety of scripts and commands in different languages and shells in a single action and map those outputs to the workflow. We’ve also heard from our community that many developers want to reuse parts of their workflows into other workflows. This new action type, composite run steps, can help developers do just that! For example, developers can abstract parts of their workflows as composite run steps actions and reuse them for other workflows.

Thus, we created this feature called composite run steps which allows users to run multiple run steps in an action!

### What does composite run steps currently support?

For each run step in a composite action, we support: 
- name
- id
- run
- env
- shell
- working-directory

In addition, we support mapping input and outputs throughout the action.

See [docs](https://docs.github.com/en/actions/creating-actions/about-actions#composite-run-steps-actions) for more info.

### What does Composite Run Steps Not Support
We don't support setting conditionals, continue-on-error, timeout-minutes, "uses", and secrets on individual steps within a composite action right now. 

(Note: we do support these attributes being set in workflows for a step that uses a composite run steps action)

## Examples of how you can use composite run steps actions
* Outputs + Inputs:
  * Workflow: https://github.com/ethanchewy/testing-actions/actions/runs/184531939/workflow
  * Action: https://github.com/ethanchewy/test-composite/blob/outputs4/action.yml
[...]

PR #1144, opened 2021-06-10
Composite Actions Support ADR

thboop:
[Rendered](https://github.com/actions/runner/blob/bc9f54f40d7cf9aaeaa15434f303ae83a1c78b43/docs/adrs/1144-composite-actions.md)
We welcome community feedback, please feel free to comment on this ADR! 

This ADR outlines the next steps we are taking to change composite run steps into composite actions, and specifically mentions how certain scenarios will work.


docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+
+We want to support the `uses` steps from workflows in composite actions, including:
+  - Container actions
+  - javascript actions
+  - and even other Composite actions (up to a limit of course!)
+  - The pre and post steps these actions can generate
+
+## Guiding Principles
+
+- Composite Actions should function as a single step or action, no matter how many steps it is composed of or how many levels of recursion it has
```
ericsciple, line 17: If we are pushing composite actions as a general step re-use mechanism, it seems like users will want an option to control whether everything is collapsed into a single step log. For now of course, let's just continue down the path of collapsing into a single step log. Simply stating that I anticipate making composite actions more useful (support all step types) will to lead to follow-up work to control the collapse behavior.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+- A workflow author should not need to understand the inner workings of a composite action in order to use it
+- Composite actions should leverage inputs to get values they need, they will not have full access to the `context` objects. The secrets context will **not** be available to composite actions, users will need to pass these values in as an input.
+- Other Actions should **just work** inside a composite action, without any code changes
+
+## Decisions
+
+### Composite Recursion Limit
+
+We will start with supporting a recursion limit of `3`
```
ericsciple, line 26: Consider starting with a much higher number like 10 (or higher; something we feel is unreasonably high).

If i'm authoring a composite action, i shouldn't have to worry about how downstream actions are implemented. Maybe it's nodejs today, but then switches to composite. Then my workflow suddenly breaks.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+      - uses action step
+      - runs action step
+```
+- We are free to bump this limit in the future, the code will be written to just require updating a variable. If the graph evaluates beyond the recursion limit, the job will fail in the pre-job phase (The `Set up job` step).
+- A composite actions interface is its inputs and outputs, nothing else is carried over when invoking recursively.
+
+### Pre/Post Steps in nested Actions
+
+- We do not plan on adding the ability to configure a customizable pre or post step for composite actions at this time. However, we will execute the pre and post steps of any actions referenced in a composite action.
+- Composite actions will generate a single pre-step and post-step for the entire composite action, even if there are multiple pre-steps and post-steps in the referenced actions.
```
ericsciple, line 41: we need to be careful to make sure intra-state doesnt get clobbered (one logical step in the UI, but multiple steps wrt intra-state)

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+    - poststep-composite1-the-second-action-referenced
+    - poststep-composite1-first-action-referenced
+  - poststep-action1
+  ```
+
+#### Set-state
+
+- While the composite action has an individual combined pre/post action, the `set-state` command will not be shared.
+- If the `set-state` command is used during a composite step, only the action that originally called `set-state` will have access to the env variable during the post run step.
+  - This prevents multible actions that set the same state from interfering with the execution of another action's post step.
```
ericsciple, line 71: sp: multible (you need to flip the b upsidedown)

docs/adrs/0000-composite-actions.md
ericsciple: ok, reading further below it looks like you address this already in the "set-state" section

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+- These values passed as input into the composite action will **not** be carried over as input into the individual steps the composite action runs.
+
+### Defaults - Not being considered at this time
+
+- In actions, we have the idea of [defaults](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun) , which allow you to specify a shell and working directory in one location, rather then on each step.
+- However, `shell` is currently required in composite run steps
+  - In regular run steps, it is optional, and defaults to a different value based on the OS.
+- We could make it optional in composite run steps, and follow the trend of workflow run steps and add the `defaults` functionality
+ - However, encouraging action authors to explicitly state the `shell` forces them to do the right thing regarding considering cross platform support.
+ - We can change this in the future as needed, but for now its largely a convenience feature that adds no new functionality, let's wait on more user feedback before we consider this further.
```
ericsciple, line 92: We should never support this. The workflow author and the action author are two different people. It will lead to a bad experience for the consumer.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+
+- The resolve actions endpoint will now validate policy to ensure that the given workflow run has access to download that action.
+  - It will return a value indicated security was checked. Older server versions would lack this boolean. For composite actions that try to resolve an action, if the server does not respond with this value, we will fail the job as we could violate access policy if we allowed the composite action to continue.
+    - Oder GHES/GHAE customers with newer runners will be locked out of composite uses steps until they upgrade their instance.
+
+### If, continue-on-error, timeout-minutes - Not being considered at this time
+
+- `if`, `continue-on-error`, `timeout-minutes` could be supported in composite run/uses steps. These values were not originally supported in our composite run steps implementation.
+  - Browsing the community forums and runner repo, there hasn't been a lot of noise asking for these features, so we will hold off on them.
+- These values passed as input into the composite action will **not** be carried over as input into the individual steps the composite action runs.
```
ericsciple, line 83: These might become more interesting after we support all action types. +1 to keep current work scoped to supporting all action types. Then composite will be more useful.

These will be required in the future as we push on composite actions to be a general step-reuse mechanism.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+
+- In actions, we have the idea of [defaults](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun) , which allow you to specify a shell and working directory in one location, rather then on each step.
+- However, `shell` is currently required in composite run steps
+  - In regular run steps, it is optional, and defaults to a different value based on the OS.
+- We could make it optional in composite run steps, and follow the trend of workflow run steps and add the `defaults` functionality
+ - However, encouraging action authors to explicitly state the `shell` forces them to do the right thing regarding considering cross platform support.
+ - We can change this in the future as needed, but for now its largely a convenience feature that adds no new functionality, let's wait on more user feedback before we consider this further.
+
+## Consequences
+
```
ericsciple, line 95: let's add a feature flag so we can make sure the experience is right before rolling out publicly

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,97 @@
...
+### Defaults - Not being considered at this time
+
+- In actions, we have the idea of [defaults](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun) , which allow you to specify a shell and working directory in one location, rather then on each step.
+- However, `shell` is currently required in composite run steps
+  - In regular run steps, it is optional, and defaults to a different value based on the OS.
+- We could make it optional in composite run steps, and follow the trend of workflow run steps and add the `defaults` functionality
+ - However, encouraging action authors to explicitly state the `shell` forces them to do the right thing regarding considering cross platform support.
+ - We can change this in the future as needed, but for now its largely a convenience feature that adds no new functionality, let's wait on more user feedback before we consider this further.
+
```
ericsciple, line 93: We should add a section to discuss the behavior for local actions. Assuming it's supported with the same limitations today (or we could be more strict). Today local actions don't support pre (ignored)

bewuethr:
As per [your comment](https://github.com/actions/runner/issues/646#issuecomment-860776457), adding a use case: I'm working on an [action](https://github.com/marketplace/actions/slack-analyzer) that regularly pulls data from a Slack workspace, processes it and posts some graphs and statistics.

Optionally, it can also send a Telegram message, for which I use [another existing action](https://github.com/marketplace/actions/telegram-message-notify). I have to describe how to use this other action and have my own action produce output that can be used with it so users can set up a workflow – if I were able to just integrate the other action, and it looks I will be, then I could hide all the complexity and achieve the same result with just my action having to be set up.

afirth:
adding a use case:
lots of microservices deployed to k8s, one per namespace. _finally_ they all deploy with the same set of commands, basically AWS login, asdf install, skaffold deploy. This workflow is then duplicated into every single repo.

same story for linting rules + actions (identical in every go repo, for example).

currently using a fork of https://github.com/varunsridharan/action-github-workflow-sync to distribute actions to each repo. effective secret handling is a must, as well as caching and log visibility. TBH the current setup is not bad, just clunky and has no versions/releases

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,94 @@
...
+    - Oder GHES/GHAE customers with newer runners will be locked out of composite uses steps until they upgrade their instance.
+
+### Local actions
+- TODO describe this more
+- We should support pre and post actions for local actions
+
+
+### If, continue-on-error, timeout-minutes - Not being considered at this time
+
+- `if`, `continue-on-error`, `timeout-minutes` could be supported in composite run/uses steps. These values were not originally supported in our composite run steps implementation.
```
thboop, line 79: Might be good to prioritize the `if` keyword sooner rather than later. Feedback from @david-gang

> I see that `if` is not planned to be supported. This could be helpful for us. for example if i have two workflows which are nearly identical but have a big chunk of identical steps. part of the steps contain if clauses. Now i can't unite them in one action and call this action from both workflows. If the if would be supported i could do it. I know that i can use a javascript action but why would i need javascript if it is just a number of bash steps.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Composite Recursion Limit
+
+- We will start with supporting a recursion limit of `10` composite actions deep
+- We are free to bump this limit in the future, the code will be written to just require updating a variable. If the graph evaluates beyond the recursion limit, the job will fail in the pre-job phase (The `Set up job` step).
+- A composite actions interface is its inputs and outputs, nothing else is carried over when invoking recursively.
+
+### Pre/Post Steps in nested Actions
+
+- We do not plan on adding the ability to configure a customizable pre or post step for composite actions at this time. However, we will execute the pre and post steps of any actions referenced in a composite action.
```
thboop, line 33: Feedback from @david-gang regarding this. Could be very helpful for cleanup scenarios. The workaround is writing a JS/container action to reference from your composite action, but that isn't ideal.

> The google cli needs a service account json. I have this json stored in a secret. so i want to write a composite action which writes a secret to a file and a post step where the file is deleted. This is not possible without post build step.

rnett:
I have a use case that really wants an explicit post step: caching.  I have a fairly complicated gradle cache setup, with post-steps requiring `if`s (i.e. stopping the gradle daemon, deleting lock files.  There are OS-specific parts, thus the `if`s).  See [here](https://github.com/rnett/krosstalk/blob/main/.github/workflows/ci.yml#L25).  Ideally I would like to move the Java setup and gradlew `chmod`s into the composite action as well, but they would need to be disable-able (more `if`s).

timharris777:
Very exciting! Our engineers have been waiting for this for a while now. We would be happy to test any early releases on our self-hosted runners and provide feedback if needed. Thanks again!

joeyparrish:
This seems like an area rife with potential for supply chain attacks.

Today, if my workflow depends on action `A/B@v1`, I can be certain that v1 will not change.  If I want to review the safety of that action, I can do it once per upgrade that I choose to undertake.

However, if action `A/B@v1` could depend on action `C/D@main`, and `C/D` is taken over by a malicious actor, my workflow could be compromised without any changes on my end.  This could lead to stolen secrets and further malicious changes to my repo.

Before actions gain the capability to include other actions, please spend some time considering the potential for attacks through indirect dependencies, and please outline steps GitHub would take to mitigate that potential.

For example, here's an idea that could help:  Require workflows to opt-in to indirect dependencies, or to allow-list specific indirect dependencies and versions.  This would help by keeping workflow maintainers aware of what they are consuming.

umarcor:
@joeyparrish isn't that a general concern with consuming any tool, since most software tools nowadays have dependencies? I can't see why composite actions increase the attack surface compared to consuming Actions. In fact, people who want to provide the features of composite actions can already achieve it if they use the Actions without the `uses` field (which is just syntactic sugar). Overall, I share your concern with security, but I don't see what specific steps you expect.

rnett:
What if instead of requiring opt-in for indirect dependencies you required an opt-in for non-static (i.e. non-tag) versions?  Think `"Composite action A depends on action B@main, which is a non static version.  You must specify a version of B to use using the versions block."`   It would prevent silent updates, which I agree is necessary, but without as much effort.

kuhnroyal:
> Today, if my workflow depends on action `A/B@v1`, I can be certain that v1 will not change.

@joeyparrish As far as I can tell that is not true in practice. A lot of actions re-tag their smaller updates based on semver so you don't even notice the action has changed. And if the `A` action is compromised then a malicious actor can also just change the `v1` tag and you don't notice it.

I agree with your points, but the problem actually starts with the first action you use.

umarcor:
To build on what @kuhnroyal said, having "moving tags" is the officially recommended approach by Github (Actions). A lof of actions re-tag their smaller updates because of a lack of proper semver support. This is also related to the recommendation of adding `node_modules` to the repo, for all JavaScript Actions. In an attempt to make it as simple as possible for unexperienced users, several bad practices are recommended as the first option. See:

- https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#commit-tag-and-push-your-action-to-github
- https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#recommendations
- https://github.com/actions/toolkit/issues/214

Many of us would be really happy with semver being understood by GitHub Action's parser/interpreter. Having the same syntax as npm or yarn seems sensible, since the ecosystem is mainly based on JS.

joeyparrish:
You are all correct.  Supply chain attacks can occur through node module deps and tag rewrites just as well as the scenario I suggested.

Perhaps I had too narrow a view and chose a poor venue.  Is there a better place to discuss these potential security issues?

umarcor:
@joeyparrish what about asking the feature to be integrated into dependabot and the regular security updates that GitHub suggests? Currently, it scans `setup.py` , `package.jon`, `go.mod`, etc. depending on the languages/package managers used in your repo. The `uses` fields of the workflows might be considered "another dependency declaration format". Then, the inheritance, notifying about vulnerabilities deep downstream, etc. should reuse the existing plumbing. When that's in place, users might be allowed to "automatically disable the workflows which have some severity level above a given type".

thboop:
> However, if action A/B@v1 could depend on action C/D@main, and C/D is taken over by a malicious actor, my workflow could be compromised without any changes on my end. This could lead to stolen secrets and further malicious changes to my repo.

For third party actions, we recommend [pinning to a sha](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions). We will update this guidance to recommend you do that for actions inside a composite action as well. This mitigates the supply-chain attack.

> allow-list specific indirect dependencies and versions.

Good news on this front! Composite actions will also respect the [actions policies](https://docs.github.com/en/github/setting-up-and-managing-your-enterprise/setting-policies-for-organizations-in-your-enterprise-account/enforcing-github-actions-policies-in-your-enterprise-account#allowing-specific-actions-to-run), so you can state which actions you want to be runnable in your org/repo/enterprise. This allows you to specify specific tags or sha's. 

> isn't that a general concern with consuming any tool, since most software tools nowadays have dependencies? I can't see why composite actions increase the attack surface compared to consuming Actions

Javascript actions have the same issue when pinning to a major version tag, if a node module updates and is compromised, you would end up having your action compromised. As posted above [pinning to a sha](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions) mitigates this.

> Perhaps I had too narrow a view and chose a poor venue. Is there a better place to discuss these potential security issues?

We have a [bug bounty](https://bounty.github.com/) program, please submit security vulnerabilities there! 

> feature to be integrated into dependabot and the regular security updates that GitHub suggests?

Let me dig and see if we can get dependabot to do automatic updates for the `action.yaml` file as well. I'll loop back on this.

> having "moving tags" is the officially recommended approach by Github (Actions)

That is correct, but we hope to improve this solution in the future. However, it's out of scope for this feature.

docs/adrs/0000-composite-actions.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Local actions
+- TODO describe this more
+- We should support pre and post actions for local actions
+
+
+### If, continue-on-error, timeout-minutes - Not being considered at this time
+
+- `if`, `continue-on-error`, `timeout-minutes` could be supported in composite run/uses steps. These values were not originally supported in our composite run steps implementation.
+  - Browsing the community forums and runner repo, there hasn't been a lot of noise asking for these features, so we will hold off on them.
```
antdking, line 80: To add a usecase, I think some of the early composite actions we'll see will be wrappers around cache + dependency management.

One of the things we do is to conditionally run `npm install` depending on if we get a cache hit against the os + node + npm versions, which saves us around 1:30m ontop of the standard npm caching.

This is a lot of boilerplate (especially if you adhere to one task per job/workflow), and easy to get wrong, so conditional support would be massive for this usecase, to avoid the complexities of having to write a full action for what should essentially be a composition of actions.

jescalan:
I just want to say that this is awesome and I am very excited to use this feature! 💖

docs/adrs/0000-composite-actions.md
zackw: Two more important use cases for `if` are executing steps only if appropriate in context (i.e. depending on the action's inputs) and capturing details of a failure.  I wrote this just now assuming it would work, and was sad when it didn't:

```
   - name: Build
      shell: bash
      if: inputs.build_targets != 'skip'
      run: make $BUILD_TARGETS
      env:
        CC: ${{ inputs.compiler }}
        BUILD_TARGETS: ${{ inputs.build_targets }}

    - name: Test
      shell: bash
      if: inputs.test_targets != 'skip'
      run: make $TEST_TARGETS
      env:
        CC: ${{ inputs.compiler }}
        TEST_TARGETS: ${{ inputs.test_targets }}

    - name: Failure details
      shell: bash
      if: ${{ failure() }}
      run: |
        dump_log () {
          if [ -s "$1" ]; then
            echo "::group::$1"
            echo '::stop-commands::resume-commands-50YEO1zJ8HSXH4Zy'
            cat "$1"
            echo '::resume-commands-50YEO1zJ8HSXH4Zy::'
            echo '::endgroup::'
          fi
        }
        dump_log config.log
        for ts in $(find . -name 'test-suite*.log' -printf '%P\n'); do
          dump_log "$ts"
        done
```

It would work (with slight adjustments) in a workflow file, but then I'd need, let's see, at least six copies of it.

ericsciple (dismissed):

thboop:
`Uses` steps are live on dotcom, for more information please see: https://github.com/actions/runner/blob/main/docs/adrs/1144-composite-actions.md

figadore:
> `Uses` steps are live on dotcom

What does this mean? I thought it meant it was ready for use, but I'm still seeing 

> Required property is missing: shell

when I try to do things like this
```
runs:
  using: "composite"
  steps:
    - name: Checkout
      uses: actions/checkout@master
```

docs/adrs/0000-composite-actions.md
prince-chrismc: This would be amazing feature https://github.com/prince-chrismc/jwt-cpp/runs/3448042058

PR #1150, opened 2021-06-11
No retries for short SHAs

### LINKED ISSUE #1037: Useless retry for short SHA
**Describe the bug**
If one asks for an action by a short SHA, the runner says it doesn't like it and then decides to try again later. And then it tries again after that time.

**To Reproduce**
1. https://github.com/jsoref/nektos-issues-composite-action/runs/2265052125?check_suite_focus=true
2. See error

.github/workflows/workflow.yml at https://github.com/jsoref/nektos-issues-composite-action/commit/763c0060eaa8623a58d636da9aabc15639187b06:
```yaml
on:
  workflow_dispatch:
  push:

jobs:
  local-action:
    runs-on: ubuntu-latest
    steps:
      - uses: jsoref/gha-debug@fae0865
```
**Expected behavior**
This is apparently a fatal error. I tried adding a branch with the same value as the short SHA, and it still failed. As such, there's really absolutely no reason to retry twice.

Version of your runner?
```
Current runner version: '2.277.1'
Operating System
  Ubuntu
  20.04.2
  LTS
Virtual Environment
  Environment: ubuntu-20.04
  Version: 20210330.1
  Included Software: https://github.com/actions/virtual-environments/blob/ubuntu20/20210330.1/images/linux/Ubuntu2004-README.md
  Image Release: https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20210330.1
```

## What's not working?
Retrying for a fatal error is kinda wasteful

## Job Log Output
```
Getting action download info
Failed to resolve action download info. Error: Unable to resolve action `jsoref/gha-debug@fae0865`, the provided ref `fae0865` is the shortened version of a commit SHA, which is not supported. Please use the full commit SHA `fae08650a96e2b93eac0db352559575da81396be` instead.
Retrying in 11.361 seconds
Failed to resolve action download info. Error: Unable to resolve action `jsoref/gha-debug@fae0865`, the provided ref `fae0865` is the shortened version of a commit SHA, which is not supported. Please use the full commit SHA `fae08650a96e2b93eac0db352559575da81396be` instead.
Retrying in 29.696 seconds
Error: Unable to resolve action `jsoref/gha-debug@fae0865`, the provided ref `fae0865` is the shortened version of a commit SHA, which is not supported. Please use the full commit SHA `fae08650a96e2b93eac0db352559575da81396be` instead.
```
If applicable, add relevant diagnostic log information.  Logs are located in the runner's `_diag` folder. The runner logs are prefixed with `Runner_` and the worker logs are prefixed with `Worker_`. Each job run correlates to a worker log.  All sensitive information should already be masked out, but please double-check before pasting here.

fhammerl:
Closes #1037

### DIFF
--- a/src/Sdk/DTWebApi/WebApi/Exceptions.cs
+++ b/src/Sdk/DTWebApi/WebApi/Exceptions.cs
@@ -2476,24 +2476,32 @@ namespace GitHub.DistributedTask.WebApi
             : base(info, context)
         {
         }
-    }
 
-    [Serializable]
-    public sealed class FailedToResolveActionDownloadInfoException : DistributedTaskException
-    {
-        public FailedToResolveActionDownloadInfoException(String message)
-            : base(message)
+        public bool IsShortSha
         {
-        }
-
-        public FailedToResolveActionDownloadInfoException(String message, Exception innerException)
-            : base(message, innerException)
-        {
-        }
-
-        private FailedToResolveActionDownloadInfoException(SerializationInfo info, StreamingContext context)
-            : base(info, context)
-        {
+            get
+            {
+                return Message.Contains("Please use the full commit SHA");
+            }
         }
     }
-}
+
+        [Serializable]
+        public sealed class FailedToResolveActionDownloadInfoException : DistributedTaskException
+        {
+            public FailedToResolveActionDownloadInfoException(String message)
+                : base(message)
+            {
+            }
+
+            public FailedToResolveActionDownloadInfoException(String message, Exception innerException)
+                : base(message, innerException)
+            {
+            }
+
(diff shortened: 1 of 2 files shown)


src/Sdk/DTWebApi/WebApi/Exceptions.cs
```diff
@@ -2476,24 +2476,32 @@ protected UnresolvableActionDownloadInfoException(SerializationInfo info, Stream
             : base(info, context)
         {
         }
-    }
 
-    [Serializable]
-    public sealed class FailedToResolveActionDownloadInfoException : DistributedTaskException
```
TingluoHuang, line 2482: we can't change this file, these files are minor from service, we need to change the server to throw a different exception.

PR #1170, opened 2021-06-23
Composite Actions

thboop:
What this does currently cover:
- Javascript actions in composite
- Passing through/evaluating default inputs for actions
- Downloading actions/building containers that are nested within composite actions
- Composite nest limit
- Hash File fixes

Later Pr's will follow up on:
- Mapping inputs throughout nesting
- Pre/post actions
- Local actions
  - Do we need to download those graphs JIT to support these?
- ENV bugfixes for composite


src/Test/L0/Worker/ActionManagerL0.cs
```diff
@@ -59,7 +59,7 @@ public async void PrepareActions_PullImageFromDockerHub_Legacy()
                 };
 
                 //Act
-                var steps = (await _actionManager.PrepareActionsAsync(_ec.Object, actions)).ContainerSetupSteps;
+                var steps = (await _actionManager.PrepareActionsV2Async(_ec.Object, actions)).ContainerSetupSteps;
```
thboop, line 62: Should probably dupe this tests instead of updating to the newest function

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -70,6 +74,13 @@ public async Task RunAsync(ActionRunStage stage)
                     // Set GITHUB_ACTION_PATH
                     step.ExecutionContext.SetGitHubContext("action_path", ActionDirectory);
 
+                    // Add Expression Functions
+                    step.ExecutionContext.ExpressionFunctions.Add(new FunctionInfo<AlwaysFunction>(PipelineTemplateConstants.Always, 0, 0));
```
thboop, line 78: I need to test this but these could end up significantly more involved then expected. What if the composite action only runs on success and a step inside it only runs on failure? Might be good to only support conditionals based on context (and `Success` the default)

src/Runner.Worker/ActionDepthOutOfRange.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace GitHub.Runner.Worker
+{
+    public class ActionDepthOutOfRange : Exception
```
ericsciple, line 5: do we catch this anywhere?

src/Runner.Worker/ActionManager.cs
```diff
@@ -986,16 +1149,16 @@ private void ConfigureAuthorizationFromContext(IExecutionContext executionContex
 
         private string GetWatermarkFilePath(string directory) => directory + ".completed";
 
-        private ActionContainer PrepareRepositoryActionAsync(IExecutionContext executionContext, Pipelines.ActionStep repositoryAction)
+        private ActionSetupInfo PrepareRepositoryActionAsync(IExecutionContext executionContext, Pipelines.ActionStep repositoryAction, Int32 depth = 0)
```
ericsciple, line 1152: nit: `int` (check the code to keep me honest)

src/Runner.Worker/ActionManager.cs
```diff
@@ -51,6 +52,168 @@ public sealed class ActionManager : RunnerService, IActionManager
         private readonly Dictionary<Guid, ContainerInfo> _cachedActionContainers = new Dictionary<Guid, ContainerInfo>();
 
         public Dictionary<Guid, ContainerInfo> CachedActionContainers => _cachedActionContainers;
+        public async Task<PrepareResult> PrepareActionsV2Async(IExecutionContext executionContext, IEnumerable<Pipelines.JobStep> steps)
```
ericsciple, line 55: notes for reader: this is the download code

ericsciple: let's check whether legacy can be deleted (depends when it shipped on GHES)

Also might be interesting to get a good answer how far back we support (latest runner, old GHES)

src/Runner.Worker/ActionManager.cs
```diff
@@ -986,16 +1149,16 @@ private void ConfigureAuthorizationFromContext(IExecutionContext executionContex
...
             var repositoryReference = repositoryAction.Reference as Pipelines.RepositoryPathReference;
             if (string.Equals(repositoryReference.RepositoryType, Pipelines.PipelineConstants.SelfAlias, StringComparison.OrdinalIgnoreCase))
             {
                 Trace.Info($"Repository action is in 'self' repository.");
                 return null;
             }
-
-            var setupInfo = new ActionContainer();
+            var setupInfo = new ActionSetupInfo();
+            var actionContainer = new ActionContainer();
```
ericsciple, line 1161: should this move into each if block

src/Runner.Worker/ActionManager.cs
```diff
@@ -1038,8 +1201,9 @@ private ActionContainer PrepareRepositoryActionAsync(IExecutionContext execution
                         var dockerFileFullPath = Path.Combine(actionEntryDirectory, containerAction.Image);
                         executionContext.Debug($"Dockerfile for action: '{dockerFileFullPath}'.");
 
-                        setupInfo.Dockerfile = dockerFileFullPath;
-                        setupInfo.WorkingDirectory = destDirectory;
+                        actionContainer.Dockerfile = dockerFileFullPath;
+                        actionContainer.WorkingDirectory = destDirectory;
+                        setupInfo.Container = actionContainer;
```
ericsciple, line 1206: might simpler to do:

```csharp
setupInfo.Container = new ActionContainer
{
...
};
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -1303,4 +1473,18 @@ public class ActionContainer
...
+    {
+        public ActionContainer Container { get; set; }
+        public List<Pipelines.ActionStep> Steps {get; set;}
+    }
+
+    public class PrepareActionsState
+    {
+        public Dictionary<string, List<Guid>> ImagesToPull;
+        public Dictionary<string, List<Guid>> ImagesToBuild;
+        public Dictionary<string, ActionContainer> ImagesToBuildInfo;
```
ericsciple, line 1487: Consider combining `ImagesToBuild` and `ImagesToBuildInfo`

src/Runner.Worker/ActionManager.cs
```diff
@@ -1068,8 +1233,11 @@ private ActionContainer PrepareRepositoryActionAsync(IExecutionContext execution
                 }
                 else if (actionDefinitionData.Execution.ExecutionType == ActionExecutionType.Composite)
                 {
-                    Trace.Info($"Action composite: {(actionDefinitionData.Execution as CompositeActionExecutionData).Steps}, no more preparation.");
-                    return null;
+                    // TODO: we need to generate unique Id's for composite steps
```
ericsciple, line 1236: new guid should be fine

Steps inside composite share the same timeline record id as their parent (ExecutionContext.CreateEmbeddedChild). Regular step ID (guid) matches the execution context ID and timeline record ID, iirc. I might be wrong, but that is my best guess because that's why those concepts exist.

So the new guid here is probably just needed to manage building/starting the image.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ericsciple: good to scrutinize this more... i think there is some sensitivity when they should be added

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -193,6 +204,7 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps)
 
                 await RunStepAsync(step);
 
+                // TODO: What if a composite action is not skipped, but all of its steps are skipped, do we mark it as skipped?
```
ericsciple, line 207: We should think about this a bit

We are leaning on composite actions to be the way to share steps. So consider a composite action that contains a a mix of nested steps: some run always; some do not.

Let's identify what we want long-term. Worried about changing behavior and breaking customers. Might want to hold off on nested `if` support altogether for now.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -208,33 +220,73 @@ private async Task RunStepAsync(IStep step)
             step.ExecutionContext.Debug($"Starting: {step.DisplayName}");
 
             await Common.Util.EncodingUtil.SetEncoding(HostContext, Trace, step.ExecutionContext.CancellationToken);
+
```
ericsciple, line 223: 👎

src/Runner.Worker/JobExtension.cs
```diff
@@ -239,7 +239,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     // Download actions not already in the cache
                     Trace.Info("Downloading actions");
                     var actionManager = HostContext.GetService<IActionManager>();
-                    var prepareResult = await actionManager.PrepareActionsAsync(context, message.Steps);
+                    PrepareResult prepareResult;
```
ericsciple, line 242: can we make an actionable error if we're connected to an old server and can't download

src/Runner.Worker/StepsRunner.cs
```diff
@@ -354,8 +355,8 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
 
             executionContext.Complete(result, resultCode: resultCode);
         }
-
-        private sealed class ConditionTraceWriter : ObjectTemplating::ITraceWriter
+    }
+    public sealed class ConditionTraceWriter : ObjectTemplating::ITraceWriter
```
ericsciple, line 359: ^ found where that extra newline was meant to be

src/Runner.Worker/action_yaml.json
```diff
@@ -130,6 +137,21 @@
                 }
             }
         },
+        "uses-step": {
```
ericsciple, line 140: compare against the schema file on the server

src/Runner.Worker/JobExtension.cs
```diff
@@ -239,7 +239,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                    // TODO should we also check DistributedTask.NewActionMetadata, 
+                    // we removed the backcompat code, so if NewActionMetadata was disabled, and Composite enabled, we would run into issues
+                    // TODO Make a real feature flag
+                    if (!string.IsNullOrEmpty(context.Global.Variables.Get("ENABLE_COMPOSITE")))
```
ericsciple, line 246: Rather than permanently leave a variable (`ENABLE_COMPOSITE`, `NEW_ACTION_METADATA`), we may want to leverage `PlanUtil.GetFeatures`. We can probably correlate plan version to GHES version.

Think about it. Sort of on the fence either way.

src/Runner.Worker/ActionDepthOutOfRange.cs
thboop: We do for test purposes

src/Runner.Worker/ActionManager.cs
ericsciple: rather than fork the code, have you considered checking using the new feature flag to error/allow nested composite references? forking the code seems less desirable if the feature flag code will not be removed any time soon

src/Runner.Worker/ActionManager.cs
thboop: Doesn't actually make a lot of sense to do here because we use it right away

src/Runner.Worker/ActionManager.cs
thboop: Going to avoid refactoring this code for now since there are other refactors in this code and I want to minimize risk

src/Runner.Worker/ActionManager.cs
```diff
@@ -179,85 +161,57 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+                    else if(setupInfo != null && setupInfo.Steps != null && setupInfo.Steps.Count > 0)
+                    {
+                        state = await PrepareActionsRecursiveAsync(executionContext, state, setupInfo.Steps, depth + 1);
+                    }
                     var repoAction = action.Reference as Pipelines.RepositoryPathReference;
                     if (repoAction.RepositoryType != Pipelines.PipelineConstants.SelfAlias)
                     {
                         var definition = LoadAction(executionContext, action);
-                        if (definition.Data.Execution.HasPre)
+                        // TODO: How should we handle pre with no Id's.
```
ericsciple, line 198: did you intend to leave this TODO comment?

src/Runner.Worker/ActionManager.cs
```diff
@@ -1068,8 +928,21 @@ private ActionContainer PrepareRepositoryActionAsync(IExecutionContext execution
...
+                    Trace.Info($"Loading Composite steps");
+                    var compositeAction = actionDefinitionData.Execution as CompositeActionExecutionData;
+                    setupInfo.Steps = compositeAction.Steps;
+
+                    foreach (var step in compositeAction.Steps)
+                    {
+                        step.Id = Guid.NewGuid();
+                        if (string.IsNullOrEmpty(executionContext.Global.Variables.Get("DistributedTask.EnableCompositeActions")) && step.Reference.Type != Pipelines.ActionSourceType.Script)
+                        {
+                            throw new Exception("Composite Actions with the `uses:` keyword are not supported on this server. Please update your GHES or GHAE instance.");
```
ericsciple, line 941: Should we just say `uses` not supported, as we would if they had an old runner that didn't support the keyword? This error message doesn't make sense for dotcom while we're in the process of rolling out the feature flag.

src/Runner.Worker/ActionManager.cs
ericsciple: did this change? it looks like it's not used right away

ericsciple: specifically the local variable `actionContainer`

asking, because i found myself scrolling back and forth to try and understand this local variable. Looks like it's only used further below within each if block.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -245,6 +245,12 @@ public void ForceTaskComplete()
 
         public void RegisterPostJobStep(IStep step)
         {
+            // TODO: Remove when we support composite post job steps
```
ericsciple, line 248: should we throw to make sure we dont miss this?

src/Runner.Worker/JobExtension.cs
```diff
@@ -239,7 +239,8 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     // Download actions not already in the cache
                     Trace.Info("Downloading actions");
                     var actionManager = HostContext.GetService<IActionManager>();
-                    var prepareResult = await actionManager.PrepareActionsAsync(context, message.Steps);
+                    PrepareResult prepareResult;
+                    prepareResult = await actionManager.PrepareActionsAsync(context, message.Steps);
```
ericsciple, line 243: either `var prepareResult =` or `PrepareResult prepareResult =`

src/Runner.Worker/Handlers/CompositeActionHandler.cs
thboop: You are correct, I think this needs to be moved down to right before each individual composite step is invoked, let me do some testing

src/Runner.Worker/action_yaml.json
```diff
@@ -195,6 +215,37 @@
...
+                "runner",
+                "env",
+                "hashFiles(1,255)"
+            ],
+            "mapping": {
+                "loose-key-type": "non-empty-string",
+                "loose-value-type": "string"
+            }
+        },
+        "uses-input": {
```
ericsciple, line 236: what is `uses-input`? looks like this should be deleted

thboop: will remove

src/Runner.Worker/ActionDepthOutOfRange.cs
```diff
@@ -0,0 +1,22 @@
...
+        }
+
+        public ActionDepthOutOfRange(string message)
+            : base(message)
+        {
+        }
+
+        private static string FormatMessage(int depth)
+        {
+            return $"Composite action depth exceeded max depth {depth-1}, please simplify your actions.'";
```
ericsciple, line 19: > `depth-1`

Should we just pass `maxDepth` instead to make this more intuitive and less fragile

ericsciple: nit: consider dropping "please simplify your actions"

It may be no fault of the author. Might be the fault of a downstream action.

src/Runner.Worker/ActionDepthOutOfRange.cs
```diff
@@ -0,0 +1,22 @@
+using System;
```
ericsciple, line 1: nit: in general we don't have specific exceptions and this is going to increase the files to grok in the Action.Worker directory. Consider just asserting the string in the unit test

Looks like ActionNotFoundException is the only one, and it's the outlier not the predominant pattern

src/Runner.Worker/ActionDepthOutOfRange.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace GitHub.Runner.Worker
```
ericsciple, line 3: class name should end in "Exception"

src/Runner.Worker/ActionDepthOutOfRange.cs
ericsciple: It looks like ActionNotFound is explicitly caught so it's not retried

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -70,6 +74,9 @@ public async Task RunAsync(ActionRunStage stage)
                     // Set GITHUB_ACTION_PATH
                     step.ExecutionContext.SetGitHubContext("action_path", ActionDirectory);
 
+                    // Add Expression Functions
```
ericsciple, line 77: Where does each nested step display name get evaluated? Is it always JIT or do we do a first-pass at the beginning of the job?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -70,6 +74,9 @@ public async Task RunAsync(ActionRunStage stage)
...
+                    step.ExecutionContext.ExpressionFunctions.Add(new FunctionInfo<HashFilesFunction>(PipelineTemplateConstants.HashFiles, 1, byte.MaxValue));
```
ericsciple, line 78: Let's move this into `RunStepsAsync` because that function should resemble StepsRunner as much as possible

ericsciple: So in between these two lines:

```
                Trace.Info($"Processing embedded step: DisplayName='{step.DisplayName}'");

                // Initialize env context
                Trace.Info("Initialize Env context for embedded step");
```

src/Runner.Worker/ActionDepthOutOfRange.cs
ericsciple: AssertThrows returns the exception which can then be used to check the message

src/Runner.Worker/ActionManager.cs
thboop: Its used in this block https://github.com/actions/runner/pull/1170/files/704b39965b40c8b04be58576fbafa66e895b16e4#diff-5b282ef924f268139b4015eb491825d3e35bfa6cbe46962e3bef3bd8fa05dfccR1170 and the else side of that block. Which sets the variable at this level, then as you go into the if conditions its used more. We could refactor those linked blocks to be in every block that has `actionContainer`, but i'm not sure that makes a lot of sense for this use case, I think this is the right level to initialize this variable.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -245,6 +245,12 @@ public void ForceTaskComplete()
...
+            if (this.IsEmbedded)
+            {
+                throw new Exception("Composite actions do not currently support post steps");
+
```
ericsciple, line 252: extra line

src/Runner.Worker/ActionManager.cs
```diff
@@ -179,85 +161,56 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
-                            Trace.Info($"Action {action.Name} ({action.Id}) from repository '{setupInfo.ActionRepository}' needs to build image '{setupInfo.Dockerfile}'");
-                            imagesToBuild[setupInfo.ActionRepository].Add(action.Id);
-                            imagesToBuildInfo[setupInfo.ActionRepository] = setupInfo;
+                            Trace.Info($"Action {action.Name} ({action.Id}) from repository '{setupInfo.Container.ActionRepository}' needs to build image '{setupInfo.Container.Dockerfile}'");
+                            state.ImagesToBuild[setupInfo.Container.ActionRepository].Add(action.Id);
+                            state.ImagesToBuildInfo[setupInfo.Container.ActionRepository] = setupInfo.Container;
                         }
                     }
-
+                    else if(setupInfo != null && setupInfo.Steps != null && setupInfo.Steps.Count > 0)
```
ericsciple, line 190: `else if (`

src/Runner.Worker/ActionManager.cs
```diff
@@ -53,30 +53,63 @@ public sealed class ActionManager : RunnerService, IActionManager
...
+            if (containerSetupSteps.Count > 0)
+            {
+                executionContext.Output("Container action is only supported on Linux, skip pull and build docker images.");
+                containerSetupSteps.Clear();
+            }
+#endif
+            return new PrepareResult(containerSetupSteps, result.PreStepTracker);
+        }
 
+        private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionContext executionContext, PrepareActionsState state, IEnumerable<Pipelines.ActionStep> actions, Int32 depth = 0)
```
ericsciple, line 106: nit: `PrepareActionsAsync`?

src/Runner.Worker/ActionManager.cs
```diff
@@ -53,30 +53,63 @@ public sealed class ActionManager : RunnerService, IActionManager
...
+            };
+            var containerSetupSteps = new List<JobExtensionRunner>();
+            IOUtil.DeleteDirectory(HostContext.GetDirectory(WellKnownDirectory.Actions), executionContext.CancellationToken);
             IEnumerable<Pipelines.ActionStep> actions = steps.OfType<Pipelines.ActionStep>();
-
-            // TODO: Deprecate the PREVIEW_ACTION_TOKEN
-            // Log even if we aren't using it to ensure users know.
-            if (!string.IsNullOrEmpty(executionContext.Global.Variables.Get("PREVIEW_ACTION_TOKEN")))
+            executionContext.Output("Prepare all required actions");
+            var result = await PrepareActionsRecursiveAsync(executionContext, state, actions, 0);
```
ericsciple, line 70: super nit, but interesting imo: if you change the name to `PrepareActionsAsync` then you could name the depth param

```
var result = await PrepareActionsAsync(executionContext, state, actions, depth: 0);
```

src/Runner.Worker/ActionManager.cs
ericsciple: `else if (setupInfo?.Steps?.Count > 0)`

src/Runner.Worker/ActionManager.cs
```diff
@@ -179,85 +161,56 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+                    {
+                        state = await PrepareActionsRecursiveAsync(executionContext, state, setupInfo.Steps, depth + 1);
+                    }
                     var repoAction = action.Reference as Pipelines.RepositoryPathReference;
                     if (repoAction.RepositoryType != Pipelines.PipelineConstants.SelfAlias)
                     {
                         var definition = LoadAction(executionContext, action);
-                        if (definition.Data.Execution.HasPre)
+                        // TODO: Support pre's in composite actions
+                        if (definition.Data.Execution.HasPre && depth < 1)
```
ericsciple, line 199: should we throw if depth >= 1?

src/Runner.Worker/ActionManager.cs
ericsciple: oh :(

src/Runner.Worker/ActionManager.cs
```diff
@@ -1068,8 +928,21 @@ private ActionContainer PrepareRepositoryActionAsync(IExecutionContext execution
...
-                    Trace.Info($"Action composite: {(actionDefinitionData.Execution as CompositeActionExecutionData).Steps}, no more preparation.");
-                    return null;
+                    // TODO: we need to generate unique Id's for composite steps
+                    Trace.Info($"Loading Composite steps");
+                    var compositeAction = actionDefinitionData.Execution as CompositeActionExecutionData;
+                    setupInfo.Steps = compositeAction.Steps;
+
+                    foreach (var step in compositeAction.Steps)
+                    {
+                        step.Id = Guid.NewGuid();
```
ericsciple, line 938: +1 looks like this isn't used for ExecutionContext.CreateEmbeddedChild but i believe this is needed for tracking "intra-state" (pre/main/post)

ericsciple: wonder if we have a bug today where all nested "main" share the same intra-state dictionary

src/Runner.Worker/ActionManager.cs
```diff
@@ -53,30 +53,63 @@ public sealed class ActionManager : RunnerService, IActionManager
         public Dictionary<Guid, ContainerInfo> CachedActionContainers => _cachedActionContainers;
         public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext executionContext, IEnumerable<Pipelines.JobStep> steps)
         {
+            // Assert inputs
```
ericsciple, line 56: What happens for "./local"?

ericsciple (approved):

src/Runner.Worker/ActionManager.cs
ericsciple: local with containers is also interesting... wonder whether we support that today 🤔

PR #1172, opened 2021-06-24
Check runner group when there is only `default` runner group.

TingluoHuang:


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -165,7 +165,7 @@ public async Task ConfigureAsync(CommandSettings command)
             List<TaskAgentPool> agentPools = await _runnerServer.GetAgentPoolsAsync();
             TaskAgentPool defaultPool = agentPools?.Where(x => x.IsInternal).FirstOrDefault();
 
-            if (agentPools?.Where(x => !x.IsHosted).Count() > 1)
+            if (agentPools?.Where(x => !x.IsHosted).Count() > 0)
```
TingluoHuang, line 168: We should do runner group lookup as long as there is a runner group.
Otherwise, the runner goes to `default` even I pass in `--runnergroup foo`.

ericsciple: What if you specify `--runnergroup foo` and you don't have any runner groups?

ericsciple (approved):

src/Runner.Listener/Configuration/ConfigurationManager.cs
ericsciple: discussed offline

ericsciple: it will error... i forgot `command.GetRunnerGroupName` is a prompt

PR #1175, opened 2021-06-28
Add notice annotation level and support more annotation fields

luketomlinson:
This PR does the following:

* Creates a new `notice` command to map to the `notice` annotation level. 
* Adds support for `title`, `endColumn`, `endLine`

### DIFF
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -598,8 +598,6 @@ namespace GitHub.Runner.Worker
             var hasLine = line != null;
             var hasEndLine = endLine != null;
 
-            Console.WriteLine($"hasColumnValue: {hasColumnValue}, hasLine: {hasLine}");
-
             if (!hasLine && hasColumnValue) 
             {
                 throw new Exception($"Invalid {command.Command} command value. 'column' and 'end_column' can only be set if 'line' value is provided."); 
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -77,6 +77,7 @@ namespace GitHub.Runner.Worker
 
             if (!ActionCommandManager.EnhancedAnnotationsEnabled(context) && actionCommand.Command == "notice")
             {
+                context.Debug($"Enhanced Annotations not enabled on the server: 'notice' command will not be processed.");
                 return false;
             }
 
@@ -528,6 +529,13 @@ namespace GitHub.Runner.Worker
             command.Properties.TryGetValue(IssueCommandProperties.Line, out string line);
             command.Properties.TryGetValue(IssueCommandProperties.Column, out string column);
 
+            if (!ActionCommandManager.EnhancedAnnotationsEnabled(context)) 
+            {
+                context.Debug("Enhanced Annotations not enabled on the server. The 'title', 'end_line', and 'end_column' fields are unsupported.");
+            }
+
+            IssueCommandExtension.ValidateLinesAndColumns(command);
+
             Issue issue = new Issue()
             {
                 Category = "General",
@@ -576,14 +584,31 @@ namespace GitHub.Runner.Worker
                 }
             }
 
-            if (!ActionCommandManager.EnhancedAnnotationsEnabled(context)) 
+            context.AddIssue(issue);
+        }
+
+        static void ValidateLinesAndColumns(ActionCommand command) 
+        {
+            command.Properties.TryGetValue(IssueCommandProperties.Line, out string line);
+            command.Properties.TryGetValue(IssueCommandProperties.EndLine, out string endLine);
+            command.Properties.TryGetValue(IssueCommandProperties.Column, out string column);
+            command.Properties.TryGetValue(IssueCommandProperties.EndColumn, out string endColumn);
+
+            var hasColumnValue = column != null || endColumn != null;
+            var hasLine = line != null;
+            var hasEndLine = endLine != null;
+
+            Console.WriteLine($"hasColumnValue: {hasColumnValue}, hasLine: {hasLine}");
+
+            if (!hasLine && hasColumnValue) 
             {
-                issue.Data.Remove(IssueCommandProperties.EndLine);
-                issue.Data.Remove(IssueCommandProperties.EndColumn);
-                issue.Data.Remove(IssueCommandProperties.Title);
+                throw new Exception($"Invalid {command.Command} command value. 'column' and 'end_column' can only be set if 'line' value is provided."); 
             }
 
-            context.AddIssue(issue);
+            if (hasEndLine && line != endLine && hasColumnValue) 
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'column' and 'end_column' cannot be set if 'line' and 'end line' are different values."); 
+            }
         }
 
         private static class IssueCommandProperties
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -537,7 +537,6 @@ namespace GitHub.Runner.Worker
             {
                 context.Debug("Enhanced Annotations not enabled on the server. The 'title', 'end_line', and 'end_column' fields are unsupported.");
             }
-
             
             Issue issue = new Issue()
             {
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -609,37 +609,37 @@ namespace GitHub.Runner.Worker
 
             if (hasEndLine && !hasStartLine)
             {
-                context.Debug($"Invalid {command.Command} command value. 'end_line' can only be set if 'line' is provided");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.EndLine}' can only be set if '{IssueCommandProperties.Line}' is provided");
                 return false;
             }
 
             if (hasEndColumn && !hasStartColumn)
             {
-                context.Debug($"Invalid {command.Command} command value. 'end_column' can only be set if 'col' is provided");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.EndColumn}' can only be set if '{IssueCommandProperties.Column}' is provided");
                 return false;
             }
 
             if (!hasStartLine && hasColumn) 
             {
-                context.Debug($"Invalid {command.Command} command value. 'column' and 'end_column' can only be set if 'line' value is provided.");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.Column}' and '{IssueCommandProperties.EndColumn}' can only be set if '{IssueCommandProperties.Line}' value is provided.");
                 return false;
             }
 
             if (hasEndLine && line != endLine && hasColumn) 
             {
-                context.Debug($"Invalid {command.Command} command value. 'column' and 'end_column' cannot be set if 'line' and 'end line' are different values.");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.Column}' and '{IssueCommandProperties.EndColumn}' cannot be set if '{IssueCommandProperties.Line}' and '{IssueCommandProperties.EndLine}' are different values.");
                 return false;
             }
 
             if (hasStartLine && hasEndLine && endLineNumber < lineNumber) 
             {
-                context.Debug($"Invalid {command.Command} command value. 'end_line' cannot be less than 'line'.");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.EndLine}' cannot be less than '{IssueCommandProperties.Line}'.");
                 return false; 
             }
 
             if (hasStartColumn && hasEndColumn && endColumnNumber < columnNumber) 
             {
-                context.Debug($"Invalid {command.Command} command value. 'end_column' cannot be less than 'col'.");
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.EndColumn}' cannot be less than '{IssueCommandProperties.Column}'.");
                 return false; 
             }
 
@@ -650,9 +650,9 @@ namespace GitHub.Runner.Worker
         {
             public const String File = "file";
             public const String Line = "line";
-            public const String EndLine = "end_line";
+            public const String EndLine = "endLine";
             public const String Column = "col";
-            public const String EndColumn = "end_column";
+            public const String EndColumn = "endColumn";
             public const String Title = "title";
         }
 
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -531,7 +531,12 @@ namespace GitHub.Runner.Worker
 
             if (ActionCommandManager.EnhancedAnnotationsEnabled(context)) 
             {
-                IssueCommandExtension.ValidateLinesAndColumns(command);
+
+                if (!IssueCommandExtension.ValidateLinesAndColumns(command, context)) 
+                {
+                    context.Debug($"Validation failed for the {command.Command}. It will not be processed.");
+                    return;
+                }
             }
             else 
             {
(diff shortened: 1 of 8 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -510,6 +511,23 @@ public void AddIssue(Issue issue, string logMessage = null)
                 }
 
                 _record.WarningCount++;
+            } else if (issue.Type == IssueType.Notice) 
+            {
+
+                // tracking line number for each issue in log file
```
thboop, line 517: We should add validation on column/character length like dotcom does so we can surface a better error message to the user. EX: cannot provide column if start/end line are different.

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -222,6 +223,13 @@ public TimelineReference Details
             set;
         }
 
+        [DataMember(Order = 60)]
```
thboop, line 226: Do we need to update order=60 to match the server changes?

src/Runner.Worker/ExecutionContext.cs
luketomlinson: oh right. I need to add that.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -75,6 +75,11 @@ public bool TryProcessCommand(IExecutionContext context, string input, Container
                 return false;
             }
 
+            if (!ActionCommandManager.EnhancedAnnotationsEnabled(context) && actionCommand.Command == "notice")
+            {
+                return false;
```
thboop, line 80: We probably want a Trace or debug message here explaining we are skipping because the ff isn't enabled.

thboop: Debug is user visible if they turn on the debug flag, so we can say something like `Skipping notice annotation as server does not support notice annotations`

Trace is for our purposes, so we can be a bit more detailed about what we say, mentioning FF's and the like.

thboop: Debug is user visible if they turn on the debug flag, so we can say something like `Skipping notice annotation as server does not support notice annotations`

Trace is for our purposes, so we can be a bit more detailed about what we say, mentioning FF's and the like.

src/Test/L0/Worker/ActionCommandManagerL0.cs
```diff
@@ -36,7 +36,7 @@ public void EnablePluginInternalCommand()
                    {
                        hc.GetTrace().Info($"{issue.Type} {issue.Message} {message ?? string.Empty}");
                    });
-
+
```
thboop, line 39: nit: remove this

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -512,6 +529,13 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
             command.Properties.TryGetValue(IssueCommandProperties.Line, out string line);
             command.Properties.TryGetValue(IssueCommandProperties.Column, out string column);
 
+            if (!ActionCommandManager.EnhancedAnnotationsEnabled(context)) 
+            {
+                context.Debug("Enhanced Annotations not enabled on the server. The 'title', 'end_line', and 'end_column' fields are unsupported.");
+            }
+
+            IssueCommandExtension.ValidateLinesAndColumns(command);
```
thboop, line 537: FF this under the same FF

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +587,36 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+
+            var hasColumnValue = column != null || endColumn != null;
+            var hasLine = line != null;
+            var hasEndLine = endLine != null;
+
+            if (!hasLine && hasColumnValue) 
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'column' and 'end_column' can only be set if 'line' value is provided."); 
+            }
+
```
thboop, line 605: Should we also check if end_column is provided but column is not? Same with line. What happens if you post a check with an end-column but no start column?

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +589,48 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+            var hasStartLine = !String.IsNullOrWhiteSpace(line);
+            var hasEndLine = !String.IsNullOrWhiteSpace(endLine);
+            var hasStartColumn = !String.IsNullOrWhiteSpace(column);
+            var hasEndColumn =  !String.IsNullOrWhiteSpace(endColumn);
+            var hasColumn = hasStartColumn || hasEndColumn;
+
+            if (hasEndLine && !hasStartLine)
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'end_line' can only be set of 'line' is provided");
```
thboop, line 607:
```suggestion
                throw new Exception($"Invalid {command.Command} command value. 'end_line' can only be set if 'line' is provided");
```

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +589,48 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+            }
+
+            if (hasEndColumn && !hasStartColumn)
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'end_column' can only be set of 'col' is provided");
```
thboop, line 612:
```suggestion
                throw new Exception($"Invalid {command.Command} command value. 'end_column' can only be set if 'col' is provided");
```

src/Runner.Worker/JobRunner.cs
```diff
@@ -110,6 +110,8 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                 Directory.CreateDirectory(toolsDirectory);
                 jobContext.SetRunnerContext("tool_cache", toolsDirectory);
 
+                jobContext.SetRunnerContext("version", BuildConstants.RunnerPackage.Version);
```
ericsciple, line 113: can we remove this for now? i think this is worth wider discussion first. There are a few different approaches to solve the problem. Maybe a small ADR to discuss the merits of each why we chose one over the others.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -106,6 +106,7 @@ public sealed class ExecutionContext : RunnerService, IExecutionContext
     {
         private const int _maxIssueCount = 10;
         private const int _throttlingDelayReportThreshold = 10 * 1000; // Don't report throttling with less than 10 seconds delay
+        private const string _noticeLogPrefix = "\u001b[31m"; //white text
```
ericsciple, line 109: we should be consistent with error/warning. iirc today that means the log view (recognizes lines that start with `[error]` or `[warning]`)

also curious if you set the ANSI color code, do you need to reset it at the end of the line? You might need to download the log and open in a text viewer to see.

ericsciple: If we switch to ANSI color codes we should be consistent and update error/warning too.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -510,6 +511,24 @@ public void AddIssue(Issue issue, string logMessage = null)
...
                 _record.WarningCount++;
+            } 
+            else if (issue.Type == IssueType.Notice) 
+            {
+
+                // tracking line number for each issue in log file
+                // log UI use this to navigate from issue to log
+                if (!string.IsNullOrEmpty(logMessage))
+                {
+                    long logLineNumber = Write(_noticeLogPrefix, logMessage);
```
ericsciple, line 522: If we prefix with ANSI color codes, should we strip color codes from the log message?

ericsciple: interesting offline discussion w/ @luketomlinson 

should we strip color codes from annotations. whether annotations supports special rendering (e.g. html/markdown) will help guide whether we should strip color codes or fix annotations

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +589,58 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+
+            var hasStartLine = int.TryParse(line, out int lineNumber);
+            var hasEndLine = int.TryParse(endLine, out int endLineNumber);
+            var hasStartColumn = int.TryParse(column, out int columnNumber);
+            var hasEndColumn = int.TryParse(endColumn, out int endColumnNumber);
+            var hasColumn = hasStartColumn || hasEndColumn;
+
+            if (hasEndLine && !hasStartLine)
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'end_line' can only be set if 'line' is provided");
```
ericsciple, line 607: should we throw or just debug-log and ignore the value?

ericsciple: personally wouldnt throw

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +589,58 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+            {
+                throw new Exception($"Invalid {command.Command} command value. 'end_column' cannot be less than 'col'."); 
+            }    
+        }
+
         private static class IssueCommandProperties
         {
             public const String File = "file";
             public const String Line = "line";
+            public const String EndLine = "end_line";
```
ericsciple, line 640: do we have precedent for compound words?

ericsciple: Wonder whether we should use `-` or camel case. For example, consider:

`::set-foo my_prop=val, my_prop2=val2:some data`

idk the right answer. Personally think camel case maybe? (2 cents)

If camel case, make sure the comparison is ordinalignorecase

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,11 +589,58 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
             public const String Column = "col";
+            public const String EndColumn = "end_column";
```
ericsciple, line 642: `endCol`

ericsciple (dismissed):

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -512,6 +529,15 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
             command.Properties.TryGetValue(IssueCommandProperties.Line, out string line);
             command.Properties.TryGetValue(IssueCommandProperties.Column, out string column);
 
+            if (ActionCommandManager.EnhancedAnnotationsEnabled(context))
```
TingluoHuang, line 532: do we need to check this again? we already skip this in `TryProcessCommand()`

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -141,6 +147,10 @@ public bool TryProcessCommand(IExecutionContext context, string input, Container
 
             return true;
         }
+
+        internal static bool EnhancedAnnotationsEnabled(IExecutionContext context) {
```
TingluoHuang, line 151: `private bool`?

TingluoHuang: or do we need a method for a single line of code? 😆

src/Runner.Worker/JobRunner.cs
TingluoHuang: +1

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: +1 make it consistent with error/warning.

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: no matter what we do, please make sure `error==warning==notice`

src/Runner.Worker/ActionCommandManager.cs
TingluoHuang: `ArgumentException` might be better than `exception`

src/Runner.Worker/ActionCommandManager.cs
luketomlinson: So `WarningCommandExtension` and `ErrorCommandExtension` both inherit from this class and are processed. There are 2 pieces that are under the same feature flag: The `notice` command and the additional field support.

src/Runner.Worker/ActionCommandManager.cs
TingluoHuang: +1, what happens when we throw?
will the job fail?
the bottom line is I don't think a misconstructed logging command should fail the job.

src/Runner.Worker/ActionCommandManager.cs
luketomlinson: lol. My only issue was the `?? false` which could be changed in one place and not the other.

luketomlinson: because this method is called in 2 places and if there would be 2 instances of `?? false`.  There is potential for them not to match if someone were to change it to `?? true` on one of them

src/Runner.Worker/ActionCommandManager.cs
luketomlinson: yep that's good feedback. I will change to just debug log and not fail

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,13 +594,67 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
+            var hasStartLine = int.TryParse(line, out int lineNumber);
+            var hasEndLine = int.TryParse(endLine, out int endLineNumber);
+            var hasStartColumn = int.TryParse(column, out int columnNumber);
+            var hasEndColumn = int.TryParse(endColumn, out int endColumnNumber);
+            var hasColumn = hasStartColumn || hasEndColumn;
+
+            if (hasEndLine && !hasStartLine)
+            {
+                context.Debug($"Invalid {command.Command} command value. '{IssueCommandProperties.EndLine}' can only be set if '{IssueCommandProperties.Line}' is provided");
+                return false;
```
ericsciple, line 613: rather than return false, have you considered setting the invalid value to zero

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -563,13 +594,67 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
             context.AddIssue(issue);
         }
 
+        public static bool ValidateLinesAndColumns(ActionCommand command, IExecutionContext context) 
+        {
+            command.Properties.TryGetValue(IssueCommandProperties.Line, out string line);
+            command.Properties.TryGetValue(IssueCommandProperties.EndLine, out string endLine);
+            command.Properties.TryGetValue(IssueCommandProperties.Column, out string column);
+            command.Properties.TryGetValue(IssueCommandProperties.EndColumn, out string endColumn);
```
ericsciple, line 602: do we need to consider cases where `endColumn` != empty string but int.tryparse fails?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1012,6 +1031,7 @@ public static class WellKnownTags
         public static readonly string Command = "##[command]";
         public static readonly string Error = "##[error]";
         public static readonly string Warning = "##[warning]";
+        public static readonly string Notice = "##[notice]";
```
ericsciple, line 1034: we might want to make the UI changes first before we start writing this.... actually i guess it doesnt matter because no one will be using this initially. So we have some time to make that change in parallel

ericsciple (dismissed):

src/Runner.Worker/ActionCommandManager.cs
luketomlinson: This is accounted for since `int.TryParse` will return false if it can't parse an int

ericsciple (approved):

aikee07 (approved):

aikee07 (approved):

Issue #849, opened 2020-12-05
Float 3.0 is misparsed as integer 3 in workflow YAML


eregon:
**Describe the bug**

`3.0` in a workflow is misparsed as integer `3`.
Or at the very least, when it's parsed and then later converted to a string (with `${{ matrix.ruby }}`), it's `3` and not the expected `3.0`.

**To Reproduce**
```yaml
jobs:
  test:
    strategy:
      matrix:
        os: [ ubuntu-latest ]
        ruby: [ 2.7, 3.0 ]
    name: ${{ matrix.os }} ${{ matrix.ruby }}
    runs-on: ${{ matrix.os }}
    steps:
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: ${{ matrix.ruby }}
```

**Expected behavior**
The input log and the job name should indicate `3.0`, but we see `3` instead:
https://github.com/eregon/setup-ruby/runs/1503368422?check_suite_focus=true#step:3:3
```
Run ./
  with:
    ruby-version: 3
    bundler-cache: true
    bundler: default
    working-directory: .
```

This is important because `ruby-version: 3.0` should pick version 3.0 but not version 3.1.
Due to this bug it would pick version 3.1.

I think in the YAML specification it's clear that floats and integers are different, so they should be kept that way too in workflow YAMLS, including when passing through `${{ matrix.ruby }}`.

Using quotes `[ 2.7, '3.0' ]` is a workaround, but it should be unnecessary and it's much more verbose with many versions.

## Runner Version and Platform
Version of your runner? 2.274.2
OS of the machine running the runner? Applies to all

maxim-lobanov:
@eregon  I have faced with the same issue in my own action https://github.com/maxim-lobanov/setup-xcode and decided that the best solution is updating my action's readme to recommend customers to use quotes in inputs.

I agree that the behavior that you have described looks a bit strange and it could be a bug. But I have found another problem if inputs are not quoted:
```
with:
    python-version: 3.10
```
This input will be parsed as `3.1` and it is definitely not that you expect to see. But based on YAML specification it is expected behavior because `3.10` is float value and trailing `0` can be removed. 

A few words from https://symfony.com/doc/current/components/yaml/yaml_format.html
> Strings in YAML can be wrapped both in single and double quotes. **In some cases, they can also be unquoted**
> ...
> Finally, there are other cases when the strings **must be quoted**, no matter if you’re using single or double quotes:
> ...
[...]

PR #1176, opened 2021-06-28
fix: Do not merge missing default value into Action inputs

### LINKED ISSUE #924: Differentiate between input variable empty or not present
Current behavior:

- If an input parameter is defined in action.yml (without default value), GitHub will set the INPUT_PARAMNAME environment variable to an empty value if the parameter is not present in the workflow

Expected behavior:

- If an input parameter is defined in action.yml (without default value), GitHub should not set the INPUT_PARAMNAME environment variable to an empty value if the parameter is not present in the workflow
Possibly related issue for JavaScript actions: https://github.com/actions/toolkit/issues/272

Resources that illustrate this issue:

- Dockerfile: https://github.com/rsenden/test-empty-var-action/blob/master/Dockerfile
- GitHub Action for this Docker image: https://github.com/rsenden/test-empty-var-action/blob/master/action.yml
- Sample workflow that invokes this Docker image both directly and through the action, with the test input parameter not set: https://github.com/rsenden/test-empty-var-action/blob/master/.github/workflows/notset.yml
- Workflow output: https://github.com/rsenden/test-empty-var-action/runs/1707506513?check_suite_focus=true

As can be seen in the workflow output, results are different between running the Docker image directly or through the GitHub Action.

This can cause issues if a Docker image (or other action implementation) differentiates between 'empty' and 'not set'. If an action wants to set an empty value as the default value, then this should be done using the default property in the action definition.

Can you please get this fixed?

shrink:
Closes #924 

At present, an Input without a default value is given a default value of an empty string, meaning there's no way for Action authors to differentiate between "no value was provided" and "an empty string was provided". As described in #924, this introduces a number of challenges for Action authors and has been acknowledged as a (low priority) bug.

The change in this Pull Request is very simple: do not set a default string value when no default is provided -- using a `NullToken` instead of a `StringToken`. As a consequence, the `INPUT_*` environment variables will now only contain a value when an input _or default_ is provided.

## Backwards Compatibility

The GitHub Actions toolkit package `core` provides a `getInput` method which checks for the presence of the environment variable and uses a default empty string if the environment variable is not found, therefore this change would not break the behaviour of the toolkit. However, it _would_ enable the introduction of `hasInput` as described in my Pull Request [feat: Check if action run hasInput #849](https://github.com/actions/toolkit/pull/849).

```ts
const val: string = process.env[`INPUT_${name.replace(/ /g, '_').toUpperCase()}`] || ''
```
https://github.com/actions/toolkit/blob/45647689407e7fb224e06d066dde6aefa67a365f/packages/core/src/core.ts#L80-L101

I've performed various tests of this against my own actions (like [actions-assert](https://github.com/pr-mpt/actions-assert)) and experienced no issues, so I'm fairly confident in saying it works without issue.

### `action.yml`

```yaml
inputs:
  example:
    description: "Example of an optional input with no default"
```

```yaml
- uses: ./
- run: printenv INPUT_EXAMPLE
```

```console
$ printenv INPUT_EXAMPLE
''
```

### `action.yml`

```yaml
inputs:
  example:
    description: "Example of an optional input with no default"
```

```yaml
- uses: ./
- run: printenv INPUT_EXAMPLE
```

```console
$ printenv INPUT_EXAMPLE

```

### DIFF
--- a/src/Test/L0/Worker/ActionManifestManagerL0.cs
+++ b/src/Test/L0/Worker/ActionManifestManagerL0.cs
@@ -45,7 +45,7 @@ namespace GitHub.Runner.Common.Tests.Worker
                 Assert.Equal("greeting", result.Inputs[0].Key.AssertString("key").Value);
                 Assert.Equal("Hello", result.Inputs[0].Value.AssertString("value").Value);
                 Assert.Equal("entryPoint", result.Inputs[1].Key.AssertString("key").Value);
-                Assert.Equal("", result.Inputs[1].Value.AssertString("value").Value);
+                Assert.Equal("", result.Inputs[1].Value.AssertNull("value").ToString());
 
                 Assert.Equal(ActionExecutionType.Container, result.Execution.ExecutionType);
 
@@ -89,7 +89,7 @@ namespace GitHub.Runner.Common.Tests.Worker
                 Assert.Equal("greeting", result.Inputs[0].Key.AssertString("key").Value);
                 Assert.Equal("Hello", result.Inputs[0].Value.AssertString("value").Value);
                 Assert.Equal("entryPoint", result.Inputs[1].Key.AssertString("key").Value);
-                Assert.Equal("", result.Inputs[1].Value.AssertString("value").Value);
+                Assert.Equal("", result.Inputs[1].Value.AssertNull("value").ToString());
 
                 Assert.Equal(ActionExecutionType.Container, result.Execution.ExecutionType);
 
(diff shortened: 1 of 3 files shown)


Malix-Labs:
I am not sure about `Assert.Equal("", result.Inputs[1].Value.AssertNull("value").ToString());`

shouldn't it be something like `Assert.Equal(null, result.Inputs[1].Value.AssertNull("value"));` ?

src/Test/L0/Worker/ActionManifestManagerL0.cs
```diff
@@ -570,7 +526,7 @@ public void Load_NodeAction_Cleanup()
                 Assert.Equal("greeting", result.Inputs[0].Key.AssertString("key").Value);
                 Assert.Equal("Hello", result.Inputs[0].Value.AssertString("value").Value);
                 Assert.Equal("entryPoint", result.Inputs[1].Key.AssertString("key").Value);
-                Assert.Equal("", result.Inputs[1].Value.AssertString("value").Value);
+                //Assert.Equal("", result.Inputs[1].Value.AssertString("value").Value);
```
Malix-Labs, line 529: isn't this a typo? it's the only odd one out

shrink: I have no recollection of creating this Pull Request. 

I'd guess there was a reason for the line to be commented out but on review, without any of the context that I might have had at the time, I can't find any. If there were a reason, I hope I would have included an explanation.

Since it is very unlikely this Pull Request will ever be merged, I'll leave it as is. I doubt this Pull Request could even be merged as-is given there is half a decade of changes since then.

PR #1179, opened 2021-06-29
Update error to say 'uninstall' not 'unconfigure'

nick-invision:
When attempting to `unconfigure` a runner via `./config.sh remove` on an installed OSX service, the error currently reads `Unconfigure service first` (which is what removing the config technically is doing) but the user must run `./svc.sh uninstall` to uninstall the service before unconfiguring it.  It's petty but the exception is unclear and a bit misleading, so update it to reflect the current action that must be taken before actually removing the configuration.


fhammerl (approved):
LGTM

PR #1191, opened 2021-07-06
fix typos in docs of JobDispatcher

rethab:


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -256,7 +258,7 @@ private async Task EnsureDispatchFinished(WorkerDispatcher jobDispatch, bool can
                     Trace.Error(ex);
 
                     jobDispatch.WorkerCancellationTokenSource.Cancel();
-                    // make sure worker process exit before we rethrow, otherwise we might leave orphan worker process behind.
+                    // make sure worker the process exits before we rethrow, otherwise we might leave orphan worker process behind.
```
fhammerl, line 261:
```suggestion
                    // make sure the worker process exits before we rethrow, otherwise we might leave orphan worker process behind.
```

fhammerl (changes requested):
Thanks for the fixes! I made a suggestion for the one last typo that remains.

fhammerl (approved):

fhammerl (approved):

Issue #960, opened 2021-02-02
Better error message when `docker` is not installed.


TingluoHuang:
Currently, we are getting `Error: File not found: 'docker'` which is not really actionable to anyone.

hross:
If you are using a container action without docker installed, this is the error. We should make it more obvious what the problem is.

PR #1210, opened 2021-07-20
Better error message when a command is not found

fhammerl:
Closes https://github.com/actions/runner/issues/960

For example, `Error: File not found: 'docker'` may be hard to make sense of, so I replaced the exception message (that gets propagated to the UI log) with the Linux default for when a command is not found.

Additionally, I included some extra info in the message the runner logs to help people who hit this problem on a self-hosted runner.

I kept the exception type `FileNotFoundException`, because it describes the problem precisely.

### DIFF
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -116,9 +116,9 @@ namespace GitHub.Runner.Sdk
             }
 
 #if OS_WINDOWS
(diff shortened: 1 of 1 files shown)


src/Runner.Sdk/Util/WhichUtil.cs
```diff
@@ -115,11 +115,15 @@ public static string Which(string command, bool require = false, ITraceWriter tr
                 }
             }
 
-            trace?.Info("Not found.");
```
fhammerl, line 118: Could be the same message with 'Path' / 'PATH'

Originally, I thought the messages for Win and Nix would be quite different

thboop (approved):
LGTM

PR #1213, opened 2021-07-20
Allow setting default severity to "notice"

CyberShadow:
Pull request #203 added a way to specify the default severity for a matcher. This allows us to define multiple matchers per severity, each matching only that specific severity.

However, currently the only allowed values for the default severity setting are "ERROR", "WARNING" and the empty string:

https://github.com/actions/runner/blob/dcda342eccd152fe69521d59de016ef09b70400c/src/Runner.Worker/IssueMatcher.cs#L350-L355

The empty string is interpreted in the same way as "ERROR":

https://github.com/actions/runner/blob/592ce1b230985aea359acdf6ed4ee84307bbedc1/src/Runner.Worker/Handlers/OutputManager.cs#L205-L208

As such, there is no way to add a matcher with the default severity of "notice".

This pull request addresses this by adding "notice" as one of the allowed options for the default matcher severity.

### DIFF
--- a/src/Runner.Worker/IssueMatcher.cs
+++ b/src/Runner.Worker/IssueMatcher.cs
@@ -350,6 +350,7 @@ namespace GitHub.Runner.Worker
                 case "":
                 case "ERROR":
                 case "WARNING":
+                case "NOTICE":
                     break;
                 default:
                     throw new ArgumentException($"Matcher '{_owner}' contains unexpected default severity '{_severity}'");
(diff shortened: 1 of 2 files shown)


luketomlinson (approved):
Thanks for the PR @CyberShadow! It looks good to me. Notice annotations aren't fully supported yet, but will be soon. We will hold off merging this one until it's rolled out.

Issue #996, opened 2021-02-23
Enhancement request to use flags for arguments instead of position for automation scripts


ruberVulpes:
When using the `create-latest-svc.sh` script I want to be able to only specify my needed options (scope/ghe_domain/labels) however since the args are pulled by position I need to also pass in the defaults for name/user. 

I was wondering if it would be possible to add support for using a flag for setting the arguments to allow the setting of labels without passing in default values for the preceding args.

https://github.com/actions/runner/blob/af198237ca6b213b77d8bdb535bd37b18a683aef/scripts/create-latest-svc.sh#L13-L34

PR #1220, opened 2021-07-22
Allow the use of flags in scripts/create-latest-svc.sh in a backwards compatible way

fhammerl:
Closes https://github.com/actions/runner/issues/996

Previously, if you only want to pass for example scope and labels, you would still have to pass all five parameters, because labels is expected to be the fifth parameter (${5}).

This PR allows for passing flags instead. If the only mandatory argument (scope) is not found, we fall back to the original, sequential parsing.


scripts/create-latest-svc.sh
```diff
@@ -27,11 +28,50 @@ set -e
...
+    esac
+    shift
+done
+else
+# process indexed args for backwards compatibility
 runner_scope=${1}
 ghe_hostname=${2}
 runner_name=${3:-$(hostname)}
 svc_user=${4:-$USER}
 labels=${5}
```
hfaulds, line 69:
```suggestion
    runner_scope=${1}
    ghe_hostname=${2}
    runner_name=${3:-$(hostname)}
    svc_user=${4:-$USER}
    labels=${5}
```

hfaulds (dismissed):
👍 other than some indentation changes

The regex isn't necessarily ideal but we do it in other scripts and it's a lot simpler than what would be required to check for flags otherwise.

TingluoHuang (approved):

PR #1229, opened 2021-07-28
Add better step telemetry and tracing for composite Actions

thboop:
This PR adds step telemetry (including embedded steps) to the JobExecutionContext. This information is then sent over to the service, and logged into our telemetry so that we can iterate on what features users are using in the runner.

This PR also does a pass over composite action tracing, to make it more in line with composite run step telemetry.

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -156,6 +156,16 @@ namespace GitHub.Runner.Worker.Handlers
             var githubContext = ExecutionContext.ExpressionValues["github"] as GitHubContext;
             ArgUtil.NotNull(githubContext, nameof(githubContext));
 
+            // Add Telemetry
+            if (stage == ActionRunStage.Main)
+            {
+                var telemetry = new ActionsStepTelemetry {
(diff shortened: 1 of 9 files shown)


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -156,6 +144,16 @@ public async Task RunAsync(ActionRunStage stage)
             var githubContext = ExecutionContext.ExpressionValues["github"] as GitHubContext;
             ArgUtil.NotNull(githubContext, nameof(githubContext));
 
+            // Add Telemetry to JobContext to send with JobCompleteMessage
+            if (stage == ActionRunStage.Main)
+            {
+                var telemetry = new ActionsStepTelemetry {
```
juliobbv, line 150: Consider adding a way to determine what shell customers are using to execute scripts.

juliobbv:
LGTM, approving

juliobbv (approved):

TingluoHuang (approved):

src/Runner.Worker/Handlers/ScriptHandler.cs
juliobbv: Yeah, not the input itself, but just the "entrypoint".

PR #1232, opened 2021-08-02
Composite Code Cleanup

thboop:
This pr fixes some minor bugs associated with composite actions and prepares the feature for release

Technical changes
- Correctly sets the condition for pre and post steps, and ensures that these conditions are evaluated when pre/post steps are ran
- When a composite action fails, we update the Job status early so we can evaluate the next composite step correctly
- Fixes an issue where we modify an existing collection crashing the for loop

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -247,7 +247,9 @@ namespace GitHub.Runner.Worker
                                 {
                                     _cachedEmbeddedPreSteps[parentStepId] = new List<Pipelines.ActionStep>();
                                 }
-                                _cachedEmbeddedPreSteps[parentStepId].Add(action);
+                                var clonedAction = action.Clone() as Pipelines.ActionStep;
+                                clonedAction.Condition = definition.Data.Execution.InitCondition;
+                                _cachedEmbeddedPreSteps[parentStepId].Add(clonedAction);
                             }
                         }
 
@@ -258,7 +260,9 @@ namespace GitHub.Runner.Worker
                                 // If we haven't done so already, add the parent to the post steps
                                 _cachedEmbeddedPostSteps[parentStepId] = new Stack<Pipelines.ActionStep>();
                             }
-                            _cachedEmbeddedPostSteps[parentStepId].Push(action);
+                            var clonedAction = action.Clone() as Pipelines.ActionStep;
+                            clonedAction.Condition = definition.Data.Execution.CleanupCondition;
+                            _cachedEmbeddedPostSteps[parentStepId].Push(clonedAction);
                         }
                     }
                 }
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -247,6 +247,7 @@ namespace GitHub.Runner.Worker
                                 {
                                     _cachedEmbeddedPreSteps[parentStepId] = new List<Pipelines.ActionStep>();
                                 }
+                                // Clone action so we can modify the condition without affecting the original
                                 var clonedAction = action.Clone() as Pipelines.ActionStep;
                                 clonedAction.Condition = definition.Data.Execution.InitCondition;
                                 _cachedEmbeddedPreSteps[parentStepId].Add(clonedAction);
@@ -260,6 +261,7 @@ namespace GitHub.Runner.Worker
                                 // If we haven't done so already, add the parent to the post steps
                                 _cachedEmbeddedPostSteps[parentStepId] = new Stack<Pipelines.ActionStep>();
                             }
+                            // Clone action so we can modify the condition without affecting the original
                             var clonedAction = action.Clone() as Pipelines.ActionStep;
                             clonedAction.Condition = definition.Data.Execution.CleanupCondition;
                             _cachedEmbeddedPostSteps[parentStepId].Push(clonedAction);
--- a/src/Runner.Worker/ActionRunner.cs
+++ b/src/Runner.Worker/ActionRunner.cs
@@ -82,12 +82,6 @@ namespace GitHub.Runner.Worker
             ActionExecutionData handlerData = definition.Data?.Execution;
             ArgUtil.NotNull(handlerData, nameof(handlerData));
 
-            if (handlerData.HasPre &&
-                Action.Reference is Pipelines.RepositoryPathReference repoAction &&
-                string.Equals(repoAction.RepositoryType, Pipelines.PipelineConstants.SelfAlias, StringComparison.OrdinalIgnoreCase))
-            {
-                ExecutionContext.Warning($"`pre` execution is not supported for local action from '{repoAction.Path}'");
-            }
             List<JobExtensionRunner> localActionContainerSetupSteps = null;
             // Handle Composite Local Actions
             // Need to download and expand the tree of referenced actions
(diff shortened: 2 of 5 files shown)


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -268,13 +277,57 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps)
                     step.ExecutionContext.Complete(TaskResult.Failed);
                 }
 
-                await RunStepAsync(step);
+                // Evaluate condition
+                step.ExecutionContext.Debug($"Evaluating condition for step: '{step.DisplayName}'");
```
thboop, line 281: This mirrors the Steps Runner functionality: https://github.com/actions/runner/blob/4e95d0d6ad46e454749cc5d69a7072df34a8e12f/src/Runner.Worker/StepsRunner.cs#L181

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -268,13 +277,57 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps)
...
+                    await RunStepAsync(step);
+                }
 
                 // Check failed or canceled
                 if (step.ExecutionContext.Result == TaskResult.Failed || step.ExecutionContext.Result == TaskResult.Canceled)
                 {
+                    Trace.Info($"Update job result with current composite step result '{step.ExecutionContext.Result}'.");
                     ExecutionContext.Result = step.ExecutionContext.Result;
-                    break;
+                    ExecutionContext.Root.Result = TaskResultUtil.MergeTaskResults(ExecutionContext.Root.Result, step.ExecutionContext.Result.Value);
```
thboop, line 329: Notice we removed the break. In the past we would short circuit running composite steps here. Now we update the JobContext, so when we evaluate the condition for the next step (defaults to `success()` it will fail).

We have to do this because we may have multiple post steps with unique conditions. If the first post step fails, and the second has a condition of `failure()` we need to ensure we actually run the second. So the job status needs to be up to date.

The goal here is to mirror how these steps would have ran if they were inside a workflow file and not a composite action.

src/Runner.Worker/ActionManager.cs
```diff
@@ -247,7 +247,9 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
                                 {
                                     _cachedEmbeddedPreSteps[parentStepId] = new List<Pipelines.ActionStep>();
                                 }
-                                _cachedEmbeddedPreSteps[parentStepId].Add(action);
+                                var clonedAction = action.Clone() as Pipelines.ActionStep;
```
thboop, line 250: Need to clone so we can update the condition without modify other copies of it.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -354,43 +354,5 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
 
             executionContext.Complete(result, resultCode: resultCode);
         }
-
-        private sealed class ConditionTraceWriter : ObjectTemplating::ITraceWriter
```
thboop, line 358: Just moved to its own file

src/Runner.Worker/ActionManager.cs
ericsciple: can you add a comment in the source code

src/Runner.Worker/ActionManager.cs
```diff
@@ -258,7 +260,9 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
                                 // If we haven't done so already, add the parent to the post steps
                                 _cachedEmbeddedPostSteps[parentStepId] = new Stack<Pipelines.ActionStep>();
                             }
-                            _cachedEmbeddedPostSteps[parentStepId].Push(action);
+                            var clonedAction = action.Clone() as Pipelines.ActionStep;
```
ericsciple, line 263: same here, comment in the source code will save future headache :)

src/Runner.Worker/ActionRunner.cs
```diff
@@ -110,6 +104,13 @@ public async Task RunAsync()
                 localActionContainerSetupSteps = prepareResult.ContainerSetupSteps;
             }
 
+            if (handlerData.HasPre &&
```
ericsciple, line 107: was this moved because it needs to come after prepare ActionsAsync when composite-local?

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -45,13 +46,17 @@ public async Task RunAsync(ActionRunStage stage)
...
                 {
-                    if (!ExecutionContext.Root.EmbeddedStepsWithPostRegistered.Contains(step.Id))
+                    if (ExecutionContext.Root.EmbeddedStepsWithPostRegistered.Contains(step.Id))
+                    {
+                        steps.Add(step);
+                    }
+                    else
                     {
-                        steps.Remove(step);
+                        Trace.Info($"Skipping executing post step id: {step.Id}, name: ${step.DisplayName}");
```
ericsciple, line 59: remind me, what is `Trace`? I thought we used either ExecutionContext or HostContext for tracing. Is `Trace` the HostContext trace writer?

ericsciple: Also, should we write this to the execution context instead (i.e. user logs)? I don't know the correct answer, but posing the question.

src/Runner.Worker/ActionRunner.cs
thboop: correct, we need to load the local  composite action to know if it has a Pre

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ericsciple: cool, does this mean we support condition on steps now?

ericsciple: i dont think this is complete. We need to register the callback so we can re-test when cancellation actually fires. This just tests before the step is run. We need to re-test when cancellation fires and the step is in-progress.

thboop: Not at the moment, but the infrastructure is there we just need to enable the keyword (and be sure its the functionality we actually want)

ericsciple: if this PR is incrementally correct, you might want to figure out the callback thing in a future PR

ericsciple: ...since this feature is behind a FF anyway

ericsciple (dismissed):

src/Runner.Worker/Handlers/CompositeActionHandler.cs
thboop: This goes into the worker diag log, which mirrors how we do this in other places. I'll think a bit about a message for the ExecutionContext, might be worth adding in addition.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
thboop: I went ahead and put it in this pr, the changes are relatively scoped and was able to do testing for it

ericsciple (approved):

PR #1239, opened 2021-08-03
Finish job when worker crashed with IOException.

TingluoHuang:
We don't want to mark job failure as infrastructure failure when the runner runs out of disk.

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -507,7 +507,20 @@ namespace GitHub.Runner.Listener
                                 {
                                     detailInfo = string.Join(Environment.NewLine, workerOutput);
                                     Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
-                                    await LogWorkerProcessUnhandledException(message, detailInfo);
+
+                                    var jobServer = HostContext.GetService<IJobServer>();
+                                    VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
+                                    VssConnection jobConnection = VssUtil.CreateConnection(systemConnection.Url, jobServerCredential);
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -507,7 +507,20 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                                 {
                                     detailInfo = string.Join(Environment.NewLine, workerOutput);
                                     Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
-                                    await LogWorkerProcessUnhandledException(message, detailInfo);
+
+                                    var jobServer = HostContext.GetService<IJobServer>();
```
TingluoHuang, line 511: these are pulled out from the `LogWorkerProcessUnhandledException()` since I need to re-use `jobServer` later.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -507,7 +507,20 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                                    VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
+                                    VssConnection jobConnection = VssUtil.CreateConnection(systemConnection.Url, jobServerCredential);
+                                    await jobServer.ConnectAsync(jobConnection);
+
+                                    await LogWorkerProcessUnhandledException(jobServer, message, detailInfo);
+
+                                    // Go ahead to finish the job with result 'Failed' if the STDERR from worker is System.IO.IOException, since it typically means we are running out of disk space.
```
TingluoHuang, line 518: I can't really inspect base on the exact error message since the message are different on Windows vs. Linux, and the message are localized base on the machine culture setting.

thboop (approved):
LGTM

PR #1241, opened 2021-08-03
Set Context Name for Pre/Post Steps

thboop:
This PR sets the context name for pre/post steps to ensure composite pre/post steps are able to correctly pull the steps context as needed. Composite actions convert these into a scope name so they can properly pull the correct steps context based on the scope.

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -272,7 +272,7 @@ namespace GitHub.Runner.Worker
                 return;
             }
 
-            step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState);
+            step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState, this.ScopeName);
             Root.PostJobSteps.Push(step);
         }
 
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/Variables.cs
```diff
@@ -70,7 +70,7 @@ public Variables(IHostContext hostContext, IDictionary<string, VariableValue> co
         public bool Retain_Default_Encoding => true;
 #endif
 
-        public bool? Step_Debug => GetBoolean(Constants.Variables.Actions.StepDebug);
+        public bool? Step_Debug => true;
```
ericsciple, line 73: was this for testing? did you mean to revert?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -272,7 +272,7 @@ public void RegisterPostJobStep(IStep step)
                 return;
             }
 
-            step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState);
+            step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState, ContextName);
```
ericsciple, line 275: do we normally set ContextName for pre/post? can pre/post actually contribute to step outputs?

PR #1243, opened 2021-08-04
Set Composite Post Step Context to the Main's step

thboop:
Currently, we generate a new scopename and associated steps context for each invocation of composite actions (pre/main/post). In post, we may have inputs that were in the steps context of the main. So, we actually need to use the steps context of the main when invoking composite actions. This pr stores that information as `SiblingStepContext` on the execution context, and computes it recursively as needed.

Each layer of composite actions will have their own step scope, you can see an example of a post run that is nested multiple deep printing out the step context below:

image

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -38,6 +38,7 @@ namespace GitHub.Runner.Worker
         Guid Id { get; }
         Guid EmbeddedId { get; }
         string ScopeName { get; }
+        string SiblingScopeName { get; }
         string ContextName { get; }
         Task ForceCompleted { get; }
         TaskResult? Result { get; set; }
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -271,12 +274,16 @@ public void RegisterPostJobStep(IStep step)
                 Trace.Info($"'post' of '{actionRunner.DisplayName}' already push to post step stack.");
                 return;
             }
+            if (step is IActionRunner runner)
```
TingluoHuang, line 277: maybe `actionRunner`

TingluoHuang:
Maybe add some L0 to assert the behavior.

TingluoHuang (approved):

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -38,6 +38,7 @@ public interface IExecutionContext : IRunnerService
         Guid Id { get; }
         Guid EmbeddedId { get; }
         string ScopeName { get; }
+        string SiblingScopeName { get; }
```
ericsciple, line 41: consider `MainScopeName` ?  trying to think of more names but it's hard 😞

PR #1250, opened 2021-08-05
Send path when resolving actions

thboop:
This PR sends up the path in the resolve actions endpoint so we can properly validate policy when JIT downloading Composite Actions

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -610,6 +610,7 @@ namespace GitHub.Runner.Worker
                     {
                         NameWithOwner = repositoryReference.Name,
                         Ref = repositoryReference.Ref,
+                        Path = repositoryReference.Path,
                     };
                 })
                 .ToList();
--- a/src/Sdk/DTWebApi/WebApi/ActionReference.cs
+++ b/src/Sdk/DTWebApi/WebApi/ActionReference.cs
@@ -18,5 +18,12 @@ namespace GitHub.DistributedTask.WebApi
             get;
             set;
         }
+
+        [DataMember]
+        public string Path
+        {
+            get;
+            set;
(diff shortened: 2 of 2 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -610,6 +610,7 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
                     {
                         NameWithOwner = repositoryReference.Name,
                         Ref = repositoryReference.Ref,
+                        Path = repositoryReference.Path,
```
ericsciple, line 613: nit: consider deleting the comment on this method (line 595)

ericsciple: `my-org/my-repo/my-folder/my-subfolder@v1`

- NameWithOwner: my-org/my-repo
- Ref: v1
- Path my-folder/my-subfolder

ericsciple (approved):

KyleJamesWalker:
@thboop or @ericsciple do you think this would be causing our composite actions to be erroring out now?

It was working yesterday just fine, and now we're seeing:
```
Error: `uses:` keyword is not currently supported.
```

I noticed this was part of the release that came out 4 hours ago, we are using a composite action with uses clauses.

thboop:
@KyleJamesWalker were you using custom self hosted runnners by any chance? Composite should never have been enabled for public use on github.com. That being said, we plan to ship the feature in the very near future.

KyleJamesWalker:
@thboop We're using public runners, not sure if it matters any but we are also on GitHub enterprise.

KyleJamesWalker:
In case anyone looks at this later it is now fully working as expected: https://github.com/actions/runner/issues/646#issuecomment-901336347

PR #1258, opened 2021-08-12
PowerShell secret masking

fhammerl:


src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -80,6 +81,24 @@ public static String TrimDoubleQuotes(String value)
             return trimmed;
         }
 
+        public static String RemovePowerShellSpecialCharacters(String value)
+        {
+            var trimmed = string.Empty;
+            if (!string.IsNullOrEmpty(value))
+            {
+                // Match any number of "+&;" characters and one extra character
+                Regex pwshSpecialChars = new Regex(@"^[+&;]*.", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
thboop, line 90:
```
^[+&;]+.
```
I think you want a plus instead of a *

thboop: Might be good to consult a guide on special characters for powershell and make sure to test them all in relation to &

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -80,6 +81,65 @@ public static String TrimDoubleQuotes(String value)
...
+                if (secretSection.Length >= 6)
+                {
+                    trimmed = secretSection;
+                }
+            }
+
+            return trimmed;
+        }
+
+        public static String PowerShellPostAmpersandPlus(String value)
```
thboop, line 118: Nit consider: `PowerShellPostAmpersandEscape` and `PowerShellPreAmpersandEscape`

Plus may or may not be in the string

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -80,6 +81,65 @@ public static String TrimDoubleQuotes(String value)
...
 
+        public static String PowerShellPreAmpersandPlus(String value)
+        {
+            // if the secret is passed to PS as a command and it causes an error, sections of it can be surrounded by color codes
+            // or printed individually. 
+            
+            // The secret secretpart1&secretpart2&secretpart3 would be split into 2 sections:
+            // 'secretpart1&secretpart2&' and 'secretpart3'. This method masks for the first section.
+            
+            // The secret secretpart1&+secretpart2&secretpart3 would be split into 2 sections:
```
thboop, line 92: great comments 👍

thboop: Is this behavior the same for powershell and powershell core (pwsh)

src/Sdk/DTLogging/Logging/ValueEncoders.cs
```diff
@@ -80,6 +81,65 @@ public static String TrimDoubleQuotes(String value)
...
+
+            var trimmed = string.Empty;
+            if (!string.IsNullOrEmpty(value) && value.Contains("&"))
+            {
+                var secretSection = string.Empty;
+                if (value.Contains("&+"))
+                {
+                    secretSection = value.Substring(0, value.IndexOf("&+") + "&+".Length);
+                }
+                else
```
thboop, line 103:
```
else if (value.Contains("&") 
{

}
```
Is a little cleaner. There isn't a functional different here due to line 109, but this sets secretSection equal to the first character for strings not containing "&"

thboop (dismissed):
Clean, minor thoughts. May want to verify powershell and powershell core work the same before merging.

src/Sdk/DTLogging/Logging/ValueEncoders.cs
fhammerl: When I tested it in powershell, it used a single color (red) for the entire error message and split the string the same way.

Issue #1268, opened 2021-08-18
incorrect runner_status in delete script for repo with multiple runners


nhsieh:
**Describe the bug**
With a repo with multiple runners, the delete script will be unable to find the correct runner and always return `fatal "Runner should be offline before removing"`.

**To Reproduce**
1. Have multiple runners in a repo
2. Run '.RUNNER_CFG_PAT=<PAT> ./delete.sh myuser/myrepo myname'
3. See error `error: Runner should be offline before removing`

**Expected behavior**
Not see an error and remove the runner.

## Runner Version and Platform
Version of your runner? v2.280.2

OS of the machine running the runner? Linux

## What's not working?
Terminal output:
```
RUNNER_CFG_PAT=${PAT} ./delete.sh ${repo} ${runner_name}
Status: nullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullnullonline
error: Runner should be offline before removing
```

nikola-jokic:
Hey @nhsieh, did you stop the runner first?

PR #1269, opened 2021-08-18
ignore empty arrays in jq select

nhsieh:
FIx for https://github.com/actions/runner/issues/1268


scripts/delete.sh
```diff
@@ -51,7 +51,7 @@ fi
 # Ensure offline
 #--------------------------------------
 runner_status=$(curl -s -X GET ${base_api_url}/${runner_scope}/actions/runners?per_page=100  -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" \
-        | jq -M -j ".runners | .[] | [select(.name == \"${runner_name}\")] | .[0].status")
+        | jq -M -j ".runners | .[] | [select(.name == \"${runner_name}\")] | select(length > 0) | .[0].status")
```
fhammerl, line 54: Hi @nhsieh, thank you very much for this PR!

Runner names are guaranteed to be unique within each repo. I don't know `jq` syntax super well and I don't see a reason to wrap the result of the select in an array like so: `[select(.name == \"${runner_name}\")]`.

What do you think of this instead?
```
| jq -M -j ".runners | .[] | select(.name == \"${runner_name}\") | .status")
```

nhsieh: Yes @fhammerl, that works better. I've pushed an update. Thank you for the review!

fhammerl (approved):
LGTM

PR #1270, opened 2021-08-18
Add extra env for the Token log-in action is going to use to request ID_TOKEN

TingluoHuang:
`ACTIONS_ID_TOKEN_REQUEST_TOKEN` currently just point to `ACTIONS_RUNTIME_TOKEN`.
With this extra indirection, we are able to change which token to use in the future, ex: switch to `GITHUB_TOKEN`, without breaking any existing customers.

https://github.com/github/c2c-actions-runtime/issues/1365


thboop (approved):
LGTM

PR #1273, opened 2021-08-18
For Main Steps, just run the step, don't check condition

### LINKED ISSUE #1271: Composite Actions: Does not respect if: `failure()` condition
**Describe the bug**
From the documentation of [failure()](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#failure), the function will return `true` if any previous step of a job fails. The step where this function is used should be executed.

It appears that using composite actions (with the `uses` key) does not work when  `if: ${{ failure() }}` is used in the step. When the action is a JavaScript Action, the action is executed as expected.  If a `run` key is used in that step instead, that works as expected as well.

**To Reproduce**
- Create a step that fails
- Add a later composite step that runs on `if: failure()`
- Notice the composite steps do not execute as expected.

**Expected behavior**
Composite steps should execute.

thboop:
Resolves #1271 

We made a change to evaluate conditions for pre/post steps for composite actions. We also evalute these conditions for main steps, but it always defaults to `success`. This works well for those pre/post steps, as the default is `always()`. However, for the main steps, if you try to run a composite step on `failure` or `cancelled`, success() will always evaluate to false, thus not running the step. This change restores the old functionality, always running main steps if the parent's `if` condition succeeded.

### DIFF
--- a/src/Runner.Worker/Handlers/CompositeActionHandler.cs
+++ b/src/Runner.Worker/Handlers/CompositeActionHandler.cs
@@ -294,99 +294,110 @@ namespace GitHub.Runner.Worker.Handlers
                 // Register Callback
                 CancellationTokenRegistration? jobCancelRegister = null;
                 try
-                {
-                    // Register job cancellation call back only if job cancellation token not been fire before each step run
-                    if (!ExecutionContext.Root.CancellationToken.IsCancellationRequested)
+                {   
+                    // For main steps just run the action
+                    if (stage == ActionRunStage.Main)
                     {
-                        // Test the condition again. The job was canceled after the condition was originally evaluated.
-                        jobCancelRegister = ExecutionContext.Root.CancellationToken.Register(() =>
                         {
-                            // Mark job as cancelled
-                            ExecutionContext.Root.Result = TaskResult.Canceled;
-                            ExecutionContext.Root.JobContext.Status = ExecutionContext.Root.Result?.ToActionResult();
+                            await RunStepAsync(step);
+                        }
+                    }
+                    // We need to evaluate conditions for pre/post steps
+                    else
+                    {
+                        // Register job cancellation call back only if job cancellation token not been fire before each step run
+                        if (!ExecutionContext.Root.CancellationToken.IsCancellationRequested)
+                        {
+                            // Test the condition again. The job was canceled after the condition was originally evaluated.
+                            jobCancelRegister = ExecutionContext.Root.CancellationToken.Register(() =>
+                            {
+                                // Mark job as cancelled
+                                ExecutionContext.Root.Result = TaskResult.Canceled;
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -295,98 +295,109 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps, ActionRunStage stage
                 CancellationTokenRegistration? jobCancelRegister = null;
                 try
                 {
-                    // Register job cancellation call back only if job cancellation token not been fire before each step run
-                    if (!ExecutionContext.Root.CancellationToken.IsCancellationRequested)
+                    // For main steps just run the action
+                    if (stage == ActionRunStage.Main)
```
thboop, line 299: This just restores the old behavior for main steps found here: https://github.com/actions/runner/blob/419ed24c1ee8c70eeb096570980e700e833f81f6/src/Runner.Worker/Handlers/CompositeActionHandler.cs#L194

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -295,98 +295,109 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps, ActionRunStage stage
                 CancellationTokenRegistration? jobCancelRegister = null;
                 try
                 {
```
thboop, line 297: Much easier to view this pr with whitespace disabled.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -295,98 +295,109 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps, ActionRunStage stage
...
-                            var conditionReTestTraceWriter = new ConditionTraceWriter(Trace, null); // host tracing only
-                            var conditionReTestResult = false;
-                            if (HostContext.RunnerShutdownToken.IsCancellationRequested)
-                            {
-                                step.ExecutionContext.Debug($"Skip Re-evaluate condition on runner shutdown.");
-                            }
-                            else
+                            await RunStepAsync(step);
+                        }
+                    }
```
TingluoHuang, line 304: do we need?
```
{
  {
       await ...
  }
}
```

thboop: good catch thank you

TingluoHuang (approved):

src/Runner.Worker/Handlers/CompositeActionHandler.cs
ericsciple: discussed offline. When we support condition on Main in the future, we will need to register the cancel-callback. For now we don't need it.

ericsciple (approved):

PR #1301, opened 2021-08-31
Allow absolute paths with uses:

heyvito:
This updates the mechanism responsible for validating the `Uses:` rule to allow actions to be provided with an absolute path along with relative paths.

We are currently replacing our old CI with GitHub Actions here at @Gympass, and we decided to include internal actions in our base runner image, which is executed by our self-hosted instances. Those images live in a folder in the root path, say `/gympass/actions`.
A workaround we found is to provide `Uses:` with a relative path to the root mountpoint, by passing several `../` before the action name. Merging this PR would allow us to provide the absolute path to our actions directory without resorting to use the hack I mentioned above.

Reviewers: I'm not sure how to write a test case for this scenario. In case this is a valid change for this application, and you intend to merge this PR, could you kindly provide instructions on how to write it?

Thank you!

### DIFF
--- a/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
+++ b/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
@@ -509,7 +509,7 @@ namespace GitHub.DistributedTask.Pipelines.ObjectTemplating
                     var image = uses.Value.Substring("docker://".Length);
                     result.Reference = new ContainerRegistryReference { Image = image };
                 }
-                else if (uses.Value.StartsWith("./") || uses.Value.StartsWith(".\\"))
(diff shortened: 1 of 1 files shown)


TingluoHuang:
I don't think we want to allow an absolute path for actions lookup, since it means we can't enforce the Actions policy you set on the server-side.
I think the right approach would be to support private actions in general.
So you can put all your internal actions into some private repository and via some configuration, you provide the credential you want to use at runtime.

@jenschelkopf might have an idea about the roadmap for support private actions.

heyvito:
Hey @TingluoHuang! Thank you for the prompt response!
I wasn't aware of actions policy, I'm sorry. I didn't mean to circumvent any policies here.

I'll close this one, but I would really like to know whether @jenschelkopf has any news regarding private actions 🌻 
For now I'll just port our action to Go and leave a binary in the base image.

Thanks!

PR #1302, opened 2021-09-01
Add attempt number to GitHub context

thyeggman:
As we work to support reruns on GitHub, we can expose the number of reruns for a given workflow.  Once it is exposed it will be easy to determine if a workflow run is a rerun or not.

### DIFF
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -32,6 +32,7 @@ namespace GitHub.Runner.Worker
             "sha",
             "workflow",
             "workspace",
+            "attempt",
         };
 
         public IEnumerable<KeyValuePair<string, string>> GetRuntimeEnvironmentVariables()
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -32,7 +32,7 @@ namespace GitHub.Runner.Worker
             "sha",
             "workflow",
             "workspace",
-            "attempt",
+            "run_attempt",
         };
 
         public IEnumerable<KeyValuePair<string, string>> GetRuntimeEnvironmentVariables()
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/GitHubContext.cs
```diff
@@ -32,6 +32,7 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "sha",
             "workflow",
             "workspace",
+            "attempt",
```
TingluoHuang, line 35: what's the value look like for `GITHUB_ATTEMPT`? `bool`/`int`/`string`

thyeggman: `int` - is there a change needed elsewhere?

TingluoHuang: We might need to change line 42 since it checks `data.Value is StringContextData`, I forgot whether an `int` will be sent down as String or not. That's the reason I ask you whether you get it working E2E in the dev environment.

thyeggman: Yeah I'm still working on that.  I'm struggling to get all three different repos to the right state without everything falling apart along the way 😞

src/Runner.Worker/GitHubContext.cs
```diff
@@ -32,7 +32,7 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "sha",
             "workflow",
             "workspace",
-            "attempt",
+            "run_attempt",
```
TingluoHuang, line 35: move to line 29?

TingluoHuang (approved):

thyeggman:
@actions/actions-runtime Do I need to do anything to get this in a deployment?  When is the next deployment planned?

Issue #899, opened 2021-01-08
Access runner name for self-hosted runners


ladislas:
We already have `${{ runner.os }}`, it would be great to also have `${{ runner.name }}` for self-hosted runners.

We have Mac laptop serving as our CI server with 2 runners running in parallel. We cache our builds but have run into issues when configuration was done on one runner and the build is done on another (our tools complain about path not being the same).

We've solved the issue by adding the runner name to our cache key.

But the workaround is ugly: `$(basename $(dirname $(dirname $(dirname $(pwd)))))`

It would be easier to just be able to access `${{ runner.name }}` to return for example `my_build_runner1`.

ladislas:
Well thanks for that! :)

NicklasWallgren:
@fhammerl How do one check if a workflow is running on a self-hosted runner?

Some sort of boolean, or access to the `runner group name` would have been handy. 

Maybe `${{ runner.group }}`.

PR #1312, opened 2021-09-07
Add name to runner context

fhammerl:
Add name and runner_group to context
Closes https://github.com/actions/runner/issues/899

Must update[ docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context)

The workaround used to be ` $(basename $(dirname $(dirname $(dirname $(pwd)))))` to access runner name.

Haven't considered if exposing this info is right / useful for hosted runners, thoughts?


src/Runner.Worker/JobRunner.cs
```diff
@@ -106,6 +106,10 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
 
                 jobContext.SetRunnerContext("os", VarUtil.OS);
 
+                var runnerSettings = HostContext.GetService<IConfigurationStore>().GetSettings();
```
TingluoHuang, line 109: we might need to re-fetch these info from the server since the runner/group can be renamed/moved via API or UI.

fhammerl: Right, I was just looking back at our earlier talk about this. I thought we fetch this info from `.runner` and I was gonna check if that updates to reflect changes in runner_group and name.

I'm gonna check how and when it changes and see where it's best to re-fetch.

TingluoHuang: We can just fetch them at the beginning of the job.

TingluoHuang: or do it on the listener side and updates the `.runner` file

TingluoHuang: When we `RenewJobRequestAsync` on the listener side, we do get the latest runner name back, we might need to update the server to return the runner group back as well, so we can avoid another round trip to server.

fhammerl: Updating `.runner` seems like the right way. I did see the problem with runner_groups, but can you also change runner names?

TingluoHuang: technically, you can via REST api.

fhammerl: I could only find `GET` and `DELETE` for the runner.
Also, I removed adding the `runner_group` from this PR.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -767,6 +768,16 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
             }
         }
 
+        private void UpdateAgentNameIfNeeded(string agentName)
```
fhammerl, line 771: @TingluoHuang What do you think about updating `.runner` like this?

fhammerl:
WIP on the failing L0s

src/Runner.Listener/JobDispatcher.cs
TingluoHuang: Let's add a try-catch around the updates, just in case we hit any file permission error, we will continue running the job.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -661,6 +661,7 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
                 try
                 {
                     request = await runnerServer.RenewAgentRequestAsync(poolId, requestId, lockToken, orchestrationId, token);
+                    UpdateAgentNameIfNeeded(request.ReservedAgent?.Name);
```
TingluoHuang, line 664: can we also move this to line 670 so we will only check the `.runner` file once per-job.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -767,6 +768,16 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
             }
         }
 
+        private void UpdateAgentNameIfNeeded(string agentName)
+        {
+            var configurationStore = HostContext.GetService<IConfigurationStore>();
```
TingluoHuang, line 773: maybe just create a global `_configurationStore` in the `Initialize(IHostContext hostContext)` (we already have `_runnerSetting`), so we don't have to pull it out everytime.

fhammerl: Right, we can rely on `_runnerSettings` and avoid another full `File.ReadAllText`. Also added `_configurationStore` as a prop 👍.

src/Runner.Listener/JobDispatcher.cs
fhammerl: Sure, it's always a good idea.

When I was writing it, I thought the permission issues are solved by deleting it first?

https://github.com/actions/runner/blob/aab936d08143dca39c6e314ec21b89f05fdbb6e1/src/Runner.Listener/SelfUpdater.cs#L364-L368

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -767,6 +771,25 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
...
+                if (_runnerSettings.AgentName != agentName)
+                {
+                    _runnerSettings.AgentName = agentName;
+                    _configurationStore.SaveSettings(_runnerSettings);
+                }
+
+            }
+            catch (IOException ex)
+            {
+                // Warning, not an Error, because the job execution continues
```
thboop, line 787: nit: is this comment from copy paste?

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -767,6 +771,25 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
             }
         }
 
+        private void UpdateAgentNameIfNeeded(string agentName)
+        {
+            try
+            {
+                if (_runnerSettings.AgentName != agentName)
```
thboop, line 778: prefer `String.Equals` here, we can provide the default StringComparison.Ordinal just to be more explicit

src/Runner.Listener/JobDispatcher.cs
fhammerl: No, I forgot to clean it up - I used to log an exception with `Trace.Warning` and wanted to leave a comment to explain why.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -767,6 +771,24 @@ public async Task RenewJobRequestAsync(int poolId, long requestId, Guid lockToke
             }
         }
 
+        private void UpdateAgentNameIfNeeded(string agentName)
+        {
+            try
+            {
+                if (string.Equals(_runnerSettings.AgentName, agentName, StringComparison.Ordinal))
```
thboop, line 778: Is this if condition correct? Shouldn't it be flipped?

thboop (approved):
LGTM

thboop (approved):
LGTM

PR #1316, opened 2021-09-08
More resilient `VssConnection` client retries in `JobServer`

juliobbv:
Re-create `VssConnection` object with each retry for resiliency. This follows an existing pattern that's used in `RunnerServer` that has been working well so far.

### DIFF
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -2,16 +2,19 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using GitHub.Runner.Sdk;
 using GitHub.Services.WebApi;
+using GitHub.Services.Common;
 
 namespace GitHub.Runner.Common
 {
     [ServiceLocator(Default = typeof(JobServer))]
     public interface IJobServer : IRunnerService
     {
-        Task ConnectAsync(VssConnection jobConnection);
+        Task ConnectAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler = null);
 
         // logging and console
         Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName, Guid planId, int logId, Stream uploadStream, CancellationToken cancellationToken);
@@ -32,20 +35,21 @@ namespace GitHub.Runner.Common
         private VssConnection _connection;
         private TaskHttpClient _taskClient;
 
-        public async Task ConnectAsync(VssConnection jobConnection)
+        public async Task ConnectAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler = null)
         {
-            _connection = jobConnection;
+            Trace.Info($"Establishing connection for JobServer");
             int attemptCount = 5;
-            while (!_connection.HasAuthenticated && attemptCount-- > 0)
+
+            while (attemptCount-- > 0)
             {
                 try
                 {
-                    await _connection.ConnectAsync();
+                    await RefreshConnectionAsync(jobServerUrl, jobServerCredential, delegatingHandler);
                     break;
                 }
                 catch (Exception ex) when (attemptCount > 0)
                 {
-                    Trace.Info($"Catch exception during connect. {attemptCount} attemp left.");
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempts left.");
                     Trace.Error(ex);
                 }
 
@@ -53,6 +57,15 @@ namespace GitHub.Runner.Common
             }
 
             _taskClient = _connection.GetClient<TaskHttpClient>();
+        }
+
+        private async Task RefreshConnectionAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler)
+        {
+            Trace.Info($"Refresh JobServer VssConnection to get on a different AFD node.");
+            _hasConnection = false;
+            _connection?.Dispose();
+            _connection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, delegatingHandler);
+            await _connection.ConnectAsync();
             _hasConnection = true;
         }
 
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -510,9 +510,8 @@ namespace GitHub.Runner.Listener
 
                                     var jobServer = HostContext.GetService<IJobServer>();
                                     VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
-                                    VssConnection jobConnection = VssUtil.CreateConnection(systemConnection.Url, jobServerCredential);
-                                    await jobServer.ConnectAsync(jobConnection);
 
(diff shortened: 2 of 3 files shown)


src/Runner.Common/JobServer.cs
```diff
@@ -2,16 +2,19 @@
...
 using GitHub.Services.WebApi;
+using GitHub.Services.Common;
 
 namespace GitHub.Runner.Common
 {
     [ServiceLocator(Default = typeof(JobServer))]
     public interface IJobServer : IRunnerService
     {
-        Task ConnectAsync(VssConnection jobConnection);
+        Task ConnectAsync(Uri jobServerUrl, VssCredentials jobServerCredential, IThrottlingReporter jobServerQueue = null);
```
TingluoHuang, line 17: rename `jobServerQueue` to `throttlingReporter`

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +60,37 @@ public async Task ConnectAsync(VssConnection jobConnection)
             _hasConnection = true;
         }
 
+        // Refresh connection is best effort. it should never throw exception
+        private void RefreshConnection(Uri jobServerUrl, VssCredentials jobServerCredential, IThrottlingReporter jobServerQueue = null)
```
TingluoHuang, line 64: rename `jobServerQueue`

src/Runner.Common/JobServer.cs
TingluoHuang: wonder whether `ConnectAsync` should just take the optional `DelegatingHandler[]` instead of the `IThrottlingReporter`?

src/Runner.Common/JobServer.cs
```diff
@@ -32,20 +35,21 @@ public sealed class JobServer : RunnerService, IJobServer
...
             int attemptCount = 5;
-            while (!_connection.HasAuthenticated && attemptCount-- > 0)
+
+            while (attemptCount-- > 0)
             {
                 try
                 {
-                    await _connection.ConnectAsync();
+                    RefreshConnection(jobServerUrl, jobServerCredential, jobServerQueue);
                     break;
```
TingluoHuang, line 48: i think we still need to call `await _connection.ConnectAsync();`

src/Runner.Common/JobServer.cs
TingluoHuang: maybe take `DelegatingHandler[]` as parameter

src/Runner.Common/JobServer.cs
TingluoHuang: or you can make `RefreshConnection` to call `_connection.ConnectAsync()`, we need the client to hit the server.
only `_connection.ConnectAsync()` and `_connection.GetClient<TaskHttpClient>();` can do that.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +60,32 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
+        // Refresh connection is best effort. it should never throw exception
+        private async Task RefreshConnectionAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler)
+        {
+            Trace.Info($"Refresh JobServer VssConnection to get on a different AFD node.");
+
+            VssConnection newConnection = null;
+
+            try
+            {
+                _hasConnection = false;
```
TingluoHuang, line 72: i think we don't need to change `_hasConnection` in the `RefreshConnectionAsync` since it will only get set to `true` in line 60 after we get a connection up and running.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +60,32 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
+            }
+            catch (Exception ex)
+            {
+                Trace.Error($"Catch exception during resetting JobServer connection.");
+                Trace.Error(ex);
+                newConnection?.Dispose();
+            }
+            finally
+            {
+                _hasConnection = true;
```
TingluoHuang, line 85: i think we can remove this as well.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +60,32 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
+            VssConnection newConnection = null;
+
+            try
+            {
+                _hasConnection = false;
+                newConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, delegatingHandler);
+                await newConnection.ConnectAsync();
+                _connection = newConnection;
+            }
+            catch (Exception ex)
```
TingluoHuang, line 77: do we need to catch the exception here? we rely on the exception to throw so `ConnectAsync` can keep retry, I think.

juliobbv: It's not necessary... I was following the pattern that was in `RunnerServer`, but this catch isn't needed here.

src/Runner.Common/JobServer.cs
juliobbv: I ended up removing the other extra `_hasConnection`, to keep the `true` and `false` pair together in one method.

src/Runner.Common/JobServer.cs
```diff
@@ -32,27 +35,40 @@ public sealed class JobServer : RunnerService, IJobServer
...
             }
 
             _taskClient = _connection.GetClient<TaskHttpClient>();
+        }
+
+        private async Task RefreshConnectionAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler)
+        {
+            Trace.Info($"Refresh JobServer VssConnection to get on a different AFD node.");
+
+            VssConnection newConnection = null;
```
TingluoHuang, line 66: maybe just `var newConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, delegatingHandler);`

src/Runner.Common/JobServer.cs
```diff
@@ -32,27 +35,40 @@ public sealed class JobServer : RunnerService, IJobServer
...
+
+            _hasConnection = false;
+            newConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, delegatingHandler);
+            await newConnection.ConnectAsync();
+            _connection = newConnection;
```
TingluoHuang, line 71: do we need to `_connection?.Dispose()` before overwrite it?

src/Runner.Common/JobServer.cs
```diff
@@ -32,27 +35,40 @@ public sealed class JobServer : RunnerService, IJobServer
...
                 await Task.Delay(100);
             }
 
             _taskClient = _connection.GetClient<TaskHttpClient>();
+        }
+
+        private async Task RefreshConnectionAsync(Uri jobServerUrl, VssCredentials jobServerCredential, DelegatingHandler[] delegatingHandler)
+        {
+            Trace.Info($"Refresh JobServer VssConnection to get on a different AFD node.");
+
```
TingluoHuang, line 65: remove these extra lines. 😄

src/Runner.Common/JobServer.cs
juliobbv: Works for me, now that we got rid of the try/catch statement.

src/Runner.Common/JobServer.cs
juliobbv: Good catch. I'll add it back.

TingluoHuang (approved):

PR #1320, opened 2021-09-09
Wait for job record updated before running steps.

TingluoHuang:
When requesting an OIDC Id_Token, the service requires the job is in the `in-progress` state.

Since timeline records update is an async background task. you might run into a race condition that we haven't marked the job as `in-progress` when you request an ID_Token.

This is normally happening to the workflow that tries to fetch Id_Token as the first step and also doesn't download any actions.

image

### DIFF
--- a/src/Runner.Worker/JobRunner.cs
+++ b/src/Runner.Worker/JobRunner.cs
@@ -145,10 +145,15 @@ namespace GitHub.Runner.Worker
                 Trace.Verbose($"Job steps: '{string.Join(", ", jobSteps.Select(x => x.DisplayName))}'");
                 HostContext.WritePerfCounter($"WorkerJobInitialized_{message.RequestId.ToString()}");
 
-                // We want to let the job is marked as in-progress before running any customer's steps as much as we can.
-                // Timeline record update background process runs every 500ms, so delay 1000ms is enough for most of the cases
-                Trace.Info($"Waiting for job to be marked as started.");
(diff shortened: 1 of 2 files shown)


thboop (dismissed):
LGTM

Do we have backoff/retries for this in the toolkit as well? Waiting a few seconds then retrying should solve the issues for anyone who isn't covered by this fix.

src/Runner.Worker/JobRunner.cs
```diff
@@ -145,6 +145,11 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
                 Trace.Verbose($"Job steps: '{string.Join(", ", jobSteps.Select(x => x.DisplayName))}'");
                 HostContext.WritePerfCounter($"WorkerJobInitialized_{message.RequestId.ToString()}");
 
+                // We want to let the job is marked as in-progress before running any customer's steps as much as we can.
+                // Timeline record update background process runs every 500ms, so delay 1000ms is enough for most of the cases
+                Trace.Info($"Waiting for job to be marked as started.");
+                await Task.WhenAny(_jobServerQueue.JobRecordUpdated.Task, Task.Delay(1000));
```
cschleiden, line 151: Is that going to delay all job invocations?

TingluoHuang: 🙇  I changed it to only delay jobs that uses OIDC.

thboop (approved):
LGTM

Issue #1337, opened 2021-09-14
`--ephemeral` doesnt "un-configure the runner" as help says


apr-1985:
**Describe the bug**
The help for `--ephemeral` says:
```
Configure the runner to only take one job and then let the service un-configure the runner after the job finishes
```

However, after the job has run you end up in a state where you cannot easily reconfigure the runner, but still need a remove token to run `config.sh remove`.

Ideally I was looking to have the process:
1. register runner with `unattended` and `ephemeral`
2. run `./run.sh`
3. Let actions give it a job
4. Let ephemeral unregister the runner
5. delete the _work folder
6. Go back to step 1

This would give a nice quick loop to being able to clean up runners without a full terminate/spin up new.  While ensuring that only a single job ran before it was cleaned.
It would also improve security by not needing to give the runners Org Admin level tokens to get registration/cleanup tokens as I could inject a registration token into the runner use that with the above flow for an hour, then terminate it and spin up new.

However because `ephemeral` isn't unconfiguring correctly I would need to remove the whole Actions folder and start by un-taring the runner code again which is waste time.

**To Reproduce**
1. configure runner `./config.sh --url https://github.com/MY_ORG --token $TOKEN --ephemeral --unattended  --runnergroup runner_group_1;`
2. let GH give it a job and let ephemeral remove it from GH
3. Delete `_work` and try to run again.  Get the error:
`Failed to create a session. The runner registration has been deleted from the server, please re-configure.` 
4. Run the configure command from step 1 and you get 
`Cannot configure the runner because it is already configured. To reconfigure the runner, run 'config.cmd remove' or './config.sh remove' first.`

5. Try to run `./config.sh remove` and get 
`Enter runner remove token:`

**Expected behavior**
I would expect that `--ephemeral` do a full `config.sh remove` so that the runner could be setup again quickly and easily.

## Runner Version and Platform
Version of your runner? [ 2.282.0](https://github.com/actions/runner/releases/tag/v2.282.0

Amazon Linux 2

pje:
Hey @apr-1985, thanks for the bug report! 

Looks like `--ephemeral` is correctly de-registering the runner on the service, but it's _not_ deleting the local registration files. In particular `.runner` looks like this: 

```
{
  "agentId": 26,
  "agentName": "gimbal",
  "poolId": 1,
  "poolName": "Default",
  "ephemeral": true,
[...]

PR #1344, opened 2021-09-15
Ephemeral runner deletes local .runner,.credentials files after completion

pje:
Currently the `--ephemeral` flag means we "clean up" _remote_ state by de-registering the runner with the actions service after completion. However it doesn't clean up _local_ state (namely the `.runner` registration config file as well as the `.credentials` file), leading to confusion in the form of  https://github.com/actions/runner/issues/1337.

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -22,6 +22,7 @@ namespace GitHub.Runner.Listener.Configuration
         bool IsConfigured();
         Task ConfigureAsync(CommandSettings command);
         Task UnconfigureAsync(CommandSettings command);
+        void DeleteLocalRunnerConfig();
         RunnerSettings LoadSettings();
     }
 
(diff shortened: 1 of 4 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -416,17 +435,7 @@ public async Task UnconfigureAsync(CommandSettings command)
...
-                if (isConfigured)
-                {
-                    _store.DeleteSettings();
-                    _term.WriteSuccessMessage("Removed .runner");
-                }
-                else
-                {
-                    _term.WriteLine("Does not exist. Skipping " + currentAction);
-                }
+                DeleteLocalRunnerConfig();
```
TingluoHuang, line 438: we better include `//delete credential config files` part to this as well since those are also staled info for the deleted runner.

pje: Done in 8e8458b

TingluoHuang (approved):

pje:
Output from a job run looks like this:

```
❍ ./run.sh

√ Connected to GitHub

2021-09-15 00:00:00Z: Listening for Jobs
2021-09-15 00:00:01Z: Running job: test
2021-09-15 00:00:02Z: Job test completed with result: Succeeded
√ Removed .credentials
√ Removed .runner
```

(note the `Removed` output at the end 😎 )

src/Test/L0/Listener/Configuration/ConfigurationManagerL0.cs
```diff
@@ -179,6 +180,7 @@ public async Task CanEnsureConfigure()
                 Assert.True(s.AgentName.Equals(_expectedAgentName));
                 Assert.True(s.PoolId.Equals(_secondRunnerGroupId));
                 Assert.True(s.WorkFolder.Equals(_expectedWorkFolder));
+                Assert.True(s.Ephemeral.Equals(true));
```
pje, line 183: I snuck this in as another ephemeral test case, but it's not strictly related to the goals of this PR.

PR #1360, opened 2021-09-21
Update the comment about the `--once` in Constants.cs

TingluoHuang:
We have no plan to deprecate `--once` at this point.

Updated message:
image

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -235,7 +235,7 @@ namespace GitHub.Runner.Listener
 
                     if (command.RunOnce)
                     {
-                        _term.WriteLine("Warning: '--once' is going to be deprecated in the future, please consider to use '--ephemeral' during runner registration.", ConsoleColor.Yellow);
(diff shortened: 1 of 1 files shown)


thboop (approved):
LGTM, should we remove the error when trying to register a --once runner

TingluoHuang:
I think we should give a special warning about `--once` instead of the generic invalid arg error. i will update the PR.

src/Runner.Listener/Runner.cs
```diff
@@ -233,8 +233,14 @@ public async Task<int> ExecuteCommand(CommandSettings command)
                     Trace.Info($"Set runner startup type - {startType}");
                     HostContext.StartupType = startType;
 
+                    if (command.RunOnce)
+                    {
+                        _term.WriteLine("Warning: '--once' is going to be deprecated in the future, please consider to use '--ephemeral' during runner registration.", ConsoleColor.Yellow);
```
thboop, line 238:
```suggestion
                        _term.WriteLine("Warning: '--once' is going to be deprecated in the future, please consider using '--ephemeral' during runner registration.", ConsoleColor.Yellow);
```

thboop (approved):
lgtm minor nit

thboop (approved):
lgtm

PR #1384, opened 2021-09-28
only cleanup runner local files on success

### LINKED ISSUE #1370: Ephemeral runner does not deregister remotely if interrupted, but does clean up local state
**Describe the bug**
https://github.com/actions/runner/pull/1344 cleans up local state (`.runner` and `.credentials` files) on exit. On an abnormal exit (e.g. if sent SIGINT), the remote state is not cleaned up, but the local state is.

**To Reproduce**
1. Configure the service with `--ephemeral`
2. Interrupt the service process
3. See that the runner is still registered but the local files are gone

**Expected behavior**
Either
1. the runner is deregistered as it would be for a clean exit, or 
2. the local state is not removed so the service can be restarted or deregistered manually with `./config.sh remove`.

1 would be my preference but 2 would be ok too - as long as we have consistency between the runner being registered remotely and the local state existing on disk.

v2.283.1
Linux

## What's not working?
The runner remains registered and cannot be removed with `./config.sh remove`

## Context

We start and stop our build servers frequently, so we need to run in ephemeral mode to avoid jobs being allocated to shutting-down runners. However, we want to restart and reuse the same server later, so it's not truly ephemeral, and this bug means we can't de-register the runner with its remove token on shutdown, and we also can't re-register the runner on startup (because it is still registered).

thboop:
Validated the happy path deletes the runner, and manually canceling runner execution leaves the file so a user can later clean those up.

Resolves #1370

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -316,6 +316,9 @@ namespace GitHub.Runner.Listener
 
                 IJobDispatcher jobDispatcher = null;
                 CancellationTokenSource messageQueueLoopTokenSource = CancellationTokenSource.CreateLinkedTokenSource(HostContext.RunnerShutdownToken);
+                
+                // Should we try to cleanup ephemeral runners
+                bool runOnceJobCompleted = false;
                 try
                 {
                     var notification = HostContext.GetService<IJobNotification>();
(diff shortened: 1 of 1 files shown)


hross (approved):

TingluoHuang (approved):

pje (approved):

hieueuro:
Ok

PR #1385, opened 2021-09-28
Use Actions Service health and api.github.com endpoints after connection failure on Actions Server and Hosted

juliobbv:
Modify the connection process so the server hits the Actions Service `_apis/health` and `api.github.com` endpoints before retrying connection.

This is done for two reasons:
1. Diagnose early runner disconnection issues (on github.com) due to any kind of networking issues
2. Hopefully work-around the issue where the call to initialize the `VssConnection` fails due to a timeout

This is how things look in the runner diag log:

```
[2021-09-30 16:03:19Z INFO JobServer] Requesting Actions Service health endpoint status
[2021-09-30 16:03:19Z INFO JobServer] Actions health status code: OK
[2021-09-30 16:03:19Z INFO JobServer] Requesting Github API endpoint status
[2021-09-30 16:03:19Z INFO JobServer] api.github.com status code: OK
```


src/Runner.Common/JobServer.cs
```diff
@@ -40,19 +41,33 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
-                    Trace.Info($"Catch exception during connect. {attemptCount} attemp left.");
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempts left.");
                     Trace.Error(ex);
                 }
 
                 await Task.Delay(100);
             }
 
             _taskClient = _connection.GetClient<TaskHttpClient>();
+
```
juliobbv, line 70: remove extraneous whitespace

src/Sdk/WebApi/WebApi/HttpClients/HealthHttpClient.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+using GitHub.Services.Common;
+using GitHub.Services.WebApi;       
+
+namespace GitHub.Services.Health.Client
+{     
+    public class HealthHttpClient : VssHttpClientBase
```
TingluoHuang, line 9: Is this something you made up or it's copied from ADN?
`/src/sdk/*` are codes we copied from ADN.

src/Runner.Common/JobServer.cs
```diff
@@ -40,19 +41,33 @@ public async Task ConnectAsync(VssConnection jobConnection)
             {
                 try
                 {
+                    try
+                    {
+                        Trace.Info("Requesting Actions health endpoint status");
+                        // Construct the HttpClient object directly, without going through GetClient<T>
+                        // as we haven't properly connected to the server at this point
+                        var client = new HealthHttpClient(_connection.Uri, _connection.InnerHandler, false);
```
TingluoHuang, line 49: I would move this into the exception catch.
If we fail to call `ConnectAsync()`, let's try the health endpoint.
Also, can we make a request to github.com as well?

src/Sdk/WebApi/WebApi/HttpClients/HealthHttpClient.cs
juliobbv: It's actually code from ADO! ADN removed that client code interestingly.

src/Runner.Common/JobServer.cs
```diff
@@ -40,13 +43,26 @@ public async Task ConnectAsync(VssConnection jobConnection)
             {
                 try
                 {
+                    if (attemptCount == 4)
```
juliobbv, line 46: Remove this temp code.

src/Runner.Common/JobServer.cs
```diff
@@ -45,8 +48,16 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
                 catch (Exception ex) when (attemptCount > 0)
                 {
-                    Trace.Info($"Catch exception during connect. {attemptCount} attemp left.");
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempts left.");
                     Trace.Error(ex);
+
+                    var configurationStore = HostContext.GetService<IConfigurationStore>();
+                    var runnerSettings = configurationStore.GetSettings();
+
+                    if (runnerSettings.IsHostedServer)
```
TingluoHuang, line 57: I am worry about Mms is not populate this setting.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +67,40 @@ public async Task ConnectAsync(VssConnection jobConnection)
             _hasConnection = true;
         }
 
+        private async Task CheckHealthEndpointsAsync()
```
TingluoHuang, line 70: Since we do not only make requests to the health endpoint, I would suggest renaming this to something like `CheckNetworkHealthAsync`

src/Sdk/WebApi/WebApi/HttpClients/HealthHttpClient.cs
TingluoHuang: We shouldn't add anything to the SDK folder if the code is not sourced in ADN.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +67,40 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
+        {
+            try
+            {
+                Trace.Info("Requesting Actions Service health endpoint status");
+                // Construct the HttpClient object directly, without going through GetClient<T>
+                // as we haven't properly connected to the server at this point
+                var baseUri = new Uri(_connection.Uri.GetLeftPart(UriPartial.Authority));
+                var client = new HealthHttpClient(baseUri, _connection.InnerHandler, false);
```
TingluoHuang, line 78: I would suggest just create `HttpClient` inline as you did for the github api endpoint.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +67,40 @@ public async Task ConnectAsync(VssConnection jobConnection)
             _hasConnection = true;
         }
 
+        private async Task CheckHealthEndpointsAsync()
+        {
+            try
```
TingluoHuang, line 72: I would break these into 2 try-catch.
- try-catch _apis/health
- try-catch api.github.com
So we can see whether a particular service is having the issue.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +67,40 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
+
+                // Call the _apis/health endpoint
+                await client.GetHealthAsync();
+
+                Trace.Info("Requesting Github API endpoint status");
+                // This is a dotcom public API... just call it directly
+                using (var httpClientHandler = HostContext.CreateHttpClientHandler())
+                using (var gitHubClient = new HttpClient(httpClientHandler))
+                {
+                    gitHubClient.BaseAddress = new Uri("https://api.github.com");
```
TingluoHuang, line 88: i would not set the `BaseAddress`, but `gitHubClient.GetAsync("https://api.github.com")`

src/Runner.Common/JobServer.cs
juliobbv: I see. Do we have an alternative way to determine whether it's a hosted server? Should I just go with the checking the URI directly for a matched prefix?

hieueuro:
Hieueuro

src/Runner.Common/JobServer.cs
TingluoHuang: no worry, I just fix the service code.

src/Sdk/WebApi/WebApi/HttpClients/HealthHttpClient.cs
juliobbv: Removed the class entirely, and calling the health endpoint directly

src/Runner.Common/JobServer.cs
juliobbv: Ah, I missed this comment. I changed it to use `UrlUtil.IsHostedServer())`.

src/Runner.Common/JobServer.cs
```diff
@@ -45,8 +48,17 @@ public async Task ConnectAsync(VssConnection jobConnection)
                 }
                 catch (Exception ex) when (attemptCount > 0)
                 {
-                    Trace.Info($"Catch exception during connect. {attemptCount} attemp left.");
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempts left.");
                     Trace.Error(ex);
+
+                    var configurationStore = HostContext.GetService<IConfigurationStore>();
```
TingluoHuang, line 54: you might want to cache the config in the outer loop, so we don't need to read the file on every retry.

src/Runner.Common/JobServer.cs
```diff
@@ -56,6 +68,50 @@ public async Task ConnectAsync(VssConnection jobConnection)
...
         }
 
+        private async Task CheckNetworkEndpointsAsync()
+        {
+            try
+            {
+                Trace.Info("Requesting Actions Service health endpoint status");
+                using (var httpClientHandler = HostContext.CreateHttpClientHandler())
+                using (var actionsClient = new HttpClient(httpClientHandler))
+                {
```
TingluoHuang, line 78: do we want to set useragent for this request as well?

juliobbv: Setting the user agent isn't required for ADN (unlike for GitHub), but yeah it'll be helpful in Kusto.

TingluoHuang (approved):

PR #1381, opened 2021-09-28
Handle keeping previous OSX versions more smoothly on Mac

thboop:
Outstanding questions
- Should we add back in deleting old runner versions in this pr or the next one?
- Need to validate these commands are all available in the base images of osx, they have man pages so they should be

PR #1410, opened 2021-10-11
Cleanup Older versions on MacOS now that we recreate node versions as needed

thboop:
Follow up to https://github.com/actions/runner/pull/1381

We originally kept a bunch of older versions of mac runners, as OSX's gatekeeper would prevent us from spawning new processes if we deleted the original process on disk. Now, we find out what version was running and dynamically recreate it, so we only need to keep one older version. This PR now readds the behavior that deletes old runner versions on upgrade, and adds telemetry for upgrading ephemeral runners as they don't use node so they don't need to recreate old versions.

### DIFF
--- a/src/Misc/layoutbin/update.sh.template
+++ b/src/Misc/layoutbin/update.sh.template
@@ -161,60 +161,14 @@ if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
             date "+[%F %T-%4N] DarwinRunnerUpgrade: Failed to find runner path. Path: $path, pgid: $procgroup, root: $rootfolder" >> "$telemetryfile" 2>&1
         fi
     else
-            date "+[%F %T-%4N] DarwinRunnerUpgrade: Failed to find runner pgid. pgid: $procgroup, root: $rootfolder" >> "$logfile" 2>&1
-            date "+[%F %T-%4N] DarwinRunnerUpgrade: Failed to find runner pgid. pgid: $procgroup, root: $rootfolder" >> "$telemetryfile" 2>&1
-    fi
-
-    if [ $attemptedtargetedfix -eq 0 ]
-    then
-
-        date "+[%F %T-%4N] DarwinRunnerUpgrade: Defaulting to old macOS service fix" >> "$logfile" 2>&1
-        date "+[%F %T-%4N] DarwinRunnerUpgrade: Defaulting to old macOS service fix" >> "$telemetryfile" 2>&1
-        if [[ ! -e "$rootfolder/externals.2.280.3/node12/bin/node" ]]
-        then
-            mkdir -p "$rootfolder/externals.2.280.3/node12/bin"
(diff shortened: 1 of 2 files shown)


src/Misc/layoutbin/update.sh.template
```diff
@@ -161,60 +161,14 @@ if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
...
-        # GHES 3.0
-        if [[ ! -e "$rootfolder/externals.2.273.5/node12/bin/node" ]]
-        then
-            mkdir -p "$rootfolder/externals.2.273.5/node12/bin"
-            cp "$rootfolder/externals/node12/bin/node" "$rootfolder/externals.2.273.5/node12/bin/node"
-        fi
+            runproc=$(ps x -o pgid,command | grep "run.sh" | grep -v grep | awk '{print $1}')
+            if [[ $? -eq 0 && -n "$runproc" ]]
+            then
+                date "+[%F %T-%4N] Running as ephemeral using run.sh, no need to recreate node folder" >> "$logfile" 2>&1
```
TingluoHuang, line 167: how did you decide the runner is an ephemeral runner?

src/Misc/layoutbin/update.sh.template
```diff
@@ -161,60 +161,14 @@ if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
...
-            cp "$rootfolder/externals/node12/bin/node" "$rootfolder/externals.2.276.1/node12/bin/node"
-        fi
-
-        # GHES 3.0
-        if [[ ! -e "$rootfolder/externals.2.273.5/node12/bin/node" ]]
-        then
-            mkdir -p "$rootfolder/externals.2.273.5/node12/bin"
-            cp "$rootfolder/externals/node12/bin/node" "$rootfolder/externals.2.273.5/node12/bin/node"
-        fi
+            runproc=$(ps x -o pgid,command | grep "run.sh" | grep -v grep | awk '{print $1}')
```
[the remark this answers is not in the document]
thboop: It is, good catch. Shouldn't have an impact but i'll push up a fix

src/Misc/layoutbin/update.sh.template
thboop: We only go into this big block if `restartinteractiverunner -eq 0`

https://github.com/actions/runner/pull/1410/files#diff-0642782eaa85c9c6cbc4e90bbd806bec100ca60e765655f435fdb613df42e627R126


restartinteractiverunner is set according to `!(!runOnce && HostContext.StartupType != StartupType.Service)`
So, if its ephemeral or its a service, its 0, otherwise its 1.

So if we get into this block and can't find the `runsvc.sh`, well, we aren't running as a service, check if we see a `run.sh`. If so, we are running an ephemeral (or once) runner. Otherwise, something went wrong and we want the telemetry because we can't tell what type of runner we have.

Issue #1409, opened 2021-10-07
Show version in the Logs


arash-bizcover:
As the runner **updates automatically and forcefully**, it would be very helpful to know what exact version is running by looking at the logs.

i.e. this  can be shown here, after printing the banner:
![image](https://user-images.githubusercontent.com/62287461/136473188-945c1765-428d-474d-bad3-8fb9787fe745.png)

PR #1441, opened 2021-10-28
Log current runner version in terminal

fhammerl:
Closes https://github.com/actions/runner/issues/1409


src/Runner.Listener/Runner.cs
```diff
@@ -312,6 +312,8 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                 }
 
                 HostContext.WritePerfCounter("SessionCreated");
+
+                _term.WriteLine($"Current runner version: '{BuildConstants.RunnerPackage.Version}'");
```
rgl, line 316: @fhammerl, why did you put the version in single quotes? Can those be removed?

PR #1443, opened 2021-10-28
Allow to optionally send job update with a custom script

fgalind1:
For analytics purposes or when running in kubernetes, it's pretty useful to know when a runner is running a specific job. Therefore this PR adds possibility to configure a custom script that sends an event with some
basic information about the status of the runner and the message of job being ran.

We've integrated this with kubernetes and this is fairly useful to send update back to the cluster about the runner/job information whenever the job starts to run. This can be used for gathering analytics about runner
usage and jobs being ran.

Example when integrating this on kubernetes. This makes it quite easy to broadcast to other environments using custom CRDs, events, or pod statuses. In my case the script is a simple bash script that does some http requests to the k8s API


ethomson:
We definitely want to do this.  This is extremely useful on (at least) two fronts:

* users can emit metrics into some system to understand their job usage
* we can drive information to ARC to avoid unnecessary API calls

I think that there are likely many other uses here.

@TingluoHuang and I have been talking about this and what the API should look like here.  Using an environment variable like this is one option; we could also do something like the git "hooks" mechanism and look for a script in a particular well-known location and invoke that.

Naively, I like the git way of doing things - first, it matches an existing pattern that people are familiar with, and second, it keeps the number of environment variables down.

I wonder if we can reduce the things in the arguments.  Instead of pulling things out of the github context and passing them as an argument, I'd like to just make the github context available to the script.  Otherwise, we'll just end up passing more and more arguments as people want information out of the context.  Instead, scripts can just parse it and pull out exactly what they want.  Are there things that we need to do to enable this?

TingluoHuang:
@ethomson I would be very careful about what we expose the extension scripts.
Things to consider:
- `GitHubContext` is huge since it contains the event payload, you have to pass it over via STDIN.
- Some information in `GitHubContext` is populated by the runner, which means on job start, that information may not get set.
- What about other contexts, ex: `RunnerContext`?

ethomson:
> GitHubContext is huge since it contains the event payload, you have to pass it over via STDIN.

Is this not on disk already though?  🤔  Or is that only the event payload that's on disk?

PR #1461, opened 2021-11-03
Add support to create containers using k8s API/kubectl

fgalind1:
One pre-requisite on this option is a shared storage solution that can be shared as a persistent volume
between the runner pod and the container pod for `externals`, `_work/_temp` and `_work/actions` folders.

It adds a new container provider implementation `KubernetesCommandManager` besides `DockerCommandManager`

Container provider solution can be chosen via `RUNNER_CONTAINER_PROVIDER` environment variable.
If not provided it defaults to "docker". Options allowed: "docker" and "kubernetes"

Persistent volume used for the container pod is taken from `RUNNER_PERSISTENT_VOLUME_CLAIM` environment variable

Example:
```
kubectl get pods -n github-actions
NAME                                                              READY   STATUS    RESTARTS   AGE
...
github.runner-9cgfx-m49b2                            1/1     Running   0          7m1s
github.runner-9cgfx-m49b2-container                  1/1     Running   0          2m53s
github.runner-9cgfx-m49b2-service-1                  1/1     Running   0          2m50s
github.runner-9cgfx-m49b2-service-2                  1/1     Running   0          2m52s
...
```

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -368,6 +368,14 @@ namespace GitHub.Runner.Common
         }
 
         /// <summary>
+        /// Register a default dynamic service.
+        /// </summary>
+        public void RegisterService(Type type, Type target)
+        {
+            _serviceTypes.TryAdd(type, target);
+        }
+
+        /// <summary>
         /// Creates a new instance of T.
         /// </summary>
         public T CreateService<T>() where T : class, IRunnerService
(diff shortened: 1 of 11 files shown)


fgalind1:
@ethomson this is the main abstraction for k8s that we discussed the other day that is working so far

drewboswell:
Holy crap. this would be insane. is this dead ?

PR #1470, opened 2021-11-09
Try to delete `portable-net45+win8` from all projects

TingluoHuang:


src/Runner.Common/Runner.Common.csproj
```diff
@@ -5,7 +5,6 @@
     <OutputType>Library</OutputType>
     <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64</RuntimeIdentifiers>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
-    <AssetTargetFallback>portable-net45+win8</AssetTargetFallback>
```
TingluoHuang, line 8: this is a legacy setting we had a long time ago, seems fine to delete it now.

fhammerl: We're no longer using the package that needed it 👍

fhammerl (approved):

NOUIY (approved):
https://github.com/actions/toolkit/blob/feature/io-update-cp-behavior/package-lock.json

PR #1471, opened 2021-11-09
Bump runtime to dotnet 6

TingluoHuang:
Dotnet 6 has been released and the runner is on dotnet 3.1 which is a 2 years old runtime.

In order to unblock macOS M1 support, I am trying to bump the dotnet runtime in the runner first.

https://github.com/github/c2c-actions-runtime/issues/1523


src/Runner.Common/Runner.Common.csproj
```diff
@@ -1,13 +1,12 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <OutputType>Library</OutputType>
     <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64</RuntimeIdentifiers>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <NoWarn>NU1701;NU1603</NoWarn>
     <Version>$(Version)</Version>
-    <TieredCompilationQuickJit>true</TieredCompilationQuickJit>
```
TingluoHuang, line 10: This is enabled by default since dotnet 5

src/Runner.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -1,4 +1,5 @@
 #if OS_WINDOWS
```
TingluoHuang, line 1: dotnet 6 introduced api platform check at compile time.
We already did our platform check using `#if OS_WINDOWS`, so we can just disable the working everywhere we have `#if OS_WINDOWS`

src/Runner.Listener/Configuration/NativeWindowsServiceHelper.cs
```diff
@@ -1,4 +1,5 @@
...
+#pragma warning disable CA1416
```
paveliak, line 2: Not sure why this pragma is needed. I didn't experience any problems on my branch with these warnings

paveliak: Taking it back. This is within `If Windows` block and I didn't test building on Windows

src/Sdk/Common/Common/Authentication/HttpRequestMessageWrapper.cs
```diff
@@ -32,7 +32,7 @@ public Uri RequestUri
         {
             get
             {
-                return m_request.Properties;
```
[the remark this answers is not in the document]
paveliak: FWIW this code is never used. You could delete these `Properties`

src/Sdk/Common/Common/VssHttpRequestSettings.cs
```diff
@@ -291,12 +291,12 @@ protected internal virtual Boolean IsHostLocal(String hostName)
         protected internal virtual Boolean ApplyTo(HttpRequestMessage request)
         {
             // Make sure we only apply the settings to the request once
-            if (request.Properties.ContainsKey(PropertyName))
+            if (request.Options.TryGetValue(new HttpRequestOptionsKey<VssHttpRequestSettings>(PropertyName), out _))
```
paveliak, line 294: There is some inconsistency in using `TryGetValue` method. In this place you use method defined on the `Options` class which takes `HttpRequestOptionsKey`. While in other places you use extension method from our code which accepts `string`. Most likely some namespaces are not imported in this file so extension method is not visible causing errors. You might consider using one or another approach throughout the code.

src/Sdk/Sdk.csproj
```diff
@@ -10,7 +10,7 @@
         <DefineConstants>TRACE</DefineConstants>
         <LangVersion>7.3</LangVersion>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-        <TieredCompilationQuickJit>true</TieredCompilationQuickJit>
+        <EnableTrimAnalyzer>false</EnableTrimAnalyzer>
```
paveliak, line 13: Trim analyzer is probably not relevant (yet) since you do not compile projects into a single file. And generally disabling it is dangerous

paveliak (approved):
LGTM

src/Sdk/Sdk.csproj
TingluoHuang: this is wrong... I forgot to remove this after I tested something else...

src/Sdk/Common/Common/Authentication/HttpRequestMessageWrapper.cs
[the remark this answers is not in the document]
TingluoHuang: 👍  will remove.

src/Sdk/Common/Common/VssHttpRequestSettings.cs
TingluoHuang: fixed.

src/Sdk/Common/Common/Authentication/HttpRequestMessageWrapper.cs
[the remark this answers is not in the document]
TingluoHuang: can't remove it.
```
error CS0535: 'HttpRequestMessageWrapper' does not implement interface member 'IHttpRequest.Properties'
```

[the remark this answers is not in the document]
paveliak: This is the interface implementation. Just remove it from both places. `IHttpRequest` is our interface (it's not a CLR one)

thboop (approved):
LGTM

dev4dev:
is this blocked? sorry for annoying, but I'm waiting for that as I want to migrate from Jenkins to GHA, and can't move because of the absence of arm support =(

Temtaime:
And then there will be another regression and it will be reverted to netcore, has been

PR #1494, opened 2021-11-18
Make `run.sh|cmd` handle update without quitting so containers using them as entrypoints don't exit on update

fhammerl:
## Context

In the current model, if you start the runner interactively using `run.sh|cmd`, in case there's an update, that process will end and then restart: `run.sh ---> Runner.Listener ---> update.sh ---> run.sh (new version)` This behaviour leads to containers terminating, potentially restarting and ending up in an infinite update loop.

## run.sh|cmd and run-herlper.sh|cmd

We decided to make `run.sh` a long-running process instead, that doesn't exit on update. Runners on containers can now update without the entrypoint exiting. The major drawback is that since we don't restart `run.sh` anymore, servicing the file and its logic is not possible. To alleviate this, I extracted the logic into the `run-helper.sh|cmd` file that we **do** restart on update.

##

On Windows, if we were to swap `run-helper.sh|cmd` as we extract the archive of the new version while it is still executing the 5 second timeout after exiting `Runner.Listener`, the execution would continue from the line it was in the old version to the same line in the new version. This `run-helper.sh|cmd` is generated by `run.sh|cmd` from a template file called `run-helper.sh|cmd.template`, after `run-helper.sh|cmd` is done with its timeout.


src/Runner.Listener/Runner.cs
```diff
@@ -408,7 +408,7 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                                     autoUpdateInProgress = true;
                                     var runnerUpdateMessage = JsonUtility.FromString<AgentRefreshMessage>(message.Body);
                                     var selfUpdater = HostContext.GetService<ISelfUpdater>();
-                                    selfUpdateTask = selfUpdater.SelfUpdate(runnerUpdateMessage, jobDispatcher, !runOnce && HostContext.StartupType != StartupType.Service, HostContext.RunnerShutdownToken);
+                                    selfUpdateTask = selfUpdater.SelfUpdate(runnerUpdateMessage, jobDispatcher, false, HostContext.RunnerShutdownToken);
```
fhammerl, line 411: If I read this right, we will no longer need to `restartInteractiveRunner` to true as we will take care of it in run.sh|cmd. 

If you agree, I'll clean up this argument and the restart depending on it in `update.sh.template`

@TingluoHuang

src/Misc/layoutroot/run.sh
```diff
@@ -11,54 +11,60 @@ fi
...
+    fi
+    
+    returnCode=$?
+    if [[ $returnCode == 0 ]]; then
+        echo "Runner listener exit with 0 return code, stop the service, no retry needed."
+        exit $returnCode
+    elif [[ $returnCode == 1 ]]; then
+        echo "Runner listener exit with terminated error, stop the service, no retry needed."
+        exit $returnCode
+    elif [[ $returnCode == 2 ]]; then
```
fhammerl, line 56: Now we would retry on `$returnCode == 2` in run.sh|cmd as well, just like in RunnerService.js

src/Runner.Listener/Runner.cs
TingluoHuang: consider an upgrade from a runner without all your `run.sh`, I think we need to keep some of the legacy code for a while...

fhammerl: You're right, it's too early to clean up `update.sh.template`.

Let's say there's `current` and `patched` runner versions and ignore ephemeral and service runners for now.

### This is how it works now:
`current` updating to `current.1`:
 - `SelfUpdate` is called with `restartInteractiveRunner=true`
 - `current` `run.sh` exits
 - `current.1` `update.sh` will start `current.1` `run.sh` because `restartInteractiveRunner=true`
`run.sh` exited, `current.1` `run.sh` running

### This is how it would migrate:
`current` updating to `patched`:
 - `SelfUpdate` is called with `restartInteractiveRunner=true` (still running `current` binaries)
 - `current` `run.sh` exits
 - `patched` (and unmodified) `update.sh` will start `patched` `run.sh` because `restartInteractiveRunner=true`

`run.sh` exited, `patched` `run.sh` running

### This is how it would work onward:
`patched` updating to `patched.1`:
 - `SelfUpdate` is called with `restartInteractiveRunner=false` (or clean up the arg)
 - `patched` `run.sh` **never** exits
 - `patched.1` `update.sh` will not start anything

`run.sh` never exits, is never replaced

There is a problem with this:

> `run.sh` never exits, is never replaced

This is fine for most updates (is it?), but how would we deliver changes to `run.sh`?

We could perhaps restart it only if the update brings changes to `run.sh`.

Or with each update, have the old `run.sh` start `run.2.284.1.sh`, then `run.2.284.2.sh`, never stopping the original?

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,43 @@ if defined VERBOSE_ARG (
...
-  rem ********************************************************************************
-  "%~dp0bin\Runner.Listener.exe" run %*
-
-  rem Return code 4 means the run once runner received an update message.
-  rem Sleep 5 seconds to wait for the update process finish and run the runner again.
-  if ERRORLEVEL 4 (
-    timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
-  )
+
```
ericsciple, line 16: regarding the ability to update run.sh/run.cmd....

i wonder whether we should put the bulk of the contents in a separate script and then run.cmd effectively does nothing but `call bin/run-helper.cmd` (in a loop?)

ericsciple: maybe you end up with the same problem, idk

src/Misc/layoutroot/run.sh
```diff
@@ -11,54 +11,60 @@ fi
...
+            ping -c 5 127.0.0.1 > /dev/null
         fi
+    else
+        sleep 5
+    fi
+}
+
+while :;
+do
+    if [[ "$1" == "localRun" ]]; then
```
ericsciple, line 43: i think localrun was deleted

src/Misc/layoutroot/run.sh
```diff
@@ -11,54 +11,60 @@ fi
...
-                    COUNT=$[$COUNT+1]
-                done
-            else
-                ping -c 5 127.0.0.1 > /dev/null
-            fi
+safe_sleep() {
+    if [ ! -x "$(command -v sleep)" ]; then
+        if [ ! -x "$(command -v ping)" ]; then
+            COUNT="0"
+            while [[ $COUNT != 5000 ]]; do
```
ericsciple, line 29: i thought the update script waits longer than 5 seconds... or do we not reach this point until after the update script finishes?

ericsciple: discussed offline, this is so the update script can perform the folder move, etc

ericsciple: ...does that mean run.cmd is still running? wondering if that's a problem since we're moving the folder... thinking...

ok i think windows will let us move a folder even if a file is still in use, so i guess cmd.exe has a handle open to the file and it shouldnt disrupt the execution of the script. might want to test that theory by running a simple cmd script in a loop and moving the folder containing the script.

ericsciple: i wonder if we *could* service it by moving it. still might be an interesting test:

running a simple cmd script that executes a loop and move the folder containing the script

src/Misc/layoutroot/run.cmd
fhammerl: Done that. 

`run.sh|cmd` is now very small and is almost only responsible for running `run-helper.sh|cmd`.

`run.sh|cmd` will potentially be very long running processes, and as such, hard for us to change: changing a running `.cmd` is as [unpredictable](https://stackoverflow.com/questions/906586/changing-a-batch-file-when-its-running) as changing a running [shell script](https://stackoverflow.com/questions/3398258/edit-shell-script-while-its-running)

Before merging this change, we should consider coming up with a way to service these files if we absolutely need to, perhaps an update could force a restart of `run.sh|cmd` through `update.sh`.

Thoughts? @ericsciple @TingluoHuang

src/Misc/layoutroot/run.sh
fhammerl: Modifying `run.cmd` while it's running is super unpredictable as it is being read byte by byte afaik by `cmd.exe`.
If we added a new line, it could be before or after the pointer, if we removed it, `cmd.exe` would see it as 'eof' and terminate `run.cmd`.

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
-    "%~dp0bin\Runner.Listener.exe" run %*
-  )
+rem ********************************************************************************
+rem Run.
+rem ********************************************************************************
+
+:launch_helper
+call "%~dp0bin\run-helper.cmd" %*
+
+rem using `if %ERRORLEVEL% EQU N` insterad of `if ERRORLEVEL N`
```
ericsciple, line 24: sp `instead`

ericsciple: also it looks like the comment is stale

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
-  if ERRORLEVEL 4 (
-    timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
-  )
+rem ********************************************************************************
+rem Run.
+rem ********************************************************************************
+
+:launch_helper
+call "%~dp0bin\run-helper.cmd" %*
```
fhammerl, line 22: I think `call` is the right way to invoke `run-helper.cmd` here, as it allows us to use `exit` in `run-helper.cmd` without terminating `run.cmd`

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
+
+rem using `if %ERRORLEVEL% EQU N` insterad of `if ERRORLEVEL N`
+rem `if ERRORLEVEL N` means: error level is N or MORE
+  
+if %ERRORLEVEL% GEQ 2 (
+  echo "Restart runner after it exited with return code '%ERRORLEVEL%'"
+  goto :launch_helper
```
ericsciple, line 29: should we pass the last exit code to the helper, so it can determine what to do next. Maybe as an env var (so it doesnt interfere with command line args).

ericsciple: discussing offline.... ok maybe if we only restart on exit code 2. The scenarios i'm concerned about is some kind of crash and then we get unexpected exit code.

ericsciple: food for thought

src/Misc/layoutbin/run-helper.cmd
```diff
@@ -0,0 +1,34 @@
+@echo off
```
ericsciple, line 1: hmmm wondering if we move the bin folder, will that disrupt run.cmd

ericsciple: i guess the question is whether we are moving bin while run-helper.cmd is still executing

ericsciple: should this file be powershell instead?

src/Misc/layoutbin/run-helper.sh
```diff
@@ -0,0 +1,55 @@
...
+    else
+        sleep 5
+    fi
+}
+
+bin/Runner.Listener run $*
+returnCode=$?
+if [[ $returnCode == 0 ]]; then
+    echo "Runner listener exit with 0 return code, stop the service, no retry needed."
+    echo "0 ORIG"
```
ericsciple, line 33: dont forget to remove this `echo "0 ORIG"` (looks like testing)

src/Misc/layoutbin/run-helper.sh
```diff
@@ -0,0 +1,55 @@
...
+        fi
+    else
+        sleep 5
+    fi
+}
+
+bin/Runner.Listener run $*
+returnCode=$?
+if [[ $returnCode == 0 ]]; then
+    echo "Runner listener exit with 0 return code, stop the service, no retry needed."
```
ericsciple, line 32: make sure with ting it's safe to echo here (doesnt influence daemon negatively)

src/Misc/layoutbin/run-helper.cmd
```diff
@@ -0,0 +1,34 @@
...
+  echo "Runner listener exit with 0 return code, stop the service, no retry needed."
+)
+
+if %ERRORLEVEL% EQU 1 (
+  echo "Runner listener exit with terminated error, stop the service, no retry needed."
+)
+
+if %ERRORLEVEL% EQU 2 (
+  echo "Runner listener exit with retryable error, re-launch runner in 5 seconds."
+  timeout /t 5 /nobreak > NUL
```
ericsciple, line 18: fyi - `start-sleep -seconds 5` in powershell

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
+) else (  
+  echo "Exit runner after it exited with return code '%ERRORLEVEL%'"
+  exit %ERRORLEVEL%
 )
+exit %ERRORLEVEL%
```
TingluoHuang, line 34: do we need this `exit %ERRORLEVEL%`? will it ever get run?

src/Misc/layoutbin/run-helper.sh
```diff
@@ -0,0 +1,55 @@
...
+elif [[ $returnCode == 4 ]]; then
+    # Sleep 5 seconds to wait for the ephemeral runner update process finish
+    echo "Runner listener exit because of updating, re-launch ephemeral runner in 5 seconds"
+    echo "4 ORIG"
+    safe_sleep
+else
+    echo "Exiting with unknown error code: ${returnCode}"
+fi
+
+exit $returnCode
```
TingluoHuang, line 55: feels like the contract between the `helper.sh` and `run.sh` is not clear.
from what i feel, the `helper.sh` might only want to return 0 or 1.
check with Eric for this.

src/Misc/layoutbin/run-helper.sh
TingluoHuang: echo to stdout should be fine...

src/Misc/layoutbin/run-helper.cmd
```diff
@@ -0,0 +1,34 @@
...
+  echo "Runner listener exit because of updating, re-launch runner in 5 seconds"
+  timeout /t 5 /nobreak > NUL
+)
+
+rem Return code 4 means the ephemeral runner received an update message.
+if %ERRORLEVEL% EQU 4 (
+  echo "Runner listener exit because of updating, re-launch ephemeral runner in 5 seconds"
+  timeout /t 5 /nobreak > NUL
+)
+
```
TingluoHuang, line 33: `run-helper.sh` has this ` echo "Exiting with unknown error code: ${returnCode}"`

src/Misc/layoutbin/run-helper.cmd
ericsciple: (as long as the bin folder isnt the working directory)

src/Misc/layoutbin/run-helper.sh
fhammerl: That would simplify it for sure - 0 for exit, 1 for restart. I felt like `run-helper.sh` giving those returnCodes would conflict with some bash standards (1 for error), but since one's not supposed to run the `helper` in isolation, I suppose it would make sense 👍

src/Misc/layoutroot/run.cmd
fhammerl: So, based on a comment `@TingluoHuang` left, I simplified the contract between `run` and `helper`, so the helper only returns 0 for exit and 1 for restart.

I understand how `helper` now goes agains exit code conventions (1 for error), but since the helper is not meant to be run in isolation, I think it's fine. Lmk what you think @ericsciple

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -54,7 +54,7 @@ public RunnerSettings LoadSettings()
             Trace.Info(nameof(LoadSettings));
             if (!IsConfigured())
             {
-                throw new InvalidOperationException("Not configured. Run config.(sh/cmd) to configure the runner.");
```
fhammerl, line 57: InvalidOperationException is mapped to resultCode = 2, which is retriable error. With the changes above, running `run.sh|cmd` on an unconfigured runner would get stuck in a loop.

I think the runner being unconfigured should not result in retriable error

TingluoHuang: We might want to exam other exceptions to make sure they don't cause a loop.

src/Misc/layoutroot/run-helper.cmd.template
```diff
@@ -0,0 +1,39 @@
+@echo off
+
+"%~dp0\bin\Runner.Listener.exe" run %*
+
+rem using `if %ERRORLEVEL% EQU N` insterad of `if ERRORLEVEL N`
+rem `if ERRORLEVEL N` means: error level is N or MORE
+  
+if %ERRORLEVEL% EQU 0 (
+  echo "Runner listener exit with 0 return code, stop the service, no retry needed."
+exit /b 0
```
TingluoHuang, line 10: space before `exit`?

src/Misc/layoutroot/run-helper.cmd.template
```diff
@@ -0,0 +1,39 @@
...
+)
+
+if %ERRORLEVEL% EQU 1 (
+  echo "Runner listener exit with terminated error, stop the service, no retry needed."
+exit /b 0
```
TingluoHuang, line 15: space before `exit`?

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
+call "%~dp0run-helper.cmd" %*
+
+rem using `if %ERRORLEVEL% EQU N` insterad of `if ERRORLEVEL N`
+rem `if ERRORLEVEL N` means: error level is N or MORE
+  
+if %ERRORLEVEL% EQU 1 (
+  echo "Restarting runner..."
+  goto :launch_helper
+) else (  
+  echo "Exit runner after it exited with return code '%ERRORLEVEL%'"
```
TingluoHuang, line 32: `"Exit runner after it exited with return code '%ERRORLEVEL%'"` might no longer accurate

src/Runner.Listener/Configuration/ConfigurationManager.cs
fhammerl: Perhaps I should remove the `retry on ERR 2` part of the helpers, it wasn't part of the `run.sh|cmd` cycle before, we could skip it for now to limit the scope of this PR.

src/Misc/layoutroot/run.cmd
```diff
@@ -13,21 +13,22 @@ if defined VERBOSE_ARG (
...
-  if ERRORLEVEL 4 (
-    timeout /t 5 /nobreak > NUL
-    "%~dp0bin\Runner.Listener.exe" run %*
-  )
+rem ********************************************************************************
+rem Run.
+rem ********************************************************************************
+
+:launch_helper
+copy run-helper.cmd.template run-helper.cmd /Y
```
TingluoHuang, line 22: why we have `cmd.template` but no `sh.template`? any reason we have to have different way for window and non-windows?

src/Runner.Listener/Runner.cs
TingluoHuang: I think this will work because both `update.cmd` and `update.sh` will restart the runner via `run.sh/cmd` which is going to be the new version you added.

TingluoHuang, line 411: i would get rid of the `false` parameter in a follow-up PR.

src/Misc/layoutroot/run.sh
```diff
@@ -1,64 +1,24 @@
...
-# Run
-shopt -s nocasematch
-if [[ "$1" == "localRun" ]]; then
-    "$DIR"/bin/Runner.Listener $*
-else
-    "$DIR"/bin/Runner.Listener run $*
-
-# Return code 3 means the run once runner received an update message.
-# Sleep 5 seconds to wait for the update process finish
+cp run-helper.sh.template run-helper.sh
```
TingluoHuang, line 12: `cp -f`

src/Runner.Listener/Runner.cs
fhammerl: For sure 👍

src/Misc/layoutroot/run-helper.cmd.template
```diff
@@ -0,0 +1,39 @@
...
+)
+
+if %ERRORLEVEL% EQU 1 (
+  echo "Runner listener exit with terminated error, stop the service, no retry needed."
+  exit /b 0
+)
+
+if %ERRORLEVEL% EQU 2 (
+  echo "Runner listener exit with retryable error, re-launch runner in 5 seconds."
+  ping 1.1.1.1 -n 1 -w 5000 >NUL
```
TingluoHuang, line 20: can we ping `127.0.0.1` instead to avoid reaching external resources?

TingluoHuang (dismissed):

TingluoHuang (approved):

PR #1497, opened 2021-11-18
Add telemetry around runner update process.

TingluoHuang:
Adding basic telemetry around the runner update process so we can easily identify issues.
- How long does the download take?
- How long does the unzip take?
- How long does the folder copy take?

https://github.com/github/c2c-actions-runtime/issues/1557

Sample telemetry in Kusto.
image


thboop (approved):
LGTM

PR #1521, opened 2021-11-30
Add masks for multiline secrets from `::add-mask::`

TingluoHuang:
When the runner gets a secret coming from the service as part of the job message, if the secret has multiline, the runner will break the multiline and add a secret masker to each line. 

However, we don't have the same logic when an action or step tries to register a new secret masker with `::add-mask::`.

We should copy the logic and do the same thing with `::add-mask::`

### DIFF
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -383,10 +383,10 @@ namespace GitHub.Runner.Worker
                 Trace.Info($"Add new secret mask with length of {command.Data.Length}");
 
                 // Also add each individual line. Typically individual lines are processed from STDOUT of child processes.
-                var split = command.Data.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+                var split = command.Data.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
                 foreach (var item in split)
                 {
-                    HostContext.SecretMasker.AddValue(item.Trim());
+                    HostContext.SecretMasker.AddValue(item);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -381,6 +381,13 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
 
                 HostContext.SecretMasker.AddValue(command.Data);
                 Trace.Info($"Add new secret mask with length of {command.Data.Length}");
+
+                // Also add each individual line. Typically individual lines are processed from STDOUT of child processes.
+                var split = command.Data.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
```
thboop, line 386: Should we also do `StringSplitOptions.TrimEntries` so we can avoid trimming each item?

src/Test/L0/Worker/ActionCommandManagerL0.cs
```diff
@@ -418,6 +418,24 @@ public void AddMatcherTranslatesFilePath()
...
 
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void AddMaskWithMultilineValue()
+        {
+            using (TestHostContext hc = CreateTestContext())
+            {
+                // Act
+                _commandManager.TryProcessCommand(_ec.Object, $"::add-mask::abc%0Ddef%0Aghi", null);
```
thboop, line 429: consider adding the crlf example that windows typically uses `\r\n` 
`abc\rdef\nghi\r\njkl`

thboop (dismissed):
LGTM nit feedback

src/Runner.Worker/ActionCommandManager.cs
TingluoHuang: `TrimEntries` is new in dotnet 5+, i am going to merge the dotnet 6 PR and then add this.

thboop (approved):
LGTM

PR #1522, opened 2021-12-01
Add boilerplate to support using kubernetes as container provider

fgalind1:
This defines the initial base layout to support additional container providers besides using docker such as kuberentes.

Feature is not yet enabled

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -376,6 +376,14 @@ namespace GitHub.Runner.Common
         }
 
         /// <summary>
+        /// Register a default dynamic service.
+        /// </summary>
+        public void RegisterService(Type type, Type target)
+        {
+            _serviceTypes.TryAdd(type, target);
+        }
+
+        /// <summary>
         /// Creates a new instance of T.
         /// </summary>
         public T CreateService<T>() where T : class, IRunnerService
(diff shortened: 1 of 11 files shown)


fgalind1:
@ethomson I've created this initial PR as discussed offline without the real implementation yet

drewboswell:
This would solve so many of our headaches. +1

ethomson:
@drewboswell Yes!  Mine, too.  😁

I regret that it's taken so long to make progress on this - I deeply appreciate the work that @fgalind1 has done here.  I'm very pleased that we have started to look at this and evaluate it.  To give some more insight into my priorities: running in Kubernetes is the top priority for the runner team as soon as we finish up the last of the work to move over the actions runtime from the soon-to-be-end-of-life Node 12 to Node 16.  Unfortunately, this work was delayed because the newest versions of Node tickled a fun kernel bug with [overlay filesystems](https://github.com/actions/runner/issues/1536) - so unfortunately affecting most Kubernetes installs in cloud providers.

It's rare that you get to say "it's a kernel bug!" when you encounter a problem, but the joy of encountering that extremely rare occurrence is wiped out by the disappointment when you realize that now you have to get everybody to patch their kernels before _your_ software works.

In any case - I apologize for the tangent.  We're excited about this and taking a serious look at both improving how our runner works in Kubernetes and with [https://github.com/actions-runner-controller/actions-runner-controller](Actions Runner Controller) in general, and this PR in particular.

Thanks for your patience!

PR #1551, opened 2021-12-10
Update dependency check for dotnet 6.

TingluoHuang:
Dotnet renames its native lib from `System.*.so` to `libSystem.*.so` since dotnet 5, we need to change our script to check the right files, otherwise, you will get annoying output like 
```
ldd: ./bin/System.Security.Cryptography.Native.OpenSsl.so: No such file or directory
ldd: ./bin/System.IO.Compression.Native.so: No such file or directory
```


thboop (approved):
lgtm

PR #1556, opened 2021-12-14
Produce trimmed down runner packages.

TingluoHuang:
Try to produce trimmed-down runner packages to speed up the runner upgrade process.
Depending on the state of the current runner, It can choose to avoid re-download the dotnet core runtime and all the externals (node12+node16).

Part of https://github.com/github/c2c-actions-runtime/issues/1517

Will create follow-up PR to actually consume these packages during runner update.

The produced packages will look like this:
image


src/Test/L0/PackagesTrimL0.cs
```diff
@@ -0,0 +1,154 @@
...
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public async Task RunnerLayoutParts_OverlapFiles()
+        {
+            using (TestHostContext hc = new TestHostContext(this))
+            {
+                Tracing trace = hc.GetTrace();
+                var runnerCoreAssertsFile = Path.Combine(TestUtil.GetSrcPath(), @"Misc/runnercoreassets");
+                var runnerDotnetRuntimeFile = Path.Combine(TestUtil.GetSrcPath(), @"Misc/runnerdotnetruntimeasserts");
```
thboop, line 60:
```suggestion
                var runnerDotnetRuntimeFile = Path.Combine(TestUtil.GetSrcPath(), @"Misc/runnerdotnetruntimeassets");
```

Super nit, but you may want to change this in the places in this file you've used it. Technically we do assert during test 🤷

src/Test/L0/PackagesTrimL0.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public async Task RunnerLayoutParts_OverlapFiles()
+        {
+            using (TestHostContext hc = new TestHostContext(this))
+            {
+                Tracing trace = hc.GetTrace();
+                var runnerCoreAssertsFile = Path.Combine(TestUtil.GetSrcPath(), @"Misc/runnercoreassets");
```
thboop, line 59: Is this meant to be "assets"? There's a few areas in this file that use asserts when "assets" is more readable to me.

.github/workflows/build.yml
```diff
@@ -57,6 +57,29 @@ jobs:
       working-directory: src
       if: matrix.runtime != 'linux-arm64' && matrix.runtime != 'linux-arm'
 
+    # Check runtime/externals hash
+    - name: Compute/Compare runtime and externals Hash
```
thboop, line 61: +1 on the ci its a great idea!

thboop (dismissed):
LGTM

src/Test/L0/PackagesTrimL0.cs
TingluoHuang: it should be `assets`... I have a hard time spelling it right... 😢

thboop (approved):
lgtm

PR #1563, opened 2021-12-16
Show service container logs on teardown

karagog:
This will export the service container logs at the end of a GitHub action right before they're torn down. This information can be very useful for debugging, and it's a pain to manually export logs on an as-needed basis.

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -338,12 +338,16 @@ namespace GitHub.Runner.Worker
 
             if (!string.IsNullOrEmpty(container.ContainerId))
             {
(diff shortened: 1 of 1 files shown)


TingluoHuang (approved):

karagog:
Note that I think this will also print the logs for the container that's running your current job action (if you're running in a container), which may be redundant b/c I think that log is already printed in a different place. But I'm not sure how to isolate just the service container logs, so I just printed all container logs.

TingluoHuang:
@karagog you can filter job container out with `if(!container.IsJobContainer){ run docker log}`.

karagog:
@TingluoHuang Done.

TingluoHuang (approved):

PR #1568, opened 2021-12-17
Use trimmed packages to speedup runner updates

TingluoHuang:
The runner update is too slow since it needs to download a relatively big package (~100MB)
However, most parts of the package are redundant cross versions most of the time.
- Externals: node 12 and node 16, we updated those once every few years)
- Dotnet Runtime: we only upgrade to the latest LTS once every 2 years.

We could make the runner only download the delta and make the update process more quick and smooth.

P95 download time cross the board (in seconds).
image

### DIFF
--- a/src/Runner.Listener/SelfUpdater.cs
+++ b/src/Runner.Listener/SelfUpdater.cs
@@ -1,5 +1,4 @@
 using GitHub.DistributedTask.WebApi;
-using GitHub.Runner.Common.Util;
 using System;
 using System.Diagnostics;
 using System.IO;
@@ -13,7 +12,7 @@ using GitHub.Services.WebApi;
 using GitHub.Services.Common;
 using GitHub.Runner.Common;
 using GitHub.Runner.Sdk;
-using System.Text;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Reflection;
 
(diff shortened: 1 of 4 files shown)


src/Runner.Listener/SelfUpdater.cs
```diff
@@ -585,5 +749,320 @@ private async Task UpdateRunnerUpdateStateAsync(string currentState)
...
+                                }
+                            };
+                            p.OutputDataReceived += (_, data) =>
+                            {
+                                if (!string.IsNullOrEmpty(data.Data))
+                                {
+                                    Trace.Info(data.Data);
+                                }
+                            };
+                            var exitCode = await p.ExecuteAsync(HostContext.GetDirectory(WellKnownDirectory.Root), newNodeBinary, "--version", null, token);
```
TingluoHuang, line 786: `node -e "console.log('RestoreTrimmedExternals')"`

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -30,13 +29,19 @@ public class SelfUpdater : RunnerService, ISelfUpdater
     {
         private static string _packageType = "agent";
         private static string _platform = BuildConstants.RunnerPackage.PackageName;
+        private static string _dotnetRuntime = "dotnetRuntime";
```
thboop, line 32: nit consider an enum for runtime/externals keys

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -59,19 +66,44 @@ public async Task<bool> SelfUpdate(AgentRefreshMessage updateMessage, IJobDispat
             {
                 var totalUpdateTime = Stopwatch.StartNew();
 
+                // Copy dotnet runtime and externals of current runner to a temp folder
+                // So we can re-use them with trimmed runner package, if possible.
+                // This process is best effort, if we can't use trimmed runner package, 
+                // we will just go with the full package.
+                var linkedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(token);
+                _cloneAndCalculateContentHashTask = CloneAndCalculateAssetsHash(_dotnetRuntimeCloneDirectory, _externalsCloneDirectory, linkedTokenSource.Token);
```
thboop, line 74: nit: I'm wondering if the time saved here is worth doing before we know if we should copy these or not. Typically disk copy speeds are much faster than downloads, are we actually saving time?

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -211,6 +301,74 @@ private async Task DownloadLatestRunner(CancellationToken token)
                 }
             }
 
+            var trimmedPackageRestoreTasks = new List<Task<bool>>();
+            if (!fallbackToFullPackage)
```
thboop, line 305: Should we consider making this a function, and just calling recursively if it fails and pass this value in? May cleanup some of the dupe create/delete code

src/Runner.Listener/SelfUpdater.cs
TingluoHuang: The reason I start the copy is we might run into errors when copying or computing the hash.
So I will just fall back to the full packages early if anything went wrong.

thboop (approved):
Left some nit comments, and a follow up in PMS, but the rest LGTM

Issue #730, opened 2020-10-01
Service name not unique for long repository names


RobCannon:
**Describe the bug**
Our company has a naming convention for repositories that creates from long names (50 characters or more) where the first 30 or so characters will be the same for all of the repos in a given project).  Service name generation seems to take the first 45 characters from the org-repo name and is only leaving the first 11 characters of runner name to combine to make a service name.  If you have several repos with similar names and that are only unique at the end, the service name generation will produce the same name for different repos.  When trying to install multiple runners on one server, this causes a conflict.

Form what I can see, service name can be 256 characters long, so the name of the repo should not be cut off at 45 characters.  I would like to name the runner name so it matches the last part of the repo name which is usually unique in 20-25 characters instead of having to generate a short unique name.

**To Reproduce**
Create two runners on the same server where the first 45 characters or the repo name are the same and the first 11 characters of the runner name are the same

**Expected behavior**
Since service names can be 256 characters long, I would not expect there to be a conflict

## Runner Version and Platform
v2.273.4 Linux

OS of the machine running the runner? OSX/Windows/Linux/...
Centos 7

## What's not working?
Creating launch runner in /etc/systemd/system/actions.runner.MyOrgName-9999_us_project-name_alliance_gh-actions.gh-actions-.service
Failed: error: exists /etc/systemd/system/actions.runner.MyOrgName-9999_us_project-name_alliance_gh-actions.gh-actions-.service

kwacky1:
:wave: @brcrista Are you able to give me any idea on if this work is scheduled or not? 

[Intel](https://github.zendesk.com/agent/tickets/1185588) are chasing me up on an update.

fgalind1:
Bumping this issue 👍 

+1 - In our company we're using also some prefixes for our runners, and we just hit the same issue described. As a WA I've updated the svc.sh to update the `SVC_NAME` and it works, but it's a bit of a pain to do this for every runner when using automation :)

PR #1577, opened 2021-12-23
Do not trim runner name from service name on unix

fgalind1:
Trimming runner name in many cases causes issues when installing multiple runners on the same machine which are just differentiated by a suffix. So far on linux without the 80 characters limitation it has been working fine

This fixes https://github.com/actions/runner/issues/730 and https://github.com/githubcustomers/Intel/issues/160

### DIFF
--- a/src/Runner.Listener/Configuration/ServiceControlManager.cs
+++ b/src/Runner.Listener/Configuration/ServiceControlManager.cs
@@ -60,12 +60,13 @@ namespace GitHub.Runner.Listener.Configuration
 
                 string runnerNameSubstring = settings.AgentName;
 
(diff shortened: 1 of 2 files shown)


src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -60,12 +60,13 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
 
                 string runnerNameSubstring = settings.AgentName;
 
+#if OS_WINDOWS
```
TingluoHuang, line 63: do we know what's the limit on Linux?

PR #1592, opened 2022-01-05
Skip adding line to console line queue if the queue is backed up.

TingluoHuang:
Avoid extra memory consumption when steps produce tons of outputs.

Might help https://github.com/github/c2c-actions-support/issues/724

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using GitHub.DistributedTask.WebApi;
-using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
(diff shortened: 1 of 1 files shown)


src/Runner.Common/JobServerQueue.cs
```diff
@@ -230,12 +242,6 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
                         stepRecordIds.Add(lineInfo.StepRecordId);
                     }
 
-                    if (!string.IsNullOrEmpty(lineInfo.Line) && lineInfo.Line.Length > 1024)
-                    {
-                        Trace.Verbose("Web console line is more than 1024 chars, truncate to first 1024 chars");
-                        lineInfo.Line = $"{lineInfo.Line.Substring(0, 1024)}...";
```
TingluoHuang, line 236: we will do the trim when we enqueue instead of dequeue.

thboop (approved):
lgtm

PR #1601, opened 2022-01-11
Generate refname for all build/pull container step.

TingluoHuang:
We are getting tons of errors on the service everyday about
`Record 7a9fa351-99bd-4bcb-9bbc-90c9d6f68d94 (name: Initialize containers, type: Task) does not have a reference name.`

This is because we didn't set a `refname` for any of the container setup steps:
- `Initialize containers`
- `Pull * container`
- `Build * container`

Errors every day...
image

This will also fix https://github.com/github/c2c-actions-runtime/issues/1602

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -122,20 +122,20 @@ namespace GitHub.Runner.Worker
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


thboop (approved):
LGTM

PR #1607, opened 2022-01-13
Adds file context for runner script with SELinux enabled

### LINKED ISSUE #1606: SELinux Enforcing stops systemd service from starting
**Describe the bug**

When SELinux is Enforcing, scripts started by `systemd` need the `initrc_exec_t` context set. It is not currently.

**To Reproduce**

1. Have SELinux Enforcing;
2. Add a runner;
3. Run `./svc.sh install`
4. Observe that `systemd` failed to start the script:

    ```bash
    sudo grep -i denied /var/log/audit/audit.log|grep -i svc|tail -1|fold -s
    ```
    ```journald
    type=AVC msg=audit(1642053001.589:304): avc:  denied  { execute } for  
    pid=14114 comm="(unsvc.sh)" name="runsvc.sh" dev="vda1" ino=25325969 
    scontext=system_u:system_r:init_t:s0 
    tcontext=unconfined_u:object_r:user_home_t:s0 tclass=file permissive=0
    ```

**Expected behavior**

- `systemd` should be able to start the script.

- Version: `2.278`
- OS: Linux (x64)

/cc: @mikedalton

krayon:
When SELinux is Enforcing, the `initrc_exec_t` context is needed in order for `systemd` to execute the script and start the runner.

This patch adds the adding and removing of this context from the script for the `svc.sh install` and `svc.sh uninstall` commands respectively.

Fixes #1606

### DIFF
--- a/src/Misc/layoutbin/systemd.svc.sh.template
+++ b/src/Misc/layoutbin/systemd.svc.sh.template
@@ -66,11 +66,9 @@ function install()
 
     # Recent Fedora based Linux (CentOS/Redhat) has SELinux enabled by default
     # We need to restore security context on the unit file we added otherwise SystemD have no access to it.
-    command -v getenforce > /dev/null
-    if [ $? -eq 0 ]
-    then
+    if command -v getenforce > /dev/null; then
         selinuxEnabled=$(getenforce)
-        if [[ $selinuxEnabled == "Enforcing" ]]
+        if [[ "${selinuxEnabled}" == "Enforcing" ]]
         then
             # SELinux is enabled, we will need to Restore SELinux Context for the service file
             restorecon -r -v "${UNIT_PATH}" || failed "failed to restore SELinux context on ${UNIT_PATH}"
@@ -85,6 +83,15 @@ function install()
     cp ./bin/runsvc.sh ./runsvc.sh || failed "failed to copy runsvc.sh"
     chown ${run_as_uid}:${run_as_gid} ./runsvc.sh || failed "failed to set owner for runsvc.sh"
     chmod 755 ./runsvc.sh || failed "failed to set permission for runsvc.sh"
+
+    if [[ "${selinuxEnabled}" == "Enforcing" ]]; then
+        # We will need to add the initrc_exec_t context for the script
+        # (if we can't add, softfail first and try modify instead)
+        semanage fcontext --add    --type initrc_exec_t "${RUNNER_ROOT}/runsvc.sh" ||\
+        semanage fcontext --modify --type initrc_exec_t "${RUNNER_ROOT}/runsvc.sh" ||\
+        failed "failed to set file context for runsvc.sh"
+        restorecon -v "${RUNNER_ROOT}/runsvc.sh"
+    fi
 
     systemctl enable ${SVC_NAME} || failed "failed to enable ${SVC_NAME}"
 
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -86,6 +84,15 @@ function install()
     chown ${run_as_uid}:${run_as_gid} ./runsvc.sh || failed "failed to set owner for runsvc.sh"
     chmod 755 ./runsvc.sh || failed "failed to set permission for runsvc.sh"
 
+    if [[ "${selinuxEnabled}" == "Enforcing" ]]; then
+        # We will need to add the initrc_exec_t context for the script
+        # (if we can't add, softfail first and try modify instead)
+        semanage fcontext --add    --type initrc_exec_t "${RUNNER_ROOT}/runsvc.sh" ||\
+        semanage fcontext --modify --type initrc_exec_t "${RUNNER_ROOT}/runsvc.sh" ||\
```
istathar, line 91: _semanage_ doesn't appear to be installed on a stock Fedora system

istathar:
```
$ sudo chcon -t initrc_exec_t runsvc.sh
```
has worked locally. Of course that's only a temporary relabelling but wanted to let you know you've got the right context.

krayon: Thanks @istathar , how annoying. CentOS and RedHat do ship with `semanage`. I guess the other option is to check for the context and bail if:
  1. Context is wrong; _and_
  2. no `semanage` is present

src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -86,6 +84,15 @@ function install()
     chown ${run_as_uid}:${run_as_gid} ./runsvc.sh || failed "failed to set owner for runsvc.sh"
     chmod 755 ./runsvc.sh || failed "failed to set permission for runsvc.sh"
 
+    if [[ "${selinuxEnabled}" == "Enforcing" ]]; then
+        # We will need to add the initrc_exec_t context for the script
+        # (if we can't add, softfail first and try modify instead)
+        semanage fcontext --add    --type initrc_exec_t "${RUNNER_ROOT}/runsvc.sh" ||\
```
TingluoHuang, line 90: should we error if we can't find `semanage` installed?

src/Misc/layoutbin/systemd.svc.sh.template
TingluoHuang: I created a new CentOS 8 Azure VM and it doesn't come with `semanage` 😢

src/Misc/layoutbin/systemd.svc.sh.template
krayon: Yes, we can either:
  - Bail if `selinux` is enabled and Enforcing but `semanage` isn't present; or
  - Report the issue to the user but continue to fail (as it does now).

Additionally, we could optionally check the context in case they've done something clever and the file has inherited the correct context anyway. In my tests with CentOS `semanage` was installed but right now we have more reports of it missing than present so it may have been how I did my installation.

JAORMX: Well, you could probably still adjust the context even if SELinux is not enforcing. I'd just check if `semanage` is installed and try do do the context change.

Issue #1605, opened 2022-01-13
External runners suddenly failing with System.Globalization.CultureNotFoundException


ramosbugs:
**Describe the bug**

We've been using external GitHub Actions runners (located on Amazon EC2) without any issue for months. Beginning this afternoon, all our GitHub Actions jobs are failing with the following error returned by the runner:

```
System.Globalization.CultureNotFoundException: Only the invariant culture is supported in globalization-invariant mode. See https://aka.ms/GlobalizationInvariantMode for more information. (Parameter 'name')
en-US is an invalid culture identifier.
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at System.Globalization.CultureInfo..ctor(String name)
   at GitHub.Runner.Common.HostContext.SetDefaultCulture(String name)
   at GitHub.Runner.Worker.Worker.SetCulture(AgentJobRequestMessage message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
```

**To Reproduce**

Run a GitHub Actions workflow with jobs specifying `runs-on` with an external runner.

**Expected behavior**

Actions job runs on external runner.

## Runner Version and Platform

Version of your runner?

2.286.0 (also fails on 2.280.2)

OS of the machine running the runner? OSX/Windows/Linux/...

Linux ARM64

## What's not working?

See stack trace above.

## Job Log Output

No output for job since it fails to start. However, the workflow run shows the following annotation:
```
System.Globalization.CultureNotFoundException: Only the invariant culture is supported in globalization-invariant mode. See https://aka.ms/GlobalizationInvariantMode for more information. (Parameter 'name')
en-US is an invalid culture identifier.
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at System.Globalization.CultureInfo..ctor(String name)
   at GitHub.Runner.Common.HostContext.SetDefaultCulture(String name)
   at GitHub.Runner.Worker.Worker.SetCulture(AgentJobRequestMessage message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
```

## Runner and Worker's Diagnostic Logs

[...]

PR #1609, opened 2022-01-14
Fix breaking change in dotnet 6 around globalization-invariant.

TingluoHuang:
Workaround the breaking change in dotnet 6
https://docs.microsoft.com/en-us/dotnet/core/compatibility/globalization/6.0/culture-creation-invariant-mode

Fix: #1605

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
           devScript: ./dev.sh
 
         - runtime: win-x64
-          os: windows-latest
+          os: windows-2019
           devScript: ./dev
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


cschleiden (dismissed):

.github/workflows/build.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           devScript: ./dev.sh
 
         - runtime: win-x64
-          os: windows-latest
+          os: windows-2019
```
TingluoHuang, line 40: `windows-latest` is using `windows-2022` which doesn't not have .net 4.5 installed, change back to `windows-2019` for now

lokesh755 (approved):

PR #1612, opened 2022-01-18
updated systemd svc.sh to accept custom service file

nikola-jokic:


src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -5,7 +5,9 @@ SVC_NAME=${SVC_NAME// /_}
 SVC_DESCRIPTION="{{SvcDescription}}"
 
 SVC_CMD=$1
-arg_2=${2}
+SVC_RUN_AS_USER=
+
+SCRIPT_NAME=$0
```
TingluoHuang, line 10: Can we just read from an environment variable and overwrite `TEMPLATE_PATH` based on that env?
Then we don't need to change the rest of the svc.sh template file, also, we probably want to make same change to `darwin.svc.sh.template` as well.

TingluoHuang (changes requested):
need to change the macOS template as well.

src/Misc/layoutbin/systemd.svc.sh.template
nikola-jokic: Sure, but then we would have to rely on users remembering that they set the path. But regardless, by using flags, it is more explicit that you want to provide your own template, and it is nice for documentation purposes.

Also, I guess it is more convenient to provide in the command you are starting a service your template than exporting or setting env var and then running it. Let's say I exported a variable to run my template, and I made a mistake. Now, to run the default svc config, I must unset the TEMPLATE_PATH env variable since it would be used again on the next run. With this, I would just run the same command without the flag.

TingluoHuang: `svc.sh` is used when you install/uninstall/start/top the service, but the `TEMPLATE_PATH` is only used during `install`.

I don't think we want the customer to install using 1 template and start the service as a different template.

Also, I am trying to minimize the change as much as possible.

nikola-jokic: Sure, we can do it that way then. I will revert to the previous state and add fix it that way 😊

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -29,7 +35,11 @@ function failed()
 }
 
 if [ ! -f "${TEMPLATE_PATH}" ]; then
-    failed "Must run from runner root or install is corrupt"
+    if [[ $IS_CUSTOM_TEMPLATE = 0 ]]; then
+        failed "Must run from runner root or install is corrupt"
+    else
+        failed "Invalid TEMPLATE_PATH env variable"
```
nikola-jokic, line 41: Added distinction between custom template and default run error message

src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -10,7 +10,10 @@ arg_2=${2}
 RUNNER_ROOT=`pwd`
 
 UNIT_PATH=/etc/systemd/system/${SVC_NAME}
-TEMPLATE_PATH=./bin/actions.runner.service.template
+TEMPLATE_PATH=$TEMPLATE_PATH
```
TingluoHuang, line 13: Should we give it a different name for the env, maybe `$GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE`?

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -17,7 +17,13 @@ RUNNER_ROOT=`pwd`
 
 LAUNCH_PATH="${HOME}/Library/LaunchAgents"
 PLIST_PATH="${LAUNCH_PATH}/${SVC_NAME}.plist"
-TEMPLATE_PATH=./bin/actions.runner.plist.template
+TEMPLATE_PATH=$TEMPLATE_PATH
```
TingluoHuang, line 20: same comment on the env name.

src/Misc/layoutbin/darwin.svc.sh.template
TingluoHuang: should we add similar code to the `systemd.svc.sh.template` as well?

nikola-jokic: Yes, I noticed. Running tests now on the repaired version. Do you want me to amend the commit? There is no reason why this one should exist.

TingluoHuang: do whatever you want. 😄

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -29,7 +35,11 @@ function failed()
 }
 
 if [ ! -f "${TEMPLATE_PATH}" ]; then
-    failed "Must run from runner root or install is corrupt"
+    if [[ $IS_CUSTOM_TEMPLATE = 0 ]]; then
+        failed "Must run from runner root or install is corrupt"
+    else
+        failed "Invalid GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE env variable"
```
TingluoHuang, line 41: Should we print out `GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE` as part of the error, so the customer can check their setting?
ex:
  `failed "Fail to find '$GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE' 
 using GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE env variable"`

src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -31,7 +37,11 @@ function failed()
 }
 
 if [ ! -f "${TEMPLATE_PATH}" ]; then
-    failed "Must run from runner root or install is corrupt"
+    if [[ $IS_CUSTOM_TEMPLATE = 0 ]]; then
+        failed "Must run from runner root or install is corrupt"
+    else
+        failed "Invalid GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE env variable"
```
TingluoHuang, line 43: same comment here. 😄

src/Misc/layoutbin/systemd.svc.sh.template
```diff
@@ -31,7 +37,11 @@ function failed()
 }
 
 if [ ! -f "${TEMPLATE_PATH}" ]; then
-    failed "Must run from runner root or install is corrupt"
+    if [[ $IS_CUSTOM_TEMPLATE = 0 ]]; then
+        failed "Must run from runner root or install is corrupt"
+    else
+        failed "Service file at '$GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE' using GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE env variable not found"
```
TingluoHuang, line 43: `is not found`

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -29,7 +35,11 @@ function failed()
 }
 
 if [ ! -f "${TEMPLATE_PATH}" ]; then
-    failed "Must run from runner root or install is corrupt"
+    if [[ $IS_CUSTOM_TEMPLATE = 0 ]]; then
+        failed "Must run from runner root or install is corrupt"
+    else
+        failed "Service file at '$GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE' using GITHUB_ACTIONS_RUNNER_SERVICE_TEMPLATE env variable not found"
```
TingluoHuang, line 41: `is not found`

TingluoHuang (approved):

nikola-jokic:
Thanks, @TingluoHuang 😊

Issue #1550, opened 2021-12-10
Please, please, please - suggest the use of Net.WebClient instead of Invoke-WebRequest in self-hosted instructions


slee-accesso:
In Settings/Actions/Runners in a repo, when creating a self-hosted runner, a series of Powershell instructions are presented. Among them is an Invoke-WebRequest call like this:

`Invoke-WebRequest -Uri https://github.com/actions/runner/releases/download/v2.285.1/actions-runner-win-x64-2.285.1.zip -OutFile actions-runner-win-x64-2.285.1.zip`

Invoke-WebRequest is terribly slow due to it updating the screen during the download. The same download can be made in a one liner like this:

`(New-Object Net.WebClient).DownloadFile("https://github.com/actions/runner/releases/download/v2.285.1/actions-runner-win-x64-2.285.1.zip", "$PWD\actions-runner-win-x64-2.285.1.zip")`

The performance difference with this ~70mb file?

Invoke-WebRequest: 154.4 seconds
Net.WebClient: 3.41 seconds

I think cutting off 151 seconds is a nice thing, no? Could you please replace the suggested command with the faster one (at least for windows?)

Thanks!

  - Sean

Temtaime:
Wouldn't it better to report this bug to powershell repo ?

slee-accesso:
Well - I thought you'd be interested in improving your user's experiences with a small change to your directions. But if you'd rather punt to Microsoft, I can't stop you.

slee-accesso:
Thank you, @ruvceskistefan : )

ruvceskistefan:
You're welcome, thank you for the reported issue and the proposed solution. Since I've created a PR that solves this problem, I'm closing this issue.

ruvceskistefan:
Hi @slee-accesso,
Unfortunately, I will close this issue for now. This issue does not seem to be a problem with the runner application, so I'll report the problem to the powershell teams, and we'll see if they can solve the problem with the underlying commands.

ruvceskistefan:
Here is the issue reported in the powershell repo: https://github.com/PowerShell/PowerShell/issues/16914

PR #1613, opened 2022-01-18
Use Net.WebClient instead of Invoke-WebRequest

ruvceskistefan:
Description: The problem was recommending the use of the Invoke-WebRequest command to download the project on a Windows operating system. The solution is to recommend a much faster command that uses Net.WebClient object. 

Related Issue: https://github.com/actions/runner/issues/1550

After this change download performance was definitely improved for me too.

### DIFF
--- a/releaseNote.md
+++ b/releaseNote.md
@@ -25,8 +25,10 @@ The following snipped needs to be run on `powershell`:
 ``` powershell
 # Create a folder under the drive root
 mkdir \actions-runner ; cd \actions-runner
+
 # Download the latest runner package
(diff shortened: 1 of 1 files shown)


ethomson:
I love that we're improving the instructions to speed things up!  Is there a more ergonomic way to do this, though?  Those instructions are becoming awfully hard to read.  🤔

ruvceskistefan:
I'm not sure what you mean, I changed one command in the `releaseNote.md`, so everything is the same in terms of readability.

thboop:
Can we file an issue to improve the `Invoke-WebRequest` in powershell instead? We would have to change this in many areas outside the runner and I don't think we want to take that on at this time.

ruvceskistefan:
Sure, here is the issue in powershell repo: https://github.com/PowerShell/PowerShell/issues/16914
Thanks @thboop !

Issue #286, opened 2020-01-16
Allow Bypassing of SSL Cert Validation on Self-Hoster Runners


Kremlin:
**Describe the enhancement**

Allow bypassing SSL certificate validation

**Additional information**

When a proxy server does SSL inspection, it does a sort of "man in the middle" action with certificates, such that the certs for github.com come back as being signed from an internal CA.
Even when using the `--sslskipcertvalidation` flag, it reports:

```
[2020-01-16 15:54:33Z] System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
 ---> System.Security.Authentication.AuthenticationException: The remote certificate is invalid according to the validation procedure.
```

TingluoHuang:
The runner doesn't support bypass SSL CA cert validation error. Customer needs to install their self-signed CA certs to certificate store on each platform. 
Linux: OpenSSL store
macOS: keychain
Windows: Windows certificates manager

snyder-riley-pfg:
@Kremlin we were running into the same issues, and following @TingluoHuang by adding our certs to the linux ssl store did not work.

What we found was by setting `SSL_CERT_DIR` to `/etc/ssl/certs` it then trusted our certs and we were able to configure the runner.

TingluoHuang:
@snyder-riley-pfg how did you add your cert to openssl store? i think `SSL_CERT_DIR` is just another way to set trusted certs for openssl, in the end the runner still use openssl for query trusted certs.

snyder-riley-pfg:
@TingluoHuang we used [update-ca-trust](https://www.mankier.com/8/update-ca-trust) to add the certs, then setting that environment variable seemed to tell the service to look where we were adding them. Perhaps before it was looking in `/etc/pki/tls` or another directory which was not where we were adding our certs.

H6LS1S:
export DOTNET_SYSTEM_NET_HTTP_USESOCKETSHTTPHANDLER=0

clairmont32:
> @Kremlin we were running into the same issues, and following @TingluoHuang by adding our certs to the linux ssl store did not work.
> 
> What we found was by setting `SSL_CERT_DIR` to `/etc/ssl/certs/` it then trusted our certs and we were able to configure the runner.
> 
> Working on RHEL 7.7

I'm running into this on Ubuntu. Did you simply do `SSL_CERT_DIR=/etc/ssl/certs` within the terminal or within one of the Actions scripts? I tried with export but that didnt work for me.

[...]

PR #1616, opened 2022-01-18
Introduce GITHUB_ACTIONS_RUNNER_TLS_NO_VERIFY=1 to skip SSL cert verification for the runner.

TingluoHuang:
Allow customers to turn off SSL validation when the runner connects to the service.

`export GITHUB_ACTIONS_RUNNER_TLS_NO_VERIFY=1` before configuring and running the runner.

Use cases
- Behind proxy that tries to decrypt SSL traffic for audit
- Self-signed cert for GHES is not compatible with a certain OS, ex macOS

https://github.com/github/c2c-actions-runtime/issues/1484

close: https://github.com/actions/runner/issues/286


thboop (approved):
LGTM

PR #1617, opened 2022-01-19
Issue 1528: use OS specific path separator

### LINKED ISSUE #1528: VssFileStorage embeds a windows path separator on other platforms
https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs#L195

This eventually results in something like:
```
/home/runner/.config/GitHub\ActionsService/8.0/Cache/34e00c7b-39b1-4ed9-b15e-f23056129642/0000005a-0000-8888-8000-000000000000/LocationServiceData.config
```

### LINKED ISSUE #1810: Embedded backslash in config dir name breaks Dropbox on Mac
The runner creates a directory, `~/.config/GitHub\ActionsService/` on Mac and Linux. Dropbox chokes on this:

DropboxError

Is the embedded backslash intended? While Dropbox ought to be able to handle it, it seems a little pathological. :-)

ruvceskistefan:
Description: Problem was usage of \\ as path separator on non-Windows platforms.

Closes #1528, closes #1810 

After this change, VssFileStorage uses path separator based on OS.

### DIFF
--- a/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs
+++ b/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs
@@ -24,7 +24,11 @@ namespace GitHub.Services.Common.ClientStorage
         private readonly VssFileStorageReader m_reader;
         private readonly IVssClientStorageWriter m_writer;
 
+#if OS_WINDOWS
         private const char c_defaultPathSeparator = '\\';
+#else
+        private const char c_defaultPathSeparator = '/';
+#endif
         private const bool c_defaultIgnoreCaseInPaths = false;
 
         /// <summary>
@@ -192,7 +196,7 @@ namespace GitHub.Services.Common.ClientStorage
                 // Windows Impersonation is being used.
 
                 // Check to see if we can find the user's local application data directory.
-                string subDir = "GitHub\\ActionsService";
+                string subDir = $"GitHub{c_defaultPathSeparator}ActionsService";
                 string path = Environment.GetEnvironmentVariable("localappdata");
                 SafeGetFolderPath(Environment.SpecialFolder.LocalApplicationData);
                 if (string.IsNullOrEmpty(path))
--- a/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs
+++ b/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs
@@ -23,12 +23,7 @@ namespace GitHub.Services.Common.ClientStorage
         private readonly string m_filePath;
         private readonly VssFileStorageReader m_reader;
         private readonly IVssClientStorageWriter m_writer;
-
-#if OS_WINDOWS
         private const char c_defaultPathSeparator = '\\';
-#else
-        private const char c_defaultPathSeparator = '/';
-#endif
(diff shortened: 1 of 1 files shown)


src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs
```diff
@@ -192,7 +196,7 @@ internal static string ClientSettingsDirectory
                 // Windows Impersonation is being used.
 
                 // Check to see if we can find the user's local application data directory.
-                string subDir = "GitHub\\ActionsService";
+                string subDir = $"GitHub{c_defaultPathSeparator}ActionsService";
```
TingluoHuang, line 199: what about delete `c_defaultPathSeparator` and use `Path.Combine()`? it should do the right thing, I think.

ruvceskistefan: That is a better solution, however this constant is later used as the default value of input parameters of several methods, in which paths are formed in different ways (eg via operator + in PathUtil).
I can explore all the different ways of forming paths in project and change them so that in all cases Path.Combine() is used.

thboop: Can we do something like `string subDir = Path.Combine("GitHub", "ActionsService")`?

ruvceskistefan: Of course, but that will require changes in several places. The problem is that the `c_defaultPathSeparator ` is used as an input parameter of one constructor and two methods, so it would be necessary to change their signatures (and all methods and constructors to which the same parameter is passed). Do you want me to change that?

fhammerl: Small drive-by comment: if we decide against `Path.Combine` and  end up merging the OS specific `c_defaultPathSeparator` solution, keep in mind that the value of `c_defaultPathSeparator` is used to generate `m_emptyPathSegment` in VssLocalFile, and I'm not sure how that affects the correctness of the invalid path check a few lines later:


https://github.com/ruvceskistefan/runner/blob/3d3388cb394453aa18db46ef50b20ea63fe8a3fd/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs#L474

```
 m_emptyPathSegment = new string(pathSeparator, 2);
```

https://github.com/ruvceskistefan/runner/blob/3d3388cb394453aa18db46ef50b20ea63fe8a3fd/src/Sdk/Common/Common/ClientStorage/VssFileStorage.cs#L483

```
if (string.IsNullOrEmpty(path) || path[0] != PathSeparator || path.IndexOf(m_emptyPathSegment, StringComparison.Ordinal) >= 0 || (!allowRootPath && path.Length == 1))                
```

ruvceskistefan: Okay, makes sense. Then I'll use the `Path.Combine` for this PR instead of OS specific `c_defaultPathSeparator`.

thboop (approved):
lgtm

Issue #1611, opened 2022-01-15
Container actions should prefer the `env` set in the action.yaml file over `set-env`


philomory:
**Describe the bug**
When a prior step in a job exports a variable value (e.g. using `core.exportVariable()`), subsequent actions of type `docker` give the exported value set for that variable name a higher priority than a value specified in the action's `action.yaml` under `runs.env`.

Actions of type `composite`, on the other hand, give precedence to variable values set as part of the action definition (via `runs.steps.*.env`), over those exported by prior steps.

**To Reproduce**
1. Fork https://github.com/philomory/gha-bug-repro
2. Go to the Actions section and run the `demo` workflow via workflow dispatch
3. Observe that the steps output (among other text):
```
Composite action says: bar
Docker action says: foo
```

**Expected behavior**
Both actions should output "bar", e.g.:
```
Composite action says: bar
Docker actions says: bar
```

Alternatively, if both actions output "foo", e.g.: 
```
Composite action says: foo
Docker actions says: foo
```
... this would be confusing, but at least _consistent_.

## Runner Version and Platform
```
Current runner version: '2.286.0'
Operating System
  Ubuntu
  20.04.3
  LTS
Virtual Environment
  Environment: ubuntu-20.04
  Version: 20220111.1
  Included Software: https://github.com/actions/virtual-environments/blob/ubuntu20/20220111.1/images/linux/Ubuntu2004-Readme.md
  Image Release: https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20220111.1
Virtual Environment Provisioner
  1.0.0.0-main-20211214-1
GITHUB_TOKEN Permissions
  Actions: write
  Checks: write
  Contents: write
  Deployments: write
  Discussions: write
  Issues: write
  Metadata: read
  Packages: write
  Pages: write
  PullRequests: write
  RepositoryProjects: write
  SecurityEvents: write
  Statuses: write
Secret source: Actions
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'actions/checkout@v2' (SHA:ec3a7ce113134d7a93b817d10a8272cb61118579)
```

## What's not working?
Given one action definition like
```yaml
...
runs:
  using: 'docker'
  env:
    VALUE: 'bar'
...
```
and another like
```yaml
...
runs:
  using: 'composite'
  steps:
    - id: echo-value
      env:
        VALUE: 'bar'
...
```
[...]

PR #1619, opened 2022-01-20
Repairing environment order in composite run set by previous step

nikola-jokic:
Environment from the script at the previous step was being set, but on subsequent steps, was not restored.

Issue: #1611


thboop:
I want to make sure I understand what we are solving here correctly

I think https://github.com/actions/runner/issues/1611 is stating that for docker actions we do set the `env` from the list of job steps last, while composite actions we prioritize the env for a step in the composite action.

I definitely get how that could be confusing, but that being said, we intentionally do it this way for composite actions, as composite actions are basically a "job" inside a job. We want to favor what the action author has set, over what the workflow author has decided. The workflow author shouldn't have to worry about the inner workings of a composite action, that problem belongs to the `action.yaml` author. That being said, this is slightly inconsistent, if we were going to make a change here I would do it for the docker actions. However, that may break some users, so I don't think we will actually make a change here, we should just close this issue. I'll follow up in the issue!

Thanks for digging into this.

PR #1621, opened 2022-01-21
Prefer node16 over node12 when running internal scripts

fhammerl:
Now that node 16 is supported, it should be used internally as well as encouraged in the docs.


src/Misc/layoutbin/update.sh.template
```diff
@@ -120,6 +120,10 @@ then
     exit 1
 fi
 
+# NODE 12: In 2.281.1 RunnerService.js was still started with node 12. In this PR (TODO:LINK) I changed it to launch with node 16.
```
thboop, line 123: nit: todo still here

src/Misc/layoutbin/update.sh.template
```diff
@@ -172,6 +176,59 @@ if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
     fi
 fi
 
+# NODE 16: DUPLICATED THE ABOVE LOGIC FOR NODE16
+# fix upgrade issue with macOS when running as a service
+attemptedtargetedfix=0
+currentplatform=$(uname | awk '{print tolower($0)}')
+if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
```
thboop, line 183: should we determine if it was started with node 12 or node 16 and only run the code for that version?

docs/adrs/0361-wrapper-action.md
```diff
@@ -23,7 +23,7 @@ Node Action Example:
  name: 'My action with pre'
  description: 'My action with pre'
  runs:
-   using: 'node12'
```
TingluoHuang, line 26: we probably don't need to change this. 😄

src/Test/L0/Worker/ActionManagerL0.cs
```diff
@@ -1451,7 +1451,7 @@ public void LoadsNodeActionDefinitionYaml()
 icon: 'hello.svg' # vector art to display in the GitHub Marketplace
 color: 'green' # optional, decorates the entry in the GitHub Marketplace
 runs:
-  using: 'node12'
```
TingluoHuang, line 1454: I would not change the test code, since we have test for both node12 and node16

src/Test/L0/Worker/ActionManifestManagerL0.cs
```diff
@@ -408,7 +408,7 @@ public void Load_NodeAction()
                 var nodeAction = result.Execution as NodeJSActionExecutionData;
 
                 Assert.Equal("main.js", nodeAction.Script);
-                Assert.Equal("node12", nodeAction.NodeVersion);
+                Assert.Equal("node16", nodeAction.NodeVersion);
```
TingluoHuang, line 411: not change L0

src/Test/TestData/nodeaction.yml
```diff
@@ -16,5 +16,5 @@ outputs:
 icon: 'hello.svg' # vector art to display in the GitHub Marketplace
 color: 'green' # optional, decorates the entry in the GitHub Marketplace
 runs:
-  using: 'node12'
```
TingluoHuang, line 19: don't change L0 tests

src/Misc/layoutbin/runsvc.sh
```diff
@@ -13,7 +13,7 @@ fi
 # insert anything to setup env when running as a service
 
 # run the host process which keep the listener alive
-./externals/node12/bin/node ./bin/RunnerService.js &
```
TingluoHuang, line 16: Do we need some environment to make everything fall back to node12 in case some customer runs into an issue with node 16?

fhammerl: Good point. Should it be a comprehensive flag that would force all new references of node16 in this PR back to node12?

src/Misc/layoutbin/update.sh.template
fhammerl: I felt vary of introducing more logic into an already complicated process. There's of course the advantage of your approach making it easier to introduce node18 in the future :)

I was thinking it's a bit easier to remove the old logic once we're done with the migration. There's already the assumption that there's only one runner service active so I didn't see a harm in duplicating it like this (should probably put it in a function at least though).

src/Test/L0/Worker/ActionManagerL0.cs
fhammerl: Shouldn't we set up the tests for that too, then?

src/Misc/layoutbin/update.sh.template
```diff
@@ -135,7 +136,13 @@ if [[ "$currentplatform" == 'darwin'  && restartinteractiverunner -eq 0 ]]; then
     then
         # inspect the open file handles to find the node process
         # we can't actually inspect the process using ps because it uses relative paths and doesn't follow symlinks
-        path=$(lsof -a -g "$procgroup" -F n | grep node12/bin/node | grep externals | tail -1 | cut -c2-)
+        nodever="node12"
+        path=$(lsof -a -g "$procgroup" -F n | grep $nodever/bin/node | grep externals | tail -1 | cut -c2-)
+        if ! [[ $? -eq 0 && -n "$path" ]]
```
thboop, line 141: Nit: double check this condition

src/Misc/layoutbin/runsvc.sh
thboop: We should do this for all node16 changes in this pr (including hash files).

`RUNNER_FORCENODEVERSION` that we check the value of and try to use. We should add docs for this as well, and indicate you may be using an unsecure node version (node 12 has an end of life in April)

fhammerl: Sounds good, I'll create a constant with that name and comment the meaning there. We could also doc this during our next release. Where else should we document it?

src/Misc/layoutbin/runsvc.sh
```diff
@@ -10,10 +10,11 @@ if [ -f ".path" ]; then
     echo ".path=${PATH}"
 fi
 
-# insert anything to setup env when running as a service
+nodever=${GITHUB_ACTIONS_RUNNER_FORCED_NODE_VERSION:-node16}
```
fhammerl, line 13: `GITHUB_ACTIONS_RUNNER` seems to be the prefix we use most often for these kind of env vars

src/Runner.Common/Constants.cs
```diff
@@ -223,6 +223,7 @@ public static class Actions
             public static class Agent
             {
                 public static readonly string ToolsDirectory = "agent.ToolsDirectory";
+                public static readonly string ForcedNodeVersion = "GITHUB_ACTIONS_RUNNER_FORCED_NODE_VERSION";
```
fhammerl, line 226: Where else should we document this var? Release notes?

TingluoHuang: we don't document these, we will let customers know if they run into issues.

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace GitHub.Runner.Common.Util
+{
+    using System;
+    using System.Collections.ObjectModel;
```
TingluoHuang, line 4: put `using`  out of the `namespace`

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,16 @@
...
+
+    public static class NodeUtil
+    {
+        public const string LatestNodeVersion = "node16";
+        public static readonly ReadOnlyCollection<string> AllowedNodeVersions = new(new[] {"node12", "node16"});
+        public static string GetForcedOrLatestNodeVersion()
+        {
+            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedNodeVersion);
```
TingluoHuang, line 12: should we make sure `forcedNodeVersion` is set to node12 or node16, avoid customer set `forcedNodeVersion=node100` etc..

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace GitHub.Runner.Common.Util
+{
+    using System;
+    using System.Collections.ObjectModel;
+
+    public static class NodeUtil
+    {
+        public const string LatestNodeVersion = "node16";
```
TingluoHuang, line 8: `private const string _latestNodeVersion`

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,16 @@
...
+        public static readonly ReadOnlyCollection<string> AllowedNodeVersions = new(new[] {"node12", "node16"});
```
TingluoHuang, line 9: private static readonly ReadOnlyCollection<string> _allowedNodeVersions

TingluoHuang, line 9: `AllowedNodeVersions` -> `BuiltInNodeVersions` ?

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,16 @@
...
+        public static string GetForcedOrLatestNodeVersion()
```
TingluoHuang, line 10: `GetForcedOrLatestNodeVersion` - > `GetNodeVersion` ?

src/Runner.Common/Util/NodeUtil.cs
fhammerl: How about _defaultNodeVersion?

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.ObjectModel;
+
+namespace GitHub.Runner.Common.Util
+{
+
```
TingluoHuang, line 6: extra line. 😄

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -0,0 +1,17 @@
...
+    public static class NodeUtil
+    {
+        private const string _defaultNodeVersion = "node16";
+        public static readonly ReadOnlyCollection<string> BuiltInNodeVersions = new(new[] {"node12", "node16"});
+        public static string GetNodeVersion()
+        {
+            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedNodeVersion);
```
TingluoHuang, line 13: I thought we also are going to control this switch based on the server-side feature flag to some degree.

bryanmacfarlane: Huge plus 1 to this.  Ultimately, we need progressive rollout.

bryanmacfarlane: FF passed down in job message to runner ...

TingluoHuang (approved):

Issue #1261, opened 2021-08-12
Canceled and Cancelled are both used in actions/runner


jsoref:
**Describe the bug**
Canceled and Cancelled are both used in actions/runner

**To Reproduce**
1. Trigger a cancellable run (e.g. an `on:push` that has a cancel option)
2. Trigger a cancel for it (e.g. by pushing a second time)
3. Visit the run https://github.com/jsoref/dhall-validator-action/actions/runs/1125145505
4. See the output

**Expected behavior**
Only use a single version of `canceled` / `cancelled` throughout a single view, and preferably a single platform. Ideally this platform would be `github`, but for the sake of argument, let's imagine that the platform is `actions/runner`.

Given that `cancelled()` is the form exposed via actions/runner, I think it has to win, unless you want to add support for `canceled()` as well, which just seems more painful: https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/docs/adrs/0274-step-outcome-and-conclusion.md#L21

## Runner Version and Platform
Version of your runner?
Current runner version: '2.280.1'

OS of the machine running the runner? OSX/Windows/Linux/...
  Ubuntu
  20.04.2
  LTS

## What's not working?
Please include error messages and screenshots.

> Status: Cancelled
> Canceling since a higher priority waiting request for 'spelling-refs/heads/test' exists
> A task was canceled.
![image](https://user-images.githubusercontent.com/2119212/129266971-b55b3377-d0c8-4ce2-8a3d-ff885823522f.png)

## Instances

https://github.com/actions/runner/search?q=canceled
28 code results in actions/runner or view all results on GitHub

https://github.com/actions/runner/search?q=cancelled
34 code results in actions/runner or view all results on GitHub

## Doc instances

On the bright side, the docs web site is using a stemming search or something and yields identical results either way, so it would be harmless to fix everything to use a single spelling:

https://docs.github.com/en?query=cancelled
[Developers / GitHub Marketplace / Marketplace API usage Handling plan cancellations: Step 1. Cancellation event](https://docs.github.com/en/developers/github-marketplace/using-the-github-marketplace-api-in-your-app/handling-plan-cancellations#step-1-cancellation-event)
[...]

PR #1624, opened 2022-01-21
Issue 1261: inconsistency of outputs (both canceled and cancelled are used)

ruvceskistefan:
Problem was the inconsistent use of words canceled and cancelled.

Related issue: https://github.com/actions/runner/issues/1261

After this change, the word cancelled is used in all traces and logs. The problem of changing enum values still remains, because it requires major protocol changes.


thboop (dismissed):
LGTM

thboop (approved):
lgtm

PR #1625, opened 2022-01-24
Problem with debugging on macOS M1

ruvceskistefan:
Description: Problem with debugging actions/runner on macOS with M1 chip.

After this change, the user is able to debug the runner on macOS M1 without any problems.
Proposed changes were also tested on macOS with Intel processor, Linux and Windows, and no problems were noticed. Therefore, such configurations in the launch.json will work for all users.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,8 @@
             ],
             "cwd": "${workspaceFolder}/src",
             "console": "integratedTerminal",
-            "requireExactSource": false
+            "requireExactSource": false,
+            "targetArchitecture": "x86_64"
         },
         {
             "name": "Run",
@@ -24,7 +25,8 @@
             ],
             "cwd": "${workspaceFolder}/src",
             "console": "integratedTerminal",
-            "requireExactSource": false
+            "requireExactSource": false,
+            "targetArchitecture": "x86_64"
         },
         {
             "name": "Configure",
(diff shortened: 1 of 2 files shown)


fhammerl:
Hi @ruvceskistefan, thanks for this PR.

So if I got this right, you can't debug the runner on M1 osx machines using the configurations defined in launch.json. This change hides these configurations from the user so they can create launch configs that do run. 

Could you explain what one of these M1 compatible configs would look like, perhaps post an example or link a related issue?

ruvceskistefan:
Hi @fhammerl, you're welcome.
You understood well there are problems in debugging the runner on macOS M1 machine due to target architecture. My additional testing showed that the proposed change will not cause any problems on macOS with Intel processor, Linux or Windows machines. Therefore, it makes sense to add the proposed change to configurations within launch.json file. I added another commit in which this was done: changes in the docs/contribute/vscode.md were dropped, and "targetArchitecture": "x86_64" was added to all configurations in the launch.json file.

fhammerl:
@ruvceskistefan Could you include an example of a working M1 launch config in this PR?

Also, [targetArchitecture](https://code.visualstudio.com/docs/cpp/launch-json-reference#_targetarchitecture) seems to be deprecated, to me that suggests there's an alternative way to do separate the launch configs by architecture. What are your thoughts?

ruvceskistefan:
@fhammerl Of course, here is the content of the launch config file: [launch.docx](https://github.com/actions/runner/files/7942563/launch.docx)

I know that targetArchitecture is deprecated, but its use on macOS M1 is necessary, because the default architecture is arm64, which causes an error when debugging (Unable to attach to CoreCLR. Unknown Error: 0x80131c3c)

thboop (approved):
LGTM

PR #1635, opened 2022-01-27
Revert "Make `run.sh|cmd` handle update without quitting so container…

thboop:
…s using them as entrypoints don't exit on update (#1494)"

d8251bf9120d0f4ee25bb7c8c8e4fdece51f2847

We rolled this revert in 2.287.1, so we should go ahead and backport this fix to main


fhammerl (approved):
Backport looks good. I'll follow up with a PR to add these changes back in once we've fixed the pathing issues.

PR #1642, opened 2022-01-31
Add support for Step Summary 

pfleidi:
This is a work in progress: I'm opening this as a draft to get some feedback while figuring out how to properly test those changes.

## What does this change?

The changes in this PR will generate a unique file name per step run and make it available via the `$GITHUB_STEP_SUMMARY` environment variable. After a step finished executing, the contents of the file are uploaded as an attachment.


src/runnerversion
```diff
@@ -1 +1 @@
-2.287.0
+2.287.1
```
pfleidi, line 1: This is here to run locally without the runner trying to download a newer version.

I'm assuming we'll have to revert this change in this PR since we don't want to push a new version once this is merged, correct?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+          if(!ActionCommandManager.StepSummaryEnabled(step.ExecutionContext))
+          {
+              Trace.Info("Step Summary is disabled; skipping attachment upload");
+              return false;
+          }
+
+          var summaryExists = File.Exists(stepSummaryFilePath);
+          if (summaryExists)
+          {
+              Trace.Info($"Summary file ({stepSummaryFilePath}) does not exist; skipping attachment upload");
```
thboop, line 389: Are these expected in any scenarios? Should we Trace.Error if not?

src/runnerversion
thboop: This should be reverted, I have a pending change that will address this, its due to a hotfix we had to release off of the releases branch that hasn't been merged into `main` yet

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+              Trace.Info($"Summary file ({stepSummaryFilePath}) is empty; skipping attachment upload");
+              return false;
+          }
+
+          return true;
+        }
+
+        private string CreateStepSummaryFile(IStep step)
+        {
+            var stepSummaryDirectory = HostContext.GetDirectory(WellKnownDirectory.StepSummary);
```
thboop, line 406: Something to consider here, if the runner does not have permission to create this file, or we run out of disk space, should we try catch here so we don't fail the job for failing to upload this data?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+          var stepSummaryFilePath = executionContext.GetGitHubContext("step_summary");
+
+          if(!ActionCommandManager.StepSummaryEnabled(step.ExecutionContext))
+          {
+              Trace.Info("Step Summary is disabled; skipping attachment upload");
+              return false;
+          }
+
+          var summaryExists = File.Exists(stepSummaryFilePath);
+          if (summaryExists)
```
thboop, line 387: nit, consider `if (File.Exists(stepSummaryFilePath))`

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+          {
+              Trace.Info($"Summary file ({stepSummaryFilePath}) does not exist; skipping attachment upload");
+              return false;
+          }
+          Trace.Info($"Summary file exists: {stepSummaryFilePath}");
+
+          var summaryFileIsEmpty = new FileInfo(stepSummaryFilePath).Length == 0;
+          if (summaryFileIsEmpty)
```
thboop, line 395: Same nit as above:

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
         {
             var executionContext = step.ExecutionContext;
 
+            if (ShouldUploadAttachment(step))
+            {
+                QueueStepSummaryUpload(step);
```
thboop, line 369: Should we error or warn the user if the step summary data size > our technical limitation of 128kb. Would be weird to have that data truncated with no indication. We could put it in the github.com docs, but more front and center to user has historically helped limit user friction.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -103,6 +104,17 @@ public async Task RunAsync(IExecutionContext jobContext)
...
 
+                    if (ActionCommandManager.StepSummaryEnabled(step.ExecutionContext))
+                    {
+                        var stepSummaryFilePath = CreateStepSummaryFile(step);
+                        step.ExecutionContext.SetGitHubContext("step_summary", stepSummaryFilePath);
+                        envContext["GITHUB_STEP_SUMMARY"] = new StringContextData(stepSummaryFilePath);
+                    }
+                    else
+                    {
+                        Trace.Info("Step Summary is disabled; skipping file creation");
```
thboop, line 115: Clean 👍 , I like this tracing for troubleshooting

src/Runner.Worker/StepsRunner.cs
```diff
@@ -352,7 +364,70 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+            return stepSummaryFilePath;
+        }
+
+        private void QueueStepSummaryUpload(IStep step)
+        {
+            var executionContext = step.ExecutionContext;
+            var parentContext = executionContext.Root;
+            var stepSummaryFilePath = executionContext.GetGitHubContext("step_summary");
+            var stepID = step.ExecutionContext.Id;
+            var attachmentName = stepID.ToString();
```
thboop, line 427: My understanding is we create the file, and the user can put arbitrary data into it because we provide them with the Path, is that correct?

If so, we will want to run the Secret Masker on this file here, and replace the contexts with the secret masked contents

src/Runner.Worker/StepsRunner.cs
thboop: That can be done via the warning and error commands on the execution context of a step, like
`step.ExecutionContext.Error("The user will see this as an annotation")`

src/Runner.Worker/StepsRunner.cs
thboop: That could be done by getting all the lines in the file, then running them through the secret masker like

            foreach (string line in File.ReadLines())
            {
                var maskedLine = HostContext.SecretMasker.MaskSecrets(telemetryItem.Message);
                // rewrite the line here
            }

src/Runner.Worker/StepsRunner.cs
thboop: Note that the above does not go into our trace/diag telemetry, so sometimes its nice to both Trace.Error and create an annotation

src/Runner.Worker/StepsRunner.cs
```diff
@@ -103,6 +104,17 @@ public async Task RunAsync(IExecutionContext jobContext)
                     // Set GITHUB_ACTION
                     step.ExecutionContext.SetGitHubContext("action", actionStep.Action.Name);
 
+                    if (ActionCommandManager.StepSummaryEnabled(step.ExecutionContext))
```
TingluoHuang, line 107: Should we follow the pattern we have for https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable which is under `src/Runner.Worker/FileCommandManager.cs`?

src/Runner.Worker/StepsRunner.cs
TingluoHuang: +1 make sure you have the proper size limit before merging the PR.

yacaovsnc: Do we need `try/catch` here?  This should be a best effort operation.  If failed, we should log an annotation but don't fail the job/step?

src/Runner.Worker/StepsRunner.cs
thboop: Following this pattern would handle all the file creation/deletion per step for you, so its a good idea 👍

src/Runner.Worker/StepsRunner.cs
robherley: added in 901a406, lemme know if we want to be more verbose with specific exceptions!

src/Runner.Worker/StepsRunner.cs
robherley: np! fixed in 901a406

src/Runner.Worker/StepsRunner.cs
robherley: fixed! 901a406

src/Runner.Worker/StepsRunner.cs
robherley: We can probably move this secret scrubbing logic to when the file is dequeued and processed in [ServerQueue](https://github.com/actions/runner/blob/67ba8a7d429d953396c08550d8590af9889c4efe/src/Runner.Common/JobServerQueue.cs#L612):

ie:

```c#
else if (String.Equals(file.Type, ChecksAttachmentType.StepSummary, StringComparison.OrdinalIgnoreCase))
{
    // read step summary line by line
    // mask secrets & rewrite
    // send to CreateAttachmentAsync
}
```

cc @pfleidi

thboop: That makes sense to me, but we want to be 100% sure we do this before merging :). If it fails in any way (cannot write to file), we should abort the upload

TingluoHuang: I would suggest the opposite, which is don't change `JobServerQueue`.
- Log files that get queued to `JobServerQueue` are already run through the secret masker, I would suggest Attachment follow the same pattern.

pfleidi: @TingluoHuang That makes sense. I already have some code in place locally to scrub secrets from the markdown files. It uses `SecretUtility.ScrubSecrets()` like this: ```
var scrubbedSummaryContent = SecretUtility.ScrubSecrets(stepSummaryContent, false);
```

I'll look into `SecretMasker`, though. Is there any preference of one vs. the other?

Comment (TingluoHuang) [reply src/Runner.Worker/StepsRunner.cs:466]: Always use `HostContext.SecretMasker` since it has all the right secrets + encoder registered.

File: src/Runner.Worker/StepsRunner.cs (hunk 9 above)
Comment (pfleidi) [reply src/Runner.Worker/StepsRunner.cs:107]: @TingluoHuang Can you elaborate on which pattern you are referring to. I might be missing some context here. :)

File: src/Runner.Worker/StepsRunner.cs (hunk 8 above)
Comment (robherley) [reply src/Runner.Worker/StepsRunner.cs:466]: @TingluoHuang 

> Log files that get queued to JobServerQueue are already run through the secret masker

Is this only when the type is `CoreAttachmentType.Log`? I'm having trouble finding the code reference, just wanted to make sure it hits the same codepath for `ChecksAttachmentType.StepSummary`.

Comment (TingluoHuang) [reply src/Runner.Worker/StepsRunner.cs:466]: yes, `PagingLogger` is for regular step log.

Comment (TingluoHuang) [reply src/Runner.Worker/StepsRunner.cs:466]: we have some bad hardcode string in there... 😢 
```
_jobServerQueue.QueueFileUpload(_timelineId, _timelineRecordId, "DistributedTask.Core.Log", "CustomToolLog", _dataFileName, true);
```

File: src/Runner.Worker/StepsRunner.cs (hunk 9 above)
Comment (pfleidi) [reply src/Runner.Worker/StepsRunner.cs:107]: We talked about this offline and decided to implement a new file command like @TingluoHuang suggested 👍

File: src/Runner.Worker/ActionCommandManager.cs [hunk 10]
```diff
@@ -191,6 +191,11 @@ internal static bool EnhancedAnnotationsEnabled(IExecutionContext context)
         {
             return context.Global.Variables.GetBoolean("DistributedTask.EnhancedAnnotations") ?? false;
         }
+
+        internal static bool StepSummaryEnabled(IExecutionContext context)
```
TingluoHuang, line 195: i would move this into `SetStepSummaryCommand` and leave `ActionCommandManager.cs` unchange.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,99 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
             return text.Substring(originalIndex, lfIndex - originalIndex);
         }
     }
+
+    public sealed class SetStepSummaryCommand : RunnerService, IFileCommandExtension
+    {
+        public string ContextName => "step_summary";
+        public string FilePrefix => "step_summary_";
+
+        private const int AttachmentSizeLimit = 128 * 1024;
```
TingluoHuang, line 268: `private const int _attachmentSizeLimit = 128 * 1024;` and move it up before all `public string xxx`

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,99 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+
+        public Type ExtensionType => typeof(IFileCommandExtension);
+
+        public void ProcessCommand(IExecutionContext context, string filePath, ContainerInfo container)
+        {
+            if (ShouldUploadAttachment(context, filePath))
```
TingluoHuang, line 274: i would suggest not creating extra methods unless you are going to unit test them, if not, just inline all the logic under `ProcessCommand()`.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,99 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+
+        private bool ShouldUploadAttachment(IExecutionContext context, string filePath)
+        {
+            if (!ActionCommandManager.StepSummaryEnabled(context))
+            {
+                Trace.Info("Step Summary is disabled; skipping attachment upload");
+                return false;
+            }
+
+            if (String.IsNullOrEmpty(filePath))
```
TingluoHuang, line 291: i would combine all file exit/empty checks into a single `if` and trace.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,99 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                }
+            }
+
+            return scrubbedFilePath;
+        }
+
+        private void QueueStepSummaryUpload(IExecutionContext context, string filePath)
+        {
+            var stepID = context.Id;
+            var attachmentName = stepID.ToString();
```
TingluoHuang, line 344: what about `var attachmentName = context.Id.ToString()`?

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,99 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+
+            Trace.Info($"Queueing file ({filePath}) for attachment upload ({attachmentName})");
+            try
+            {
+                context.QueueAttachFile(ChecksAttachmentType.StepSummary, attachmentName, filePath);
+            }
+            catch (Exception e)
+            {
```
TingluoHuang, line 352: should customer see any warning or message if there step summary is not process properly.

src/Runner.Worker/FileCommandManager.cs
pfleidi: @TingluoHuang I can do that. Just out of curiosity: Is style consistency your main concern in that regard or is there something else I'm missing?

TingluoHuang: style consistency

src/Runner.Worker/FileCommandManager.cs
pfleidi: @TingluoHuang I found the separate trace statements useful when testing/debugging. I guess now that I have unit tests for all of them, they might not be as useful any more. Is that what you're aiming at?

TingluoHuang: we can revisit this when you inline all the methods. my main goal is to get the minimal key tracing, ex: `we don't upload the summary because it's not there or empty.`

src/Runner.Worker/FileCommandManager.cs
pfleidi: @TingluoHuang Good question! You mean like adding a warning visible to the customer when something went wrong? I think that's a great idea!

TingluoHuang: warning or just info might want to double-check with PM.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,73 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
             return text.Substring(originalIndex, lfIndex - originalIndex);
         }
     }
+
+    public sealed class SetStepSummaryCommand : RunnerService, IFileCommandExtension
```
TingluoHuang, line 263: maybe rename this to `CreateStepSummaryCommand`?

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,73 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                return;
+            }
+
+            if (String.IsNullOrEmpty(filePath) || !File.Exists(filePath))
+            {
+                Trace.Info($"Step Summary file ({filePath}) does not exist; skipping attachment upload");
+                return;
+            }
+
+            var fileSize = new FileInfo(filePath).Length;
```
TingluoHuang, line 286: `new FileInfo()` can throw on permission error with the file.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,73 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+
+                var attachmentName = context.Id.ToString();
+
+                Trace.Info($"Queueing file ({filePath}) for attachment upload ({attachmentName})");
+                context.QueueAttachFile(ChecksAttachmentType.StepSummary, attachmentName, scrubbedFilePath);
+            }
+            catch (Exception e)
+            {
+                Trace.Error($"Error while processing file file ({filePath}): {e}");
+                context.Error($"Error while enqueueing GITHUB_STEP_SUMMARY: {e.Message}");
```
TingluoHuang, line 327: What about `Fail to create step summary using 'GITHUB_STEP_SUMMARY': {e.Message}`

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,73 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                }
+
+                var attachmentName = context.Id.ToString();
+
+                Trace.Info($"Queueing file ({filePath}) for attachment upload ({attachmentName})");
+                context.QueueAttachFile(ChecksAttachmentType.StepSummary, attachmentName, scrubbedFilePath);
+            }
+            catch (Exception e)
+            {
+                Trace.Error($"Error while processing file file ({filePath}): {e}");
```
TingluoHuang, line 326: extra `file`. 😄

src/Runner.Worker/FileCommandManager.cs
pfleidi: I've moved the `try` block up by a few lines so exceptions that might be happening here are caught as well.

src/Runner.Worker/FileCommandManager.cs
pfleidi: Good catch 😆

src/Runner.Common/ExtensionManager.cs
```diff
@@ -60,6 +60,7 @@ public sealed class ExtensionManager : RunnerService, IExtensionManager
                 case "GitHub.Runner.Worker.IFileCommandExtension":
                     Add<T>(extensions, "GitHub.Runner.Worker.AddPathFileCommand, Runner.Worker");
                     Add<T>(extensions, "GitHub.Runner.Worker.SetEnvFileCommand, Runner.Worker");
+                    Add<T>(extensions, "GitHub.Runner.Worker.SetStepSummaryCommand, Runner.Worker");
```
TingluoHuang, line 63: `CreateStepSummaryCommand`

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -259,4 +259,76 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                if (fileSize > _attachmentSizeLimit)
+                {
+                    context.Error($"$GITHUB_STEP_SUMMARY supports content up a size of {_attachmentSizeLimit / 1024}k got {fileSize / 1024}k");
+                    Trace.Info($"Step Summary file ({filePath}) is too large ({fileSize} bytes); skipping attachment upload");
+
+                    return;
+                }
+
+                Trace.Verbose($"Step Summary file exists: {filePath} and has a file size of {fileSize} bytes");
+                var scrubbedFilePath = filePath + "-scrubbed";
```
TingluoHuang, line 304: maybe `filePath+Guid.NewGuid().ToString()` so you don't have to have 
```
if (File.Exists(scrubbedFilePath))
{
    File.Delete(scrubbedFilePath);
}
```

src/Runner.Common/ExtensionManager.cs
pfleidi: Good catch!

src/Runner.Worker/FileCommandManager.cs
pfleidi: @TingluoHuang This will make testing this code exponentially harder because I can't make assumptions about the file name any more. If it's not a big blocker for you I'd rather leave the file check in place. 🤔 

Any concerns?

TingluoHuang: you can keep what you have, but I think we can still remove 
```
if (File.Exists(scrubbedFilePath))
{
    File.Delete(scrubbedFilePath);
}
```
since `StreamWriter` should overwrite the existing file. You might want to double checkout in case I am wrong.

TingluoHuang (approved):

Issue #775, opened 2020-10-29
Allow to specify docker network options, such as mtu


sharonovd:
**Describe the enhancement**
It would be great to set up custom options for networks created for `container` jobs.
 As of now, it is only possible to pass options to `docker create` command via `jobs.<job_id>.container.options`

**Code Snippet**
This could look this way:
```
container:
  image: centos:7
  network:
    options:
      mtu: 1400
```

**Additional information**
See, according to https://github.com/moby/moby/issues/34981 , docker daemon options, such as mtu, are not passed to new bridge networks. So my CI jobs fail while trying to e. g. run `yum -y update`.

NOTE: if the feature request has been agreed upon then the assignee will create an ADR.   See docs/adrs/README.md

kim0:
Hello @sharonovd .. The lack of possibility to set MTU has bit me hard the past few days. Is there any progress on this front? Any workarounds available today to pass custom options at the `docker network create` step ? Thanks!

lasse-aagren:
Hack'ish workaround: if you are creating your own runners, like us, You can wrap every call to `docker network create...` like:

```
#!/usr/bin/env bash
$
# move current docker to docker.bin
# mv /path/to/bin/docker /path/to/bin/docker.bin
# place this script in /path/to/bin/docker

MTU=1460

if [[ $1 = "network" ]] && [[ $2 = "create" ]] 
then
    shift; shift #pop 2 first parameters
    /path/to/bin/docker.bin network create --opt com.docker.network.driver.mtu=$MTU "${@}"
else
    #just call docker as normal if not network create
    /usr/local/bin/docker.bin "${@}"
fi
```
Not a pretty hack - but it works :)

kim0:
Thanks for the hack! This was breaking pretty consistently for me, so I ended up up'ing the VPC MTU to 1500. Glad GCP allows that now. Things are OK for me now.

NicklasWallgren:
Thanks for the workaround!, we encountered the same issue using k3s and default flannel configuration.

michaelfindlater:
Hi, we've encountered the same issue and are unfortunately having to use the work around.

Would be excellent to see some more flexibility in the form of allowing options to be configured/passed (so we can specify a MTU) when creating Docker networks for workflows jobs.

mzwennes:
In the README it specifies the option to pass an optional MTU to the `RunnerDeployment`:

```
# Optional Docker containers network MTU
[...]

PR #1650, opened 2022-02-02
Introduce GITHUB_ACTIONS_RUNNER_DOCKER_NETWORK_MTU to override the default MTU on the docker network driver.

NicklasWallgren:
Allow customers to override the default MTU (of 1500) for networks created for the container jobs.

Fixes https://github.com/actions/runner/issues/775

### DIFF
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -280,10 +280,15 @@ namespace GitHub.Runner.Worker.Container
 
         public async Task<int> DockerNetworkCreate(IExecutionContext context, string network)
         {
+            int mtu;
+            if (!int.TryParse(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_DOCKER_NETWORK_MTU") ?? string.Empty, out mtu))
+            {
+                mtu = 1500;
+            }
 #if OS_WINDOWS
-            return await ExecuteDockerCommandAsync(context, "network", $"create --label {DockerInstanceLabel} {network} --driver nat", context.CancellationToken);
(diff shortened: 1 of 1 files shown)


NicklasWallgren:
@TingluoHuang Is there any chance that we will see this getting merged?

The code change is pretty crude, and it would probably be a better solution to pass some sort of docker context/configuration down to `DockerCommandManager`, but I guess it conforms to how env variables are handled in this repository.

This PR fixes a pretty common issue with MTU using the `actions-runner-controller` in k8s clusters.
https://github.com/actions-runner-controller/actions-runner-controller/issues?q=label%3A%22docker+mtu+issue%22+is%3Aclosed

NicklasWallgren:
An alternative is to derive the actual MTU via `docker network inspect bridge` and then use while creating the networks.

```
    MTU=$(docker network inspect bridge --format '{{index .Options "com.docker.network.driver.mtu"}}' 2>/dev/null); 
```

DawnGKoz904 (approved):

michaelfindlater:
Is it possible for someone to please take a look at this? 🙏

nebuk89:
I will follow up with @benwells :)

thboop:
Thanks digging into this @NicklasWallgren,

While we see the use case for this pr, we really don't want the runner to be a pass through for default configuration settings for docker. There are [other network options](https://docs.docker.com/engine/reference/commandline/network_create/) which could be helpful, but we don't want to expose them as runner features in this way.

Customizable default settings for the docker instance would be better set as a feature request to docker daemon itself. In the meantime, you have a few options:

- Shim the docker binary to set the mtu
- Modify the mtu of the vpc or network you are using to use the default mtu
- https://docs.github.com/en/actions/hosting-your-own-runners/customizing-the-containers-used-by-jobs

Issue #1536, opened 2021-12-06
fs.copyFileSync hangs for file created within same layer in overlay filesystem


ericsciple:
Working on a better description...

Since upgrading node 12, we are seeing fs.copyFileSync hang during some container scenarios.

We should revert the node 12 update and continue the investigation. It's worth noting node 16 has the same issue.

Related issues:
- https://github.com/docker/setup-buildx-action/issues/117
- https://github.com/nodejs/node/issues/40200

crazy-max:
Thanks @ericsciple!

ericsciple:
Working on understanding the repro now..

ericsciple:
Two immediate action items:
- Revert to older version of node 12
- Create simplest possible repro we can provide for nodejs

ericsciple:
fyi - the new runner finished rolling out to all accounts. We reverted back to the previous version of node12.

ericsciple:
OK here's the simplest repro that @ethomson and I could come up with:

- Follow [these three instructions](https://kubernetes.io/docs/tasks/debug-application-cluster/get-shell-running-container/#getting-a-shell-to-a-container) to get a shell to a container running in kubernetes
- Copy paste commands from [this script](https://github.com/ericsciple/testing/blob/main/copy-file/repro.sh)

Key details about the repro:
- Reproduces on AKS and EKS (not minikube and not docker)
- Uses [copy_file_range](https://man7.org/linux/man-pages/man2/copy_file_range.2.html) to perform the copy
- Copies a file from non-overlay file system to overlay file system, then from overlay file system to another location on the overlay file system (hangs)

ericsciple:
Opened an AKS [issue here](https://github.com/Azure/AKS/issues/2715)

thboop:
@ericsciple and I went through this and it appears to be fixed in AKS as well, and it is not reproducible on `ubuntu-latest`, so we should be good to re-update node12 and close this issue

PR #1537, opened 2021-12-06
revert node12 version due to fs.copyFileSync hang

ericsciple:
Related issue https://github.com/actions/runner/issues/1536

PR #1651, opened 2022-02-02
Revert "revert node12 version due to fs.copyFileSync hang https://git…

thboop:
…hub.com/actions/runner/issues/1536 (#1537)"

bef164a12fbe5cbd1e5c8b6ee43c827303d59ffa

Now that the kernel is fixed, lets update node12 again.

Resolves https://github.com/actions/runner/issues/1536

### DIFF
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -3,7 +3,7 @@ PACKAGERUNTIME=$1
 PRECACHE=$2
 
 NODE_URL=https://nodejs.org/dist
-NODE12_VERSION="12.13.1"
+NODE12_VERSION="12.22.7"
 NODE16_VERSION="16.13.0"
 
 get_abs_path() {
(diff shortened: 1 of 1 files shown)


TingluoHuang (approved):

TingluoHuang (approved):

jeremyd2019:
the release notes say "Bump node12 version to latest" but 12.22.7 is not the latest, I think 12.22.10 may be the latest (but I'm no node.js expert).

crazy-max:
This PR breaks users since `v2.288.0`: https://github.com/docker/setup-buildx-action/issues/117#issuecomment-1055780080. The node update has been already reverted in https://github.com/actions/runner/pull/1537 (cc @ericsciple) but this PR basically puts the exact same version back with the `fs.copyFileSync` hanging issue.

TingluoHuang:
@crazy-max base on https://github.com/actions/runner/issues/1536
the Linux kernel bug has been patched in 5.4, and both AWS and Azure have this patch in their latest VM image.

crazy-max:
@TingluoHuang Ok thanks! Maybe this user uses the runner on a kernel that is not patched yet. Sorry for the noise.

crazy-max:
Looks like this is tracked in https://bugs.launchpad.net/ubuntu/+source/linux-base/+bug/1953199

crazy-max:
@TingluoHuang I wonder if the runner instance could be smart enough to detect kernels that can cause issues like this. Maybe a warning message or set a minimal required kernel version in the runner config that says to upgrade the kernel before making an update of the runner?

TingluoHuang:
@crazy-max it's kernel + overlay filesystem combined that exposes the bug. 😢  it might be harder to detect the filesystem when you are inside it.

crazy-max:
@TingluoHuang Ah yeah true might be tricky

PR #1654, opened 2022-02-03
Allow mocked updates for E2E testing

fhammerl:
With this change, you can replace the incoming updateMessage with a custom json map `mock_update_messages.json`.

### Usage
- Add files like 'runner2.281.2.tar.gz' or 'runner2.283.0.zip' depending on your platform and mock version number in your runner root folder. 
- Set up a json map in `mock_update_messages.json`, where the key is **the version this mock applies to** and the value is the mocked update message object (only the targetVersion property matters at the moment), see example test json
- Set the GITHUB_ACTIONS_RUNNER_IS_MOCK_UPDATE env var to '1'
- Watch the terminal and update logs to confirm that the mock happened.


src/Runner.Listener/SelfUpdater.cs
```diff
@@ -266,14 +266,57 @@ private async Task DownloadLatestRunner(CancellationToken token)
...
+                    else 
+                    {
+                        archiveFile = null;
+                        _terminal.WriteLine($"Mock runner archive not found at {archiveFile} for target version {targetVersion}, proceeding with download instead");
+                        _updateTrace.Enqueue($"Mock runner archive not found at {archiveFile} for target version {targetVersion}, proceeding with download instead");
+                    }
                 }
+#endif
+                // archiveFile is not null only if we mocked it above
+                if (archiveFile == null)
```
thboop, line 310: Maybe IsNullOrEmpty to be consistent with the pattern in this file

thboop (approved):
LGTM

PR #1667, opened 2022-02-07
Use job execution context instead of step for adding summary attachments

robherley:
In #1642, the execution context used for `QueueAttachFile` was originally the [Job context](https://github.com/actions/runner/blob/e552c0cced4b841c91f4796578451d29e537ff14/src/Runner.Worker/StepsRunner.cs#L403) but was incorrectly refactored to use the [Step context](https://github.com/actions/runner/blob/2946801fb6c2afd960ba4d86e5269ca8ecf012ac/src/Runner.Worker/FileCommandManager.cs#L320). This is incorrect, since the API will use the job context as per: https://github.com/github/actions-dotnet/pull/10957

Also, attaching to the step context is not supported according to: https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Runner.Worker/DiagnosticLogManager.cs#L106

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -317,7 +317,8 @@ namespace GitHub.Runner.Worker
                 var attachmentName = context.Id.ToString();
 
                 Trace.Info($"Queueing file ({filePath}) for attachment upload ({attachmentName})");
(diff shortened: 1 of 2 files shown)


pfleidi (approved):
This looks great. Thanks for making that change!

TingluoHuang (approved):

Issue #1673, opened 2022-02-09
safe_sleep function...


mathieu-aubin:
**Describe the bug**
Seems like the function does achieve correct sleep times when neother sleep nor ping are present

i have taken the liberty of writing a little script showing the effects of the current code and a suggestion on how to essentially fix, under recent bash version, using SECONDS bash internal timer

the script i wrote can be found here https://0x0.st/oXL1.txt

Feel free to do whatever with this but as i said, it doesnt seem to work currently (if both ping and sleep are missing) so thats a good enough reasom to either FIX or remove useless code

Thanks!

mathieu-aubin:
oh, thats in
src/Misc/layoutroot/run-helper.sh.template

ruvceskistefan:
Hi @mathieu-aubin,
Thanks for filing this issue! I'm reproducing it now and I'll get back to you as soon as possible!

mathieu-aubin:
[![asciicast](https://asciinema.org/a/HSzeAOlQEqaNppQA7gqMRrzfQ.svg)](https://asciinema.org/a/HSzeAOlQEqaNppQA7gqMRrzfQ)

ruvceskistefan:
Thanks @mathieu-aubin for the very detailed description of the problem and the proposed solution! We're considering a solution so I'll probably create PR that will close this issue.

mathieu-aubin:
I could do a PR quick but honestly, i wanted to have confirmation beforehand. If the function is useless since ANY and EVERY places runners can be installed does for sure contain `sleep` command, there just is no need to keep the code.

Otherwise, remove the multiple IF checks and just use the method i proposed as a "no dependency, internal solution".

SECONDS variable has been in bash for... well.. 25+ years so i think you are safe with such a solution.

ruvceskistefan:
I created a PR that should close this issue. Thank you @mathieu-aubin for your contribution!

ruvceskistefan:
Hi @mathieu-aubin,
Due to the [feature request](https://github.com/actions/runner/issues/1698) to use `safe_sleep` instead of standard `sleep` function in all bash scripts, I created a new [PR](https://github.com/actions/runner/pull/1707) (already merged to the main). This PR contains changes that were in the previously linked PR (just in the new bash script), as well as those additional requests. Therefore, I will close this issue. Please feel free to reopen it if you think it is necessary.
Thanks again for your contribution and suggestions!

mathieu-aubin:
[...]

PR #1674, opened 2022-02-09
Issue 1673: modifications in safe_sleep function

ruvceskistefan:
Problem was the implementation of sleep in the case when both ping and sleep are missing in safe_sleep function.

Related issue: https://github.com/actions/runner/issues/1673

After this change, the problem described in the related issue has been fixed.

### DIFF
--- a/src/Misc/layoutroot/run-helper.sh.template
+++ b/src/Misc/layoutroot/run-helper.sh.template
@@ -11,18 +11,10 @@ fi
 shopt -s nocasematch
 
 safe_sleep() {
-    if [ ! -x "$(command -v sleep)" ]; then
-        if [ ! -x "$(command -v ping)" ]; then
-            SECONDS=0
-            while [[ $SECONDS != 5 ]]; do
-                :
-            done
-        else
-            ping -c 5 127.0.0.1 > /dev/null
-        fi
-    else
-        sleep 5
-    fi
+    SECONDS=0
+    while [[ $SECONDS != 5 ]]; do
+        :
+    done
 }
 
 SOURCE="${BASH_SOURCE[0]}"


mathieu-aubin:
the PR is somewhat okay... i just want to specify something:

the reason i used the `trap` method and the `DEBUG` signal is that when both are combined, the trap is executed everytime the/a function starts - which was great for making certain i had the most accurate times without adding extra commands time

That being said...

You should instead reset the `SECONDS=0` variable before you enter the while loop. This way, it is ran only when needed and no trap will run exessively - maybe read on that?

As i said, also, having the rest of the code in there becomes rather useless since the `SECONDS` (with bash, only) is internal and guaranteed. Checking for other stuff is a waste and so is using anything else, specially when uncertain of their existence.

`function safe_sleep() { SECONDS=0; while [[ $SECONDS != 5 ]]; do :; done }`

also... it would be a good habbit to comment unusual code like `trap` and `SECONDS` for future references by others. Its certainly obscure for MANY and would benefit from precisions.

Enjoy

ruvceskistefan:
Thanks for the reply, I replaced the `trap` method with resetting `SECONDS=0` before the while loop. As for other if conditions and unnecessary code, let them stay for now just in case, because `SECONDS` isn't POSIX-compliant, so POSIX-features-only shells (dash on Ubuntu) may have some problems.

ruvceskistefan:
@fhammerl I think you're right, then I'll get the unnecessary code out of `safe_sleep` function.
Thank you both @fhammerl and @mathieu-aubin for your suggestions and contributions!

fhammerl (approved):

fhammerl:
Approved! 

@ruvceskistefan if you're up for some follow-up, it may be worth looking into other usages of `sleep` where we don't have a safe fallback.

ruvceskistefan:
Thanks @fhammerl! Of course, I'll move on that after I solve the priority issues.

ruvceskistefan:
This does not have to be merged now, since the [PR#1707](https://github.com/actions/runner/pull/1707) has already been merged. That PR is an upgrade to this (in accordance with the @fhammerl's feature request from the comments), so there is no need for this PR to be merged. Therefore, this PR can be closed.

PR #1675, opened 2022-02-09
Add ability to specify runner group when creating service

samcook:
Add ability to specify runner group when creating service.

Simply passes through the runner group to the underlying config script, if specified.


fhammerl (approved):
@samcook Thanks for this enhancement! Looks good, I just ran a workflow on a runner configured with the script you updated. 

Nice parameter substitution.

PR #1679, opened 2022-02-10
Standardize casing for x64 and x86

ethanchewy:
When a user creates a self-hosted runner, it selects an architecture to use for the runner. 

We use this value in the runner details page. 

`X64` should be `x64` and `X86` should be `x86` for consistency.


TingluoHuang:
Will this change break any existing functionality during job assignments?

PR #1680, opened 2022-02-11
Move JobTelemetry and StepsTelemetry into GlobalContext.

TingluoHuang:
We introduced `GlobalContext` to `ExecutionContext` for holding shared objects of all `ExecutionContext` (Job + Steps).

Both `ActionsStepsTelemetry` and `JobTelemetry` fail into this `shared objects` category, so I am moving both into the `GlobalContext`.

Moving forward, you will not need to worry about which context (job vs. step) to use in order to get the right object.

image

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -15,8 +15,8 @@ using GitHub.DistributedTask.Pipelines;
 using GitHub.DistributedTask.Pipelines.ContextData;
 using GitHub.DistributedTask.Pipelines.ObjectTemplating;
 using GitHub.DistributedTask.WebApi;
-using GitHub.Runner.Common.Util;
 using GitHub.Runner.Common;
+using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
 using GitHub.Runner.Worker.Container;
 using GitHub.Services.WebApi;
@@ -52,8 +52,7 @@ namespace GitHub.Runner.Worker
         Dictionary<string, string> IntraActionState { get; }
         Dictionary<string, VariableValue> JobOutputs { get; }
         ActionsEnvironmentReference ActionsEnvironment { get; }
-        List<ActionsStepTelemetry> ActionsStepsTelemetry { get; }
-        List<JobTelemetry> JobTelemetry { get; }
+        ActionsStepTelemetry StepTelemetry { get; }
         DictionaryContextData ExpressionValues { get; }
         IList<IFunctionInfo> ExpressionFunctions { get; }
         JobContext JobContext { get; }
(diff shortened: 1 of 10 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -52,8 +52,7 @@ public interface IExecutionContext : IRunnerService
         Dictionary<string, string> IntraActionState { get; }
         Dictionary<string, VariableValue> JobOutputs { get; }
         ActionsEnvironmentReference ActionsEnvironment { get; }
-        List<ActionsStepTelemetry> ActionsStepsTelemetry { get; }
-        List<JobTelemetry> JobTelemetry { get; }
+        ActionsStepTelemetry StepTelemetry { get; }
```
TingluoHuang, line 55: Each `ExecutionContext` will hold a single `ActionsStepTelemetry` object, the object will get append to the `GlobalContext.ActionsStepsTelemetry` when `ExecutionContext.Complete()` is called.

src/Runner.Worker/JobRunner.cs
```diff
@@ -279,14 +279,13 @@ private async Task<TaskResult> CompleteJobAsync(IJobServer jobServer, IExecution
...
 
             // Make sure we don't submit secrets as telemetry
-            MaskTelemetrySecrets(jobContext.JobTelemetry);
-
-            Trace.Info("Raising job completed event.");
-            var jobCompletedEvent = new JobCompletedEvent(message.RequestId, message.JobId, result, jobContext.JobOutputs, jobContext.ActionsEnvironment, jobContext.ActionsStepsTelemetry, jobContext.JobTelemetry);
+            MaskTelemetrySecrets(jobContext.Global.JobTelemetry);
 
+            var jobCompletedEvent = new JobCompletedEvent(message.RequestId, message.JobId, result, jobContext.JobOutputs, jobContext.ActionsEnvironment, jobContext.Global.StepsTelemetry, jobContext.Global.JobTelemetry);
+            Trace.Info($"Raising job completed event: {StringUtil.ConvertToJson(jobCompletedEvent)}");
```
TingluoHuang, line 288: revert the trace after testing...

src/Runner.Worker/Handlers/Handler.cs
```diff
@@ -150,6 +118,40 @@ public override void Initialize(IHostContext hostContext)
             ActionCommandManager = hostContext.CreateService<IActionCommandManager>();
         }
 
+        protected string GetActionRef()
```
TingluoHuang, line 121: this method is only get moved down and changed from `public virtual` to `protected`

src/Runner.Worker/ExecutionContext.cs
thboop: For composite child steps we don't call `complete()` so I don't think you will be getting that data currently

TingluoHuang: thanks for pointing this out, i will take a look.

TingluoHuang: I am adding a new `PublishStepTelemetry()` to executionContext, so composite actions can use that to publish step telemetry.

thboop (approved):
LGTM

PR #1686, opened 2022-02-14
Force node16 for js actions if FF is set

fhammerl:


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -673,6 +673,11 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedNodeVersion, "node12");
             }
 
+            if (Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode16") ?? false)
+            {
+                Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion, "node16");
```
TingluoHuang, line 678: I would suggest we don't use an environmental variable to control this, since we can fully control the switch via feature flag from the service side.

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,18 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
             {
                 handler = HostContext.CreateService<INodeScriptActionHandler>();
-                (handler as INodeScriptActionHandler).Data = data as NodeJSActionExecutionData;
+                var nodeData = data as NodeJSActionExecutionData;
+                
+                // With node12 EoL in 03/2022, we want to be able to uniformly upgrade JS actions to node16
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode16") ?? false)
+                {             
+                    // The user can opt out of this behaviour by setting this variable to true
+                    if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion)))
```
thboop, line 64: I think this only applies to self hosted runners. We may need to also check job env context so non-self hosted runners can opt out: https://github.com/actions/runner/blob/main/src/Runner.Worker/ActionCommandManager.cs#L167

Do we have access to that here?

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -5,8 +5,8 @@
 using System.Threading.Tasks;
 using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Common;
+using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
-using Pipelines = GitHub.DistributedTask.Pipelines;
```
thboop, line 9: What do these changes do?

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,20 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
-                (handler as INodeScriptActionHandler).Data = data as NodeJSActionExecutionData;
+                var nodeData = data as NodeJSActionExecutionData;
+                
+                // With node12 EoL in 03/2022, we want to be able to uniformly upgrade all JS actions to node16 from the server
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode16") ?? false)
+                {   
+                    // The user can opt out of this behaviour by setting this variable to true, either in 'env' in their workflow or as an environment variable on their machine
+                    executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion, out var optOutWorkflowEnv);                    
+                    var optOutMachineEnv = Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion);
+                    if (!StringUtil.ConvertToBoolean(optOutWorkflowEnv ?? optOutMachineEnv))
```
thboop, line 66: we should convert these to bool individually and check if either is true imo.

WE could have a case here where  optOutWorkflowEnv is specifically set to "false" so we ignore the machineEnv. I don't envision it being likely, but may be better to cover the scenario anyways.

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,20 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
             else if (data.ExecutionType == ActionExecutionType.NodeJS)
             {
                 handler = HostContext.CreateService<INodeScriptActionHandler>();
-                (handler as INodeScriptActionHandler).Data = data as NodeJSActionExecutionData;
+                var nodeData = data as NodeJSActionExecutionData;
+                
+                // With node12 EoL in 03/2022, we want to be able to uniformly upgrade all JS actions to node16 from the server
```
thboop, line 60: super super super nit 04/2022

src/Runner.Worker/Handlers/HandlerFactory.cs
fhammerl: So, my logic here was to see if there's a `optOutWorkflowEnv` set, then evaluate it. If not, fall back to machine env vars and do the same.

> WE could have a case here where optOutWorkflowEnv is specifically set to "false" so we ignore the machineEnv.

Doesn't that happen here, though? If `optOutWorkflowEnv` is `"false"`, then `optOutMachineEnv` would be ignored

src/Runner.Common/Constants.cs
```diff
@@ -219,14 +219,15 @@ public static class Actions
                 public static readonly string AllowUnsupportedStopCommandTokens = "ACTIONS_ALLOW_UNSECURE_STOPCOMMAND_TOKENS";
                 public static readonly string RunnerDebug = "ACTIONS_RUNNER_DEBUG";
                 public static readonly string StepDebug = "ACTIONS_STEP_DEBUG";
+                public static readonly string AllowActionsUseUnsecureNodeVersion = "ALLOW_ACTIONS_USE_UNSECURE_NODE_VERSION";
```
thboop, line 222: nit can we do `ACTIONS_ALLOW...` as all of our other env's start with `ACTIONS_`

TingluoHuang: `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION`

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                {
+                    // The user can opt out of this behaviour by setting this variable to true, either setting 'env' in their workflow or as an environment variable on their machine
+                    executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion, out var workflowOptOut);
+                    var isWorkflowOptOutSet = !string.IsNullOrEmpty(workflowOptOut);
+                    if (isWorkflowOptOutSet)
+                    {
+                        var isWorkflowOptOut = StringUtil.ConvertToBoolean(workflowOptOut);
+                        if (!isWorkflowOptOut)
+                        {
+                            nodeData.Script = "node16";
```
TingluoHuang, line 71: `nodeData.NodeVersion = "node16"`

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                    else
+                    {
+                        // Fallback to local environment variable opt out
+                        var isLocalEnvOptOut = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion));
+                        if (!isLocalEnvOptOut)
+                        {
+                            nodeData.Script = "node16";
+                        }
+                    }
+                }
```
TingluoHuang, line 83: `nodeData.NodeVersion = "node16"`

TingluoHuang:
Can we split this PR into 2?
Looks like we are doing 2 different things in the PR.
- some cleanup work with `ForcedInternalNodeVersion`
- Adding new change to force all actions to use `node16`

fhammerl:
@TingluoHuang 
Closing this in favour of
Internal Node 16: https://github.com/actions/runner/pull/1715
JS action Node 16: https://github.com/actions/runner/pull/1716

Issue #1596, opened 2022-01-07
Unhelpful error message when writing malformed data to $GITHUB_ENV


jbzdarkid:
**Describe the bug**
When writing a particular string to `$GITHUB_ENV`, the runner throws a nullref, i.e. `Object reference not set to an instance of an object.`

**To Reproduce**
See https://github.community/t/nullref-when-writing-to-github-env-from-python/221606 . The following python code reproduces the issue.
```py
import environ
with open(environ['GITHUB_ENV'], 'a') as f:
  subcomment = """\
Please verify the following diffs:
- [ ] All articles ran in 0:00:56.697371: [en](https://wiki.tf/d/3106540) [ar](https://wiki.tf/d/3106518) [cs](https://wiki.tf/d/3106519) [da](https://wiki.tf/d/3106520) [de](https://wiki.tf/d/3106521) [es](https://wiki.tf/d/3106522) [fi](https://wiki.tf/d/3106523) [fr](https://wiki.tf/d/3106524) [hu](https://wiki.tf/d/3106525) [it](https://wiki.tf/d/3106526) [ja](https://wiki.tf/d/3106527) [ko](https://wiki.tf/d/3106528) [nl](https://wiki.tf/d/3106529) [no](https://wiki.tf/d/3106530) [pl](https://wiki.tf/d/3106531) [pt](https://wiki.tf/d/3106532) [pt-br](https://wiki.tf/d/3106533) [ro](https://wiki.tf/d/3106534) [ru](https://wiki.tf/d/3106535) [sv](https://wiki.tf/d/3106536) [tr](https://wiki.tf/d/3106537) [zh-hans](https://wiki.tf/d/3106538) [zh-hant](https://wiki.tf/d/3106539)
- [ ] Wanted templates ran in 0:00:37.264911: [en](https://wiki.tf/d/3106541)"""
  f.write('GITHUB_COMMENT<<EOF')
  f.write(subcomment)
  f.write('EOF')
```

**Expected behavior**
The bug here is about a lack of newline after the EOF marker. So, I would expect an error which says something like "malformed EOF marker", or "EOF marker too long". A very precient error might say "EOF marker missing newline".

## Runner Version and Platform
Runner version 2.285.1, Ubuntu 20.04.3

## What's not working?
https://github.com/jbzdarkid/TFWiki-scripts/runs/4728629242?check_suite_focus=true

## Job Log Output
```
2022-01-06T15:26:24.2881206Z ##[group]Run python master.py pull_request
[...]

PR #1687, opened 2022-02-14
Issue 1596: Runner throws null ref exception when new line after EOF is missing

ruvceskistefan:
Description: Runner throws null ref exception when new line after EOF is missing. The problem occurs when the string being written to the variable contains new line and a new line is missing after the EOF marker.

Related issue: #1596 

After this change, the described problem was removed and additional tests covering this part of the code were implemented.


thboop (approved):
LGTM

Issue #1677, opened 2022-02-09
Systemd service doesn't fail when runner listener crashes


bionade24:
Due to a permission issue the runner listener crashes, but the systemd service still shows up as running.
```
$ sudo systemctl status actions.runner.name-robotics-name_ci.image-build.service
● actions.runner.name-robotics-name_ci.image-build.service - GitHub Actions Runner (name-robotics-name_ci.image-build)
     Loaded: loaded (/etc/systemd/system/actions.runner.name-robotics-name_ci.image-build.service; enabled; vendor preset: enabled)
     Active: active (running) since Wed 2022-02-09 20:37:58 UTC; 8min ago
   Main PID: 20082 (runsvc.sh)
      Tasks: 8 (limit: 18966)
     Memory: 6.6M
     CGroup: /system.slice/actions.runner.name-robotics-name_ci.image-build.service
             ├─20082 /bin/bash /home/gh_actions_runner/runsvc.sh
             └─20085 ./externals/node12/bin/node ./bin/RunnerService.js

Feb 09 20:45:57 image-build runsvc.sh[20085]:    at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallo>
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at System.IO.Strategies.OSFileStreamStrategy..ctor(String path, FileMode mode, FileAccess access, FileShare share, FileOptions options, Int64 preallocati>
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at GitHub.Runner.Common.HostTraceListener.CreatePageLogWriter()
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at GitHub.Runner.Common.HostTraceListener..ctor(String logFileDirectory, String logFilePrefix, Int32 pageSizeLimit, Int32 retentionDays)
Feb 09 20:45:57 image-build runsvc.sh[20085]:    at GitHub.Runner.Common.HostContext..ctor(String hostType, String logFile)
[...]

PR #1693, opened 2022-02-17
Enhancement: RunnerService.js added logic to fail on N attempts if env variable exported

nikola-jokic:
Added behaviour so that system service could exit after a specified number of failures, depending on the `GITHUB_ACTIONS_SERVICE_EXIT_AFTER_N_FAILURES` env variable.

This feature should aid customers that have their own service templates to exit the service and leave restarting and notifying logic to the service they are using.

This change should not affect previous runners, that don't have this environment variable exported.

Added restart logic on status code 4 which is not caught before on the RunnerService.js, but is returned on update runOnce.

This change relates to #1677

### DIFF
--- a/src/Misc/layoutbin/RunnerService.js
+++ b/src/Misc/layoutbin/RunnerService.js
@@ -4,7 +4,6 @@
 
 var childProcess = require("child_process");
 var path = require("path");
-const { exit } = require("process");
 
 var supported = ["linux", "darwin"];
 
@@ -27,7 +26,7 @@ if (exitServiceAfterNFailures <= 0) {
 
 var consecutiveFailureCount = 0;
 
-var gracefulShutdown = function (code) {
+var gracefulShutdown = function () {
   console.log("Shutting down runner listener");
   stopping = true;
   if (listener) {
@@ -109,7 +108,7 @@ var runService = function () {
             console.error(
               `${messagePrefix}, exiting service after ${consecutiveFailureCount} consecutive failures`
             );
-            gracefulShutdown(5);
+            gracefulShutdown();
             return;
           } else {
             console.log(`${messagePrefix}, re-launch runner in 5 seconds.`);
@@ -130,9 +129,9 @@ runService();
 console.log("Started running service");
 
 process.on("SIGINT", () => {
-  gracefulShutdown(0);
+  gracefulShutdown();
 });
 
 process.on("SIGTERM", () => {
-  gracefulShutdown(0);
+  gracefulShutdown();
 });
--- a/src/Misc/layoutbin/RunnerService.js
+++ b/src/Misc/layoutbin/RunnerService.js
@@ -3,94 +3,136 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 var childProcess = require("child_process");
-var path = require("path")
+var path = require("path");
+const { exit } = require("process");
 
-var supported = ['linux', 'darwin']
+var supported = ["linux", "darwin"];
 
 if (supported.indexOf(process.platform) == -1) {
-    console.log('Unsupported platform: ' + process.platform);
-    console.log('Supported platforms are: ' + supported.toString());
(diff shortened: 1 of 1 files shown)


fhammerl:
Thanks, I brought these changes to main under https://github.com/actions/runner/tree/fhammerl/feature/1677 to run some e2e service upgrade tests, I'll get back to reviewing this PR after.

nikola-jokic:
Amazing! Thank you @fhammerl

src/Misc/layoutbin/RunnerService.js
```diff
@@ -3,94 +3,136 @@
...
+
+if (exitServiceAfterNFailures <= 0) {
+  exitServiceAfterNFailures = NaN;
 }
 
-runService();
-console.log('Started running service');
+var consecutiveFailureCount = 0;
 
 var gracefulShutdown = function (code) {
```
fhammerl, line 30: I know this was here before your changes, but do you know why we have `code` as a parameter here?

src/Misc/layoutbin/RunnerService.js
```diff
@@ -3,94 +3,136 @@
...
+      listener.on("error", (err) => {
+        console.log(`Runner listener fail to start with error ${err.message}`);
+      });
+
+      listener.on("close", (code) => {
+        console.log(`Runner listener exited with error code ${code}`);
+
+        if (code === 0) {
+          console.log(
+            "Runner listener exit with 0 return code, stop the service, no retry needed."
```
fhammerl, line 84: NIT / non-blocking: Normally I'm not a fan of long lines, but I feel with `console.log` especially it hurts readability a bit + leads to a noticeably longer function. How would you feel about marginally wider lines?

src/Misc/layoutbin/RunnerService.js
nikola-jokic: It is not used. I just did not want to change it if someone wanted later to do something with it. It is useless complitely

src/Misc/layoutbin/RunnerService.js
nikola-jokic: This was formatted by the prettier in order to follow prettier conventions. I guess that it is just JS convention and if you ask me, I would follow the formatting style of other tools, just like `go fmt` or c# formatter.
I could change it to maybe standard js, but regardless, I would follow the formatting of some external tooling, rather than formatting myself, for consistency 😊

src/Misc/layoutbin/RunnerService.js
nikola-jokic: Since the function is accepting a code, I wanted to send a status code if it is decided in the future that the code passed would be logged. I have noticed in several places that the code was passed as 0. My thought process was to pass the correct code from this function just in case any future modification would like to include the status code in the error message or logging itself.

src/Misc/layoutbin/RunnerService.js
```diff
@@ -3,94 +3,136 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 var childProcess = require("child_process");
-var path = require("path")
+var path = require("path");
+const { exit } = require("process");
```
fhammerl, line 7: Why do we need to import process now if we didn't use it before?

src/Misc/layoutbin/RunnerService.js
fhammerl: In that case, I'd say let's remove the arguments, and change the signature to `function () {`. 

Exit codes from the Runner.Listener do have different meanings and this might be confused to be one of them.

nikola-jokic: Of course

src/Misc/layoutbin/RunnerService.js
nikola-jokic: We don't. My bad... I will fix it now

fhammerl (approved):
LGTM!

Issue #1657, opened 2022-02-03
Step Order value for Post Steps can be incorrect if a job contains composite actions


thboop:
**Describe the bug**
We provide an Order field on steps that dictate what order the steps ran in , this helps the UI populate the step order. For jobs with a composite action and a number of post job steps, occasionally that order field can collide on post steps. This is not intentional, but is typically handled by server side code to figure out what order the steps should be in.

**To Reproduce**
Create an action in `local_action/action.yml`
```
name: test
description: test
runs:
  using: composite
  steps:
    - name: ok
      shell: bash
      run: |
        echo "OK"
```

Create a workflow
```
name: CI
on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
  workflow_dispatch:

jobs:
  build:
    runs-on: [ self-hosted ]
    steps:
      - uses: actions/checkout@v2
      - uses: ./local_action
      - uses: actions/setup-node@v2
```

Both of the post steps with end with order: 10, this is because our composite action handler is actually [iterating the order](https://github.com/actions/runner/blob/main/src/Runner.Worker/ExecutionContext.cs#L336) on steps despite it not being a step with a timeline record that is sent back to the server. We should probably use a dummy childTimelineRecord order for Embedded steps to fix this.

**Expected behavior**

## Runner Version and Platform
2.287.1
All OS

nikola-jokic:
Hi @thboop,

Thanks for filing this issue! I am reproducing it now and I will get back to you as soon as possible!

Jaynor:
Hi @nikola-jokic 
May I ask you if there is an update on this?

nikola-jokic:
Hey @Jaynor,

We have a PR fixing this problem which is currently in review.

Jaynor:
@nikola-jokic Thank you so much for the update!

ruffsl:
@github-actions , bad bot! say away! This remains still quite the blocker for:
- https://github.com/actions/runner/issues/1478#issuecomment-1532027544

neilime:
This issue is also applies to composite action using external action which have their own post action. The LIFO order is not respected.

Example with this action (https://github.com/hoverkraft-tech/ci-github-container/blob/main/actions/docker/build-image/action.yml) where: 

1. docker/setup main runs early
2. buildkit-cache-dance main runs later
3. job finishes
4. docker/setup-buildx-action cleanup runs first randomly
5. buildkit-cache-dance post runs after randomly

[...]

PR #1697, opened 2022-02-18
Fixed Post Steps order timeline

nikola-jokic:
Related: #1657

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -136,6 +136,8 @@ namespace GitHub.Runner.Worker
         private Guid _detailTimelineId;
         private bool _expandedForPostJob = false;
         private int _childTimelineRecordOrder = 0;
+        private int _childPostActionCounter = 0;
+
         private CancellationTokenSource _cancellationTokenSource;
         private TaskCompletionSource<int> _forceCompleted = new TaskCompletionSource<int>();
         private bool _throttlingReported = false;
@@ -1043,7 +1045,7 @@ namespace GitHub.Runner.Worker
             }
 
             var newGuid = Guid.NewGuid();
-            return CreateChild(newGuid, displayName, newGuid.ToString("N"), null, null, ActionRunStage.Post, intraActionState, _childTimelineRecordOrder - Root.PostJobSteps.Count, siblingScopeName: siblingScopeName);
+            return CreateChild(newGuid, displayName, newGuid.ToString("N"), null, null, ActionRunStage.Post, intraActionState, _childTimelineRecordOrder - (_childPostActionCounter++), siblingScopeName: siblingScopeName);
         }
     }
 
--- a/src/Test/L0/Worker/ExecutionContextL0.cs
+++ b/src/Test/L0/Worker/ExecutionContextL0.cs
@@ -381,6 +381,111 @@ namespace GitHub.Runner.Common.Tests.Worker
         [Fact]
         [Trait("Level", "L0")]
         [Trait("Category", "Worker")]
+        public void CreatePostChild_TimelineCheck()
+        {
+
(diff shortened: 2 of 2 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1043,7 +1045,7 @@ private IExecutionContext CreatePostChild(string displayName, Dictionary<string,
             }
 
             var newGuid = Guid.NewGuid();
-            return CreateChild(newGuid, displayName, newGuid.ToString("N"), null, null, ActionRunStage.Post, intraActionState, _childTimelineRecordOrder - Root.PostJobSteps.Count, siblingScopeName: siblingScopeName);
```
TingluoHuang, line 1046: add some L0 tests?

thboop, line 1046: I'm not sure this covers all the edge cases. The problem in question is this [line](https://github.com/actions/runner/blob/ec9830836b214bb7564277c3d2b53e1303e191ae/src/Runner.Worker/ExecutionContext.cs#L351). In particular, the call from [CreateEmbeddedChild](https://github.com/actions/runner/blob/ec9830836b214bb7564277c3d2b53e1303e191ae/src/Runner.Worker/ExecutionContext.cs#L375)

Emedded Children shouldn't have an timeline order that matters, I think we can just define a constant (like 99999), and pass it into the [Createchild recordOrder](https://github.com/actions/runner/blob/ec9830836b214bb7564277c3d2b53e1303e191ae/src/Runner.Worker/ExecutionContext.cs#L375) call here, that way we enter this [side](https://github.com/actions/runner/blob/ec9830836b214bb7564277c3d2b53e1303e191ae/src/Runner.Worker/ExecutionContext.cs#L347) of this if condition, rather then the side that iterates `_childTimelineRecordOrder`

nikola-jokic: While tracking the execution, Children would be created first, and then, when you hit [line](https://github.com/actions/runner/blob/15061a558c6d86a2b0da509f23b2e8e97b1d80f1/src/Runner.Worker/ActionRunner.cs#L139), it will attempt to call RegisterPostJobStep, which calls CreatePostChild, where the change was made. The way it worked is that you would set the timeline as  `2*NUMBER_OF_STEPS - NUMBER_OF_POST_STEPS`, which is a read-only property. That way, every POST_STEP would have the same timeline record and that is where the issue was found. By doing this, we are always setting the timeline to `2*NUMBER_OF_STEPS` and then, since the post steps should be in reverse order, and are called in the reverse order, we would just have one counter which would create a timeline record from `2*NUMBER_OF_STEPS-MAIN_STEP_INDEX`. For example, if we have:

1. checkout (has post action)
2. custom_action (does not have post action)
3. setup-node (has post action)

the timeline would be:
job: 1, checkout step: 2, custom_step: 3, setup_node: 4, checkout_post: 8, setup_node_post: 7

I did not want to change the way the logic works a lot, but since the previous code was a constant `(2 * NUMBER_OF_STEPS - NUMBER_OF_POST_STEPS) `, checkout_post and setup_node_post would both get the 8 and both hit the execution path you have described with [Createchild recordOrder](https://github.com/actions/runner/blob/ec9830836b214bb7564277c3d2b53e1303e191ae/src/Runner.Worker/ExecutionContext.cs#L375). This way, we are still hitting that execution path, but we are setting timeline attribute in reverse order from `2*NUMBER_OF_STEPS` to `NUMBER_OF_STEPS`

cread:
Is there a reason this has been closed? We've been waiting for a while for a fix for #1657 ...

nikola-jokic:
Yes, this fix was not covering all cases, so we decided to close this PR. @thboop will be implementing the entire fix so you should be expecting it soon. Thank you for showing interest @cread!

Issue #1698, opened 2022-02-18
Use safe_sleep instead of sleep function


ruvceskistefan:
**Describe the enhancement**
Bash script `run-helper.sh.template` contains implementation of `safe_sleep` function. However, this function is not used in all bash scripts: there are a couple of places where the standard `sleep` function is used. It would be good to adjust the `safe_sleep` function so that it receives one parameter (sleep time in seconds) just like the standard `sleep` function, and then use it wherever necessary.

**Additional information**
Script that uses standard `sleep` function is`update.sh.template`.

PR #1707, opened 2022-02-22
Issue 1698: Use safe_sleep executable in bash scripts

ruvceskistefan:
Problems were:
- the implementation of `sleep` in the case when both ping and sleep are missing in `safe_sleep` function
- usage of standard `sleep` function instead of `safe_sleep` in bash script

Related issue: #1698

After this change, `safe_sleep` is implemented as separate executable bash script, and scripts `run-helper.sh` and `update.sh` are using that function instead of standard `sleep` function.


src/Misc/layoutroot/run-helper.sh.template
```diff
@@ -44,17 +28,17 @@ elif [[ $returnCode == 1 ]]; then
     exit 0
 elif [[ $returnCode == 2 ]]; then
     echo "Runner listener exit with retryable error, re-launch runner in 5 seconds."
-    safe_sleep
+    ./bin/safe_sleep.sh 5
```
fhammerl, line 31: I'm always a bit iffy on relative paths, does is work fine if I start run.sh with an absolute path from a different directory?

ruvceskistefan: i changed to absolute paths and changed the location of the `safe_sleep` script, so it is not in bin folder.

Mart-Bogdan:
How this got merged? This is definitely endless loop.

https://github.com/actions/runner/blob/0ebdf9e83d51b0ab8c168a4869954466f3e2e88f/src/Misc/layoutroot/safe_sleep.sh#L3-L6


**EDIT**.  Oh, SECONDS is autoincremented by bash. didn't know that.

elupus:
Why is this not attempting to yield using a potentially available sleep call on the systems where that is available. Ie close to all systems. Could just do that inside the new loop.

Busy waiting is always really bad.

PR #1708, opened 2022-02-22
Use better exit codes and comparison

fhammerl:
Exiting with `exit 1` in `run-helper.sh` and then reacting to it with `if [[ $returnCode == 1 ]]; then` will also restart on generic errors occuring in run-helper, like how we `exit 1` in `run-helper.sh` (line 7) if it's run with sudo.

I suggest we indicate a restart request with `exit 2`, then react to it with `if [[ $returnCode -eq 2 ]]; then`

Also changes `==` to `-eq` for numerical comparison.

### DIFF
--- a/src/Misc/layoutroot/run.sh
+++ b/src/Misc/layoutroot/run.sh
@@ -15,7 +15,7 @@ while :;
 do
     "$DIR"/run-helper.sh $*
     returnCode=$?
-    if [[ $returnCode == 1 ]]; then
+    if [[ $returnCode -eq 2 ]]; then
         echo "Restarting runner..."
     else
         echo "Exiting runner..."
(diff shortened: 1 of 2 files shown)


src/Misc/layoutroot/run.sh
```diff
@@ -15,7 +15,7 @@ while :;
 do
     "$DIR"/run-helper.sh $*
     returnCode=$?
-    if [[ $returnCode == 1 ]]; then
```
thboop, line 18: Should we just change the exit code here? https://github.com/actions/runner/pull/1708/files#diff-869ba9f7fc8ba69126cf55bb734923c0a217d1b731213e48243fffb297b45b30R7

thboop: Choose something like `3` for non-retryable exit codes and use that convention as needed so we can create an a denylist for what errors are not retryable

thboop: We spoke in person

thboop (approved):
LGTM

PR #1709, opened 2022-02-23
Repaired escaping on container arguments

### LINKED ISSUE #1435: GitHub cannot handle the github context for pull_requests with html. `toJSON` fails.
**Describe the bug**
When a Pull Request is created with HTML in it, f.e. a Dependabot PR, the github event can not be parsed. The `toJson` method is not escaping the HTML parts correctly.

**To Reproduce**
1. Pass the github context to an action, with the `toJSON` function. Screenshot 2021-10-26 at 16 27 12
2. Try parsing the github context when passed as in a Pull_request workflow with `jq`. Screenshot 2021-10-26 at 16 28 10

I've created a repository with this setup. See PR.
1. Go to https://github.com/JeroenKnoops/test-github-pull-request-event/pull/3
2. Two workflows run, one on commit, and one on Pull Request. The commit is correct, the Pull Request fails because of the html in the request.
3. Look at the error-log of the Pull Request: In [line 735](https://github.com/JeroenKnoops/test-github-pull-request-event/runs/4010841146?check_suite_focus=true#step:3:735) you will see the double `\` in the html part of the body of the PR. This can not be decoded by `jq`, that's why the workflow fails. 
Screenshot 2021-10-26 at 16 23 39

**Expected behavior**
The toJSON function should encode the HTML content of the PR body correctly.

Using the GitHub cloud runners.

## What's not working?
```
parse error: Invalid escape at line 1, column 9651
```
See screenshot in step 3.

This has impact on **ALL** dependabot PRs made which triggering a pull_request event. You cannot use the `github` context in those cases.

nikola-jokic:
Closes #1435


src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -130,7 +130,7 @@ public async Task RunAsync(ActionRunStage stage)
                 {
                     if (!string.IsNullOrEmpty(arg))
                     {
-                        container.ContainerEntryPointArgs = container.ContainerEntryPointArgs + $" \"{arg.Replace("\"", "\\\"")}\"";
+                        container.ContainerEntryPointArgs = container.ContainerEntryPointArgs + $" \"{arg.Replace("\\", "\\\\").Replace("\"", "\\\"")}\"";
```
nikola-jokic, line 133: The issue here is when the JSON string was passed to the container. This is only escaping quotes, but if those quotes were previously escaped: `\"`, if we are not escaping the `\`, prior to the fix, the argument would be `\\"`, which would break the JSON string and the command would be wrong.

thboop: Does this fix the `HashFiles` Scenario? 

Can we set up a simple workflow in a [job container](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container)
that

```
- uses: actions/checkout@v2
- runs: echo ${{hashFiles("somefolderintherepo")}}
```
and validate if we are seeing it hash correctly inside/outside of job containers. I'm curious what the root bug here is, I don't really understand the fix or issue, it would be good to be a reproduction.

nikola-jokic: I am sorry, I linked the wrong issue on the PR. I have updated it and [here](https://github.com/actions/runner/issues/1435) is an issue I am fixing. Usecase would be [this](https://github.com/JeroenKnoops/test-github-pull-request-event/tree/use-action-with-toJSON) branch provided by the customer. So summarize, github context is passed to `toJson`, and later used as an argument. With escaping only ("), we are getting to the issue where we have json similar to this:
`"some_key": "some \" value",`
As you can see, " would be escaped but \ would not, which would cause (\\") which would break the JSON passed to the container

thboop: I'm worried that escaping `"` may break other users, we need to escape "\" to prevent injection style attacks, but we don't have the same issue with '"'. 

If a user needs to full escaping capabilities, we should likely redirect them to passing their JSON formatted strings into their action as an env variable like [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepswith). 

If we are going to JSON escape the args, should we handle other formatted data types like XML? My tendency here is to lean no, and let users figure out how to escape, or let the system (env variabels) do the escaping.

nikola-jokic: I understand, but the current filter in place is incomplete in escaping it for the injection style attack. 
File:
Screenshot 2022-03-07 at 12 00 02

Would cause docker command to be:
`/usr/bin/docker run --name b27b3f4b753ffb04cbe49c2746650_33475c --label 873830 --workdir /github/workspace --rm -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_RUN_ATTEMPT -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_REF_NAME -e GITHUB_REF_PROTECTED -e GITHUB_REF_TYPE -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_ACTION_REPOSITORY -e GITHUB_ACTION_REF -e GITHUB_PATH -e GITHUB_ENV -e GITHUB_STEP_SUMMARY -e RUNNER_OS -e RUNNER_ARCH -e RUNNER_NAME -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/nikola/Projects/work/runner/_layout/_work/_temp/_github_home":"/github/home" -v "/home/nikola/Projects/work/runner/_layout/_work/_temp/_github_workflow":"/github/workflow" -v "/home/nikola/Projects/work/runner/_layout/_work/_temp/_runner_file_commands":"/github/file_commands" -v "/home/nikola/Projects/work/runner/_layout/_work/step-order/step-order":"/github/workspace" 873830:486b27b3f4b753ffb04cbe49c2746650  "\\" -v /home/nikola:/home/nikola #"`

That is why I think that we should escape the `\` character as well. It does not have anything to do with JSON, but the way the escape is working when passed to the command.

thboop, line 133: Can we update this pr to make this same change to all the `ScriptHandlers` and lock it behind the feature flag
`DistributedTask.EnhancedRunnerEscaping`

PR #1710, opened 2022-02-24
Lets allow up to 150 characters for services on linux/mac

thboop:
This helps support users that use self hosted runners with very long runner/repository names to avoid collisions

Why 150?

Filenames in linux/mac are max of 255 characters, systemd is limited to 255 characters for the service name as well. While we could go all the way there, doubling the length from ~80 to ~150 should be plenty because **Runner names can only be 64 characters in length.** 

So after this change on linux/mac we should only trim the service name if the repository + org name exceeds 70 characters. That isn't a huge deal, as ~64 characters of runner name and 70 characters of org+repository should be unique enough where there are no collisions.

We can't do this on windows, as windows services have max name lengths that are very low, which is why we added this limit in the first place.

Fixing: https://github.com/actions/runner/issues/730

### DIFF
--- a/src/Runner.Listener/Configuration/ServiceControlManager.cs
+++ b/src/Runner.Listener/Configuration/ServiceControlManager.cs
@@ -51,8 +51,8 @@ namespace GitHub.Runner.Listener.Configuration
             if (serviceName.Length > MaxServiceNameLength)
             {
                 Trace.Verbose($"Calculated service name is too long (> {MaxServiceNameLength} chars). Trying again by calculating a shorter name.");
-                // Subtract 5 to add -xxxx random number on the end
-                int exceededCharLength = serviceName.Length - MaxServiceNameLength - 5;
+                // Add 5 to add -xxxx random number on the end
+                int exceededCharLength = serviceName.Length - MaxServiceNameLength + 5;
                 string repoOrOrgNameSubstring = StringUtil.SubstringPrefix(repoOrOrgName, MaxRepoOrgCharacters);
 
                 exceededCharLength -= repoOrOrgName.Length - repoOrOrgNameSubstring.Length;
--- a/src/Runner.Listener/Configuration/ServiceControlManager.cs
+++ b/src/Runner.Listener/Configuration/ServiceControlManager.cs
@@ -65,19 +65,12 @@ namespace GitHub.Runner.Listener.Configuration
                 {
                     runnerNameSubstring = StringUtil.SubstringPrefix(settings.AgentName, settings.AgentName.Length - exceededCharLength);
                 }
-                #pragma warning disable CS0162
-                if (AdditionalDigits > 0)
-                {
-                    var random = new Random();
-                    var num = random.Next((int)Math.Pow(10, AdditionalDigits-1),(int)Math.Pow(10, AdditionalDigits)).ToString();
-                    runnerNameSubstring +=$"-{num}";
-                    serviceName = StringUtil.Format(serviceNamePattern, repoOrOrgNameSubstring, runnerNameSubstring);
-                }
-                else
-                {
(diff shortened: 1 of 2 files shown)


src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -49,12 +49,12 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
 
             serviceName = StringUtil.Format(serviceNamePattern, repoOrOrgName, settings.AgentName);
 
-            if (serviceName.Length > 80)
```
TingluoHuang, line 52: should we always add some random chars once we start trimming names?
We don't know how customers are going to name their repository and their runner, so we don't have confidence about the service name will still be unique on the machine even after we trimmed X from NWO and Y from the runner name.

My idea is always to add 4 random chars to the end once we trim anything.

thboop: Yeah, I'll add a 4 digit number!

src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -65,13 +65,33 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
...
             Trace.Info($"Service name '{serviceName}' display name '{serviceDisplayName}' will be used for service configuration.");
         }
+        #if (OS_LINUX || OS_OSX)
+            const int MaxServiceNameLength = 150;
+            const int MaxRepoOrgCharacters = 70;
+            const int AdditionalDigits = 4;
+        #elif OS_WINDOWS
+            const int MaxServiceNameLength = 80;
+            const int MaxRepoOrgCharacters = 45;
+            const int AdditionalDigits = 0;
```
TingluoHuang, line 94: should we also do this for windows?

src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -65,13 +65,33 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
                 {
                     runnerNameSubstring = StringUtil.SubstringPrefix(settings.AgentName, settings.AgentName.Length - exceededCharLength);
                 }
-
-                serviceName = StringUtil.Format(serviceNamePattern, repoOrOrgNameSubstring, runnerNameSubstring);
+                #pragma warning disable CS0162
+                if (AdditionalDigits > 0)
+                {
+                    var random = new Random();
+                    var num = random.Next((int)Math.Pow(10, AdditionalDigits-1),(int)Math.Pow(10, AdditionalDigits)).ToString();
```
TingluoHuang: what about `IOUtil.GetSha256Hash(HostContext.GetDirectory(WellKnownDir.Root)).SubString(0, 4)`?
Or just `random.Next(1000, 9999).ToString();`

src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -66,12 +66,25 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
...
             }
 
             serviceDisplayName = StringUtil.Format(serviceDisplayNamePattern, repoOrOrgName, settings.AgentName);
 
             Trace.Info($"Service name '{serviceName}' display name '{serviceDisplayName}' will be used for service configuration.");
         }
+        #if (OS_LINUX || OS_OSX)
+            const int MaxServiceNameLength = 150;
+            const int MaxRepoOrgCharacters = 70;
+            const int AdditionalDigits = 4;
```
TingluoHuang, line 83: move this out the `#if` since both have it?

TingluoHuang (approved):

Issue #519, opened 2020-06-07
macOS launctl services sets incorrect userName


kylef:
**Describe the bug**

The generated launctl plist from `./svc.sh install` is generated with incorrect userName. The userName of the account the script was run from is not used.

**To Reproduce**
1. `ssh` into the machine via a personal account (perhaps administrator)
2. use `su` to change user to a service account where you wish to run the actions runner
3. go to your github actions runner directory
4. Run `./svc.sh install`

**Expected behavior**

I have a generated, working launctl plist which will run as the current user.

**Actual behaviour**

## Runner Version and Platform

`actions-runner-osx-x64-2.263.0`

## What's not working?

`UserName` key is set to `kyle` not `palaver` as current user, see output below.

## Runner and Worker's Diagnostic Logs

```
$ id
uid=502(palaver) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),703(com.apple.sharepoint.group.3),702(com.apple.sharepoint.group.2),100(_lpoperator),701(com.apple.sharepoint.group.1)

$ ./svc.sh install
Creating launch runner in /Users/palaver/Library/LaunchAgents/actions.runner.kylef-palaver-ios.applesauce.plist
Creating /Users/palaver/Library/Logs/actions.runner.kylef-palaver-ios.applesauce
Creating /Users/palaver/Library/LaunchAgents/actions.runner.kylef-palaver-ios.applesauce.plist
Creating runsvc.sh
Creating .service
svc install complete

$ cat /Users/palaver/Library/LaunchAgents/actions.runner.kylef-palaver-ios.applesauce.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>Label</key>
    <string>actions.runner.kylef-palaver-ios.applesauce</string>
    <key>ProgramArguments</key>
    <array>
      <string>/Users/palaver/actions-runner/runsvc.sh</string>
    </array>
    <key>UserName</key>
    <string>kyle</string>
    <key>WorkingDirectory</key>
    <string>/Users/palaver/actions-runner</string>
    <key>RunAtLoad</key>
    <true/>
    <key>StandardOutPath</key>
    <string>/Users/palaver/Library/Logs/actions.runner.kylef-palaver-ios.applesauce/stdout.log</string>
[...]

PR #1714, opened 2022-02-25
Prefer user who initiated install before

nikola-jokic:
Related: #519

### DIFF
--- a/src/Misc/layoutbin/darwin.svc.sh.template
+++ b/src/Misc/layoutbin/darwin.svc.sh.template
@@ -63,7 +63,7 @@ function install()
     mkdir -p "${log_path}" || failed "failed to create ${log_path}"
 
     echo Creating ${PLIST_PATH}
-    sed "s/{{User}}/${SUDO_USER:-$USER}/g; s/{{SvcName}}/$SVC_NAME/g; s@{{RunnerRoot}}@${RUNNER_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
+    sed "s/{{User}}/${USER:-$SUDO_USER}/g; s/{{SvcName}}/$SVC_NAME/g; s@{{RunnerRoot}}@${RUNNER_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
     mv "${TEMP_PATH}" "${PLIST_PATH}" || failed "failed to copy plist"
 
     # Since we started with sudo, runsvc.sh will be owned by root. Change this to current login user.


nikola-jokic:
Hey, I just wanted to point out that this fix is related to using the current user to create a service. This script will fail if directory permission is not set properly on the runner directory itself. While reviewing, if you get permission denied, it is most likely due to directory permissions on the runner project directory, since the copying and rewriting using `sed` is done, and the correct file and directory permissions should be in place for this fix to be working well.

I did not change anything about that since the same thing would apply to the `$SUDO_USER`. What this script does is registers this service for the user invoking ./svc.sh, but additional steps should be taken by the user who wanted to allow other users to register the service in order for this script to take place. As I pointed out, if the user who initiated the `sudo` (`$SUDO_USER`) does not have the same permissions, the same problem would arise.

src/Misc/layoutbin/darwin.svc.sh.template
```diff
@@ -63,7 +63,7 @@ function install()
     mkdir -p "${log_path}" || failed "failed to create ${log_path}"
 
     echo Creating ${PLIST_PATH}
-    sed "s/{{User}}/${SUDO_USER:-$USER}/g; s/{{SvcName}}/$SVC_NAME/g; s@{{RunnerRoot}}@${RUNNER_ROOT}@g; s@{{UserHome}}@$HOME@g;" "${TEMPLATE_PATH}" > "${TEMP_PATH}" || failed "failed to create replacement temp file"
```
thboop, line 66: This [syntax](https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html) indicates to use `SUDO_USER` if it is set, otherwise we should be fallback to `USER`. I think that is the correct behavior here,  Is sudo user not getting set for some reason?

nikola-jokic: No, you are right, it is getting set. We need sudo for the systemd service to run install, but we don't on darwin. The issue here is that the behaviour is confusing to the user issuing the installation of the service.
If let's say I am user `nikola`, and I want to switch to the user `test` to install this service, I would issue the command: `sudo su test`. Now I am a test user, but my `SUDO_USER` would be nikola. When I run `./svc.sh install`, the service would be installed for user `nikola` who issued this command. It would be clearer, since we don't need sudo permissions to enable service on the darwin architecture, to install service as the user issuing the `./svc.sh install`.

To install this service as the user `test`, I would need to issue another sudo command changing the `SUDO_USER`, or to unset the `SUDO_USER`. This behaviour makes it a bit tricky for users who want to issue installation of the service, to install it as another user.

This fix tries to prefer the user invoking the installation, so when I switch to user test, and invoke installation, I would install this service as the user who invoked the service. No the one who ran `sudo su test`. We can do it differently, like in systemd, to pass an argument to install service as another user, but this seemed like a quicker fix, without changing the code that much.

PR #1716, opened 2022-02-25
Force JS Actions Node version to 16 if FF is on unless user opted out

fhammerl:


src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
             else if (data.ExecutionType == ActionExecutionType.NodeJS)
             {
                 handler = HostContext.CreateService<INodeScriptActionHandler>();
-                (handler as INodeScriptActionHandler).Data = data as NodeJSActionExecutionData;
```
TingluoHuang, line 58: do you mind adding some L0 tests?

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                var nodeData = data as NodeJSActionExecutionData;
+
+                // With node12 EoL in 04/2022, we want to be able to uniformly upgrade all JS actions to node16 from the server
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode16") ?? false)
+                {
+                    // The user can opt out of this behaviour by setting this variable to true, either setting 'env' in their workflow or as an environment variable on their machine
+                    executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion, out var workflowOptOut);
+                    var isWorkflowOptOutSet = !string.IsNullOrEmpty(workflowOptOut);
```
TingluoHuang, line 65: `StringUtil.ConvertToBoolean();` will check `string.IsNullOrEmpty()` for you

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                    if (isWorkflowOptOutSet)
+                    {
+                        var isWorkflowOptOut = StringUtil.ConvertToBoolean(workflowOptOut);
+                        if (!isWorkflowOptOut)
+                        {
+                            nodeData.NodeVersion = "node16";
```
TingluoHuang, line 71: should we only force if the `NodeVersion == node12`?

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -55,7 +55,33 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
             else if (data.ExecutionType == ActionExecutionType.NodeJS)
             {
                 handler = HostContext.CreateService<INodeScriptActionHandler>();
-                (handler as INodeScriptActionHandler).Data = data as NodeJSActionExecutionData;
+                var nodeData = data as NodeJSActionExecutionData;
+
+                // With node12 EoL in 04/2022, we want to be able to uniformly upgrade all JS actions to node16 from the server
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode16") ?? false)
+                {
```
TingluoHuang, line 62: What about something like:
```C#
var optOut = false;
executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion, out var workflowOptOutString);
var workflowOptOut = StringUtil.ConvertToBoolean(workflowOptOutString);
var envOptOut = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion));
optOut = workflowOptOut || envOptOut;
if(!optOut && string.equal(nodeData.NodeVersion, "node12", stringcompare.orgnalIgnoreCase))
{
  nodeData.NodeVersion = "node16";
}

```

src/Runner.Worker/Handlers/HandlerFactory.cs
fhammerl: Yes, but it will also convert `null` or `""` to false. We want to be able to differentiate between unset workflow env and workflow env set to false, because:
- if it's unset we just fallback to machine env
- it it's set to false we ignore machine env

Lmk what you think or if I could make this more obvious in the code.

https://github.com/actions/runner/pull/1686#discussion_r809393166

src/Runner.Worker/Handlers/HandlerFactory.cs
fhammerl: Fair 👍

TingluoHuang (approved):

Issue #353, opened 2020-03-02
Default non-Windows bash is not invoked as documented


johnstevenson:
**Describe the bug**
The documentation at https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsrun implies that when a shell is not specified on a non-Windows platform, bash (if found) will be invoked as `bash --noprofile --norc -eo pipefail {0}`.

> Fail-fast behavior using set -e o pipefail: Default for bash and built-in shell. It is also the default when you don't provide an option on non-Windows platforms.

However, this is not exactly the case.

**To Reproduce**
1. Create an action without specifying a shell value: 
```yml
    - name: No shell specified (default bash)
      run: echo Hello, world!
```
2. Observe that it is executed as `shell: /bin/bash -e {0}` as per this example at https://github.com/johnstevenson/runner-actions-test/actions/runs/48283923 which also shows the documented behaviour when the shell is specified.

![action-error-bash](https://user-images.githubusercontent.com/881777/75700964-d1117780-5caa-11ea-8da9-1d5fd13cc874.png)

**Expected behavior**
For the command to be executed as `shell: /bin/bash --noprofile --norc -e -o pipefail {0}`

## Runner Version and Platform
Current runner version: 2.165.2
Tested on ubuntu-latest and macos-latest

## What's not working?
Regardless of whether this is a documentation problem (or my incorrect interpretation), there is no consistency when invoking bash with and without a shell value.

The `ScriptHandler:RunAsync` code (also duplicated in `ScriptHandler:PrintActionDetails`) sets `shellCommand` as `sh` even if bash is found, resulting in the wrong format string being used:

https://github.com/actions/runner/blob/d80ab095a55324be2777de87a3c70791ae917885/src/Runner.Worker/Handlers/ScriptHandler.cs#L152-L168

So any fix would be:

```cs
#if OS_WINDOWS
                shellCommand = "pwsh";
                commandPath = WhichUtil.Which(shellCommand, require: false, Trace, prependPath);
                if (string.IsNullOrEmpty(commandPath))
                {
                    shellCommand = "powershell";
                    Trace.Info($"Defaulting to {shellCommand}");
                    commandPath = WhichUtil.Which(shellCommand, require: true, Trace, prependPath);
                }
                ArgUtil.NotNullOrEmpty(commandPath, "Default Shell");
#else
                shellCommand = "bash";
                commandPath = WhichUtil.Which(shellCommand, require: false, Trace, prependPath)
[...]

PR #1717, opened 2022-02-28
defaulting to bash in Non-Windows script handler

nikola-jokic:
Closes: #353

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -85,11 +85,15 @@ namespace GitHub.Runner.Worker.Handlers
                     }
                 }
 #else
-                shellCommand = "sh";
-                if (validateShellOnHost)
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -194,8 +198,16 @@ public async Task RunAsync(ActionRunStage stage)
                 }
                 ArgUtil.NotNullOrEmpty(commandPath, "Default Shell");
 #else
-                shellCommand = "sh";
-                commandPath = WhichUtil.Which("bash", false, Trace, prependPath) ?? WhichUtil.Which("sh", true, Trace, prependPath);
+                shellCommand = "bash";
```
thboop, line 201: LGTM, can we add an L0 for this?

thboop:
Can we update the docs to describe this behavior, making this change now will break some users.

PR #1726, opened 2022-03-02
worker: expose `github.triggering_actor` as an env-var

aybabtme:
Reruns will distinguish between the executing (`github.actor`, whom created the first run) and triggering (`github.triggering_actor`, whom requested the rerun) soon, and the `github.actor` will always be stable across reruns. To allow users to act differently based on the triggering actor, we're adding support for `github.triggering_actor`. Since `"actor"` is an allowed field that ends up in the env vars exposed to workflow steps, we concluded that `triggering_actor` also should be allowed.

### DIFF
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -13,6 +13,7 @@ namespace GitHub.Runner.Worker
             "action_ref",
             "action_repository",
             "actor",
+            "triggering_actor",
             "api_url",
             "base_ref",
             "env",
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/GitHubContext.cs
```diff
@@ -13,6 +13,7 @@ public sealed class GitHubContext : DictionaryContextData, IEnvironmentContextDa
             "action_ref",
             "action_repository",
             "actor",
+            "triggering_actor",
```
TingluoHuang, line 16: move it down a few lines to make these orders right. 😄

TingluoHuang (approved):

PR #1730, opened 2022-03-04
Send postlines via websocket if we can

yaananth:
For https://github.com/github/c2c-actions-service/issues/2828

- Service: https://github.com/github/actions-dotnet/pull/11327 adds new websocket controller
- This: https://github.com/actions/runner/pull/1730 using the socket

Please check the service PR for context and risk analysis

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -3,10 +3,13 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Sdk;
+using Newtonsoft.Json;
 using Pipelines = GitHub.DistributedTask.Pipelines;
 
 namespace GitHub.Runner.Common
@@ -17,7 +20,7 @@ namespace GitHub.Runner.Common
         TaskCompletionSource<int> JobRecordUpdated { get; }
         event EventHandler<ThrottlingEventArgs> JobServerQueueThrottling;
         Task ShutdownAsync();
-        void Start(Pipelines.AgentJobRequestMessage jobRequest);
+        void Start(Pipelines.AgentJobRequestMessage jobRequest, ServiceEndpoint serviceEndpoint);
         void QueueWebConsoleLine(Guid stepRecordId, string line, long? lineNumber = null);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
         void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRecord);
@@ -71,7 +74,7 @@ namespace GitHub.Runner.Common
 
         // Web console dequeue will start with process queue every 250ms for the first 60*4 times (~60 seconds).
         // Then the dequeue will happen every 500ms.
-        // In this way, customer still can get instance live console output on job start, 
+        // In this way, customer still can get instance live console output on job start,
         // at the same time we can cut the load to server after the build run for more than 60s
         private int _webConsoleLineAggressiveDequeueCount = 0;
         private const int _webConsoleLineAggressiveDequeueLimit = 4 * 60;
@@ -79,15 +82,23 @@ namespace GitHub.Runner.Common
         private bool _webConsoleLineAggressiveDequeue = true;
         private bool _firstConsoleOutputs = true;
 
+        private ClientWebSocket _websocketClient = null;
+
+        private ServiceEndpoint _serviceEndPoint;
+
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _jobServer = hostContext.GetService<IJobServer>();
         }
 
-        public void Start(Pipelines.AgentJobRequestMessage jobRequest)
+        public void Start(Pipelines.AgentJobRequestMessage jobRequest, ServiceEndpoint serviceEndpoint)
         {
             Trace.Entering();
+
+            this._serviceEndPoint = serviceEndpoint;
+
+            InitializeWebsocket();
 
             if (_queueInProcess)
             {
@@ -155,6 +166,9 @@ namespace GitHub.Runner.Common
             Trace.Verbose("Draining timeline update queue.");
             await ProcessTimelinesUpdateQueueAsync(runOnce: true);
             Trace.Info("Timeline update queue drained.");
+
+            Trace.Info($"Disposing websocket client ...");
+            this._websocketClient?.Dispose();
 
             Trace.Info("All queue process tasks have been stopped, and all queues are drained.");
         }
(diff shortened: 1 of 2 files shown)


src/Runner.Common/JobServerQueue.cs
```diff
@@ -17,7 +20,7 @@ public interface IJobServerQueue : IRunnerService, IThrottlingReporter
         TaskCompletionSource<int> JobRecordUpdated { get; }
         event EventHandler<ThrottlingEventArgs> JobServerQueueThrottling;
         Task ShutdownAsync();
-        void Start(Pipelines.AgentJobRequestMessage jobRequest);
+        void Start(Pipelines.AgentJobRequestMessage jobRequest, ServiceEndpoint serviceEndpoint);
```
yaananth, line 23: from @TingluoHuang : jobRequest already has the endpoint!

src/Runner.Common/JobServerQueue.cs
```diff
@@ -391,6 +437,35 @@ private async Task ProcessFilesUploadQueueAsync(bool runOnce = false)
...
+            if (_serviceEndPoint.Authorization != null &&
+                _serviceEndPoint.Authorization.Parameters.TryGetValue(EndpointAuthorizationParameters.AccessToken, out var accessToken) &&
+                !string.IsNullOrEmpty(accessToken))
+            {
+                if (_serviceEndPoint.Data.TryGetValue("FeedStreamUrl", out var feedStreamUrl) && !string.IsNullOrEmpty(feedStreamUrl))
+                {
+                    Trace.Info($"Creating websocket client ..." + feedStreamUrl);
+                    this._websocketClient = new ClientWebSocket();
+                    this._websocketClient.Options.SetRequestHeader("Authorization", $"Bearer {accessToken}");
+                    Task.Run(async () =>
```
yaananth, line 451: from: @TingluoHuang : store the task and then `await` later

src/Runner.Common/JobServerQueue.cs
```diff
@@ -391,6 +437,35 @@ private async Task ProcessFilesUploadQueueAsync(bool runOnce = false)
             }
         }
 
+        private void InitializeWebsocket()
```
yaananth, line 440: @TingluoHuang : random back off / timespan parameter in here

src/Runner.Common/JobServerQueue.cs
```diff
@@ -292,14 +306,46 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
...
+                                    Trace.Info($"Sending to websocket: {jsonData}");
+                                    try
+                                    {
+                                        // It should be okay to wait for the result since we are already doing this in the background and doing it one by one
+                                        await this._websocketClient.SendAsync(new ArraySegment<byte>(Encoding.UTF8.GetBytes(jsonData)), WebSocketMessageType.Text, false, default(CancellationToken));
+                                        pushedLinesViaWebsocket = true;
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        Trace.Info("Catch exception during append web console line to websocket, let's fallback to sending via non-websocket call.");
```
yaananth, line 327: @TingluoHuang : may be let's also short circuit when things bail out, give some max attempts

src/Runner.Common/JobServerQueue.cs
```diff
@@ -292,14 +306,46 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
                         {
                             try
                             {
-                                // we will not requeue failed batch, since the web console lines are time sensitive.
-                                if (batch[0].LineNumber.HasValue)
+                                bool pushedLinesViaWebsocket = false;
+                                if (this._websocketClient?.State == WebSocketState.Open)
                                 {
-                                    await _jobServer.AppendTimelineRecordFeedAsync(_scopeIdentifier, _hubName, _planId, _jobTimelineId, _jobTimelineRecordId, stepRecordId, batch.Select(logLine => logLine.Line).ToList(), batch[0].LineNumber.Value, default(CancellationToken));
+                                    var jsonData = JsonConvert.SerializeObject(new
```
yaananth, line 312: @TingluoHuang : `StringUtil.ConvertToJson` use this https://cs.github.com/actions/runner?q=StringUtil.ConvertToJson instead

yaananth (commented):
@yaananth : How do we let customers know about websockets requirement? (someone might block it)

@TingluoHuang : Add a check for websocket, doc: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github

src/Runner.Common/JobServerQueue.cs
```diff
@@ -30,6 +34,12 @@ public sealed class JobServerQueue : RunnerService, IJobServerQueue
         private static readonly TimeSpan _delayForWebConsoleLineDequeue = TimeSpan.FromMilliseconds(500);
         private static readonly TimeSpan _delayForTimelineUpdateDequeue = TimeSpan.FromMilliseconds(500);
         private static readonly TimeSpan _delayForFileUploadDequeue = TimeSpan.FromMilliseconds(1000);
+        private static readonly TimeSpan _minDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(1);
+        private static readonly TimeSpan _maxDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(500);
+
+        // In our hosted environment
```
yaananth, line 40: remove this comment

src/Runner.Common/JobServerQueue.cs
```diff
@@ -292,14 +317,62 @@ private async Task ProcessWebConsoleLinesQueueAsync(bool runOnce = false)
...
+                                if (this._websocketClient != null)
+                                {
+                                    var linesWrapper =  batch[0].LineNumber.HasValue? new TimelineRecordFeedLinesWrapper(stepRecordId, batch.Select(logLine => logLine.Line).ToList(), batch[0].LineNumber.Value):
+                                        new TimelineRecordFeedLinesWrapper(stepRecordId, batch.Select(logLine => logLine.Line).ToList());
+                                    var jsonData = StringUtil.ConvertToJson(linesWrapper);
+                                    try
+                                    {
+                                        totalBatchedLinesAttemptedByWebsocket++;
+                                        // It should be okay to wait for the result since we are already doing this in the background and doing it one by one
+                                        await this._websocketClient.SendAsync(new ArraySegment<byte>(Encoding.UTF8.GetBytes(jsonData)), WebSocketMessageType.Text, true, default(CancellationToken));
```
yaananth, line 336: I think we might have to break this into 1024 bytes each and then finally send `true` as the last message

thboop (commented):
Minor thoughts

src/Runner.Common/JobServerQueue.cs
```diff
@@ -71,14 +82,21 @@ public sealed class JobServerQueue : RunnerService, IJobServerQueue
...
-        // In this way, customer still can get instance live console output on job start, 
+        // In this way, customer still can get instance live console output on job start,
         // at the same time we can cut the load to server after the build run for more than 60s
         private int _webConsoleLineAggressiveDequeueCount = 0;
         private const int _webConsoleLineAggressiveDequeueLimit = 4 * 60;
         private const int _webConsoleLineQueueSizeLimit = 1024;
         private bool _webConsoleLineAggressiveDequeue = true;
         private bool _firstConsoleOutputs = true;
 
+        private int totalBatchedLinesAttemptedByWebsocket = 0;
```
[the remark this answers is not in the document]
yaananth: >for long jobs one instance of failures will have us suddenly using the old (slower) method for all calls until that job ends.

we have 50% failure rate to consider here (with 10 minimum calls), I think in the end, we try our best to do websocket, which should be true in most cases, but it's not the end of the world if we couldn't for few jobs as blips happen, as we fallback to what we were doing before (so any concerns with this being a DDOS already exists today, but even if this happens where we can't use websocket, we are limiting the affect since this is scoped to few jobs).

src/Runner.Common/JobServerQueue.cs
```diff
@@ -71,14 +82,21 @@ public sealed class JobServerQueue : RunnerService, IJobServerQueue
...
+        private int failedAttemptsToPostBatchedLinesByWebsocket = 0;
+
+        private ClientWebSocket _websocketClient = null;
```
TingluoHuang, line 96: Should we bake `_websocketClient` into `_jobServer`?

yaananth: https://github.com/actions/runner/pull/1755

PR #1735, opened 2022-03-07
Add annotations if Node 12 action is found and FF is on

fhammerl:
We eventually plan on deprecating node 12 actions, as node12 is approaching end of life in April. This PR adds the plumbing to warn users when they run node12 actions, so that they can update their actions to node16, or update to use the latest version. When this goes live in production, it will be accompanied by a blogpost describing how to migrate in more detail, along with a transition period to update actions.

Users should be warned when running actions that use node 12 because of its end of life: https://nodejs.org/en/about/releases/
Warnings:
image

Long Warnings:
image

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -146,7 +146,7 @@ namespace GitHub.Runner.Common
                 public const int RunOnceRunnerUpdating = 4;
             }
 
-            public static class FeatureFlags
+            public static class Features
             {
                 public static readonly string DiskSpaceWarning = "runner.diskspace.warning";
                 public static readonly string Node12Warning = "DistributedTask.AddWarningToNode12Action";
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -146,9 +146,10 @@ namespace GitHub.Runner.Common
                 public const int RunOnceRunnerUpdating = 4;
             }
 
-            public static class Features
+            public static class FeatureFlags
             {
                 public static readonly string DiskSpaceWarning = "runner.diskspace.warning";
+                public static readonly string Node12Warning = "DistributedTask.Node12Warning";
             }
 
             public static readonly string InternalTelemetryIssueDataKey = "_internal_telemetry";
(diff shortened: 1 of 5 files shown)


src/Runner.Common/Constants.cs
```diff
@@ -146,9 +146,10 @@ public static class ReturnCode
                 public const int RunOnceRunnerUpdating = 4;
             }
 
-            public static class Features
+            public static class FeatureFlags
```
fhammerl, line 149: Perhaps we could consider putting future FFs in this class? It's probably preferable over literals.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -246,6 +247,13 @@ public async Task RunAsync(IExecutionContext jobContext)
 
                 Trace.Info($"Current state: job state = '{jobContext.Result}'");
             }
+
+            if (jobContext.JobContext.ContainsKey("Node12ActionsWarnings"))
```
fhammerl, line 251: @thboop let me know what you think of creating an `IActionCommandManager` here and printing the warning.

It's outside the main `while` loop and I wonder if it can have unforeseen consequences.

Otherwise it's fairly safe, the `if` body will only execute if the FF is on because `"Node12ActionsWarnings"` is only set if the FF is on.

src/Runner.Worker/JobRunner.cs
```diff
@@ -257,6 +258,12 @@ private async Task<TaskResult> CompleteJobAsync(IJobServer jobServer, IExecution
                 }
             }
 
+            if (jobContext.JobContext.ContainsKey("Node12ActionsWarnings"))
+            {
+                var actions = string.Join(", ", jobContext.JobContext["Node12ActionsWarnings"].AssertArray("Node12ActionsWarnings").Select(action => action.ToString()));
+                jobContext.Warning(string.Format(Constants.Runner.Node12DetectedAfterEndOfLife, actions));
```
fhammerl, line 264: Replaced `ActionCommandManager` with `jobContext.Warning`

src/Runner.Common/Constants.cs
```diff
@@ -158,6 +159,7 @@ public static class Features
...
             public static readonly string UnsupportedStopCommandTokenDisabled = "You cannot use a endToken that is an empty string, the string 'pause-logging', or another workflow command. For more information see: https://docs.github.com/actions/learn-github-actions/workflow-commands-for-github-actions#example-stopping-and-starting-workflow-commands or opt into insecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_STOPCOMMAND_TOKENS` environment variable to `true`.";
             public static readonly string UnsupportedSummarySize = "$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of {0}k, got {1}k. For more information see: https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions#adding-a-markdown-summary";
+            public static readonly string Node12DetectedAfterEndOfLife = "One or more actions in this job ran using Node v12 which is coming to its End-oflife on 2022-04-30. It is recommended to update all affected action definitions to use Node v16. Actions ran on Node v12: {0}";
```
thboop, line 162:
```suggestion
            public static readonly string Node12DetectedAfterEndOfLife = "Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: {0}";
```

Thoughts on ^. Node12 is end of life, which is our motivation for deprecating node12 actions, but we can make this message a little more concise. Node.js 12 actions are deprecated is short and sweet

In the future, we can add `More information can be found at: `

thboop (approved):
LGTM

PR #1738, opened 2022-03-09
Added ability to run Dockerfile.SUFFIX ContainerAction

### LINKED ISSUE #1416: Bug when parsing runs.image value in action.yml
I have been redirected here by GH support.

It looks like there is a bug connected to a value of `runs.image` in `action.yml` and execution of `docker run` (inside workflow step containing created action). I did below tests on `v2.283.3` of `actions/runner`.

### Typical usage
When creating `Docker contained action` we use such directory schema as a standard example:

```
|-- .github
    |-- action.yml
    |-- Dockerfile
    |-- entrypoint.sh
```

Action file might look like this:

```
# action.yaml
# ...
runs:
  using: 'docker'
  image: 'Dockerfile'
# ...
```

Content of `Dockerfile` and `entrypoint.sh` doesn't matter in these cases.

### More complex usage
Let's imagine a situation when we want to keep two or more actions with a similar profile in one repository.

We can divide directories using the following directory schema:

```
|-- .github
     |-- action-a
          |-- action.yml
     |-- action-b
          |-- action.yml
     |-- docker
          |-- action-a
               |-- entrypoint.sh
          |-- action-b
               |-- entrypoint.sh
          |-- Dockerfile.a
          |-- Dockerfile.b
          |-- shared.sh
```

You might ask why we extract some files into separate `docker` directory?
This separation allows to load shared files (such as `shared.sh`) in both `entrypoint.sh` files. Shared file is a level higher in directory hierarchy and as we know Docker only allows adding (`COPY`, `ADD`) files from within `Dockerfile` directory and its children.

However, this is not the key problem in the whole case. With the above layout both `action.yml` files need to have such entry to work properly:

```
  using: 'docker'
  image: './../docker/Dockerfile.a'
  # OR
  # image: './../docker/Dockerfile.b'
```

Executing above actions ends with the following error (example):

```
/usr/bin/docker run --name dockerDockerfilea_336f1e --label cf4eaa ...
docker: invalid reference format: repository name must be lowercase.
```

### Lowercasing above complex usage
It seems that the given image path (`./../docker/Dockerfile.a`) has been reformatted to a name (`dockerDockerfilea_336f1e`). **Docker doesn't like uppercases in repository name**.

So I changed `Dockerfile.a` name to `dockerfile.a` and put `./../docker/dockerfile.a` into `image` property. In parallel, I made the same change for the second `Dockerfile.b`.

```
  using: 'docker'
  image: './../docker/dockerfile.a'
  # OR
  # image: './../docker/dockerfile.b'
```

Sadly, executing changed actions ends with the following error (example):

```
/usr/bin/docker run --name dockerdockerfilea_0ed722 ...
docker: invalid reference format.
```

### Fixing above complex usage
Interestingly, there is a simple workaround for the above problem. **You need to reverse the file name with file extension**. Instead of:

```
  using: 'docker'
  image: './../docker/dockerfile.a'
  # OR
  # image: './../docker/dockerfile.b'
```

I used this:

```
  using: 'docker'
  image: './../docker/a.Dockerfile'
  # OR
  # image: './../docker/b.Dockerfile'
```

My directory structure looks like this:

```
|-- .github
     |-- action-a
          |-- action.yml
     |-- action-b
          |-- action.yml
     |-- docker
          |-- action-a
               |-- entrypoint.sh
          |-- action-b
               |-- entrypoint.sh
          |-- a.Dockerfile
          |-- b.Dockerfile
          |-- shared.sh
```

**_This change caused that action has been executed correctly_**. As you can see `Dockerfile` extension started with a capital letter. **In this approach it doesn't matter, whether the first letter is capitalized or not** (case insensitivity).

### Conclusion
In the examples we have two naming conventions for multiple `Dockerfile`s cases:
- `Dockerfile.sufix-extension`, which is **[allowed](https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f)** by Docker, but **doesn't work** in GitHub Actions
- `prefix-naming.Dockerfile`, which **is not prohibited** by Docker and **works** in GitHub Actions

It looks like a bug. However, it is possible that such an implementation is due to some specific reasons. **Where do such discrepancies result from?**

nikola-jokic:
Closes #1416 

This is an enhancement to the runner allowing it to run Dockerfile.SUFFIX files, which is [allowed](https://docs.docker.com/engine/reference/commandline/build/#specify-a-dockerfile--f) by Docker, but doesn't work in GitHub Actions


src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -44,29 +44,34 @@ public async Task RunAsync(ActionRunStage stage)
...
-                    dockerFile,
-                    Directory.GetParent(dockerFile).FullName,
-                    imageName);
-                ExecutionContext.Output("##[endgroup]");
-
-                if (buildExitCode != 0)
+            else if (Data.Image.Split('/').Length > 0)
+            {
+                var imagePathSplit = Data.Image.Split('/');
+                var image = imagePathSplit[imagePathSplit.Length-1];
```
fhammerl, line 50: NIT: Formatting of `Length-1`

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -44,29 +44,34 @@ public async Task RunAsync(ActionRunStage stage)
...
-                var buildExitCode = await dockerManager.DockerBuild(
-                    ExecutionContext,
-                    ExecutionContext.GetGitHubContext("workspace"),
-                    dockerFile,
-                    Directory.GetParent(dockerFile).FullName,
-                    imageName);
-                ExecutionContext.Output("##[endgroup]");
-
-                if (buildExitCode != 0)
+            else if (Data.Image.Split('/').Length > 0)
```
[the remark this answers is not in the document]
nikola-jokic: Thanks! I will notify you as soon as I fix this

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -232,5 +229,20 @@ public async Task RunAsync(ActionRunStage stage)
             }
 #endif
         }
+
+        private bool IsDockerFile(string image)
+        {
+
```
fhammerl, line 235: Super NIT: empty line

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -232,5 +229,20 @@ public async Task RunAsync(ActionRunStage stage)
...
+            var imagePathSplit = Data.Image.Split('/');
+            var imageWithoutPath = imagePathSplit[imagePathSplit.Length - 1];
+            if (imageWithoutPath.StartsWith("Dockerfile") || imageWithoutPath.StartsWith("dockerfile") || imageWithoutPath.EndsWith("Dockerfile") || imageWithoutPath.EndsWith("dockerfile"))
```
fhammerl, line 238: Do we need the `EndsWith` conditions here? It made sense for a full path, but now this would match 'foo-bar.Dockerfile', which is not what we want I think.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -232,5 +229,20 @@ public async Task RunAsync(ActionRunStage stage)
             }
 #endif
         }
+
+        private bool IsDockerFile(string image)
+        {
+
+            var imagePathSplit = Data.Image.Split('/');
+            var imageWithoutPath = imagePathSplit[imagePathSplit.Length - 1];
```
fhammerl, line 237: NIT: how about `var imageWithoutPath = Data.Image.Split('/').Last()` ?

src/Runner.Worker/Handlers/ContainerActionHandler.cs
nikola-jokic: I left it because of the comment at the end of the issue:

> prefix-naming.Dockerfile, which is not prohibited by Docker and works in GitHub Actions

So I am guessing that people are using it that way

fhammerl (dismissed):

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -44,12 +44,9 @@ public async Task RunAsync(ActionRunStage stage)
             {
                 Data.Image = Data.Image.Substring("docker://".Length);
             }
-            else if (Data.Image.EndsWith("Dockerfile") || Data.Image.EndsWith("dockerfile"))
+            else if (IsDockerFile(Data.Image))
             {
-                // ensure docker file exist
                 var dockerFile = Path.Combine(ActionDirectory, Data.Image);
```
fhammerl, line 49: Style mini NIT: Up to you, `IsDockerFile(Data.Image)` -> `TryParseDockerFile(Data.Image, out var dockerFile`) might be more idiomatic.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -232,5 +229,18 @@ public async Task RunAsync(ActionRunStage stage)
             }
 #endif
         }
+
+        private bool IsDockerFile(string image)
+        {
+            var imageWithoutPath = Data.Image.Split('/').Last();
+            if (imageWithoutPath.StartsWith("Dockerfile") || imageWithoutPath.StartsWith("dockerfile") || imageWithoutPath.EndsWith("Dockerfile") || imageWithoutPath.EndsWith("dockerfile"))
```
fhammerl, line 236: There is similar code in `ActionManager`, (`.EndsWith("dockerfile")`), do we need this change there too?

src/Runner.Worker/ActionManager.cs
```diff
@@ -1004,7 +1004,7 @@ private ActionSetupInfo PrepareRepositoryActionAsync(IExecutionContext execution
                 if (actionDefinitionData.Execution.ExecutionType == ActionExecutionType.Container)
                 {
                     var containerAction = actionDefinitionData.Execution as ContainerActionExecutionData;
-                    if (containerAction.Image.EndsWith("Dockerfile") || containerAction.Image.EndsWith("dockerfile"))
+                    if (DockerUtil.IsDockerfile(containerAction.Image))
```
thboop, line 1007: Can we update the error on line 1029

thboop: should we swap this if condition to be safer?

So do

```
if (containerAction.Image.StartsWith("docker://" ...)
{

} else if(DockerUtil.IsDockerfile()
{

}
else
{
   throw
}

```

src/Runner.Worker/Container/DockerUtil.cs
```diff
@@ -61,5 +62,11 @@ public static string ParseRegistryHostnameFromImageName(string name)
             }
             return "";
         }
+
+        public static bool IsDockerfile(string image)
+        {
+            var imageWithoutPath = image.Split('/').Last();
```
thboop, line 68: Are we overthinking this? If we swap the order on the if statement, should we just check if the path contains "dockerfile"?

src/Test/L0/Container/DockerUtilL0.cs
```diff
@@ -144,5 +144,45 @@ public void ParseRegistryHostnameFromImageName(string input, string expected)
...
+        [InlineData("dockerhub/repo", false)]
+        [InlineData("debian:latest", false)]
+        [InlineData("something/dockerfileimage", false)]
+        [InlineData("ghcr.io/docker/dockerfile", true)] // should be false but might break the current workflows
+        [InlineData("Dockerfile", true)]
+        [InlineData("Dockerfile.", true)]
+        [InlineData(".Dockerfile", true)]
+        [InlineData(".Dockerfile.", false)]
+        [InlineData("dockerfile", true)]
+        [InlineData("dockerfile.", true)]
```
thboop, line 160: can we add a `docker://` test case

src/Runner.Worker/ActionManager.cs
nikola-jokic: I think that we will break some of the workflows by doing that if someone specified `image: ubuntu:latest`. Right now, the logic tries to see if we start with `docker://`, and if not, we try to parse it as a Dockerfile. If Dockerfile is not provided, we are falling back to Docker Hub pull image. If we do this, someone specifying `image: ubuntu:latest` and not docker:// is going to have their workflow throwing exceptions. I checked if I define it as `image: ubuntu:latest`, inside ContainerActionHandler, we would get `ubuntu:latest`

src/Runner.Worker/Container/DockerUtil.cs
nikola-jokic: My view is that we should not change the order of the if statement. Also, if the image is named `nikola-jokic/dockerfile:latest`, and if we check only for existence of the dockerfile, we would treat this as a file, but it is not.
We are running those dockerfiles with `docker run -f [path to file]...`. This means that in theory, we could name the file whatever we want, let's say `ContainerDeffinition` and put it inside `/dockerfile/ContainerDeffinition`, and it would work (because of the `-f` flag). But I am worried if the image name contains dockerfile, we would treat it as a file to build an image from, and we would not try to pull it from the docker hub registry.
This way, we clearly allow prefix `dockerfile.` or endswith `dockerfile`.
But if you feel like we could allow anything if the path contains dockerfile, should we check if the last part is a tag (For example `:latest` or `:20.04`)? If so, that is **most likely** not a dockerfile

src/Test/L0/Container/DockerUtilL0.cs
nikola-jokic: Sure. I extended our test to check if starts with `docker://` since it is clearly not a dockerfile. This would not introduce a bug to the ContainerActionHandler, since the docker:// is checked first, but might be better for the IsDockerfile check to account for this case

src/Runner.Worker/Container/DockerUtil.cs
fhammerl: So, the changes as they are currently would skip actual local dockerfiles with the following names:

```
/foo/bar/nikola-jokic/dockerfile:latest
./dockerfile:latest
```

I don't think using contains is the way to go, testing contains(dockerfile) for the element after the last `/`, we would treat URLs like these as file paths:

```
nikola-jokic/dockerfile:latest                            # dockerhub URL
nikola-jokic/app-dockerfile-test:latest
```

I don't think we can have a perfect solution here. Probably the most user friendly way would be to try and treat `image` as a path, if it exists, build it, if not, fallback and try treating it as a URL. We have already discussed why we don't want to do that.

src/Runner.Worker/ActionManager.cs
thboop: Don't we just [throw](https://github.com/actions/runner/pull/1738/files#diff-5b282ef924f268139b4015eb491825d3e35bfa6cbe46962e3bef3bd8fa05dfccL1029) if it doesn't start with `docker://`

Do we catch that throw somewhere or something?

nikola-jokic: I am only worried that some stuff in [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-running-a-job-within-a-container) are explaining it without the `docker://`
Usually, you specify docker:// when you want to talk to the Docker socket, and if you want to specify an online registry, you would have to do it like `docker://gcr.io/nikola-jokic/my-immage` let's say

fhammerl:
```
if (containerAction.Image.StartsWith("docker://" ...)
{

} else if(DockerUtil.IsDockerfile()
{

}
else
{
   throw
}
```

This above would throw on the config below wouldn't it?

```
container:
      image: node:14.16
```

fhammerl:
@thboop Do you think we should wrap this one in a FF?

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -68,6 +67,10 @@ public async Task RunAsync(ActionRunStage stage)
 
                 Data.Image = imageName;
             }
+            else
```
fhammerl, line 70: Currently, we allow local (local as in referenced by path in `uses: ./action-path`) container actions to specify images like so: `image: ubuntu:latest`. This goes against the documentation, so we're fixing that here. 

We will now throw on local container actions that specify the image in a format like `image: ubuntu:latest`, because now we require `image: docker://ubuntu:latest` strictly, in accordance with the [documentation](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-using-public-docker-registry-container).

thboop (approved):
lgtm

PR #1741, opened 2022-03-09
Issue 1739: Fixing null reference exception during configuring runner with invalid repo URL or token

### LINKED ISSUE #1739: Runner throws a null ref exception during configuration for an invalid token or link to the repository
**Describe the bug**
During configuration, the runner throws a null ref exception for an invalid token or GitHub repo URL.

**To Reproduce**
1. Run the runner configuration
2. Pass the invalid token or repository URL
3. See error

**Expected behavior**
Configuration should be terminated with a meaningful message.

v2.288.1

ruvceskistefan:
The problem was that the runner crasher with null reference exception during configuring with invalid repository URL or register token.

Closes #1739

After this change, the runner will not crash with a null reference exception but will exit with an appropriate error message.

### DIFF
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -642,8 +642,9 @@ namespace GitHub.Runner.Listener.Configuration
 
                             if(response.StatusCode == System.Net.HttpStatusCode.NotFound)
                             {
-                                _term.WriteError("Invalid repository URL or register token");
-                                return StringUtil.ConvertFromJson<GitHubRunnerRegisterToken>(errorResponse);
+                                // No need to retry
+                                retryCount = 3;
+                                throw new Exception("Invalid repository URL or register token");
                             }
                             else
                             {
@@ -712,8 +713,9 @@ namespace GitHub.Runner.Listener.Configuration
 
                             if(response.StatusCode == System.Net.HttpStatusCode.NotFound)
                             {
-                                _term.WriteError("Invalid repository URL or register token");
-                                return StringUtil.ConvertFromJson<GitHubAuthResult>(errorResponse);
+                                // No need to retry
+                                retryCount = 3;
+                                throw new Exception("Invalid repository URL or register token");
                             }
                             else
                             {
--- a/src/Runner.Listener/Configuration/ConfigurationManager.cs
+++ b/src/Runner.Listener/Configuration/ConfigurationManager.cs
@@ -624,9 +624,12 @@ namespace GitHub.Runner.Listener.Configuration
                     httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("basic", base64EncodingToken);
                     httpClient.DefaultRequestHeaders.UserAgent.AddRange(HostContext.UserAgents);
                     httpClient.DefaultRequestHeaders.Accept.ParseAdd("application/vnd.github.v3+json");
+                    
+                    var responseStatus = System.Net.HttpStatusCode.OK;
                     try
                     {
                         var response = await httpClient.PostAsync(githubApiUrl, new StringContent(string.Empty));
+                        responseStatus = response.StatusCode;
 
                         if (response.IsSuccessStatusCode)
                         {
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -634,28 +634,33 @@ private async Task<GitHubRunnerRegisterToken> GetJITRunnerTokenAsync(string gith
...
                         {
                             _term.WriteError($"Http response code: {response.StatusCode} from 'POST {githubApiUrl}'");
                             var errorResponse = await response.Content.ReadAsStringAsync();
                             _term.WriteError(errorResponse);
-                            response.EnsureSuccessStatusCode();
+
+                            if(response.StatusCode == System.Net.HttpStatusCode.NotFound)
+                            {
+                                _term.WriteError("Invalid repository URL or register token");
+                                return StringUtil.ConvertFromJson<GitHubRunnerRegisterToken>(errorResponse);
```
TingluoHuang, line 646: `ConvertFromJson` will return useless data if the `errorResponse` is not a contract of `GitHubRunnerRegisterToken`

ruvceskistefan: I changed this in the new commit.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -634,28 +634,34 @@ private async Task<GitHubRunnerRegisterToken> GetJITRunnerTokenAsync(string gith
...
-                            // It doesn't make sense to retry in this case, so just stop
-                            break;
-                        }
                         else
                         {
                             _term.WriteError($"Http response code: {response.StatusCode} from 'POST {githubApiUrl}'");
                             var errorResponse = await response.Content.ReadAsStringAsync();
                             _term.WriteError(errorResponse);
-                            response.EnsureSuccessStatusCode();
+
```
thboop, line 642: Can we just try storing the response code and using that to determine if we should throw or not.

```
                    var responseStatus = HttpStatusCode.OK;
                    try
                    {
                        var response = await httpClient.PostAsync(githubApiUrl, new StringContent(string.Empty));
                        responseStatus = response.StatusCode;
                        if (response.IsSuccessStatusCode)
                        {
                            Trace.Info($"Http response code: {response.StatusCode} from 'POST {githubApiUrl}'");
                            var jsonResponse = await response.Content.ReadAsStringAsync();
                            return StringUtil.ConvertFromJson<GitHubRunnerRegisterToken>(jsonResponse);
                        }
                        else
                        {
                            _term.WriteError($"Http response code: {response.StatusCode} from 'POST {githubApiUrl}'");
                            var errorResponse = await response.Content.ReadAsStringAsync();
                            _term.WriteError(errorResponse);
                            response.EnsureSuccessStatusCode();
                        }
                    }
                    catch(Exception ex) when (retryCount < 2 && responseStatus != HttpStatusCode.NotFound)
                    {
                        retryCount++;
                        Trace.Error($"Failed to get JIT runner token -- Atempt: {retryCount}");
                        Trace.Error(ex);
                        Trace.Info("Retrying in 5 seconds");
                    }
```

thboop: We may want to add others to this not retry list, and looking at the response code is the easiest way to do that

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -699,29 +705,34 @@ private async Task<GitHubAuthResult> GetTenantCredential(string githubUrl, strin
...
-                        }
                         else
                         {
                             _term.WriteError($"Http response code: {response.StatusCode} from 'POST {githubApiUrl}'");
                             var errorResponse = await response.Content.ReadAsStringAsync();
                             _term.WriteError(errorResponse);
-                            // Something else bad happened, let's go to our retry logic
-                            response.EnsureSuccessStatusCode();
+
+                            if(response.StatusCode == System.Net.HttpStatusCode.NotFound)
```
thboop, line 714: Same as above!

src/Runner.Listener/Configuration/ConfigurationManager.cs
ruvceskistefan: Sure, makes sense.

ruvceskistefan: I just won't trace delay time there, because it's randomly determined outside of catch.

thboop (approved):
lgtm

Issue #1719, opened 2022-02-28
Corrupted artifact: actions-runner-osx-x64-2.288.0.tar.gz


0x2b3bfa0:
The latest macOS release file [`actions-runner-osx-x64-2.288.0.tar.gz`](https://github.com/actions/runner/releases/download/v2.288.0/actions-runner-osx-x64-2.288.0.tar.gz) is corrupted.

### Steps to reproduce the behavior
1. Go to the [Releases](https://github.com/actions/runner/releases) page
2. Download [`actions-runner-osx-x64-2.288.0.tar.gz`](https://github.com/actions/runner/releases/download/v2.288.0/actions-runner-osx-x64-2.288.0.tar.gz) from the latest ([v2.288.0](https://github.com/actions/runner/releases/tag/v2.288.0)) release
3. Try to extract the file with GNU [`tar`](https://www.gnu.org/software/tar) or any other compatible tool
4. Extraction fails

### Expected behavior
Extraction succeeds

## Version

Only happens with `v2.288.0` (latest) and this specific macOS artifact.

## Issue

```console
$ curl --silent --location https://github.com/actions/runner/releases/download/v2.288.0/actions-runner-osx-x64-2.288.0.tar.gz | gunzip > /dev/null
gzip: stdin: invalid compressed data--format violated
```

thboop:
Thanks for the report we are working on a fix

thboop:
We regenerated this release and confirmed it is no longer corrupted, going to close this issue but please let me know if you see any more issues!

Thanks for the report.

0x2b3bfa0:
Block 1444 in the [DEFLATE](https://datatracker.ietf.org/doc/html/rfc1951) stream was corrupted, and NLEN (34710) was not the [one's complement](https://en.wikipedia.org/wiki/One%27s_complement) of LEN (22286) as [expected](https://datatracker.ietf.org/doc/html/rfc1951#page-11); see Go [`compress/flate`](https://github.com/golang/go/blob/eb8198d2f67477517e7a735faa49dfd7c0fb3622/src/compress/flate/inflate.go#L637) for more information.

Given that `1000 0111 1001 0110` doesn't have any plausible resemblance to the bitwise negation of `0101 0111 0000 1110`, it's hard to believe in a bit flip. 🤔 Anyway, it doesn't matter now.

```console
$ go run main.go < actions-runner-osx-x64-2.288.0.tar.gz > /dev/null
panic: flate: corrupt input before offset 34572188
[...]

PR #1745, opened 2022-03-11
Validate packages hash before uploading to github release in CD workflow.

TingluoHuang:
To avoid future recurrence of https://github.com/actions/runner/issues/1719, we are going to validate the runner package checksum before uploading to GitHub releases.

More context:

The runner release is created by executing workflow `.github/workflows/release.yml`

The workflow mainly has 3 stages:
- Check 
    - some pre-release validation
- Build
    - compile and create runner packages for all platforms
    - calculate SHA256 for all created `tar.gz`/`zip`
    - upload all `tar.gz` and `zip` as workflow artifact
- Release
    - download workflow artifacts created by the `Build` stage
    - generate release note
    - **(NEW) Validate checksum of all downloaded `tar.gz` and `zip` with SHA256 from the `Build` stage**
    - create a GitHub release and upload runner packages as release assets.

If the validation failed, the release process will stop.
Ex:
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -260,6 +260,17 @@ jobs:
           console.log(releaseNote)
           core.setOutput('version', runnerVersion);
           core.setOutput('note', releaseNote);
+
+    - name: Validate Packages HASH
+      working-directory: _package
+      run: |
+        ls -l
+        echo "${{needs.build.outputs.win-x64-sha}}  actions-runner-win-x64-${{ steps.releaseNote.outputs.version }}.zip" | shasum -a 256 -c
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -260,6 +260,17 @@ jobs:
           console.log(releaseNote)
           core.setOutput('version', runnerVersion);
           core.setOutput('note', releaseNote);
+
+    - name: Validate Packages HASH
+      working-directory: _package
```
fhammerl, line 265: Not sure how the artifacts got corrupted in the first place, so this might be irrelevant, but should we validate the trim files in `_package/trim*/**` folders?

TingluoHuang: the runner will validate the hash when it's try to use any of the trimmed packages, it will fall back to the full package if the validation failed.
So we only need to validate hash for the full packages as a minimum requirement.

thboop (approved):
LGTM

PR #1751, opened 2022-03-14
ADR: Runner Job Started/Completed Hooks

thboop:
[Rendered](https://github.com/actions/runner/blob/thboop/hookADR/docs/adrs/1751-runner-job-hooks.md)

This ADR details how we plan to support Job Started / Job Completed hooks for self hosted runner admins.


docs/adrs/1751-runner-job-hooks.md
```diff
@@ -0,0 +1,83 @@
...
+
+
+### Environment Files
+We will also enable [Environment Files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files) to support setup scenarios for the runner environment.
+
+While a self hosted runner admin can [set env variables](https://docs.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners#using-a-env-file-to-set-the-proxy-configuration), these apply to all jobs. By enabling the ability to `add a path` and `set an env` we give runner admins the ability to do this dynamically based on the [workflows environment variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) to empower setup scenarios.
+
+
+### Exit codes
+These are **syncronous** hooks, so they will block job execution while they are being run. Exit code 0 will indicate a successful run of the hook and we will proceed with the job, any other exit code will fail the job with an appropriate annotation.
```
ethomson, line 62:
```suggestion
These are **synchronous** hooks, so they will block job execution while they are being run. Exit code 0 will indicate a successful run of the hook and we will proceed with the job, any other exit code will fail the job with an appropriate annotation.
```

docs/adrs/1751-runner-job-hooks.md
```diff
@@ -0,0 +1,83 @@
...
+## Key Decisions
+- We will expose 2 variables that users can set to enable hooks  
+  - `ACTIONS_RUNNER_HOOK_JOB_STARTED`
+  - `ACTIONS_RUNNER_HOOK_JOB_COMPLETED`
+- Users can set these variables to the path of a `.sh` or `.ps1` file, which we will execute when Jobs are started or completed.
+  - Output from these will be added to a new step at the start/end of a job named `Pre Job Hook` or `Post Job Hook`. 
+    - These steps will only be generated on runs with these hooks
+- These hooks `always()` execute if the env variable is set
+- These files will execute as the Runner user, outside of any container specification on the job
+- These are **syncronous** hooks
```
ethomson, line 74:
```suggestion
- These are **synchronous** hooks
```

docs/adrs/1751-runner-job-hooks.md
```diff
@@ -0,0 +1,83 @@
...
+We will execute `pwsh` (fallback to `powershell`) or `bash` (fallback to `sh`) as appropriate.
+- `.sh` files will execute with the args `-e {pathtofile}`
+- `.ps1` files will execute with the args `-command \". '{pathtofile}'\"`
+
+We will **not** set the [standard flags we typically set](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) for `runs` commands. So, if you want to set `pipefail` on `bash` for example, you will need to do that in your script.
+
+### UI
+We want to ensure the experience for users invoking workflows is good, if hooks take too long, you may feel your job is delayed or broken. So, much like `Set Up Job`, we will generate two new steps automatically in your job, one for each configured hook: 
+- `Pre Job Hook`  
+- `Post Job Hook`
```
ethomson, line 37: I'd like to tweak the verbiage here a little bit. I think that the runner admins who set these jobs up will understand what this means, but the end-user who is writing/executing a workflow on that runner may not.

We already have "Set up job" and "Complete job".  🤔 

I tend towards "Set up runner" and "Complete runner" to align? I don't _love_ "complete runner", because it's vague, but I like that it's matching.

j3parker: From above it seems like this feature is going to always be exclusive to self-hosted runners, right?

If that's the case, speaking as someone who is responsible for an org that uses self-hosted runners a bunch, it might help to have it be something like "Set up _self-hosted_ runner".

This is a minor point, but our users are not necessarily clear on where our infrastructure ends and where GitHub's begins... if they see weird stuff in the logs for a step about self-hosted runners they'll know who to blame quicker :) This isn't a big deal though, when they ask about something going wrong the people who can diagnose this look under "Set up job" to see who's runner it is, etc.

thboop: > If that's the case, speaking as someone who is responsible for an org that uses self-hosted runners a bunch, it might help to have it be something like "Set up self-hosted runner".

That's definitely a scenario I've been thinking about a lot. The user who owns the self hosted runners may not always be the same user who is writing the workflows.

If I (as a workflow author) had my job fail because the runner admin injected a step that isn't working correctly, its not necessarily clear who I should contact to fix that. `Set up runner` sounds like it could be an issue with the runner application, I may not have context here.

To help, we are planning to add text that says `Running the 'Set up runner' step configured by the self-hosted runner administrator` as the first line of the step logs, to make it more clear  why this step was injected, and what users can do if it starts failing.

tyrken:
Thanks @thboop, I had a chance to play with the Beta and it's mostly good for my use case - the only issue I can see so far is the precise environment variables present when the hook script runs.  In [my earlier PR I have a section](https://github.com/actions/runner/pull/1469/files#diff-e10dd2daf26c47f8e2914b189bbbc8043bdcd073c633c9a2d29d67f0ec3b4581R620-R631) that grabs the standard global env-vars then adds the main ones from the job's definition.  This allows the workflow writer to configure the hook script slightly, e.g. via the `env:` block in their `job` definition or [at the top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#env).

Would it be possible to add a similar tweak here?  There doesn't seem any other way of getting at them (i.e. they aren't in the event.json nor anywhere else I can think of yet)...

j3parker:
This is working well for us 👍 Thanks!

fhammerl (approved):
LGTM

tyrken:
Hello?  Any comments on my feedback or the PR to implement a fix @thboop ?

dimitar-hristov:
Hi @thboop :wave: Reading the ADR, I noticed that you included:
> A runner administrator is able to grab custom telemetry of jobs running on their self hosted runner
...
Runner admins have the ability to tie into the runner job execution to publish their own telemetry or perform their own cleanup or setup

I was wondering how the started/completed hooks tie with the telemetry? I am sorry if it is a silly question, but I struggle to understand it and I couldn't find any documentation about it. :disappointed:

nickolashkraus:
>Runner admins can execute a background process for async hooks if they want

How does one do this?

Issue #827, opened 2020-11-25
Runner ignores DOCKER_HOST variable when starting container


jeromewaibel:
**Describe the bug**
I tried to run a self hosted github runner with [rootless docker](https://docs.docker.com/engine/security/rootless/). I added
```
XDG_RUNTIME_DIR=/home/runner/.docker/run
DOCKER_HOST=unix:///home/runner/.docker/run/docker.sock
```
to .env and `/home/runner/bin` to .path. In general github actions and rootless docker work fine.

When the container is initialized it runs
```
/home/runner/bin/docker create --name 395a5f555e924b28bf52524b85bb3eee_docker1903dind_710b8c --label d94681 --workdir /__w/actionstest/actionstest --network github_network_1c96635676ce470c8b9cb5f86ec01f5c
  -e "HOME=/github/home" -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/actions-runner/_work":"/__w" -v "/home/runner/actions-runner/externals":"/__e":ro -v "/home/runner/actions-runner/_work/_temp":"/__w/_temp" -v "/home/runner/actions-runner/_work/_actions":"/__w/_actions"
 -v "/home/runner/actions-runner/_work/_tool":"/__w/_tool" -v "/home/runner/actions-runner/_work/_temp/_github_home":"/github/home" -v "/home/runner/actions-runner/_work/_temp/_github_workflow":"/github/workflow" --entrypoint "tail" docker:19.03-dind "-f" "/dev/null"
```

As you can see it tries to volume mount the non existing /var/run/docker.sock from the host, which prevents any dind image from working. It seems this value is [hardcoded here](https://github.com/actions/runner/blob/4e85b8f3b7493aa1adf9dd0a574a9fa2bf1add71/src/Runner.Worker/Container/ContainerInfo.cs#L52) and DOCKER_HOST is ignored.

**To Reproduce**
1. Create a self hosted runner using rootless docker
2. Start a build with a step running inside a container

**Expected behavior**
/var/run/docker.sock should not be hard coded. DOCKER_HOST should be used if set.

## Runner Version and Platform
Version of your runner? 2.274.2

OS of the machine running the runner? Linux, Ubuntu 18.04

## Job Log Output

Runner starts the build step with

`/home/runner/bin/docker create -[...] -v "/var/run/docker.sock":"/var/run/docker.sock" [...]`

which makes this job
```
  dind-test:
[...]

Issue #434, opened 2020-04-17
Self-hosted runner with Docker step creates files that trip up the checkout step


j3parker:
**Describe the bug**
When using self-hosted runners, git checkouts are cached between runs (this is nice, because it greatly speeds up our builds).

However, if a docker-based step writes a file to the workspace it will (possibly) be owned by the root user. If the permissions don't give the (non-root) action runner user +w permission then a checkout step in a future workflow run will fail to remove this file. The first time, the error will look like this:

```
##[group]Cleaning the repository
[command]/usr/bin/git clean -ffdx
warning: could not open directory 'foo/': Permission denied
warning: failed to remove foo/: Directory not empty
##[endgroup]
##[warning]Unable to clean or reset the repository. The repository will be recreated instead.
Deleting the contents of '/home/jparker/actions-runner/_work/self-hosted-runner-permissions-issue-repro/self-hosted-runner-permissions-issue-repro'
##[error]Command failed: rm -rf "/home/jparker/actions-runner/_work/self-hosted-runner-permissions-issue-repro/self-hosted-runner-permissions-issue-repro/foo"
rm: cannot remove '/home/jparker/actions-runner/_work/self-hosted-runner-permissions-issue-repro/self-hosted-runner-permissions-issue-repro/foo': Permission denied
```

So `git clean -ffdx` tried to `stat()` this `foo/` directory (created via a container in a previous build) but failed. It was then unable to remove the directory because it wasn't empty. It tried to fall back to `rm -rf` which failed for the same reasons.

In future builds it goes straight to `rm -rf` because the `.git` folder did get cleaned up. It continues to fail in the same way for all future builds. Here's a screenshot:

![image](https://user-images.githubusercontent.com/597907/79594979-0f70c380-80ac-11ea-96fb-66d5c2f394af.png)

**To Reproduce**

I've created a repo that reproduces the error: https://github.com/Brightspace/self-hosted-runner-permissions-issue-repro

Here's an example of a workflow failing: https://github.com/Brightspace/self-hosted-runner-permissions-issue-repro/runs/596011452?check_suite_focus=true

**Expected behavior**

I guess I'd expect all the files to be owned by the runner user. Maybe that can be done with user namespace maps? [Documentation](https://docs.docker.com/engine/security/userns-remap/#enable-userns-remap-on-the-daemon). Not sure what that would entail though or if it makes sense for what the runner is doing.

[...]

PR #1754, opened 2022-03-15
Adding support for custom DOCKER_HOST / Rootless docker

srini-hv:
We are in process of adopting [Rootless Docker](https://docs.docker.com/engine/security/rootless/) for all our self-hosted runners for better security and isolation. While testing our workflows and custom actions, we noticed that Runner is not acknowledging the `DOCKER_HOST` Environment variable for Docker-in-Docker Actions and Docker Container Jobs, even when it is set on `.env` file on the runner directory. 

In our case, `DOCKER_HOST=unix:///run/user/2022/docker.sock`

I noticed that docker socket `/var/run/docker.sock` is hard-coded in the runner code for both these scenarios and I tried to make these dynamic by reading it from Environment variable in this PR.
This change would be technically fixing #827 

PS: I'm new to C# and if there is any better way to do this, I'm happy to update the PR.


fhammerl:
Hi @srini-hv 👋  Thanks for the PR. We'll have to evaluate how this might affect users who may have `DOCKER_HOST` set to something on their environment, but have come to expect the runner to ignore it and use `/var/run/docker.sock` instead.

NunoMaga:
+1

kolpav:
> but have come to expect the runner to ignore it and use /var/run/docker.sock instead

Judging by the amount of issues and comments here and elsewhere I would argue its nobody expectation.

Current state of things is that you are forced to run as root for no apparent reason while accumulating garbage after each run even though you are using docker then you get error because the runner is not even able to clean up after itself due to permissions. Issues mentioning these problems are open for years pilling on more comments with workarounds and replies questioning why is the self hosted runner doing all these things.

After all of that your expectations as a user trying to set up self hosted runner are so low that `DOCKER_HOST` doing what its supposed to do would surprise but in positive way. I think all the people in #434 and many issues linking to it would be also positively surprised.

Kurt-von-Laven:
For folks wanting to run Docker in rootless mode on GitHub Actions, I recommend giving [ScribeMD/rootless-docker](https://github.com/marketplace/actions/rootless-docker) a try. Full disclosure: I am the author of this action.

jmcrfp:
We are already running Docker in rootless mode. The issue is when trying to run docker in docker(rootless)

NunoMaga:
> Apologies for my careless reading. We recently began encountering this same issue with rootless Docker (not dind). We were able to work around it by proxying bidirectionally between `unix:///var/run/docker.sock` and the rootless Docker socket (e.g., `unix:///run/user/1001/docker.sock` or more generally `unix://$XDG_RUNTIME_DIR/docker.sock`). Here is [the relevant snippet of rootless-docker](https://github.com/ScribeMD/rootless-docker/blob/e6a76cfc95b2a2327ec36437134a42af5b9ddf30/action.yaml#L63-L74) in case it helps you or others. You may alternatively be able to simply run the rootless-docker action and then perform any additional needed Docker setup afterwards.

our issue is that we are unable to mount the rootless docker socket in the docker run command. Are you suggesting we proxy the socket at runner install? and if so, how did you go about doing so? (cannot open the provided relevant snippet)

Kurt-von-Laven:
What do you mean by runner install? I don't know whether I accurately understand what you are trying to do, but it should be quick to try bidirectionally proxying the rootful and rootless Docker sockets, and I can't think of a specific reason it would fix our problem but not yours.

rootless-docker is a public repository, and I am able to open the link in an incognito window. What happens when you click on the link? Here is the snippet in any case:

```yaml
    - name: Proxy bidirectionally between rootful and rootless Docker sockets.
      if: steps.rootless-docker.outputs.in-use != 'true'
      run: >
        sudo systemd-run
        --unit=docker-proxy.service
        --description="Bidirectional proxy between rootful and rootless Docker sockets"
        --service-type=exec
        --property=Requires=docker.socket
        --property=PrivateNetwork=true
        --property=PrivateTmp=true
        /lib/systemd/systemd-socket-proxyd "$XDG_RUNTIME_DIR/docker.sock"
      shell: bash
```

john-jam:
We also need to have the docker socket configurable when running docker in rootless mode. Allowing to bind a specific socket rather than the proxy socket workaround is a better solution imho.

Kurt-von-Laven:
Yes, the only reason to proxy sockets is that works today. It would be preferable to allow binding to a different socket.

nathanweeks:
Support for setting the path to the docker socket via DOCKER_HOST environment variable would benefit rootless podman as well (e.g., see this Red Hat support article for allowing rootless podman to be used with docker-compose (v1) by setting DOCKER_HOST: https://access.redhat.com/solutions/7011472)

Eddman:
Any new on this PR?

We're running ARC + DinD and the official ARC docs are stating that we should change the socket location as it has to be mounted to the [second container through an `emptyDir` volume](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/deploying-runner-scale-sets-with-actions-runner-controller#using-docker-in-docker-mode).

However with hardcoded socket path we can't use actions that require Docker inside containers. The docs also mention setting `DOCKER_HOST` variable, but this is ignored by the runner.

nodesocket:
I’m running “containerd” in a self hosted runner, how can I specify a custom path to the socket? Getting an error that /var/run/docker.sock does not exist.

carl-reverb:
As a workaround for running a job container with the GHA runner-scale-set-controller and an autoscalingrunnerset with `dind` mode configured, I figured out that I could modify the `job.<id>.container` to mount `/run/docker:/run/docker` and set `DOCKER_HOST=unix:///run/docker/docker.sock`.

jcpunk:
This patch would simplify things for me.

aseggemu-yoc (approved):

srini-hv:
Hi @aseggemu-yoc , I just updated the MR with the latest main branch. I'm guessing you might have to approve again for the CI to run. Can you take a look?

aseggemu-yoc (approved):

paulcostinean:
cc @nikola-jokic - maybe there's some time between horse races in the off season... 😆 
Bumping for visibility as it affects all rootless podman/docker users.

@BPouw's solution only allows one linux user to run an action and there's no way around it short of LD_PRELOAD hacks..

davidhjp01:
Bump up, this is also needed for us

jwinarske:
@aseggemu-yoc is this going to get merged?

wrigtim:
Bump bump

thiagocrepaldi:
> Yes, the only reason to proxy sockets is that works today. It would be preferable to allow binding to a different socket.

@Kurt-von-Laven FYI, I've tried using your action and it didnt work. 

```yaml
    - name: Print docker information
      run: |
        docker info --format "{{ .ClientInfo.Context }}"
    - name: Use Docker in rootless mode.
      uses: ScribeMD/rootless-docker@0.1.7
    - name: Print docker information
      run: |
        docker info --format "{{ .ClientInfo.Context }}"
    - name: Build HTML using Sphinx
      uses: ammaraskar/sphinx-action@master
```

I am using a different action (ammaraskar/sphinx-action instead of actions/runner). Under the hood, it calls docker like this:

```yaml
runs:
  using: 'docker'
  image: 'Dockerfile'
```

which will generate a call to docker similar to `/usr/bin/docker run --name NAME -v "/var/run/docker.sock":"/var/run/docker.sock" IMAGE

It seems `ScribeMD/rootless-docker` doesn't solve the problem that https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runsusing-for-docker-container-actions calls `docker run -v "/var/run/docker.sock":"/var/run/docker.sock"`

mkushakov:
What is blocking this PR to merge? We recently faced similar issue when running dockerized action in ARC self-hosted was failing to mound docker socket. This PR seems should solve our issue.

yaroslavsadin:
It's been two years, rootless podman is now widely used, this needs to be solved

tgarr:
Bumping this again to see if it gets attention. This PR would be really useful. Currently relying on a dirty workaround.

PR #1755, opened 2022-03-15
postlines: refactor per feedback

yaananth:
reacting to feedback at https://github.com/actions/runner/pull/1730 post merge

https://github.com/github/c2c-actions-service/issues/2828

### DIFF
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -143,8 +143,8 @@ namespace GitHub.Runner.Common
 
         public ValueTask DisposeAsync()
         {
-            Trace.Info($"Disposing websocket client ...");
-            this._websocketClient?.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Shutdown", CancellationToken.None);
+            _websocketClient?.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Shutdown", CancellationToken.None);
+            GC.SuppressFinalize(this);
             return ValueTask.CompletedTask;
         }
 
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -217,6 +217,8 @@ namespace GitHub.Runner.Common
                 await _websocketConnectTask;
             }
 
+            // "_websocketClient != null" implies either: We have a successful connection OR we have to attempt sending again and then reconnect
+            // ...in other words, if websocket client is null, we will skip sending to websocket and just use rest api calls to send data
             if (_websocketClient != null)
             {
                 var linesWrapper =  startLine.HasValue? new TimelineRecordFeedLinesWrapper(stepId, lines, startLine.Value):  new TimelineRecordFeedLinesWrapper(stepId, lines);
@@ -247,6 +249,7 @@ namespace GitHub.Runner.Common
                         {
                             Trace.Info($"Exhausted websocket allowed retries, we will not attempt websocket connection for this job to post lines again.");
                             _websocketClient?.CloseOutputAsync(WebSocketCloseStatus.InternalServerError, "Shutdown due to failures", CancellationToken.None);
+                            // By setting it to null, we will ensure that we never try websocket path again for this job
                             _websocketClient = null;
                         }
                     }
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -15,7 +15,7 @@ using Newtonsoft.Json;
 namespace GitHub.Runner.Common
 {
     [ServiceLocator(Default = typeof(JobServer))]
-    public interface IJobServer : IRunnerService
+    public interface IJobServer : IRunnerService, IAsyncDisposable
     {
         Task ConnectAsync(VssConnection jobConnection);
 
@@ -31,8 +31,6 @@ namespace GitHub.Runner.Common
         Task RaisePlanEventAsync<T>(Guid scopeIdentifier, string hubName, Guid planId, T eventData, CancellationToken cancellationToken) where T : JobEvent;
         Task<Timeline> GetTimelineAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, CancellationToken cancellationToken);
         Task<ActionDownloadInfoCollection> ResolveActionDownloadInfoAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid jobId, ActionReferenceList actions, CancellationToken cancellationToken);
-
-        void Dispose();
     }
 
     public sealed class JobServer : RunnerService, IJobServer
@@ -48,7 +46,7 @@ namespace GitHub.Runner.Common
         private int failedAttemptsToPostBatchedLinesByWebsocket = 0;
 
 
-        private static readonly TimeSpan _minDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(1);
+        private static readonly TimeSpan _minDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(100);
         private static readonly TimeSpan _maxDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(500);
         private static readonly int _minWebsocketFailurePercentageAllowed = 50;
         private static readonly int _minWebsocketBatchedLinesCountToConsider = 5;
@@ -143,10 +141,11 @@ namespace GitHub.Runner.Common
              InitializeWebsocketClient(TimeSpan.Zero);
         }
 
-        public void Dispose()
+        public ValueTask DisposeAsync()
         {
             Trace.Info($"Disposing websocket client ...");
             this._websocketClient?.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Shutdown", CancellationToken.None);
+            return ValueTask.CompletedTask;
         }
 
         private void CheckConnection()
@@ -170,21 +169,7 @@ namespace GitHub.Runner.Common
                     Trace.Info($"Creating websocket client ..." + feedStreamUrl);
                     this._websocketClient = new ClientWebSocket();
                     this._websocketClient.Options.SetRequestHeader("Authorization", $"Bearer {accessToken}");
-                    this._websocketConnectTask = Task.Run(async () =>
-                    {
-                        try
-                        {
-                            Trace.Info($"Attempting to start websocket client with delay {delay}.");
-                            await Task.Delay(delay);
-                            await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), default(CancellationToken));
-                            Trace.Info($"Successfully started websocket client.");
-                        }
-                        catch(Exception ex)
-                        {
-                            Trace.Info("Exception caught during websocket client connect, fallback of HTTP would be used now instead of websocket.");
-                            Trace.Error(ex);
-                        }
-                    });
+                    this._websocketConnectTask = ConnectWebSocketClient(feedStreamUrl, delay);
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


src/Runner.Common/JobServer.cs
```diff
@@ -16,24 +19,41 @@ public interface IJobServer : IRunnerService
...
         Task<TaskLog> CreateLogAsync(Guid scopeIdentifier, string hubName, Guid planId, TaskLog log, CancellationToken cancellationToken);
         Task<Timeline> CreateTimelineAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, CancellationToken cancellationToken);
         Task<List<TimelineRecord>> UpdateTimelineRecordsAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, IEnumerable<TimelineRecord> records, CancellationToken cancellationToken);
         Task RaisePlanEventAsync<T>(Guid scopeIdentifier, string hubName, Guid planId, T eventData, CancellationToken cancellationToken) where T : JobEvent;
         Task<Timeline> GetTimelineAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, CancellationToken cancellationToken);
         Task<ActionDownloadInfoCollection> ResolveActionDownloadInfoAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid jobId, ActionReferenceList actions, CancellationToken cancellationToken);
+
+        void Dispose();
```
TingluoHuang, line 35: add `IDisposable` to the interface so we don't need this?

TingluoHuang: or maybe `IAsyncDisposable`

src/Runner.Common/JobServer.cs
```diff
@@ -16,24 +19,41 @@ public interface IJobServer : IRunnerService
     {
         Task ConnectAsync(VssConnection jobConnection);
 
+        void InitializeWebsocketClient(ServiceEndpoint serviceEndpoint);
```
TingluoHuang, line 22: I am trying to think whether we can merge `InitializeWebsocketClient` with `ConnectAsync` since they are both doing some kind of setup before the JobServer can be used.

src/Runner.Common/JobServer.cs
```diff
@@ -125,6 +157,46 @@ private void CheckConnection()
...
+                 _serviceEndpoint.Authorization.Parameters.TryGetValue(EndpointAuthorizationParameters.AccessToken, out var accessToken) &&
+                 !string.IsNullOrEmpty(accessToken))
+            {
+                if (_serviceEndpoint.Data.TryGetValue("FeedStreamUrl", out var feedStreamUrl) && !string.IsNullOrEmpty(feedStreamUrl))
+                {
+                    // let's ensure we use the right scheme
+                    feedStreamUrl = feedStreamUrl.Replace("https://", "wss://").Replace("http://", "ws://");
+                    Trace.Info($"Creating websocket client ..." + feedStreamUrl);
+                    this._websocketClient = new ClientWebSocket();
+                    this._websocketClient.Options.SetRequestHeader("Authorization", $"Bearer {accessToken}");
```
TingluoHuang, line 172: What's the different between `SetRequestHeader` vs set `Options.Credentials`?
I saw there is a Proxy property under the `ClientWebSocket.Options`, do we need to set that?

src/Runner.Common/JobServer.cs
```diff
@@ -125,6 +157,46 @@ private void CheckConnection()
...
+                    this._websocketConnectTask = Task.Run(async () =>
```
TingluoHuang, line 173: i would suggest we create a method and move away from `Task.Run()` (We don't use `Task.Run` in almost everywhere of the runner)
```C#
private async Task ConnectWebSocketClient(string feedStreamUrl)
{
    try
    {
        Trace.Info($"Attempting to start websocket client with delay {delay}.");
        await Task.Delay(delay);
        await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), default(CancellationToken));
        Trace.Info($"Successfully started websocket client.");
    }
    catch(Exception ex)
    {
        Trace.Info("Exception caught during websocket client connect, fallback of HTTP would be used now instead of websocket.");
        Trace.Error(ex);
    }
}
```

Then, you will do `this._websocketConnectTask = ConnectWebSocketClient(feedStreamUrl)`

src/Runner.Common/JobServer.cs
```diff
@@ -125,6 +157,46 @@ private void CheckConnection()
...
+                        try
+                        {
+                            Trace.Info($"Attempting to start websocket client with delay {delay}.");
+                            await Task.Delay(delay);
+                            await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), default(CancellationToken));
+                            Trace.Info($"Successfully started websocket client.");
+                        }
+                        catch(Exception ex)
+                        {
+                            Trace.Info("Exception caught during websocket client connect, fallback of HTTP would be used now instead of websocket.");
```
TingluoHuang, line 184: do we need to set `_websocketClient = null` if we fail to connect?

src/Runner.Common/JobServer.cs
```diff
@@ -135,16 +207,69 @@ public Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName,
...
+                        await _websocketClient.SendAsync(chunk, WebSocketMessageType.Text, endOfMessage:lastChunk, CancellationToken.None);
+                    }
+
+                    pushedLinesViaWebsocket = true;
+                }
+                catch (Exception ex)
+                {
+                    Trace.Info($"Caught exception during append web console line to websocket, let's fallback to sending via non-websocket call (total calls: {totalBatchedLinesAttemptedByWebsocket}, failed calls: {failedAttemptsToPostBatchedLinesByWebsocket}, websocket state: {this._websocketClient?.State}).");
+                    Trace.Error(ex);
+                    failedAttemptsToPostBatchedLinesByWebsocket++;
```
TingluoHuang, line 241: `failedAttemptsToPostBatchedLinesByWebsocket++;` before the trace()?

src/Runner.Common/JobServer.cs
```diff
@@ -16,24 +19,41 @@ public interface IJobServer : IRunnerService
...
         private TaskHttpClient _taskClient;
+        private ClientWebSocket _websocketClient;
+
+        private ServiceEndpoint _serviceEndpoint;
+
+        private int totalBatchedLinesAttemptedByWebsocket = 0;
+        private int failedAttemptsToPostBatchedLinesByWebsocket = 0;
+
+
+        private static readonly TimeSpan _minDelayForWebsocketReconnect = TimeSpan.FromMilliseconds(1);
```
TingluoHuang, line 51: i would do 100ms, 1ms seems too aggressive

src/Runner.Common/JobServer.cs
```diff
@@ -135,16 +207,69 @@ public Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName,
...
+            if (_websocketConnectTask != null)
+            {
+                await _websocketConnectTask;
+            }
 
-        public Task AppendTimelineRecordFeedAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, Guid timelineRecordId, Guid stepId, IList<string> lines, long startLine, CancellationToken cancellationToken)
-        {
-            CheckConnection();
-            return _taskClient.AppendTimelineRecordFeedAsync(scopeIdentifier, hubName, planId, timelineId, timelineRecordId, stepId, lines, startLine, cancellationToken: cancellationToken);
+            if (_websocketClient != null)
```
TingluoHuang, line 219: does `_websocketClient!=null` mean we connect succeed?

src/Runner.Common/JobServer.cs
yaananth: `Credentials` is essentially cookie based ones etc., it doesn't cover bearer.

I don't think we need to set `proxy` property, I tested this using proxy and it worked fine.

src/Runner.Common/JobServer.cs
yaananth: we use that `null` to determine whether to attempt reconnect again, so not here

src/Runner.Common/JobServer.cs
yaananth: not necessarily, if it's `null` it means we shouldn't proceed at all - for sending/reconnect.

If it's not null, then we might have to reconenct/in the attempting to send to websocket scenario.

src/Runner.Common/JobServer.cs
yaananth: ...also I think we initialize this in jobrunner etc., places, where we don't need to create websocket. I think explicit method for websocket can help to ensure we kick start that only when we need to

yaananth: ...also I think we initialize this in jobrunner etc., places, where we don't need to create websocket. I think explicit method for websocket can help to ensure we kick start that only when we need to

src/Runner.Common/JobServer.cs
```diff
@@ -135,16 +208,72 @@ public Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName,
...
+                    failedAttemptsToPostBatchedLinesByWebsocket++;
+                    Trace.Info($"Caught exception during append web console line to websocket, let's fallback to sending via non-websocket call (total calls: {totalBatchedLinesAttemptedByWebsocket}, failed calls: {failedAttemptsToPostBatchedLinesByWebsocket}, websocket state: {this._websocketClient?.State}).");
+                    Trace.Error(ex);
+                    if (totalBatchedLinesAttemptedByWebsocket > _minWebsocketBatchedLinesCountToConsider)
+                    {
+                        // let's consider failure percentage
+                        if (failedAttemptsToPostBatchedLinesByWebsocket * 100 / totalBatchedLinesAttemptedByWebsocket > _minWebsocketFailurePercentageAllowed)
+                        {
+                            Trace.Info($"Exhausted websocket allowed retries, we will not attempt websocket connection for this job to post lines again.");
+                            _websocketClient?.CloseOutputAsync(WebSocketCloseStatus.InternalServerError, "Shutdown due to failures", CancellationToken.None);
```
thboop, line 251: Should we pass in the actual CancellationToken here?

src/Runner.Common/JobServer.cs
```diff
@@ -135,16 +208,72 @@ public Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName,
...
+                            // By setting it to null, we will ensure that we never try websocket path again for this job
+                            _websocketClient = null;
```
thboop, line 253: nit: consider a private function that calls that we can reuse for `DisposeAsync` and here
```
private TeardownWebsocketClient ( reason, message, cancellationToken)
{
   _websocketClient?.CloseOutputAsync( reason, message, cancellationToke
    _websocketClient = null;
}
```

yaananth: we will be sending all those params, and we won't need to make it `null` in the other case, not totally sure if it's worth it, if you have strong opinion, let me know, I can refactor that into method!

src/Runner.Common/JobServerQueue.cs
```diff
@@ -178,8 +160,8 @@ public async Task ShutdownAsync()
             await ProcessTimelinesUpdateQueueAsync(runOnce: true);
             Trace.Info("Timeline update queue drained.");
 
-            Trace.Info($"Disposing websocket client ...");
-            this._websocketClient?.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Shutdown", CancellationToken.None);
+            Trace.Info($"Disposing job server ...");
+            await _jobServer.DisposeAsync();
```
TingluoHuang, line 164: do you need to call `DisposeAsync()` or the GC will do it for you?

src/Runner.Common/JobServer.cs
```diff
@@ -117,6 +135,19 @@ private async Task CheckNetworkEndpointsAsync(int attemptsLeft)
...
 
+        public void InitializeWebsocketClient(ServiceEndpoint serviceEndpoint)
+        {
+             this._serviceEndpoint = serviceEndpoint;
+             InitializeWebsocketClient(TimeSpan.Zero);
+        }
+
+        public ValueTask DisposeAsync()
+        {
+            Trace.Info($"Disposing websocket client ...");
```
TingluoHuang, line 146: don't trace, since the `Trace` may already gone.

src/Runner.Common/JobServer.cs
```diff
@@ -117,6 +135,19 @@ private async Task CheckNetworkEndpointsAsync(int attemptsLeft)
...
+            this._websocketClient?.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, "Shutdown", CancellationToken.None);
+            return ValueTask.CompletedTask;
```
TingluoHuang, line 148: do you need to call `GC.SuppressFinalize(this);`?

TingluoHuang: I remember we call it in other files of the runner.

yaananth: done! Kind of surprised though, I never thoughts we will be poking with `GC`, but looks like the pattern here, so added

TingluoHuang (approved):

PR #1757, opened 2022-03-16
Added continue on error to composite action

### LINKED ISSUE #1457: Add ability to use continue-on-error from composite Action steps
**Describe the enhancement**
Add the ability to use ``continue-on-error`` option for steps of composite actions.

**Code Snippet**
````yaml
runs:
  using: "composite"
  steps:
    - name: Unstable Step
      uses: risky-step/risky-step@v1.0.0
      continue-on-error: true
      with:
         output_log_location: /var/log/my_risky_build.txt

    - name: Check Result
      shell: bash
      run: 
        cat /var/log/my_risky_build.txt 
````

### PR DESCRIPTION
**Pull request by nikola-jokic**
Fixes #1457

### DIFF
--- a/src/Runner.Worker/action_yaml.json
+++ b/src/Runner.Worker/action_yaml.json
@@ -129,6 +129,7 @@
                         "required": true
                     },
                     "env": "step-env",
+                    "continue-on-error": "boolean-strategy-context",
                     "working-directory": "string-steps-context",
                     "shell": {
                         "type": "non-empty-string",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Runner.Worker/action_yaml.json [hunk 1]
```diff
@@ -201,6 +203,15 @@
             ],
             "string": {}
         },
+        "boolean-strategy-context": {
```
fhammerl, line 206: use `boolean-steps-context` from workflow.json

PR #1762, opened 2022-03-17
composite action github.action_path set based on the StepHost

### LINKED ISSUE #1221: Composite actions which contains external bash script doesn't run if the jobs are executed with container
## Bug!!!
I am trying to create my own composite action and run in my workflow as described [here](https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#running-local-scripts)

When I run my workflow job with **container** then the bash script then the bash script is not accessible to run.
The error in the logs I see as:

```
/__w/_temp/75ecb274-82d9-459d-8632-54e0099fe49e.sh: line 1: /home/xxx/actions-runner/_work/_actions/abc/ooo-action/v1.0/deploy.sh: No such file or directory
Error: Process completed with exit code 127.
```
But on the other hand, if I see in the runner, the script exists in the specified path.

**To Reproduce**
1. Setup the composite action as described [here](https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action)
2. Write your workflow to run the action but run the jobs on **container**
```
runs-on: [self-hosted]    
    container: 
      image: myregistry/xxx/abc/abc-image:1.0
      credentials:
        username: ${{ github.actor }}
        password: ${{ secrets.GITHUB_TOKEN }}
```
3. The action will fail

**Expected behavior**
Bash script in the composite action should have been executed but it resulted in error stating _there is no such file_.

Runner version - 2.278.0
Running on Linux - x64

## What's not working?
The script in composite action is not running with below error:
```
/__w/_temp/75ecb274-82d9-459d-8632-54e0099fe49e.sh: line 1: /home/xxx/actions-runner/_work/_actions/abc/ooo-action/v1.0/deploy.sh: No such file or directory
Error: Process completed with exit code 127.
```

nikola-jokic:
Closes #1221

### DIFF
--- a/src/Runner.Worker/Handlers/CompositeActionHandler.cs
+++ b/src/Runner.Worker/Handlers/CompositeActionHandler.cs
@@ -86,7 +86,7 @@ namespace GitHub.Runner.Worker.Handlers
 
                 ExecutionContext.StepTelemetry.HasPreStep = Data.HasPre;
                 ExecutionContext.StepTelemetry.HasPostStep = Data.HasPost;
-                
+
                 ExecutionContext.StepTelemetry.HasRunsStep = hasRunsStep;
                 ExecutionContext.StepTelemetry.HasUsesStep = hasUsesStep;
                 ExecutionContext.StepTelemetry.StepCount = steps.Count;
@@ -158,7 +158,7 @@ namespace GitHub.Runner.Worker.Handlers
                     step.ExecutionContext.ExpressionValues["github"] = gitHubContext;
 
                     // Set GITHUB_ACTION_PATH
-                    step.ExecutionContext.SetGitHubContext("action_path", ActionDirectory);
+                    step.ExecutionContext.SetGitHubContext("action_path", StepHost.ResolvePathForStepHost(ActionDirectory));
 
                     embeddedSteps.Add(step);
                 }


thboop (dismissed):
LGTM

ChristopherHX:
> it this not getting converted for some reason?

You are only converting the environment variables, no single expressions contexts has ever been translated to a container path.
Even paths of custom environment variables get converted
e.g.
The following prints host path
```yaml
- run: echo '${{github.workspace}}'
  shell: bash
```
The following two prints the container path
```yaml
- run: echo  $WP
  env:
    WP: ${{github.workspace}}
  shell: bash
```
```yaml
- run: echo  $GITHUB_WORKSPACE
  shell: bash
```

`GITHUB_ACTION_PATH` already holds the container path, but the composite docs suggest to use `${{ github.action_path }}` in a run step to call a bash script, which leads to problems.

thboop:
So, this does solve the issue for this specific context variable, however this problem exists for ALL path based context variables. For example consider the composite action:

```
  using: "composite"
  steps:
    - run: printenv
      shell: bash
    - run: |
        echo ${{github.action_path}}
        echo ${{github.event_path}}
        echo ${{github.workspace}}
      shell: bash

```

The env's will get translated, the `echo` statements will not. What we should really do is in [Actionsrunner](https://github.com/actions/runner/blob/d081289ed53841c9d78c3e796c429965c4fe5d58/src/Runner.Worker/ActionRunner.cs#L175) where we evaluate step inputs, use the step host to modify the context

```
            // TODO: Translate expression values into container expression values and pass that in here.
            // TODO: Temp create another dictionary for expression values and feed it in here, we need to map the path based containers to push it in.
            var inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
```

src/Sdk/DTPipelines/Pipelines/ContextData/DictionaryContextData.cs
```diff
@@ -1,8 +1,8 @@
 ﻿using System;
+using System.Runtime.Serialization;
```
thboop, line 2: nit: revert this

TingluoHuang: +1 since this was copied from the service side code.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -165,14 +164,14 @@ public async Task RunAsync()
...
+
             // Setup File Command Manager
             var fileCommandManager = HostContext.CreateService<IFileCommandManager>();
             fileCommandManager.InitializeFiles(ExecutionContext, null);
 
             // Load the inputs.
             ExecutionContext.Debug("Loading inputs");
-            var templateEvaluator = ExecutionContext.ToPipelineTemplateEvaluator();
-            var inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
+            var inputs = EvaluateStepInputs(stepHost);
```
TingluoHuang, line 174: I would prefer to wrap this change behind a feature flag so we can easily roll back. Breaking inputs are not fun.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -299,6 +298,44 @@ public bool TryEvaluateDisplayName(DictionaryContextData contextData, IExecution
...
+            var templateEvaluator = ExecutionContext.ToPipelineTemplateEvaluator();
+            var inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, expressionValues, ExecutionContext.ExpressionFunctions);
+
+            return inputs;
+        }
+
+        private void UpdatePathsInExpressionValues(string contextName, DictionaryContextData expressionValues, IStepHost stepHost)
+        {
+            var dict = expressionValues[contextName].AssertDictionary($"expected context {contextName} to be a dictionary");
+            var output = dict.Clone().AssertDictionary($"expected dictionary clone in {contextName} to be dictionary");
```
TingluoHuang, line 327: do we need to have `output`? can we update `dict` by looping through `dict.keys`?
ex:
```C#
foreach(var key in dict.Keys)
{
   var value = dict[key]?.ToString();
   if(!string.IsNullOrEmpty(value))
  {
     dict[key] = new StringContextData(stepHost.ResolvePathForStepHost(value));
  }
}
```

src/Runner.Worker/ActionRunner.cs
```diff
@@ -299,6 +298,44 @@ public bool TryEvaluateDisplayName(DictionaryContextData contextData, IExecution
...
+            foreach (var pair in dict)
+            {
+                var value = pair.Value?.ToString();
+                if (pair.Value?.Type == TokenType.String && !String.IsNullOrEmpty(value) && pair.Key != null)
```
TingluoHuang, line 331: is `!String.IsNullOrEmpty(value)` good enough for the `if`? will `pair.Key` ever be null?

TingluoHuang:
Can we update the PR description to include an output of dumping `github` and `runner` context with your changes? want to make sure all values are expected.  🙇

nikola-jokic:
@TingluoHuang of course! I will fix all of the suggestions as soon as I can 👍

src/Runner.Worker/ActionRunner.cs
fhammerl: [Feature flag example here](https://github.com/actions/runner/blob/909b05eb66a81e278a6f95e80532f972bacaf2b0/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs#L117)

src/Runner.Common/Constants.cs
```diff
@@ -150,6 +150,8 @@ public static class Features
             {
                 public static readonly string DiskSpaceWarning = "runner.diskspace.warning";
                 public static readonly string Node12Warning = "DistributedTask.AddWarningToNode12Action";
+
```
TingluoHuang, line 153: remove extra space. 😄

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -34,6 +29,7 @@ public interface IStepHost : IRunnerService
                                bool killProcessOnCancel,
                                bool inheritConsoleHandler,
                                CancellationToken cancellationToken);
+        DictionaryContextData GetExpressionValues(IExecutionContext executionContext);
```
TingluoHuang, line 32: add an empty line before this.

TingluoHuang: seems weird that we add an interface method that will eventually just call `ResolvePathForStepHost()`

nikola-jokic: Thanks for the review!  

My thoughts are that the logic is only dependent on the execution context and the step host. Since the IExecutionContext is being used in a lot of places, and StepHost does the translation, it is clear when we call it to see that we want expression values for that StepHost. We pass the execution context, and we get expression values that should be used when the template is evaluated. That way, wherever we use template evaluation and we have a step host, we just call the step host to translate those values for the template.

Also, if the new StepHost is introduced, it should also know what expressionValues should be for it to work properly

fhammerl: I'm inclined to agree the logic in `GetExpressionValues` and private method `UpdatePathsInExpressionValues` belongs in `ActionRunner.cs`, the translation of ExpressionValues like this feels like a general case. 

In StepHost, we eventually do just call `ResolvePathForStepHost`. Most of this method is prepwork for `Container.TranslateToContainerPath`. This 'prepwork' is not operating on any properties of 'StepHost' other than 'Container', so it doesn't present a problem to cross reference from `GetExpressionValues` to `ResolvePathForStepHost`.

Alternatively, we could modify ResolvePathForStepHost and move all the the logic between `ArgUtil.NotNull(Container, nameof(Container));` and `Container.MountVolumes.Exists` to a private method `isPathContainerTranslatable`. `isPathContainerTranslatable` can then be shared between ResovePathForStepHost and GetExpressionValues?

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -239,5 +241,27 @@ public async Task<string> DetermineNodeRuntimeVersion(IExecutionContext executio
                                                          cancellationToken: cancellationToken);
             }
         }
+
+        public DictionaryContextData GetExpressionValues(IExecutionContext executionContext)
```
thboop, line 245: Can we move this to an extension method on execution context. It adds a a dependency on the step host (which we don't want for execution context) but its okay for an extension method.

We should be able to l0 test the extension method using these existing tests, and we can call it in Action Runner

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -239,5 +241,27 @@ public async Task<string> DetermineNodeRuntimeVersion(IExecutionContext executio
...
+            return expressionValues;
+        }
+
+        private void UpdatePathsInExpressionValues(string contextName, DictionaryContextData expressionValues) 
+        {
+            var dict = expressionValues[contextName].AssertDictionary($"expected context {contextName} to be a dictionary");
+            foreach (var key in dict.Keys.ToList())
+            {
+                var value = dict[key]?.ToString();
+                if (!string.IsNullOrEmpty(value))
```
thboop, line 259: We need to only modify it if it is a string context data, otherwise we are going to blow up the `github.event` which is an object

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1064,6 +1061,66 @@ private IExecutionContext CreatePostChild(string displayName, Dictionary<string,
             var newGuid = Guid.NewGuid();
             return CreateChild(newGuid, displayName, newGuid.ToString("N"), null, null, ActionRunStage.Post, intraActionState, _childTimelineRecordOrder - Root.PostJobSteps.Count, siblingScopeName: siblingScopeName);
         }
+
+        public DictionaryContextData GetExpressionValues(IStepHost stepHost)
```
fhammerl, line 1075: Can we move it to an extensionmethod in `ExecutionContextExtension`

fhammerl (approved):
LGTM!

faubion-hbo:
Hi,
i'm running on the latest github runner (2.290.1) that should have this fix in it, but am still seeing a difference between the outputs of $GITHUB_ACTION_PATH and ${{ github.action_path }} when running a job within a container where ${{ github.action_path }} doesn't map to the right directory. i.e.
```
${{ github.action_path }} == /home/runner/work/_actions/<org>/<action>/<tag>
$GITHUB_ACTION_PATH == /__w/_actions/<org>/<action>/<tag>
```
where the container image is being pulled from ECR

AllanOricil:
does not work

zachspar:
The problem still exists on self-hosted runners (actions-runner-controller) using defined container value.

PR #1765, opened 2022-03-17
Development environment: Add extensions and initialisation steps for a smoother codespace / devcontainer experience

fhammerl:
This PR adds gitlens and csharp extensions to the codespace / devcontainer you check out.
It also triggers dotnet restore on the projects and symlinks the image's dotnet sdk to runner/_dotnetsdk (so we don't have to download it on first `dev.sh l`.

When opening this repo locally (as a devcontainer), the whole runner directory is mounted, including _layout and _dotnetsdk, so if you're not on a Linux host machine, those folders must be purged first.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,32 @@
...
+	"settings": {
+		"files.watcherExclude": {
+			"**/_layout/**": true
+		}
+	},
+	"remoteUser": "codespace",
+	"overrideCommand": false,
+	"mounts": ["source=codespaces-linux-var-lib-docker,target=/var/lib/docker,type=volume"],
+	"runArgs": [
+		"--cap-add=SYS_PTRACE",
```
thboop, line 14: What do these args do? Do we need them all?

.devcontainer/init.sh
```diff
@@ -0,0 +1,12 @@
+### Create _dotnetsdk
+echo 'Creating symlink for embedded sdk...'
+ln -s /home/codespace/.dotnet/sdk/ _dotnetsdk 
+echo .6.0.100 >> _dotnetsdk/6.0.100/.6.0.100 
+ln -s /home/codespace/.dotnet/dotnet _dotnetsdk/6.0.100/dotnet
+
+### Restore 
+dotnet restore src/Runner.Listener &
```
thboop, line 8: Are we able to just restore at the root recursively rather then having to maintain a list of all projects here

.devcontainer/init.sh
```diff
@@ -0,0 +1,12 @@
+### Create _dotnetsdk
+echo 'Creating symlink for embedded sdk...'
+ln -s /home/codespace/.dotnet/sdk/ _dotnetsdk 
+echo .6.0.100 >> _dotnetsdk/6.0.100/.6.0.100
```
thboop, line 4: Do we need to update this when we update the dotnet version?

PR #1767, opened 2022-03-18
Added input context to shell in composite run-step

### LINKED ISSUE #835: unable to inject shell to composite action
**Describe the bug**
Unable to pass the shell via input parameter to the composite action.

**To Reproduce**
1. create following composite action:
````
name: 'test actions'
description: 'test action description'
inputs:
  shell:
    description: 'shell to use'
    required: false
    default: bash
runs:
  using: "composite"
  steps:
    - name:
      run: echo "bla bla bla"
      shell: ${{ inputs.shell}}
````
2. use the action in some workflow
3. Job fails on the 'Set up job' phase, printing the following errors:
````
Current runner version: '2.274.2'
Operating System
Virtual Environment
Prepare workflow directory
Prepare all required actions
Getting action download info
Download action repository 'msys2/setup-msys2@v2'
Download action repository 'actions/checkout@main'
Download action repository 'myci-actions/add-pacman-repo@master'
Error: myci-actions\add-pacman-repo\master\action.yml (Line: 25, Col: 14):
Error: myci-actions\add-pacman-repo\master\action.yml (Line: 25, Col: 14): Unrecognized named-value: 'inputs'. Located at position 1 within expression: inputs.shell
Error: GitHub.DistributedTask.ObjectTemplating.TemplateValidationException: The template is not valid. myci-actions\add-pacman-repo\master\action.yml (Line: 25, Col: 14): Unrecognized named-value: 'inputs'. Located at position 1 within expression: inputs.shell
   at GitHub.DistributedTask.ObjectTemplating.TemplateValidationErrors.Check()
   at GitHub.Runner.Worker.ActionManifestManager.ConvertRuns(IExecutionContext executionContext, TemplateContext templateContext, TemplateToken inputsToken, String f[...]
Error: Fail to load myci-actions\add-pacman-repo\master\action.yml
````

where `(Line: 25, Col: 14)` corresponds to the last line of the action from step 1. I.e. `shell: ${{ inputs.shell}}`

**Expected behavior**
The action works fine and uses the shell passed in as input argument to the action.

`windows-latest`

### PR DESCRIPTION
**Pull request by nikola-jokic**
Fixes #835

### DIFF
--- a/src/Runner.Worker/action_yaml.json
+++ b/src/Runner.Worker/action_yaml.json
@@ -131,7 +131,7 @@
                     "env": "step-env",
                     "working-directory": "string-steps-context",
                     "shell": {
-                        "type": "non-empty-string",
+                        "type": "string-shell-context",
                         "required": true
                     }
                 }
@@ -198,6 +198,12 @@
                 "runner",
                 "env",
                 "hashFiles(1,255)"
+            ],
+            "string": {}
+        },
+        "string-shell-context": {
+            "context": [
+                "inputs"
             ],
             "string": {}
         },

### CODE REVIEW

File: src/Runner.Worker/action_yaml.json [hunk 1]
```diff
@@ -131,7 +131,7 @@
                     "env": "step-env",
                     "working-directory": "string-steps-context",
                     "shell": {
-                        "type": "non-empty-string",
+                        "type": "string-shell-context",
```
thboop, line 134: I think we should just do `string-steps-context` here. While most of those contexts may not be useful, there is also no reason to fail if someone finds a way to use `set-env` to set the `env` context to  the shell they want.

We will want to test all of these contexts in that list work as intended though, and we evaluate them properly

nikola-jokic: Updated the context and tested. It seems like all of those contexts are available to us in shell now 😊

fhammerl (approved):
LGTM, could you link to a (screenshot) of a workflow of this?

vlad-nn:
Guys, can you expedite this fix? Really needed ASAP.

PR #1768, opened 2022-03-18
Added Kotlin as a shell option

### LINKED ISSUE #813: "Custom shell" (kotlin) should allow custom file name extension (main.kts)
**Describe the enhancement**
Actions allows to define a [custom shell](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#custom-shell) in worflow runs.
Kotlin has support for [scripting](https://github.com/Kotlin/kotlin-script-examples/blob/master/jvm/main-kts/MainKts.md).
I've tried to make use of Kotlin for runninng a script in Actions.
The issue is that kotlin expects a file name ending in `.main.kts`, but the runner creates a file without extensions for "shells" that are not hardcoded in https://github.com/actions/runner/blob/a2e32170fd83851569804fa7b27d2cae235a5225/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs#L38

File name construction: https://github.com/actions/runner/blob/a2e32170fd83851569804fa7b27d2cae235a5225/src/Runner.Worker/Handlers/ScriptHandler.cs#L233

**Code Snippet**

Consider this example workflow: 

```
name: CI
on:
  push:
  workflow_dispatch:
jobs:
  build:
    runs-on: ubuntu-20.04
    steps:
      - uses: actions/checkout@v2
      - uses: fwilhe2/setup-kotlin@main
      - run: |
            #!/usr/bin/env kotlin
            println("hello")
        shell: kotlin {0}
```

I would expect to be able to run the given script in kotlin (the cli compiler is installed on the runner via `fwilhe2/setup-kotlin`).

**Additional information**

Result of the given workflow script:

image

This is because kotlin requires the `main.kts` suffix to know this should be run as a script.

I would need some way to provide a map of `kotlin` to `.main.kts` to `ScriptHandlerHelpers.GetScriptFileExtension()`

Compare to running the script locally with the right or the wrong filename

```
~$ cat foo.main.kts 
#!/usr/bin/env kotlin
println("hello")
~$ kotlin foo.main.kts 
hello
~$ mv foo.main.kts foo
~$ kotlin foo
error: could not find or load main class foo
```

Did I overlook something? Is this already possible? If not, I'd be glad if some option to make this would be provided, as it would make the "custom shell" feature much more interesting.

Thanks!

ruvceskistefan:
Closes #813

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
@@ -13,7 +13,8 @@ namespace GitHub.Runner.Worker.Handlers
             ["powershell"] = "-command \". '{0}'\"",
             ["bash"] = "--noprofile --norc -e -o pipefail {0}",
             ["sh"] = "-e {0}",
-            ["python"] = "{0}"
+            ["python"] = "{0}",
+            ["kotlin"] = "{0}"
         };
 
         private static readonly Dictionary<string, string> _extensions = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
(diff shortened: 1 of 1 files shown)


thboop:
The bar we have for new shell options is very high, I don't think we are going to support Kotlin at this time until we hear more feedback from the community. Currently you can write your own kotlin actions using composite actions or via scripts, so user workarounds are available.

PR #1775, opened 2022-03-22
Fix for enhancement 1218: Added repository name and workflow file name

### LINKED ISSUE #1218: Add the ability to output additional job details to the console

**Describe the enhancement**
We have multiple repositories sharing our self-hosted runners and it is difficult to know which one is running by the job name alone.  It would be great if there was additional information, such as RepoName and WorkFlowName, to be able to determine which is currently running.  We currently have to go in through the UI and do a lot of investigating to see which repo's jobs are running.

**Code Snippet**
If applicable, add a code snippet.

**Additional information**
Example of console output:
image

ruvceskistefan:
Closes #1218

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -366,7 +366,7 @@ namespace GitHub.Runner.Listener
                 string additionalInfo = string.Empty;
                 if (!string.IsNullOrEmpty(workflowName))
                 {
-                    additionalInfo += $"(workflow \"{workflowName}\"";
+                    additionalInfo += $"(workflow \"{Pipelines.Validation.NameValidation.Sanitize(workflowName, allowHyphens: true, allowDots: true)}\"";
                     if (!string.IsNullOrEmpty(repositoryName))
                     {
                         additionalInfo += $" in repository \"{Pipelines.Validation.NameValidation.Sanitize(repositoryName, allowHyphens: true, allowSlashes: true)}\")";
--- a/src/Sdk/DTPipelines/Pipelines/Validation/NameValidation.cs
+++ b/src/Sdk/DTPipelines/Pipelines/Validation/NameValidation.cs
@@ -35,7 +35,8 @@ namespace GitHub.DistributedTask.Pipelines.Validation
         public static String Sanitize(
             String name,
             Boolean allowHyphens = false,
-            Boolean allowSlashes = false)
+            Boolean allowSlashes = false,
+            Boolean allowDots = false)
         {
             if (name == null)
             {
@@ -50,7 +51,8 @@ namespace GitHub.DistributedTask.Pipelines.Validation
                     (name[i] >= '0' && name[i] <= '9' && sb.Length > 0) ||
                     (name[i] == '_') ||
                     (allowHyphens && name[i] == '-' && sb.Length > 0) ||
-                    (allowSlashes && name[i] == '/' && sb.Length > 0))
+                    (allowSlashes && name[i] == '/' && sb.Length > 0) ||
+                    (allowDots && name[i] == '.' && sb.Length > 0))
                 {
                     sb.Append(name[i]);
                 }


jonatanlr98 (commented):
ruvceskistefan:stefan-ruvceski/feature-fix/1218

jonatanlr98:
ruvceskistefan:stefan-ruvceski/feature-fix/1218

jonatanlr98 (approved):

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                if(message.Variables.TryGetValue("system.workflowFilePath", out var workflowPath))
+                {
+                    workflowName = workflowPath.Value.Split('/').LastOrDefault().Trim();
+                }
+
+                string repositoryName = string.Empty;
+                if(message.ContextData.TryGetValue("github", out var context))
+                {
+                    var githubContext = context as Pipelines.ContextData.DictionaryContextData;
+                    if(githubContext.TryGetValue("repository", out var repository))
```
thboop, line 360: We should fail silently if these casts fail, `githubContext?.TryGetValue(`

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                    {
+                        repositoryName = (repository as Pipelines.ContextData.StringContextData).Value.Trim();
```
thboop, line 362: Same here, if these values are null/unexpected we shouldn't crash the runner

src/Runner.Listener/JobDispatcher.cs
jonatanlr98: Ok

jonatanlr98 (approved):

jonatanlr98 (approved):

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: Done

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: Done

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                 }
 
                 var term = HostContext.GetService<ITerminal>();
-                term.WriteLine($"{DateTime.UtcNow:u}: Running job: {message.JobDisplayName}");
+
+                string workflowName = string.Empty;
+                if(message.Variables.TryGetValue("system.workflowFilePath", out var workflowPath))
```
TingluoHuang, line 351: this might not work if the job is coming from a callable workflow since `system.workflowFilePath` is always pointed to the entry workflow.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                {
+                    workflowName = workflowPath.Value.Split('/').LastOrDefault().Trim();
+                }
+
+                string repositoryName = string.Empty;
+                if(message.ContextData.TryGetValue("github", out var context))
```
TingluoHuang, line 357: you might want to format the file, I think you miss a space between `if` and `(message`. 😄

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                {
+                    var githubContext = context as Pipelines.ContextData.DictionaryContextData;
+                    if(githubContext?.TryGetValue("repository", out var repository) == true)
+                    {
+                        repositoryName = (repository as Pipelines.ContextData.StringContextData)?.Value?.Trim();
+                    }
+                }
+                
+                string additionalInfo = string.Empty;
```
TingluoHuang, line 366: should we just send [Job_Workflow_ref](https://cs.github.com/github/actions-dotnet/blob/b668258657a87d3fcc44a41a6e5b525323d51c7c/Actions/Runtime/Sdk/Server/TaskHub.cs#L5236) from service so you don't need any of these string magic?

ruvceskistefan: I agree that this is also an option, but I wanted the change to be exclusively on the runner.

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: Done.

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: In that case, the name of the entry workflow will be displayed, ie the workflow that calls the reusable workflow. Also, the job name from the entry workflow will be displayed, as well as the job name from the reusable workflow (entry workflow job name / reusable workflow job name). Here is an example of output in case of reusable workflow:
```
2022-03-29 07:36:15Z: Running job: "call-workflow / first_job" (workflow "use_callable_workflow.yml" in repository "ruvceskistefan/git-runner-test")
```
`call-workflow` is job name in entry workflow,
`first_job` is job name in callable workflow,
`use_callable_workflow.yml` is entry workflow and
`ruvceskistefan/git-runner-test` is repository name for entry workflow.

In my opinion this works properly, because the user can easily check which reusable workflow is called from entry workflow.

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: I can't access the link you sent, i get 404

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -346,7 +346,38 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                    if (!string.IsNullOrEmpty(repositoryName))
+                    {
+                        additionalInfo += $" in repository \"{repositoryName}\")";
+                    }
+                    else
+                    {
+                        additionalInfo += ")";
+                    }
+                }
+
```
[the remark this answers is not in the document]
ruvceskistefan: `NameValidation.Sanitize` will remove characters '/' from repo name, which I guess isn't very good. I can change the `NameValidation.Sanitize` method so that '/' is valid character, but I guess it can make the problem elsewhere.

Im2451soon (approved):

Im2451soon (dismissed):

src/Runner.Listener/JobDispatcher.cs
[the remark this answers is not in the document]
ruvceskistefan: I can add an optional argument `allowSlashes` (like already existing optional argument `allowHyphens`) to the `NameValidation.Sanitize` method. I think that in this way I will cover all valid characters in the repository name, and I will not change the remaining method `NameValidation.Sanitize` calls.

PR #1781, opened 2022-03-24
Added warning in case of invalid combination of command and flags and/or arguments

### LINKED ISSUE #1731: Runner should warn if wrong flags are passed to the runner
**Describe the enhancement**
With regards to Issue #1729, the runner should warn if the unknown flag is passed, so the user invoking the execution of the runner can know that the invocation of the runner is wrong.

There were multiple issues trying to pass `--disableupdate` to the run.sh, which caused the runner to execute normally, with updates if `--disableupdate` was not passed in the configuration step. 

It would be better if the invocation of the runner exits saying that the flag is not recognized, as the behaviour of most command-line tools on Linux for example.

ruvceskistefan:
Closes #1731


src/Runner.Listener/CommandSettings.cs
```diff
@@ -17,43 +17,65 @@ public sealed class CommandSettings
...
-            Constants.Runner.CommandLine.Args.RunnerGroup,
-            Constants.Runner.CommandLine.Args.StartupType,
-            Constants.Runner.CommandLine.Args.Token,
-            Constants.Runner.CommandLine.Args.Url,
-            Constants.Runner.CommandLine.Args.UserName,
-            Constants.Runner.CommandLine.Args.WindowsLogonAccount,
-            Constants.Runner.CommandLine.Args.WindowsLogonPassword,
-            Constants.Runner.CommandLine.Args.Work
+            // Valid configure flags and args
+            [Constants.Runner.CommandLine.Commands.Configure] =
```
thboop, line 24: Can we make a general list that is 

```
                    Constants.Runner.CommandLine.Flags.Help,
                    Constants.Runner.CommandLine.Flags.Version,
                    Constants.Runner.CommandLine.Flags.Commit,
                    Constants.Runner.CommandLine.Flags.Check
```
and we check if the validOptions[command] or the general list includes your options. That way we don't have to repeat all of this everywhere.

src/Test/L0/Listener/CommandSettingsL0.cs
```diff
@@ -708,33 +708,86 @@ public void ValidateCommands()
...
-
-        [Fact]
-        [Trait("Level", "L0")]
-        [Trait("Category", nameof(CommandSettings))]
-        public void ValidateArgs()
+                Assert.Contains(flag, command.Validate());
+            }
+        }
+
+        [Theory]
```
thboop, line 744: Love the tests!

src/Runner.Listener/CommandSettings.cs
```diff
@@ -17,43 +17,56 @@ public sealed class CommandSettings
...
+            // Valid configure flags and args
+            [Constants.Runner.CommandLine.Commands.Configure] = 
+                new string[] 
+                {
+                    Constants.Runner.CommandLine.Flags.DisableUpdate,
+                    Constants.Runner.CommandLine.Flags.Ephemeral,
+                    Constants.Runner.CommandLine.Flags.Replace,
+                    Constants.Runner.CommandLine.Flags.RunAsService,
+                    Constants.Runner.CommandLine.Flags.Unattended,
+                    Constants.Runner.CommandLine.Args.Auth,
```
TingluoHuang, line 41: Do we use `Auth` anywhere?

src/Runner.Listener/CommandSettings.cs
```diff
@@ -17,43 +17,56 @@ public sealed class CommandSettings
...
+                    Constants.Runner.CommandLine.Args.Labels,
+                    Constants.Runner.CommandLine.Args.MonitorSocketAddress,
```
TingluoHuang, line 43: do we still use `MonitorSocketAddress`?

src/Runner.Listener/CommandSettings.cs
thboop: I think so for backcompat for codedev environments

src/Runner.Listener/CommandSettings.cs
```diff
@@ -17,43 +17,56 @@ public sealed class CommandSettings
...
+                    Constants.Runner.CommandLine.Args.UserName,
+                    Constants.Runner.CommandLine.Args.WindowsLogonAccount,
+                    Constants.Runner.CommandLine.Args.WindowsLogonPassword,
+                    Constants.Runner.CommandLine.Args.Work
+                },
+            // Valid remove flags and args
+            [Constants.Runner.CommandLine.Commands.Remove] =
+                new string[]
+                {
+                    Constants.Runner.CommandLine.Args.Token
```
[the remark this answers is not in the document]
thboop: we should!

src/Runner.Listener/CommandSettings.cs
thboop: We still have the code for it in the runner, i'm happy to leave it as a config option until we clean that code up. Not sure if it is actively used currently.

thboop:
Could we update the error message to provide more context.

1. If we can easily get the name of the command being run (or at least the first command if there are multiple we could do
```
terminal.WriteError($"Unrecognized command-line input arguments for command ${command}: '{string.Join(", ", unknownCommandlines)}'. For usage refer to: .\\config.cmd --help or ./config.sh --help");
```
2. I don't see an easy way to get the the command name in program.cs , so we could also:
```
terminal.WriteError($"This command does not recognize the command-line input arguments: '{string.Join(", ", unknownCommandlines)}'. For usage refer to: .\\config.cmd --help or ./config.sh --help");
```

ruvceskistefan:
> Could we update the error message to provide more context.
> 
> 1. If we can easily get the name of the command being run (or at least the first command if there are multiple we could do
> 
> ```
> terminal.WriteError($"Unrecognized command-line input arguments for command ${command}: '{string.Join(", ", unknownCommandlines)}'. For usage refer to: .\\config.cmd --help or ./config.sh --help");
> ```
> 
>  2. I don't see an easy way to get the the command name in program.cs , so we could also:
> 
> ```
> terminal.WriteError($"This command does not recognize the command-line input arguments: '{string.Join(", ", unknownCommandlines)}'. For usage refer to: .\\config.cmd --help or ./config.sh --help");
> ```

I can implement a method in the `CommandSettings` class that will return the command's name based on the flags `Configure`, `Remove`, `Run`, and `Warmup`, and then use that method in `Program.cs`.

thboop (approved):
lgtm

PR #1790, opened 2022-03-28
FeedStream 289 patch: handle websocket close failures

yaananth:
To handle https://github.com/github/c2c-actions-service/issues/2906 gracefully on the runner side.

Essentially, we are attempting to send close frame to a websocket that might not be open yet. 

So, we were bailing out, let's not bail out, we should be able to proceed even if we failed to close the websocket, the server should have timeouts set up already to close it eventually.

We are **patching** since we need this runner for **GHES** 🙇 

**Main**: https://github.com/actions/runner/pull/1789


thboop (approved):
lgtm

TingluoHuang (approved):

Thana43 (approved):

PR #1792, opened 2022-03-28
Mark run as Cancelled/Failed upon HostContext.RunnerShutdownToken state

soelinn:
github/c2c-actions-support#883

- Currently when Actions Runner shuts down for any `ShutdownReason`, Workflow Run ends as `Cancelled`.
- `ShutdownReason.OperatingSystemShutdown` only occurs for Windows platform, and shutdown causes on other platforms are hard to detect accurately.
- **Requirement Change**
    - **✅ SAME:** User Cancellation from the GitHub UI remains the same ➡️ `TaskResult.Cancelled`.
    - **⚠️ CHANGE:** Mark Workflow Run as `TaskResult.Failed` after receiving a runner shutdown signal (**e.g.,** `SIGTERM`)
        - This `TaskResult.Failed` state can only be occur if the Runner process can report its shutdown state back to the Actions Service **successfully** before exiting.
        - I coded this behavior change by depending on the `HostContext.RunnerShutdownToken.IsCancellationRequested` when the Runner process receives a `Ctrl-C` or `Ctrl-Break` signal. It will only be able to report back to the Actions Service upon `SIGTERM` as `SIGKILL` will be immediate, and won't allow graceful shutdown.
    - If Runner process exiting due to `Ctrl-C` SIGTERM by user manually ➡️ `TaskResult.Failed`

### DIFF
--- a/src/Runner.Worker/StepsRunner.cs
+++ b/src/Runner.Worker/StepsRunner.cs
@@ -134,8 +134,10 @@ namespace GitHub.Runner.Worker
                             // Test the condition again. The job was cancelled after the condition was originally evaluated.
                             jobCancelRegister = jobContext.CancellationToken.Register(() =>
                             {
-                                // Mark job as cancelled
-                                jobContext.Result = TaskResult.Canceled;
+                                // Mark job as Cancelled or Failed depending on HostContext shutdown token's cancellation
+                                jobContext.Result = HostContext.RunnerShutdownToken.IsCancellationRequested
+                                                    ? TaskResult.Failed
+                                                    : TaskResult.Canceled;
                                 jobContext.JobContext.Status = jobContext.Result?.ToActionResult();
 
                                 step.ExecutionContext.Debug($"Re-evaluate condition on job cancellation for step: '{step.DisplayName}'.");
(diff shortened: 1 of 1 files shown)


TingluoHuang:
FYI, you might only get `OperatingSystemShutdown` when you are on Windows and configure the runner as a Windows service.

soelinn:
I expected it'll be cases like this, and I might not be able to handle it reliably and distinguish between UserCancellation vs OS shutdown. I will probably need to consult @chrispat.

src/Runner.Common/HostContext.cs
```diff
@@ -647,5 +647,7 @@ public enum ShutdownReason
     {
         UserCancelled = 0,
```
TingluoHuang, line 648: Do we no longer use `UserCancelled`?

soelinn: It is used it in several key locations; so it is still needed.

- https://github.com/actions/runner/blob/main/src/Runner.Listener/Runner.cs#L264
- https://github.com/actions/runner/blob/main/src/Runner.Listener/Runner.cs#L291
- https://github.com/actions/runner/blob/main/src/Runner.Listener/JobDispatcher.cs#L576
- https://github.com/actions/runner/blob/main/src/Runner.Worker/JobRunner.cs#L78
- https://github.com/actions/runner/blob/main/src/Runner.Worker/Worker.cs#L120
- https://github.com/actions/runner/blob/main/src/Runner.Worker/Worker.cs#L217

TingluoHuang (approved):

PR #1794, opened 2022-03-29
fix: GITHUB_ENV in composite

### LINKED ISSUE #789: GITHUB_ENV does not update when running composite actions multiple times
**Describe the bug**
I've separated repeating shell logic to its own composite action. On each run, I'm setting a label into GITHUB_ENV to re-use it down the line. It works the first time, but subsequent calls to the composite action can't seem to update that value, as mentioned in the docs. After trying to set it, it just keeps the old value.

**To Reproduce**
1. Setup a composite action
2. Add step run: echo "label=${{ inputs.label }}-$GITHUB_SHA" >> $GITHUB_ENV
3. Call this action multiple times in your job with different labels

**Expected behavior**
The environment variable should be updated properly on every call.

Version of your runner? Where do I see that?

OS of the machine running the runner? ubuntu-latest

## What's not working?
There is no output whatsoever

### LINKED ISSUE #1899: Composite Actions fail to overwrite env if it is already set
**Describe the bug**
Setting an env that is already set in a composite action causes the value to not be set as expected.
**To Reproduce**
Setup a composite action
```
  using: "composite"
  steps:
    - name: set env
      shell: bash
      run: |
        VAR1=$RANDOM
        echo "MY_VAR1=$(echo $VAR1)" >> $GITHUB_ENV
        echo $VAR1
        echo 'create finished'
  
    # Show ref information
    - name: Show ref information
      run: |
          echo 'random1'
          echo $MY_VAR1
          echo 'env-random1'
          echo "${{ env.MY_VAR1 }}"
          echo 'read finished'
      shell: bash
```
Call it twice locally using `uses: ./{path}`
Notice the second time the value it reads is incorrect.
**Expected behavior**

2.291.1 linux

ChristopherHX:
This PR just shows how you could fix issue #789. This doesn't includes any tests and I'm not affiliated with GitHub.

Minimal repro of the bug, which this change aims to solve.
action.yml
```yaml
runs:
  using: composite
  steps:
  - run: echo "myenvvar=val2" > $GITHUB_ENV
    shell: bash
  # The following step fails on actions/runner prior these changes
  - run: exit 1
    if: env.myenvvar != 'val2'
    shell: bash
```

.github/workflows/main.yml
```yaml
on: push
jobs:
  test:
    runs-on: [self-hosted, linux]
    steps:
    - uses: actions/checkout@v3
    - run: echo "myenvvar=val1" > $GITHUB_ENV
      shell: bash
    - uses: ./
```

You see that myenvvar isn't specified as an env parameter of uses: ./, so the env variable should be updated within composite. Actually the actions/runner runs it like the workflow would be, which is expected to always fail
.github/workflows/main.yml
```yaml
on: push
jobs:
  test:
    runs-on: [self-hosted, linux]
    steps:
    - uses: actions/checkout@v3
    - run: echo "myenvvar=val1" > $GITHUB_ENV
      shell: bash
    - uses: ./
      env:
        myenvvar: val1
```

The problem is that the composite step env context is already merged with the old global variable set by GITHUB_ENV.

CC @nikola-jokic
Resolves #789
Resolves #1899


pragmaticivan:
🙏🏼 @thboop @yaananth @soelinn

ruvceskistefan:
Hi @ChristopherHX,
I don't think that this PR solves the problem. If you try to execute workflow described in [this issue](https://github.com/actions/runner/issues/1899), you'll see that the same thing is still happening.

ChristopherHX:
@ruvceskistefan How did you test it? I create a public run and it works.

Steps I did
- Download test artifacts https://github.com/ChristopherHX/runner.server/actions/runs/2341816395
- extract the under `_package`
- `./config.sh --url <runnerrepo> --token <runnertoken> --disableupdate`, did you omit `--disableupdate`? **Otherwise it switches to an unpatched runner**

Here a public repro, which did work for me
- https://github.com/ChristopherHX/newcomposite-sample/tree/github-env-composite-test
- https://github.com/ChristopherHX/newcomposite-sample/runs/6562334669?check_suite_focus=true

ruvceskistefan:
@ChristopherHX you're definitely right, I accidentally tested using runner without `--disableupdate`. I can confirm that this is working properly, so it is legitimate as far as I'm concerned. I think we'll merge this soon, thanks for your contribution!

bauergeorg:
@ruvceskistefan what is the current state?

ivan-santos-eq:
@ruvceskistefan Is there any chance someone from GH could merge this PR soon? the tech debt is growing by the day and we will need to remove a lot of the workarounds we've created.

DonDebonair:
Can this PR be reviewed and merged?

sagi-shimoni:
Please merge this PR soon ,
We are making a lot of workarounds due to this issue...

dodwmd:
+1

thboop (approved):
LGTM thanks for the contribution!

thboop:
Thanks for your patience while we evaluated the options to solve this problem. I've gone ahead and merged this PR, it will rollout in the next runner release, which should occur in a few weeks! 🎉

bauergeorg:
@thboop Do you know when the next version will be released?

tlsimmons88:
@thboop When is this fix going to be released.  As of Aug 1, 2022 I am still seeing the issue.

bauergeorg:
thanks for [release 2.295.0](https://github.com/actions/runner/releases/tag/v2.295.0)

PR #1803, opened 2022-04-04
Enhancecment: added defaults shell and working-directory to composite

### LINKED ISSUE #836: add possibility to use defaults inside of composite actions
**Describe the enhancement**
Add the `defaults:` option to the composite actions

**Code Snippet**
````
runs:
  using: "composite"
  defaults:
    run:
      shell: bash
  steps:
    - run: echo "bla bla 1"
    - run: echo "bla bla 2"
    - run: echo "bla bla 3"
````

### PR DESCRIPTION
**Pull request by nikola-jokic**
Closes #836 

This also removes the required shell parameter. That would not break existing workflows but will execute using default behaviour if `shell` and `defaults` are not specified.

### DIFF
--- a/src/Runner.Worker/ActionManifestManager.cs
+++ b/src/Runner.Worker/ActionManifestManager.cs
@@ -372,6 +372,7 @@ namespace GitHub.Runner.Worker
             var postEntrypointToken = default(StringToken);
             var postIfToken = default(StringToken);
             var steps = default(List<Pipelines.Step>);
+            var defaultsToken = default(MappingToken);
 
             foreach (var run in runsMapping)
             {
@@ -422,6 +423,9 @@ namespace GitHub.Runner.Worker
                         steps = PipelineTemplateConverter.ConvertToSteps(templateContext, stepsToken);
                         templateContext.Errors.Check();
                         break;
+                    case "defaults":
+                        defaultsToken = run.Value.AssertMapping("defaults");
+                        break;
                     default:
                         Trace.Info($"Ignore run property {runsKey}.");
                         break;
@@ -451,7 +455,7 @@ namespace GitHub.Runner.Worker
                         };
                     }
                 }
-                else if (string.Equals(usingToken.Value, "node12", StringComparison.OrdinalIgnoreCase)||
+                else if (string.Equals(usingToken.Value, "node12", StringComparison.OrdinalIgnoreCase) ||
                          string.Equals(usingToken.Value, "node16", StringComparison.OrdinalIgnoreCase))
                 {
                     if (string.IsNullOrEmpty(mainToken?.Value))
@@ -486,7 +490,8 @@ namespace GitHub.Runner.Worker
                             PostSteps = new Stack<Pipelines.ActionStep>(),
                             InitCondition = "always()",
                             CleanupCondition = "always()",
-                            Outputs = outputs
+                            Outputs = outputs,
+                            JobDefaults = defaultsToken,
                         };
                     }
                 }
--- a/src/Runner.Worker/Handlers/CompositeActionHandler.cs
+++ b/src/Runner.Worker/Handlers/CompositeActionHandler.cs
@@ -1,8 +1,6 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using GitHub.DistributedTask.Expressions2;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/Runner.Worker/action_yaml.json [hunk 1]
```diff
@@ -252,6 +256,34 @@
                 "loose-key-type": "non-empty-string",
                 "loose-value-type": "string"
             }
+        },
+        "job-defaults": {
```
thboop, line 260: can we have the idea of a `composite-defaults` keyword here instead of `job-defaults`

src/Runner.Worker/action_yaml.json
```diff
@@ -252,6 +256,34 @@
...
+          "mapping": {
+            "properties": {
+              "run": "job-defaults-run"
+            }
+          }
+        },
+        "job-defaults-run": {
```
thboop, line 267: same as above

src/Runner.Worker/action_yaml.json
```diff
@@ -252,6 +256,34 @@
...
+          "context": [
+            "github",
```
thboop, line 269: We should provide inputs here as well!

src/Runner.Worker/action_yaml.json
thboop: We currently have two defaults

`workflow-defaults` which apply to the entire run, and `job-defaults` which apply to a job. we should have a new idea of `composite-defaults` which apply only to the current composite action (not its children or its parents).

So if you set a default shell any of your immediate run steps can use it, but it won't use the job defaults, workflow defaults or parent composite action defaults.

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -86,13 +83,21 @@ public async Task RunAsync(ActionRunStage stage)
...
+
                 ExecutionContext.StepTelemetry.HasRunsStep = hasRunsStep;
                 ExecutionContext.StepTelemetry.HasUsesStep = hasUsesStep;
                 ExecutionContext.StepTelemetry.StepCount = steps.Count;
             }
             ExecutionContext.StepTelemetry.Type = "composite";
 
+            // save job defaults run to be restored after the composite run
+            IDictionary<string, string> jobDefaultsRun = new Dictionary<string, string>();
+            if (ExecutionContext.Global.JobDefaults.ContainsKey("run"))
```
thboop, line 95: Lets not overwrite job defaults, let's use a new key called `CompositeDefaults`, to minimize risk of these changes and make it clearer to read the code where these values are coming from.

nikola-jokic: I changed it but I am not sure if this is what you have meant.

Im2451soon (approved):

Im2451soon (approved):

src/Runner.Worker/action_yaml.json
```diff
@@ -133,7 +137,7 @@
                     "working-directory": "string-steps-context",
                     "shell": {
                         "type": "string-steps-context",
-                        "required": true
+                        "required": false
```
thboop, line 140: we should still fail here if composite defaults aren't set and a shell isn't set for that step. Do we currently do that?

roryabraham:
Would love to see this PR make it over the finish line 😄

src/Runner.Worker/action_yaml.json
nikola-jokic: Yes, in that case, the action would fail with this output:
image

Bentley617617 (dismissed):

src/Runner.Worker/ActionManifestManager.cs
```diff
@@ -476,14 +480,41 @@ public ActionDefinitionData Load(IExecutionContext executionContext, string mani
...
+                                if (step.Reference.Type != Pipelines.ActionSourceType.Script)
+                                {
+                                    continue;
+                                }
+                                var mapping = step.Inputs.AssertMapping("inputs");
+                                var foundShell = false;
+                                foreach (var pair in mapping)
+                                {
+                                    // Expression key
+                                    if (pair.Key is StringToken && pair.Key.AssertString("inputs key").Value.Equals("shell"))
```
AvaStancu, line 497: [minor] instead of using the `foundShell` var can we use negation on this line and throw directly the error without the extra assignments to `foundShell`?

nikola-jokic: How can I do that? I'm going through all mappings trying to validate that the `shell` is present. Only if no mapping exists for shell, I need to throw

AvaStancu: Was thinking of something like
```
foreach (var pair in mapping)
{
	if !(pair.Key is StringToken && pair.Key.AssertString("inputs key").Value.Equals("shell"))
	{
		throw new TemplateValidationException("Composite action must set the default shell, or the shell parameter in each step is required.");
	}
}
```
but I can live without this 😃

PR #1804, opened 2022-04-04
Added retries to worker invocation

### LINKED ISSUE #1795: Add retries to worker invocation
**Describe the enhancement**
When the `Runner.Listener` gets a job message, it creates a worker and sends that job message over stdin to that process, with a 30 second timeout. If that timeout is exceeded, we cancel the job.

For most cases, this makes a lot of sense, and very very very few users run into issues with that, as writing a few bytes to stdin can be done pretty quickly. However, for under-provisioned or over-utilized machines, this can exceed the timeout and fail their job.

We should be more resilient here, can we add a few retries? If we fail 3 times we can fail the job, otherwise

**Code Snippet**
[This code is the general area
](https://github.com/actions/runner/blob/eb9a604b63d9d656beee60c84e03d3cd87485496/src/Runner.Listener/JobDispatcher.cs#L459)

If we fail to write this to stdin, kill the current worker and create a new worker and retry.

ruvceskistefan:
Closes #1795

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -352,49 +352,50 @@ namespace GitHub.Runner.Listener
                 TaskCompletionSource<int> firstJobRequestRenewed = new TaskCompletionSource<int>();
                 var notification = HostContext.GetService<IJobNotification>();
 
-                int retryCount = 0;
-                while (retryCount < 3)
+                // lock renew cancellation token.
+                using (var lockRenewalTokenSource = new CancellationTokenSource())
                 {
-                    // lock renew cancellation token.
-                    using (var lockRenewalTokenSource = new CancellationTokenSource())
-                    using (var workerProcessCancelTokenSource = new CancellationTokenSource())
+                    long requestId = message.RequestId;
+                    Guid lockToken = Guid.Empty; // lockToken has never been used, keep this here of compat
+
+                    // start renew job request
+                    Trace.Info($"Start renew job request {requestId} for job {message.JobId}.");
+                    Task renewJobRequest = RenewJobRequestAsync(_poolId, requestId, lockToken, orchestrationId, firstJobRequestRenewed, lockRenewalTokenSource.Token);
+
+                    // wait till first renew succeed or job request is cancelled
+                    // not even start worker if the first renew fail
+                    await Task.WhenAny(firstJobRequestRenewed.Task, renewJobRequest, Task.Delay(-1, jobRequestCancellationToken));
+
+                    if (renewJobRequest.IsCompleted)
                     {
-                        long requestId = message.RequestId;
-                        Guid lockToken = Guid.Empty; // lockToken has never been used, keep this here of compat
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -352,246 +352,128 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
                                     await processChannel.SendAsync(
-                                        messageType: messageType,
-                                        body: string.Empty,
-                                        cancellationToken: csSendCancel.Token);
+                                        messageType: MessageType.NewJobRequest,
+                                        body: JsonUtility.ToString(message),
+                                        cancellationToken: csSendJobRequest.Token);
                                 }
+                                // stop retrying
+                                retryCount = 3;
```
thboop, line 470: Can we add the right before this line. I think we can scope down the changes a little, and just retry the process invocation.

```
using (var processChannel = HostContext.CreateService<IProcessChannel>())
```

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -601,48 +483,182 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                                    Trace.Info($"Stop renew job request for job {message.JobId}.");
+                                    // stop renew lock
+                                    lockRenewalTokenSource.Cancel();
+                                    // renew job request should never blows up.
+                                    await renewJobRequest;
+                                    
+                                    // not finish the job request since the job haven't run on worker at all, we will not going to set a result to server.
+                                    return;
+                                }
+                                else
```
thboop, line 498: We might need to clean out workerOutput if we have any retries remaining. Can we get an e2e demo setup for this and send over job logs? I'm curious what the experience looks like. Happy to pair on how we can accomplish that, we may need to use the debugger or add code to the worker to throw

src/Runner.Listener/JobDispatcher.cs
ruvceskistefan: Of course, only in that case we have to move this line of code
`using (var workerProcessCancelTokenSource = new CancellationTokenSource())` inside of `while` loop, like this:
```
int retryCount = 0;
while (retryCount < 3)
{
    using (var workerProcessCancelTokenSource = new CancellationTokenSource())
    using (var processChannel = HostContext.CreateService<IProcessChannel>())
    using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
    {
       // ...
```

Im2451soon: 도와줘

src/Runner.Listener/JobDispatcher.cs
Im2451soon: 도와줘

ruvceskistefan: I agree, I'll add cleanup of that list. I'll think about best way of doing e2e demo and I’ll do that too.

PR #1818, opened 2022-04-11
Using the hashFiles function from the toolkit

### LINKED ISSUE #1805: `actions/glob` implements a hashfiles implementation, we should use that
**Describe the bug**
The `@actions/glob` package implements [hashfiles](https://github.com/actions/toolkit/blob/main/packages/glob/src/glob.ts), we should use that implementation for our hashfiles implementation by consolidating our implementation with that one.

### Checklist
- [ ] Compare the two versions and see what functionality differs
- [ ] Port any changes we need to over to the toolkit
- [ ] Release a new toolkit version if needed
- [ ] Consume the hashfiles toolkit implementation.

ruvceskistefan:
Closes #1805

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,0 @@
-{
-  "name": "runner",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {}
-}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "runner",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
--- a/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts
+++ b/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts
@@ -14,7 +14,11 @@ async function run(): Promise<void> {
 
   console.log(`Match Pattern: ${matchPatterns}`)
 
-  const result = await glob.hashFiles(matchPatterns, {followSymbolicLinks})
+  const result = await glob.hashFiles(
+    matchPatterns,
+    {followSymbolicLinks},
+    true
+  )
   console.error(`__OUTPUT__${result}__OUTPUT__`)
 }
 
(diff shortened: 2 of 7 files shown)


package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
thboop, line 1: Do we need this, or can we delete it? The root of the runner shouldn't be a node project

src/Misc/expressionFunc/hashFiles/src/hashFiles.ts
```diff
@@ -18,38 +13,9 @@ async function run(): Promise<void> {
...
-  }
-  result.end()
 
-  if (hasMatch) {
-    console.log(`Found ${count} files to hash.`)
-    console.error(`__OUTPUT__${result.digest('hex')}__OUTPUT__`)
-  } else {
-    console.error(`__OUTPUT____OUTPUT__`)
-  }
+  const result = await glob.hashFiles(matchPatterns, {followSymbolicLinks})
```
thboop, line 17: We lose some logging by doing this, can we submit a pr to the toolkit to add a verbose mode to `glob.hashfiles`. Happy to walk through that as needed!

package-lock.json
ruvceskistefan: My mistake, it's unnecessary, I deleted it

src/Misc/expressionFunc/hashFiles/src/hashFiles.ts
ruvceskistefan: We can use `core.info()` instead of `core.debug()` in toolkit implementation, if that won't cause a problem in other places where the `hashFiles()` is used.

DawnGKoz904 (dismissed):

src/Misc/layoutbin/hashFiles/index.js
```diff
@@ -1057,7 +1057,9 @@ var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, ge
     });
 };
 Object.defineProperty(exports, "__esModule", { value: true });
+exports.hashFiles = exports.create = void 0;
 const internal_globber_1 = __webpack_require__(297);
```
thboop, line 1061: I think modifying this file is causing the test to fail for some reason, can we put in a fix for that test as a part of this pr?

PR #1823, opened 2022-04-13
Capture telemetry when git errors on unsafe repository.

TingluoHuang:
A recent [Git release](https://github.blog/2022-04-12-git-security-vulnerability-announced/) has an enforcement check on the repository folder's ownership.

This may break workflow run when customers are using containers within their job that have some interaction with the local git repository (push commit, etc) since the mapped in `github_workspace` will have different owner, (root inside container and regular user on the runner machine).

We are collecting telemetry in order to provide feedback to the `git` maintainer.


thboop (approved):
lgtm

PR #1833, opened 2022-04-19
Fix package corruption for 2.289.x runner

TingluoHuang:
We made some big changes to `run.sh`, However, the changed `run.sh` didn't package into the runner release properly.
The script is polluted as part of the L0 test step.

Since we never ship the change to GitHub.com, we are going to revert all related changes to take us back to a working state.

In this way, we can safely ship this runner package to GHES without worrying about further damage.


thboop (approved):
lgtm

PR #1839, opened 2022-04-20
Update Actions Summary limit to 1MiB

yacaovsnc:
We have seen users already hit 128KiB limit in limited beta.  In order to have a better user experience, we are relaxing the limit to 1MiB per step summary.

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -266,7 +266,7 @@ namespace GitHub.Runner.Worker
 
     public sealed class CreateStepSummaryCommand : RunnerService, IFileCommandExtension
     {
-        private const int _attachmentSizeLimit = 1024 * 1024;
+        public const int AttachmentSizeLimit = 1024 * 1024;
 
         public string ContextName => "step_summary";
         public string FilePrefix => "step_summary_";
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -266,7 +266,7 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
 
     public sealed class CreateStepSummaryCommand : RunnerService, IFileCommandExtension
     {
-        private const int _attachmentSizeLimit = 128 * 1024;
+        public const int AttachmentSizeLimit = 1024 * 1024;
```
TingluoHuang, line 269: do you want to change this to pass down the limit from service? so you don't need to update the runner every time you need to bump the limit, especially if you need to bump the limit for certain customers.

yacaovsnc: I don't want to manage limits per users if it can be helped at all... 😅 

Yeah passing it down from server may be a good choice if we find we need to do this again in the future.  Having the limit defined in a public repo is still a good communication tool IMHO though - we could point others to this number and they will understand the limit right away.  😄   However if we find we need to update this often, then yeah... 

Maybe next time we should visit this decision?

TingluoHuang: I am fine with it.

TingluoHuang (approved):

yacaovsnc:
👋🏼  @TingluoHuang I don't have write permission in this repo, do you mind merge it for me?  🙇🏼‍♂️

Issue #1840, opened 2022-04-21
hashFiles() couldn't finish within 120 seconds.


Carsten-MaD:
(filing here since I understand that `hashFile()` is owned by https://github.com/actions/runner)

Hey!

We have migrated a semi-large Unity project from SVN to Git / Github and I am working on setting up Actions for a build pipeline.
Unfortunately, a step to create a cache for Unity-specific files is giving me a headache. If I understand the error message correctly, I am running into a 120 second timeout for that step.

As I see it, the timeout for `hasFiles()` is set here and there is no way to increase it, is that correct?
https://github.com/actions/runner/blob/100c99f2635be35c7ea690c58fa7d203725cac9e/src/Runner.Worker/Expressions/HashFilesFunction.cs#L16

Runner version: 2.290.1
OS: Windows 10 64bit self hosted runner

Action configuration:
`      - uses: actions/cache@v3
        with:
          path: Library
          key: Library-${{ hashFiles('Assets\**', 'Packages\**', 'ProjectSettings\**') }}
          restore-keys: |
            Library-`

Error message with debug logging enabled:
`[looong list of files that are processed, omitted]
##[debug]Sending CTRL_C to process 10936.
##[debug]Successfully send CTRL_C to process 10936.
##[debug]Waiting for process exit or 7.5 seconds after CTRL_C signal fired.
##[debug]Ignore Ctrl+C to current process.
##[debug]STDOUT/STDERR stream read finished.
##[debug]STDOUT/STDERR stream read finished.
##[debug]Process exit successfully.
##[debug]Process cancelled successfully through Ctrl+C/SIGINT.
##[debug]Process Cancellation finished.
##[debug]Finished process 10936 with exit code -1073741510, and elapsed time 00:02:00.0198012.
Error: .github/workflows/main.yml (Line: 43, Col: 16):
Error: The template is not valid. .github/workflows/main.yml (Line: 43, Col: 16): hashFiles('Assets\**, Packages\**, ProjectSettings\**') couldn't finish within 120 seconds.
##[debug]GitHub.DistributedTask.ObjectTemplating.TemplateValidationException: The template is not valid. .github/workflows/main.yml (Line: 43, Col: 16): hashFiles('Assets\**, Packages\**, ProjectSettings\**') couldn't finish within 120 seconds.
##[debug]   at GitHub.DistributedTask.ObjectTemplating.TemplateValidationErrors.Check()
[...]

PR #1844, opened 2022-04-22
Make timeout for `hashFiles()` configureable

Carsten-MaD:
As reported here https://github.com/actions/runner/issues/1840, in scenarios where a lot of files need to be hashed to build up a cache, the process running `hashFiles()` may run into the fixed 120 second timeout and be cancelled prematurely.

This pull request introduces a new parameter `--timeout=` that is mean to be passed the same way as the already existing optional parameter `--follow-symbolic-links`.

Example configuration for actions/cache:
```
- uses: actions/cache@v3
  with:
    key: ${{ runner.os }}-${{ hashFiles('--timeout=600', '**') }}
```

## Points of discussion:
- It may be controversial to pass a parameter that affects the timeout of a process like this, as it is not *directly* related to the hashFiles() command itself, but rather to how it is run. If you see an alternative way of injecting such a parameter, please let me know.
- For this, I removed `firstParameter` and the connected logic. Instead, the logic now looks for if a parameter name starts with 
`--` and tries to match one of the two expected parameters. Please let me know if you strongly feel the logic that these parameters need to be first is needed, I will then put something in.
- There are two different exceptions thrown for when the timeout number could not be parsed (e.g. a string passed), and when it was empty. This could be made into one more generic exception to handle and explain both.

### DIFF
--- a/src/Runner.Worker/Expressions/HashFilesFunction.cs
+++ b/src/Runner.Worker/Expressions/HashFilesFunction.cs
@@ -33,24 +33,40 @@ namespace GitHub.Runner.Worker.Expressions
             string githubWorkspace = workspaceData.Value;
             bool followSymlink = false;
             List<string> patterns = new List<string>();
-            var firstParameter = true;
             foreach (var parameter in Parameters)
             {
                 var parameterString = parameter.Evaluate(context).ConvertToString();
-                if (firstParameter)
+                if (parameterString.StartsWith("--"))
                 {
-                    firstParameter = false;
-                    if (parameterString.StartsWith("--"))
+                    if (string.Equals(parameterString, "--follow-symbolic-links", StringComparison.OrdinalIgnoreCase))
                     {
-                        if (string.Equals(parameterString, "--follow-symbolic-links", StringComparison.OrdinalIgnoreCase))
+                        followSymlink = true;
+                        continue;
+                    }
+                    else if (parameterString.StartsWith("--timeout=", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var timeoutSecondsString = parameterString.Split("=")[1];
+                        if (!String.IsNullOrEmpty(timeoutSecondsString))
                         {
-                            followSymlink = true;
-                            continue;
+                            bool parseSuccess = int.TryParse(timeoutSecondsString, out int timoutSeconds);
+                            if (parseSuccess)
+                            {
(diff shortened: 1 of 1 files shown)


Carsten-MaD:
Not entirely sure, is there anything I need to do so the PR is reviewed, am I missing a step?

@TingluoHuang sorry if I tag you directly, maybe you can advise?

src/Runner.Worker/Expressions/HashFilesFunction.cs
```diff
@@ -33,25 +33,41 @@ public sealed class HashFilesFunction : Function
...
+                        continue;
+                    }
+                    else if (parameterString.StartsWith("--timeout=", StringComparison.OrdinalIgnoreCase))
+                    {
+                        var timeoutSecondsString = parameterString.Split("=")[1];
+                        if (!String.IsNullOrEmpty(timeoutSecondsString))
                         {
-                            followSymlink = true;
-                            continue;
+                            bool parseSuccess = int.TryParse(timeoutSecondsString, out int timoutSeconds);
```
JJ, line 51: Please check the typo, `timout` → `timeout`
If we're going to get all nitpicky about it, `parseSuccess` → `parsingSuccess`

JJ (approved):
Just a typo.

src/Runner.Worker/Expressions/HashFilesFunction.cs
Carsten-MaD: Fixed, thanks!

shubhank008:
This Pr is really needed for us large gamedev project users, any update on this ?

gotchipete:
> This Pr is really needed for us large gamedev project users, any update on this ?

Agree!  (realizing I'm replying to a comment from 10 mo ago)

haoming29:
Any updates for this issue or this should be closed? We caught the same error and suspect our runner fails because we have large files to hash and it could go over 120s in many cases.

BE-Digital-Solutions-Bart (approved):
Code seems good
This merge should solve issue #1840

alb3rtuk:
Hey guys, we're running into the same issue with the 120 second `hashFiles()` timeout. Can we merge this please?

alb3rtuk (approved):

Propagant:
So what's the status on this?

AryanJ-NYC:
Bump?

guicaulada:
Running into this as well.

PR #1853, opened 2022-04-28
Runner container hooks Beta

fhammerl:
[ADR](https://github.com/github/c2c-actions-runtime/pull/1748)


src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,215 @@
...
+                    EnvironmentVariables = environmentVariables,
+                    PrependPath = prependPath,
+                    WorkingDirectory = workingDirectory,
+                },
+                State = GetHookStateInJson(context)
+            };
+
+            var response = await ExecuteHookScript(context, input, ActionRunStage.Main);
+            SaveHookState(context, response.State);
+
```
ruvceskistefan, line 153: NIT: remove empty line

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,215 @@
...
+            // TODO: consider the case when the responseFile doesn't exist - should we throw or noop on response?
+            var response = IOUtil.LoadObject<HookResponse>(input.ResponseFile);
+            IOUtil.DeleteFile(input.ResponseFile);
+            return response;
+        }
+
+        private string GenerateResponsePath()
+        {
+            return $"{Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Temp), ResponseFolderName)}/{Guid.NewGuid()}.json";
```
ruvceskistefan, line 195: Can we use Path.Combine for whole path? Something like this:
Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Temp), ResponseFolderName, $"{Guid.NewGuid()}.json")

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    public class HookContainer : HookArgs
+    {
+        public string Id { get; set; }
+        public string DisplayName { get; set; }
+        public string Network { get; set; }
+        public string NetworkAlias { get; set; }
+        public string Image { get; set; }
+        public string Name { get; set; }
+        public string EntryPointArgs { get; set; }
```
thboop, line 15: This should be a List<string>, rather then a string

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,214 @@
...
+            handler.PrepareExecution(stage); // TODO: find out stage, we only use Start in pre, but double check
+
+            IOUtil.CreateEmptyFile(input.ResponseFile);
+            await handler.RunAsync(stage);
+            if (handler.ExecutionContext.Result == TaskResult.Failed)
+            {
+                throw new Exception("Hook failed"); // TODO: better exception
+            }
+
+            // TODO: consider the case when the responseFile doesn't exist - should we throw or noop on response?
```
fhammerl, line 186: I'd say let's not force the users to manage a response file if they don't want to, container scenarios can and do work without this postback.

src/Runner.Worker/ActionManager.cs
```diff
@@ -101,38 +101,41 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
             IEnumerable<Pipelines.ActionStep> actions = steps.OfType<Pipelines.ActionStep>();
             executionContext.Output("Prepare all required actions");
             var result = await PrepareActionsRecursiveAsync(executionContext, state, actions, depth, rootStepId);
-            if (state.ImagesToPull.Count > 0)
+            if (!FeatureFlagManager.IsContainerHooksEnabled(executionContext))
```
fhammerl, line 104: Note: no need to prebuild / pull, the hooks will take care of it

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -50,72 +55,15 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
-            if (dockerVersion.ClientVersion < requiredDockerEngineAPIVersion)
-            {
-                throw new NotSupportedException($"Min required docker engine API client version is '{requiredDockerEngineAPIVersion}', your docker ('{_dockerManager.DockerPath}') client version is '{dockerVersion.ClientVersion}'");
+                // Initialize the containers
+                containers.ForEach(container => UpdateRegistryAuthForGitHubToken(executionContext, container));
+                containers.ForEach(container => MountWellKnownDirectories(executionContext, container));
+                await _containerHookManager.PrepareJobAsync(executionContext, containers);
+                return;
             }
+            await AssertCompatibleOS(executionContext);
```
fhammerl, line 66: ☝️ The above 'removed' code moved to this function

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -58,18 +63,26 @@ public Task<string> DetermineNodeRuntimeVersion(IExecutionContext executionConte
...
                                             Encoding outputEncoding,
                                             bool killProcessOnCancel,
                                             bool inheritConsoleHandler,
+                                            string standardInInput,
                                             CancellationToken cancellationToken)
         {
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
             {
+                Channel<string> redirectStandardIn = null;
+                if (standardInInput != null)
```
fhammerl, line 81: `!= null`, because `IsNullOrEmpty` would not write `""`, which is _technically_ a valid input (someone could be waiting for it to finish on the other side)

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,244 @@
...
+    {
+        Task PrepareJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task CleanupJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task ContainerStepAsync(IExecutionContext context, ContainerInfo container);
+        Task ScriptStepAsync(IExecutionContext context, ContainerInfo container, string arguments, string fileName, IDictionary<string, string> environment, string prependPath, string workingDirectory);
+    }
+
+    public class ContainerHookManager : RunnerService, IContainerHookManager
+    {
+        private const string ResponseFolderName = "_hook_responses";
```
thboop, line 28: super nit, we do `work/_temp/_runner_file_commands/step_summary_d058fe18-f53a-4d7d-b1ce-0dd9cf80f799` for file commands, should we do `_runner_hook_responses`

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,244 @@
...
+        private string HookIndexPath;
+
+        public override void Initialize(IHostContext hostContext)
+        {
+            base.Initialize(hostContext);
+            HookIndexPath = $"{Environment.GetEnvironmentVariable(Constants.Hooks.ContainerHooksPath)}";
```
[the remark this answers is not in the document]
fhammerl: I'll put it in FeatureFlagManager

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,260 @@
...
+            };
+
+            var prependPath = GetPrependPath(context);
+            var response = await ExecuteHookScript(context, input, ActionRunStage.Pre, prependPath);
+            if (response == null)
+            {
+                return;
+            }
+
+            jobContainer.IsAlpine = (bool) response.IsAlpine;
```
thboop, line 62: Do we need to try and evaluate this some how? Why the cast?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        private void SaveHookState(IExecutionContext context, JToken hookState, HookInput input)
+        {
+            // TODO: consider JTokenContextData
+            if (hookState == null)
+            {
+                Trace.Info($"No 'state' property found in response file for '{input.Command}'. Context variable 'hook_state' will not be updated.");
+                return;
+            }
+            context.JobContext["hook_state"] = new StringContextData(JsonUtility.ToString(hookState));
```
thboop, line 256: is this going to show on the `job` context? do we want that?

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    public class HookContainer : HookArgs
+    {
+        public string Id { get; set; }
+        public string DisplayName { get; set; }
+        public string Network { get; set; }
```
thboop, line 11: Are network or network alias needed? Are they sent over?

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+        public string NetworkAlias { get; set; }
+        public string Image { get; set; }
+        public string Name { get; set; }
```
thboop, line 14: Do we need Name?

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -330,6 +256,42 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
+            container.AddPathTranslateMapping(tempHomeDirectory, "/github/home");
+            container.ContainerEnvironmentVariables["HOME"] = container.TranslateToContainerPath(tempHomeDirectory);
+
+            var tempWorkflowDirectory = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Temp), "_github_workflow");
+            Directory.CreateDirectory(tempWorkflowDirectory);
+            container.MountVolumes.Add(new MountVolume(tempWorkflowDirectory, "/github/workflow"));
+            container.AddPathTranslateMapping(tempWorkflowDirectory, "/github/workflow");
+
+            container.ContainerWorkDirectory = container.TranslateToContainerPath(workingDirectory);
+            if (!FeatureFlagManager.IsContainerHooksEnabled(executionContext))
```
thboop, line 288: Can we add a comment here saying container hooks are responsible for these

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -523,5 +485,74 @@ private void UpdateRegistryAuthForGitHubToken(IExecutionContext executionContext
                 container.RegistryAuthPassword = executionContext.GetGitHubContext("token");
             }
         }
+
+        private async Task AssertCompatibleOS(IExecutionContext executionContext)
+        {
+                   // Check whether we are inside a container.
```
thboop, line 491: nit spacing

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -523,5 +485,74 @@ private void UpdateRegistryAuthForGitHubToken(IExecutionContext executionContext
...
+            // Our container feature requires to map working directory from host to the container.
+            // If we are already inside a container, we will not able to find out the real working direcotry path on the host.
+#if OS_WINDOWS
+#pragma warning disable CA1416
+            // service CExecSvc is Container Execution Agent.
```
thboop, line 496: all of this is unchanged just moved right?

src/Runner.Worker/Handlers/StepHost.cs
thboop: I don't understand this comment, could you elaborate more? We should always provide some input right?

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -94,11 +107,14 @@ public sealed class ContainerStepHost : RunnerService, IContainerStepHost
...
 
-        public string ResolvePathForStepHost(string path)
+        public string ResolvePathForStepHost(IExecutionContext executionContext, string path)
         {
             // make sure container exist.
             ArgUtil.NotNull(Container, nameof(Container));
-            ArgUtil.NotNullOrEmpty(Container.ContainerId, nameof(Container.ContainerId));
+            if (!FeatureFlagManager.IsContainerHooksEnabled(executionContext))
+            {
+                ArgUtil.NotNullOrEmpty(Container.ContainerId, nameof(Container.ContainerId));
```
thboop, line 116: Should we just consider moving this check? Maybe a todo in the future

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -234,5 +281,28 @@ public async Task<string> DetermineNodeRuntimeVersion(IExecutionContext executio
...
+
+        private string CheckPlatformForAlpineContainer(IExecutionContext executionContext, string preferredVersion)
+        {
+            string nodeExternal = preferredVersion;
+            if (!Constants.Runner.PlatformArchitecture.Equals(Constants.Architecture.X64))
+            {
+                var os = Constants.Runner.Platform.ToString();
+                var arch = Constants.Runner.PlatformArchitecture.ToString();
+                var msg = $"JavaScript Actions in Alpine containers are only supported on x64 Linux runners. Detected {os} {arch}";
+                throw new NotSupportedException(msg);
```
thboop, line 293: should we remove this check if running in hooks is an interesting question. Can we add a todo? May make sense to be more permissive (we don't care if the runner is alpine if we are running in k8s)

src/Test/L0/Worker/ExecutionContextL0.cs
```diff
@@ -760,6 +759,7 @@ public void GetExpressionValues_ContainerStepHost()
                 stepHost.Container = containerInfo;
 
                 var ec = new Runner.Worker.ExecutionContext();
+
```
thboop, line 762: can we remove this line? I'm okay with removing the above line too but I don't think this space should exist

thboop:
This pr looks good, lets see what testing we can add (particularly on container translation and any other places we can find to store it)

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+        public IEnumerable<string> EntryPointArgs { get; set; }
+        public string EntryPoint { get; set; }
+        public string WorkingDirectory { get; set; }
+        public string CreateOptions { get; private set; }
+        public string RuntimePath { get; set; }
```
thboop, line 19: What is `RuntimePath`?

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,62 @@
...
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public ContainerRegistry Registry { get; set; }
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> EnvironmentVariables { get; set; }
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> PortMappings { get; set; }
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public IEnumerable<MountVolume> MountVolumes { get; set; }
+        public HookContainer() { } // For Json deserializer
```
thboop, line 28: 🏅 on the comment

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
```diff
@@ -0,0 +1,18 @@
+using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    public class HookResponse
+    {
+        public ResponseContext Context { get; set; }
```
thboop, line 8: Something to consider, we should think about parsing each hook response as a particular type, like PrepareJobResponse, and throwing if fields that are required are missing. May be easier to build on in the future. We do a lot of assuming things are stuff like strings currently, and i'm curious if that is going to cause us issues.

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
```diff
@@ -0,0 +1,18 @@
...
+        public JToken State { get; set; }
+        public bool? IsAlpine { get; set; }
+    }
+
+    public class ResponseContext
+    {
+        public HookContainer Container { get; set; }
+        public IList<HookContainer> Services { get; set; }
```
thboop, line 16: Are these going to be type`Hook Container`?

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -58,18 +63,26 @@ public Task<string> DetermineNodeRuntimeVersion(IExecutionContext executionConte
             return Task.FromResult<string>(preferredVersion);
         }
 
-        public async Task<int> ExecuteAsync(string workingDirectory,
+        public async Task<int> ExecuteAsync(IExecutionContext context,
```
thboop, line 66: why did we add context here?

src/Runner.Worker/Handlers/StepHost.cs
thboop: Or just pass in a bool rather then passing in the full context

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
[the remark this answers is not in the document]
fhammerl: I've made the changes in `FeatureFlagManager`.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
fhammerl: `response.IsAlpine` is nullable `bool?` so that after deserialising it we can throw if it's not present. 
`jobContainer.IsAlpine` is non-nullable `bool` because it's mandatory.

I'll replace it with `response.IsAlpine.Value` as at this point IsAlpine is guaranteed to be there.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
fhammerl: Should we make a conscious effort to hide it?

It does allow the hooks to make arbitrary information, **potentially unregistered secrets** available for printing in the workflow. Someone unfamiliar with the implementation would not expect this.

Any other concerns? Otherwise, I see how hiding `hook_state` from the job context would be a good 'least privilege / access'  type of improvement.

src/Runner.Worker/ContainerOperationProvider.cs
fhammerl: Correct 👍

src/Runner.Worker/Handlers/StepHost.cs
fhammerl: What do you think of passing `executionContext.Globals.Variables`? It gives us the access to server sent FFs.

src/Runner.Worker/Handlers/StepHost.cs
fhammerl: As a TODO: container hooks, for all we know, could be implemented such that the containers are running on a completely different machine. Should we even have _any_ restrictions?

src/Test/L0/Worker/ExecutionContextL0.cs
fhammerl: Done!

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
fhammerl: I've got the same feeling, I'll look into splitting responses between hooks

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
fhammerl: I'd expect the hook authors to use the same container interface for their output as their input - with potentially a number of unset properties. Would you suggest we use a more specific class?

src/Runner.Worker/Handlers/StepHost.cs
fhammerl: To give the FFManager access to FF vars given by the server. I'll change it to type `Variables` instead, from `context.Global.Variables` like discussed above.

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
thboop: This just maps to the [job context](https://docs.github.com/en/actions/learn-github-actions/contexts#example-contents-of-the-job-context)

We don't allow you to set anything outside of fields in that context, so I don't think there is a point in accepting them, or am I missing something?

src/Runner.Worker/Handlers/StepHost.cs
thboop: We can always loosen restrictions as we go forward, we can leave this for now.

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
thboop: Is that helpful to pass to the hooks? I don't think its in the ADR. Should we just remove the field? Ideally I think we should only add fields we are planning to use.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
thboop: I think we should be very deliberate about what we have in the context, as that is an interface between our system and workflow authors. Unless we have a reason to put it in there, I don't think we should. We can always add it in the future based on feedback.

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
```diff
@@ -0,0 +1,64 @@
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    public class HookContainer : HookArgs
+    {
+        public string Id { get; set; }
+        public string DisplayName { get; set; }
```
thboop, line 10: A lot of these fields aren't in the [ADR](https://github.com/actions/runner/blob/thboop/ContainerHookADR/docs/adrs/1891-container-hooks.md), so we don't plan on using them in our hook examples. 

I think we should remove them from this. We can always add them back later if we need them, but its very hard to remove them later.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -11,6 +11,7 @@
 using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
 using GitHub.Runner.Worker.Handlers;
+using GitHub.Runner.Worker.Container;
```
TingluoHuang, line 14: do we need this `using`?

src/Runner.Worker/Container/FeatureFlagManager.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Worker.Container
```
TingluoHuang, line 6: we might want to move this to a higher namespace since the `FeatureFlagManager` sounds like a general concept that is not scoped to the container.

src/Test/L0/Worker/ExecutionContextL0.cs
```diff
@@ -100,7 +100,6 @@ public void ApplyContinueOnError_CheckResultAndOutcome()
         {
             using (TestHostContext hc = CreateTestContext())
             {
-
```
TingluoHuang, line 103: undo this change? 😄

src/Sdk/DTWebApi/WebApi/ActionsStepTelemetry.cs
```diff
@@ -56,5 +56,7 @@ public ActionsStepTelemetry()
 
         [DataMember(EmitDefaultValue = false)]
         public int? ExecutionTimeInSeconds { get; set; }
+        [DataMember(EmitDefaultValue = false)]
+        public bool IsContainerHookStep { get; set; }
```
TingluoHuang, line 60: is this used anywhere?

src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
```diff
@@ -82,7 +84,7 @@ internal static (string shellCommand, string shellArgs) ParseShellOptionString(s
             }
         }
 
-        internal static string GetDefaultShellForScript(string path, Common.Tracing trace, string prependPath)
+        internal static string GetDefaultShellForScript(string path, Common.Tracing trace, string prependPath, IHostContext hostContext = null)
```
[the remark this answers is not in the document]
TingluoHuang: if we want to support using `js` in the jobhook.

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -58,18 +63,26 @@ public Task<string> DetermineNodeRuntimeVersion(IExecutionContext executionConte
...
+                                            string standardInInput,
                                             CancellationToken cancellationToken)
         {
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
             {
+                Channel<string> redirectStandardIn = null;
+                if (standardInInput != null)
+                {
+                    redirectStandardIn = Channel.CreateUnbounded<string>(new UnboundedChannelOptions() { SingleReader = true, SingleWriter = true });
+                    await redirectStandardIn.Writer.WriteAsync(standardInInput, cancellationToken);
```
TingluoHuang, line 84:
```
                var redirectStandardIn = Channel.CreateUnbounded<string>(new UnboundedChannelOptions() { SingleReader = true, SingleWriter = true });
                redirectStandardIn.Writer.TryWrite(standardInInput);
```

`await WriteAsync()` might be a blocking call when the channel is busted...

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
fhammerl: Will work on syncing these 👍

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
fhammerl: Will use more specific classes 👍

src/Sdk/DTWebApi/WebApi/ActionsStepTelemetry.cs
fhammerl: Will be, it's in a PR against the fork@main

src/Runner.Worker/ActionRunner.cs
fhammerl: I think we need it for OSX

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -16,18 +19,20 @@ public interface IStepHost : IRunnerService
         event EventHandler<ProcessDataReceivedEventArgs> OutputDataReceived;
         event EventHandler<ProcessDataReceivedEventArgs> ErrorDataReceived;
 
-        string ResolvePathForStepHost(string path);
+        string ResolvePathForStepHost(IExecutionContext executionContext, string path);
```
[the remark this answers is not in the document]
ruvceskistefan: Should we add `IExecutionContext` prop to `IStepHost` interface or `IDefaultStepHost` because in `ContainerStepHost` class we already have `ContainerInfo` prop which will be redundant when we add `IExecutionContext` prop to it?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
fhammerl: What do you think of putting it in `ExecutionContext.Global`? We use `GlobalContext` for a lot of non-customer facing info, like steps and job telemetries.

src/Runner.Worker/Container/ContainerHooks/HookContainer.cs
fhammerl: I've put these fields (network, id, ports) into a class called `ResponseContainer`, to be read from `response.json` and used to populate the fields mentioned above in JobContext

src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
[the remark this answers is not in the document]
fhammerl: `HostContext` and `HostContextExtension` are in `GitHub.Runner.Common`, `ScriptHandlerHelpers` is in `GitHub.Runner.Worker.Handlers`, it doesn't feel right to have a dependency in that direction.

How about we make it an extension method of `HostContext`, but keep it in the same location / class (`ScriptHandlerHelpers`)?

`internal static string GetDefaultShellForScript(this IHostContext hostContext, string path, Common.Tracing trace, string prependPath)`

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            Trace.Entering();
+            var responsePath = GenerateResponsePath();
+            var input = new HookInput
+            {
+                Command = HookCommand.RunScriptStep,
+                ResponseFile = responsePath,
+                Args = new ScriptStepArgs
+                {
+                    Container = container.GetHookContainer(),
```
thboop, line 108: I don't think we need this container

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,259 @@
...
+            return string.Join(Path.PathSeparator.ToString(), context.Global.PrependPath.Reverse<string>()); ;
+        }
+
+        private T GetResponse<T>(HookInput input) where T: HookResponse
+        {
+            T response = null;
+
+            if (!string.IsNullOrEmpty(input.ResponseFile) && File.Exists(input.ResponseFile))
+            {
+                response = IOUtil.LoadObject<T>(input.ResponseFile);
```
thboop, line 176: we should validate this throws if they write xml or something

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,259 @@
...
+                IOUtil.DeleteFile(input.ResponseFile);
+                Trace.Info($"Response file for the hook script at '{HookIndexPath}' running command '{input.Command}' successfully processed and deleted.");
+                response?.Validate();
+            }
+            else
+            {
+                Trace.Info($"Response file for the hook script at '{HookIndexPath}' running command '{input.Command}' not found.");
+                if (input.Command == HookCommand.PrepareJob)
+                {
+                    throw new Exception($"Response file is required but not found for the hook script at '{HookIndexPath}' running command '{input.Command}'");
```
thboop, line 186: lets somehow catch all of these exceptions and bubble them up to the user is a super friendly way.

Something like
```
Custom container implementation failed with error: {error}. Please contact your self hosted runner administrator.
```

src/Runner.Worker/Container/ContainerHooks/HookInput.cs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public class ScriptStepArgs : HookArgs
+    {
+        public IEnumerable<string> EntryPointArgs { get; set; }
+        public string EntryPoint { get; set; }
+        public IDictionary<string, string> EnvironmentVariables { get; set; }
+        public string PrependPath { get; set; }
+        public string WorkingDirectory { get; set; }
+        public HookContainer Container { get; internal set; }
```
thboop, line 46: nit remove this

src/Runner.Worker/Container/ContainerHooks/HookInput.cs
```diff
@@ -0,0 +1,96 @@
...
+        [EnumMember(Value = "run_container_step")]
+        RunContainerStep,
+    }
+    public class HookArgs { }
+
+    public class PrepareJobArgs : HookArgs
+    {
+        public HookContainer Container { get; set; }
+        public IList<HookContainer> Services { get; set; }
+        public string Network { get; set; }
```
thboop, line 36: What is this used for?

src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
```diff
@@ -82,7 +84,7 @@ internal static (string shellCommand, string shellArgs) ParseShellOptionString(s
             }
         }
 
-        internal static string GetDefaultShellForScript(string path, Common.Tracing trace, string prependPath)
+        internal static string GetDefaultShellForScript(this IHostContext hostContext, string path, Common.Tracing trace, string prependPath)
```
thboop, line 87: 🏅

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -58,18 +63,26 @@ public Task<string> DetermineNodeRuntimeVersion(IExecutionContext executionConte
...
+                                            string standardInInput,
                                             CancellationToken cancellationToken)
         {
             using (var processInvoker = HostContext.CreateService<IProcessInvoker>())
             {
+                Channel<string> redirectStandardIn = null;
+                if (standardInInput != null)
+                {
+                    redirectStandardIn = Channel.CreateUnbounded<string>(new UnboundedChannelOptions() { SingleReader = true, SingleWriter = true });
+                    redirectStandardIn.Writer.TryWrite(standardInInput);
```
thboop, line 84: do we need a timeout here. What if the hook never reads input. Max 60 seconds or something?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,259 @@
...
+            }
+            context.JobContext["hook_state"] = new StringContextData(JsonUtility.ToString(hookState));
+            Trace.Info($"Context variable 'hook_state' updated successfully for '{input.Command}' with data found in 'state' property of the response file.");
+        }
+
+        private void UpdateJobContext(IExecutionContext context, ContainerInfo jobContainer, List<ContainerInfo> serviceContainers, PrepareJobResponse response)
+        {
+            if (response.Context == null)
+            {
+                Trace.Info($"The response file does not contain a context. The fields 'jobContext.Container' and 'jobContext.Services' will not be set.");
```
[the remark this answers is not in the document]
ruvceskistefan: Should we trace whole context or just some info? If we need to trace the whole context in which format?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
ruvceskistefan: Currently we are validating if `IsAlpine` is null. Should we validate the rest of the properties?

src/Runner.Worker/Handlers/StepHost.cs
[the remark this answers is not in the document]
fhammerl: I think this is a bigger refactoring effort, it might obfuscate the PR a bit too much.

After we sort out redundancies like that ☝️ though I'd be more than happy to take a look

src/Sdk/DTWebApi/WebApi/ActionsStepTelemetry.cs
```diff
@@ -56,5 +56,8 @@ public ActionsStepTelemetry()
 
         [DataMember(EmitDefaultValue = false)]
         public int? ExecutionTimeInSeconds { get; set; }
+
+        [DataMember(EmitDefaultValue = false)]
+        public bool? IsContainerHook { get; set; }
```
TingluoHuang, line 61: need to update this to match the server-side changes.

src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
TingluoHuang, line 87: Move this method to `Class HostContextExtension` in `HostContext.cs`

src/Runner.Worker/ActionManager.cs
```diff
@@ -101,38 +101,41 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
             IEnumerable<Pipelines.ActionStep> actions = steps.OfType<Pipelines.ActionStep>();
             executionContext.Output("Prepare all required actions");
             var result = await PrepareActionsRecursiveAsync(executionContext, state, actions, depth, rootStepId);
-            if (state.ImagesToPull.Count > 0)
+            if (!FeatureFlagManager.IsContainerHooksEnabled(executionContext.Global.Variables))
```
TingluoHuang, line 104: I think we still want to do this after the feature flag is enabled when you are not using container hooks.

src/Runner.Worker/ActionRunner.cs
```diff
@@ -158,8 +158,12 @@ public async Task RunAsync()
             // Setup container stephost for running inside the container.
             if (ExecutionContext.Global.Container != null)
             {
-                // Make sure required container is already created.
-                ArgUtil.NotNullOrEmpty(ExecutionContext.Global.Container.ContainerId, nameof(ExecutionContext.Global.Container.ContainerId));
+                // Make sure required container is already created
+                // Container hooks do not necesseraly set 'ContainerId'
+                if (!FeatureFlagManager.IsContainerHooksEnabled(ExecutionContext.Global.Variables))
```
TingluoHuang, line 163: I think the condition should be `ff enabled` + `use container hook`

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,302 @@
...
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    [ServiceLocator(Default = typeof(ContainerHookManager))]
+    public interface IContainerHookManager : IRunnerService
+    {
+        Task PrepareJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task CleanupJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task ContainerStepAsync(IExecutionContext context, ContainerInfo container, string dockerFile);
+        Task ScriptStepAsync(IExecutionContext context, ContainerInfo container, string arguments, string fileName, IDictionary<string, string> environment, string prependPath, string workingDirectory);
```
TingluoHuang, line 23: `ScriptStepAsync` makes it sounds like only takes care of `-run` step.

src/Runner.Worker/ActionRunner.cs
fhammerl: `FeatureFlagManager.IsContainerHooksEnabled` has been changed to check on the FF and the ENV 👍

src/Runner.Worker/ActionManager.cs
fhammerl: That should also be checked in `FeatureManager. IsContainerHooksEnabled`

src/Runner.Common/HostContext.cs
```diff
@@ -641,6 +642,30 @@ public static HttpClientHandler CreateHttpClientHandler(this IHostContext contex
             var handlerFactory = context.GetService<IHttpClientHandlerFactory>();
             return handlerFactory.CreateClientHandler(context.WebProxy);
         }
+
+        public static string GetDefaultShellForScript(this IHostContext hostContext, string path, Common.Tracing trace, string prependPath)
```
TingluoHuang, line 646: since you have the `hostContext` here, do we still need to pass in a `trace` object from the caller?

you can also create a new one via `var trace = hostContext.GetTrace(nameof(GetDefaultShellForScript));`

src/Runner.Worker/GlobalContext.cs
```diff
@@ -22,5 +23,6 @@ public sealed class GlobalContext
         public StepsContext StepsContext { get; set; }
         public Variables Variables { get; set; }
         public bool WriteDebug { get; set; }
+        public JToken ContainerHookState { get; set; }
```
TingluoHuang, line 26: do we care whether this is a JToken, JObject, or just a string?

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -318,21 +325,24 @@ public async Task RunAsync(ActionRunStage stage)
 
             ExecutionContext.Debug($"{fileName} {arguments}");
 
+            Inputs.TryGetValue("standardInInput", out var standardInInput); // string
```
TingluoHuang, line 328: do we need to wrap this inside an `if(IsContainerHooksEnabled){}` check?

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
```diff
@@ -0,0 +1,37 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    public abstract class HookResponse
+    {
+        public JToken State { get; set; }
+        public virtual void Validate() { }
+    }
+    public class PrepareJobResponse : HookResponse
```
TingluoHuang, line 12: do we need the abstract class if we only have `PrepareJobResponse` that overrides it?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+                    PrependPath = prependPath,
+                    WorkingDirectory = workingDirectory,
+                },
+                State = context.Global.ContainerHookState
+            };
+
+            PrepareJobResponse response;
+            try
+            {
+                response = await ExecuteHookScript<PrepareJobResponse>(context, input, ActionRunStage.Pre, prependPath);
```
TingluoHuang, line 150: same question about `PrepareJobResponse`

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+
+namespace GitHub.Runner.Worker.Container.ContainerHooks
+{
+    [ServiceLocator(Default = typeof(ContainerHookManager))]
+    public interface IContainerHookManager : IRunnerService
+    {
+        Task PrepareJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task CleanupJobAsync(IExecutionContext context, List<ContainerInfo> containers);
+        Task RunContainerStepAsync(IExecutionContext context, ContainerInfo container, string dockerFile);
+        Task RunScriptStepAsync(IExecutionContext context, ContainerInfo container, string arguments, string fileName, IDictionary<string, string> environment, string prependPath, string workingDirectory);
```
TingluoHuang, line 23:
```suggestion
        Task RunScriptStepAsync(IExecutionContext context, ContainerInfo container, string workingDirectory, string fileName, string arguments, IDictionary<string, string> environment, string prependPath);
```

rerange the args to match stepHost.executeasync

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+        {
+            base.Initialize(hostContext);
+            HookIndexPath = $"{Environment.GetEnvironmentVariable(Constants.Hooks.ContainerHooksPath)}";
+            ValidateHookExecutable();
+        }
+
+        public async Task PrepareJobAsync(IExecutionContext context, List<ContainerInfo> containers)
+        {
+            Trace.Entering();
+            var jobContainer = containers.Where(c => c.IsJobContainer).FirstOrDefault();
```
TingluoHuang, line 42:
```suggestion
            var jobContainer = containers.FirstOrDefault(c => c.IsJobContainer);
```

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            var serviceContainers = containers.Where(c => c.IsJobContainer == false).ToList();
```
TingluoHuang, line 43:
```suggestion
            var serviceContainers = containers.Where(c => !c.IsJobContainer).ToList();
```

src/Runner.Worker/Container/ContainerHooks/HookInput.cs
```diff
@@ -0,0 +1,98 @@
...
+        [EnumMember(Value = "prepare_job")]
+        PrepareJob,
+        [EnumMember(Value = "cleanup_job")]
+        CleanupJob,
+        [EnumMember(Value = "run_script_step")]
+        RunScriptStep,
+        [EnumMember(Value = "run_container_step")]
+        RunContainerStep,
+    }
+    public class HookArgs { }
```
TingluoHuang, line 30: it's a little bit weird to have an empty class...

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            var input = new HookInput
+            {
+                Command = HookCommand.CleanupJob,
+                ResponseFile = GenerateResponsePath(),
+                State = context.Global.ContainerHookState,
+            };
+            var prependPath = GetPrependPath(context);
+            try
+            {
+                await ExecuteHookScript<PrepareJobResponse>(context, input, ActionRunStage.Pre, prependPath);
```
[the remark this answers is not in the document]
TingluoHuang: also, we didn't use the return value, is it expected?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            var scriptDirectory = Path.GetDirectoryName(HookIndexPath);
+            var stepHost = HostContext.CreateService<IDefaultStepHost>();
+            Dictionary<string, string> inputs = new()
+            {
+                ["standardInInput"] = JsonUtility.ToString(input),
+                ["path"] = HookIndexPath,
+                ["shell"] = HostContext.GetDefaultShellForScript(HookIndexPath, Trace, prependPath)
+            };
+
+            var handlerFactory = HostContext.GetService<IHandlerFactory>();
```
TingluoHuang, line 180: feels weird that we are using Handler to execute this, should we use processInvoker directly?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+        private string GetPrependPath(IExecutionContext context)
+        {
+            return string.Join(Path.PathSeparator.ToString(), context.Global.PrependPath.Reverse<string>()); ;
+        }
+
+        private void ValidateHookExecutable()
+        {
+            if (!string.IsNullOrEmpty(HookIndexPath) && !File.Exists(HookIndexPath))
+            {
+                throw new Exception($"File not found at '{HookIndexPath}'. Set {Constants.Hooks.ContainerHooksPath} to the path of an existing file.");
```
TingluoHuang, line 217: FileNotFoundException or FileNotExistException?

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            {
+                Trace.Info($"Response file for the hook script at '{HookIndexPath}' running command '{input.Command}' not found.");
+                if (input.Command == HookCommand.PrepareJob)
+                {
+                    throw new Exception($"Response file is required but not found for the hook script at '{HookIndexPath}' running command '{input.Command}'");
+                }
+                return null;
+            }
+
+            T response = IOUtil.LoadObject<T>(input.ResponseFile);
```
TingluoHuang, line 239: do we need some protection or error handling for invalid JSON from the hook script?

src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -128,47 +158,71 @@ public async Task<string> DetermineNodeRuntimeVersion(IExecutionContext executio
...
+                                fileName,
+                                arguments,
+                                environment,
+                                requireExitCodeZero,
+                                outputEncoding,
+                                killProcessOnCancel,
+                                inheritConsoleHandler,
+                                cancellationToken);
+        }
+        public async Task<int> ExecuteAsync(string workingDirectory,
```
TingluoHuang, line 216: should we merge the 2 `ExecuteAsync` here? line 206 and line 177

src/Runner.Worker/GlobalContext.cs
fhammerl: Thanks for mentioning that, `JObject` is the most precise: State is never a `JArray` or a `JProperty`, and we don't need to stringify it here.

`JToken` works, but it's a bit sloppy as it's just the abstract superclass of all the above.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
fhammerl: We do catch them in the hook methods individually, I think that should be brought down to ExecuteHookScript

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            try
+            {
+                response = await ExecuteHookScript<PrepareJobResponse>(context, input, ActionRunStage.Pre, prependPath);
+            }
+            catch (Exception ex)
+            {
+                Trace.Error(ex);
+                throw new Exception($"Custom container implementation failed with error: {ex.Message}. Please contact your self hosted runner administrator.");
+            }
+            if (response == null)
```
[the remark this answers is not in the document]
fhammerl: Response is not necessary for anything but PrepareJob

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
fhammerl: This would throw if there isn't a jobContainer which is not what we want. SingleOrDefault is more precise though, so I used that and made sure that PrepareJobAsync can handle a null `jobContainer`

src/Runner.Worker/GlobalContext.cs
fhammerl: Refactored it, using JObject now

src/Runner.Worker/Container/ContainerHooks/HookResponse.cs
fhammerl: Should have been `HookResponse` :/ Fixed it.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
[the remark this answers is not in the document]
fhammerl: Should be HookResponse, fixed it.

Aside from PrepareJobResponse, we only use the response to update `State`. After cleanupJob, we don't need to update it anymore because the job is finished.

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+                Command = HookCommand.RunContainerStep,
+                ResponseFile = GenerateResponsePath(),
+                State = hookState
+            };
+
+            var prependPath = GetPrependPath(context);
+            PrepareJobResponse response;
+            try
+            {
+                response = await ExecuteHookScript<PrepareJobResponse>(context, input, ActionRunStage.Pre, prependPath);
```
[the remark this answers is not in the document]
fhammerl: They shouldn't...

src/Runner.Worker/Container/ContainerHooks/ContainerHookManager.cs
```diff
@@ -0,0 +1,308 @@
...
+            {
+                throw new Exception($"The hook script at '{HookIndexPath}' running command '{input.Command}' did not execute successfully."); // TODO: better exception
+            }
+            var response = GetResponse<T>(input);
+            return response;
+        }
+
+        private string GenerateResponsePath()
+        {
+            return Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Temp), ResponseFolderName, $"{Guid.NewGuid()}.json");
```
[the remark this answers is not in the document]
fhammerl: I refactored them into expression bodies, but I think it makes sense not to repeat their content for each hook, even if the methods are short.

src/Runner.Worker/Container/ContainerHooks/HookInput.cs
fhammerl: Refactored the class hierarchy a bit, thanks for pointing this out.

src/Runner.Common/HostContext.cs
thboop: Lets do this!

src/Runner.Worker/ActionManager.cs
```diff
@@ -101,38 +101,41 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
                 }
-            }
 
 #if !OS_LINUX
-            if (containerSetupSteps.Count > 0)
-            {
-                executionContext.Output("Container action is only supported on Linux, skip pull and build docker images.");
-                containerSetupSteps.Clear();
-            }
+                if (containerSetupSteps.Count > 0)
```
thboop, line 132: nit: indents here are off I think

src/Runner.Worker/Handlers/StepHost.cs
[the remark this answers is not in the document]
thboop: I like this idea, but we can take it as a refactor item after this pr merges

src/Runner.Worker/ActionRunner.cs
```diff
@@ -158,8 +158,12 @@ public async Task RunAsync()
             // Setup container stephost for running inside the container.
             if (ExecutionContext.Global.Container != null)
             {
-                // Make sure required container is already created.
-                ArgUtil.NotNullOrEmpty(ExecutionContext.Global.Container.ContainerId, nameof(ExecutionContext.Global.Container.ContainerId));
+                // Make sure required container is already created
+                // Container hooks do not necesseraly set 'ContainerId'
```
thboop, line 162:
```suggestion
                // Container hooks do not necessarily set 'ContainerId'
```

src/Runner.Worker/ActionRunner.cs
```diff
@@ -158,8 +158,12 @@ public async Task RunAsync()
             // Setup container stephost for running inside the container.
             if (ExecutionContext.Global.Container != null)
             {
-                // Make sure required container is already created.
-                ArgUtil.NotNullOrEmpty(ExecutionContext.Global.Container.ContainerId, nameof(ExecutionContext.Global.Container.ContainerId));
+                // Make sure required container is already created
```
thboop, line 161:
```suggestion
                // Make sure the required container is already created
```

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -50,72 +55,16 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
-            {
-                throw new NotSupportedException($"Min required docker engine API client version is '{requiredDockerEngineAPIVersion}', your docker ('{_dockerManager.DockerPath}') client version is '{dockerVersion.ClientVersion}'");
+                executionContext.StepTelemetry.ContainerHookData = _containerHookManager.GetContainerHookData();
+                // Initialize the containers
+                containers.ForEach(container => UpdateRegistryAuthForGitHubToken(executionContext, container));
+                containers.ForEach(container => MountWellKnownDirectories(executionContext, container));
+                await _containerHookManager.PrepareJobAsync(executionContext, containers);
+                return;
             }
+            await AssertCompatibleOS(executionContext);
```
TingluoHuang, line 67: is this just copy-paste? or there are changes for all the things we checked.
I would suggest to undo any refact changes when you work on massive changes to reduce the amount of diff for reviewing.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -166,6 +115,14 @@ public async Task StopContainersAsync(IExecutionContext executionContext, object
             List<ContainerInfo> containers = data as List<ContainerInfo>;
             ArgUtil.NotNull(containers, nameof(containers));
 
+            bool isContainerHooksEnabled = FeatureManager.IsContainerHooksEnabled(executionContext.Global.Variables);
+            if (isContainerHooksEnabled)
```
TingluoHuang, line 119:
```suggestion
            if (FeatureManager.IsContainerHooksEnabled(executionContext.Global.Variables))
```

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -238,35 +195,7 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
...
-
-                var tempWorkflowDirectory = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Temp), "_github_workflow");
-                Directory.CreateDirectory(tempWorkflowDirectory);
-                container.MountVolumes.Add(new MountVolume(tempWorkflowDirectory, "/github/workflow"));
-                container.AddPathTranslateMapping(tempWorkflowDirectory, "/github/workflow");
-
-                container.ContainerWorkDirectory = container.TranslateToContainerPath(workingDirectory);
-                container.ContainerEntryPoint = "tail";
-                container.ContainerEntryPointArgs = "\"-f\" \"/dev/null\"";
+                MountWellKnownDirectories(executionContext, container);
```
TingluoHuang, line 198: same comment about whether this is just copy-paste? undo if so, to reduce the diff in code review.

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -237,6 +239,11 @@ public async Task RunAsync(ActionRunStage stage)
                 ExecutionContext.StepTelemetry.Action = shellCommand;
             }
 
+            if (ExecutionContext.JobContext.Container != null)
+            {
+                ExecutionContext.StepTelemetry.ContainerHookData = HostContext.GetService<IContainerHookManager>().GetContainerHookData();
```
TingluoHuang, line 244: should we do this in a single place inside the ExecutioContext? so each caller doesn't have to do this by themselves.

src/Runner.Worker/ActionManager.cs
fhammerl: This condition is now part of the `if (!FeatureManager.IsContainerHooksEnabled(executionContext.Global.Variables))` outer condition, it looks correct when you look at it in an editor

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -90,6 +92,12 @@ public async Task RunAsync(ActionRunStage stage)
             }
             ExecutionContext.StepTelemetry.Type = "composite";
 
+            if (ExecutionContext.JobContext.Container != null)
```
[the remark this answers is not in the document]
fhammerl: I realised perhaps the best place to do it is right before hook execution in ContainerHookManager. We have access to the executioncontext there, and we are 100% sure it's a container hook step at that point.

edbighead:
looking forward to test it but [ADR](https://github.com/github/c2c-actions-runtime/pull/1748) is private :(

src/Runner.Worker/ActionManager.cs
Gam3356: Thank you guys! plese help me Get really of it can't beta getting all the funds

timharris777:
I can confirm [ADR](https://github.com/github/c2c-actions-runtime/pull/1748) is private

jb-2020:
Do we need to enable `DistributedTask.AllowRunnerContainerHooks` somehow?  Or is simply setting `ACTIONS_RUNNER_CONTAINER_HOOKS` sufficient?  

Any insight into how to toggle these feature flags?

PR #1871, opened 2022-05-04
(delete.sh) Loggin repaired and made runner_name optional defaulting to hostname

### LINKED ISSUE #1867: Stale sample usage `./delete.sh myorg`
This assumes one can pass a single argument:
https://github.com/actions/runner/blob/777ce5a0dc7ff840d0456870026e4351617a1dff/scripts/delete.sh#L11

This clearly indicates that isn't legal:
https://github.com/actions/runner/blob/777ce5a0dc7ff840d0456870026e4351617a1dff/scripts/delete.sh#L37-L38

And this code is called too early (it shouldn't be called until after the preceding excerpt):
https://github.com/actions/runner/blob/777ce5a0dc7ff840d0456870026e4351617a1dff/scripts/delete.sh#L29

nikola-jokic:
Closes #1867


scripts/delete.sh
```diff
@@ -12,7 +12,7 @@ set -e
 #
 # Usage:
 #     export RUNNER_CFG_PAT=<yourPAT>
-#     ./delete.sh scope name
+#     ./delete.sh <scope> <name>
```
jsoref, line 15:
```suggestion
#     ./delete.sh <scope> [<name>]
```

fhammerl (approved):
I like the move of `echo "Deleting runner ${runner_name} @ ${runner_scope}"` below the validation


LGTM

PR #1889, opened 2022-05-12
Save original, pre-parsed string from workflow input for the user's custom volume mounts 

fhammerl:
We can later use this original, pre-parsed string for logging and to separate mountvolumes coming from the user vs runner defaults.

### DIFF
--- a/src/Runner.Worker/Container/ContainerInfo.cs
+++ b/src/Runner.Worker/Container/ContainerInfo.cs
@@ -1,16 +1,16 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
-using GitHub.Runner.Common.Util;
 using GitHub.Runner.Common;
 using GitHub.Runner.Sdk;
 using Pipelines = GitHub.DistributedTask.Pipelines;
+using System.Collections.ObjectModel;
+using System.Linq;
 
 namespace GitHub.Runner.Worker.Container
 {
     public class ContainerInfo
     {
-        private IDictionary<string, string> _userMountVolumes;
         private List<MountVolume> _mountVolumes;
         private IDictionary<string, string> _userPortMappings;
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -192,13 +192,12 @@ private async Task StartContainerAsync(IExecutionContext executionContext, Conta
             {
                 Trace.Info($"User provided port: {port.Value}");
             }
-            foreach (var volume in container.UserMountVolumes)
+            foreach (var mount in container.UserMountVolumes)
             {
-                Trace.Info($"User provided volume: {volume.Value}");
-                var mount = new MountVolume(volume.Value);
+                Trace.Info($"User provided volume: {mount.UserProvidedValue}");
```
fhammerl, line 197: I tested we still print the same `[2022-05-13 13:00:38Z INFO ContainerOperationProvider] User provided volume: /home/test:/home/test/` message

src/Runner.Worker/Container/ContainerInfo.cs
```diff
@@ -260,18 +260,27 @@ private void UpdateWebProxyEnv(RunnerWebProxy webProxy)
 
     public class MountVolume
     {
+        public string UserProvidedValue { get; set; }
```
fhammerl, line 263: Naming suggestions welcome

nikola-jokic: I like it, to be honest. It is the string value user provided after all :relaxed:

thboop: Should we just maintain a list of the system provided mounts and do .Where not in that list?

thboop (approved):
lgtm

PR #1903, opened 2022-05-19
Set timeout when sending live console log to service

TingluoHuang:
The `websocketClient.SendAsync()` doesn't have a default timeout, which means if the WebSocket is in a bad state, the client (runner) won't abort the request until the service kills the connection after 10 minutes (Max connection time control by AFD).

I am adding a CancellationToken to auto-fire after 60 seconds to avoid the runner sitting there and waiting for the 10 minutes timeout.

### DIFF
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -1,5 +1,4 @@
-using GitHub.DistributedTask.WebApi;
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -9,6 +8,7 @@ using System.Net.WebSockets;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Sdk;
 using GitHub.Services.Common;
 using GitHub.Services.WebApi;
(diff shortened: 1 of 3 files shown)


src/Runner.Common/JobServer.cs
```diff
@@ -1,5 +1,4 @@
-using GitHub.DistributedTask.WebApi;
```
TingluoHuang, line 1: hide whitespace when reviewing.

src/Runner.Common/JobServer.cs
```diff
@@ -274,7 +274,7 @@ public async Task AppendTimelineRecordFeedAsync(Guid scopeIdentifier, string hub
                 }
             }
 
-            if (!pushedLinesViaWebsocket)
+            if (!pushedLinesViaWebsocket && !cancellationToken.IsCancellationRequested)
```
TingluoHuang, line 277: failback to HttpClient only if we haven't been cancelled.

src/dir.proj
```diff
@@ -48,7 +48,7 @@
 
     <Target Name="Test" DependsOnTargets="GenerateConstant">
         <Exec Command="dotnet build Test/Test.csproj -c $(BUILDCONFIG) /p:PackageRuntime=$(PackageRuntime)" ConsoleToMSBuild="true" />
-        <Exec Command="dotnet test Test/Test.csproj --no-build --logger:trx" ConsoleToMSBuild="true" />
+        <Exec Command="dotnet test Test/Test.csproj -c $(BUILDCONFIG) --no-build --logger:trx" ConsoleToMSBuild="true" />
```
TingluoHuang, line 51: i noticed `./dev.sh test Release` is not working... this should fix it.

thboop (approved):
lgtm

PR #1907, opened 2022-05-23
  support messages from broker

ericsciple:
Working on broker flow. Introducing a new message type received from the broker (message ref). Upon receiving a message ref, the Runner calls the Run service to retrieve the full job message.


src/Runner.Listener/Runner.cs
```diff
@@ -457,6 +461,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
+                                }
+                                else
+                                {
+                                    var messageRef = StringUtil.ConvertFromJson<MessageRef>(message.Body);
+
+                                    // Create connection
+                                    var credMgr = HostContext.GetService<ICredentialManager>();
+                                    var creds = credMgr.LoadCredentials();
+
+                                    // todo: add retries
```
ericsciple, line 480: todo before merge

ericsciple: actually, i wonder whether we should just let the other while loop handle it. Thoughts?

src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
```diff
@@ -703,6 +704,33 @@ public TaskAgentHttpClientBase(Uri baseUrl, HttpMessageHandler pipeline, bool di
...
+        public virtual Task<Pipelines.AgentJobRequestMessage> GetJobMessageAsync(
+            string id,
+            object userState = null,
+            CancellationToken cancellationToken = default)
+        {
+            HttpMethod httpMethod = new HttpMethod("GET");
+            Guid locationId = new Guid("25adab70-1379-4186-be8e-b643061ebe3a");
+
+            List<KeyValuePair<string, string>> queryParams = new List<KeyValuePair<string, string>>();
+            queryParams.Add("id", id);
```
TingluoHuang, line 723: it's a little bit weird that we are using query parameters instead of PATH. 😄

src/Runner.Listener/MessageRef.cs
```diff
@@ -0,0 +1,11 @@
+using System.Runtime.Serialization;
+
+namespace GitHub.Runner.Listener
+{
+    [DataContract]
+    public sealed class MessageRef
```
TingluoHuang, line 6: `RunnerJobRequestRef`?

src/Runner.Common/RunServer.cs
```diff
@@ -0,0 +1,76 @@
...
+                {
+                    await connection.ConnectAsync();
+                    return connection;
+                }
+                catch (Exception ex) when (attemptCount > 0)
+                {
+                    Trace.Info($"Catch exception during connect. {attemptCount} attempt left.");
+                    Trace.Error(ex);
+
+                    await HostContext.Delay(TimeSpan.FromMilliseconds(100), CancellationToken.None);
```
TingluoHuang, line 54: do you want some random backoff instead of always 100ms?

src/Runner.Listener/Runner.cs
TingluoHuang: If you want to use the while loop, you need to catch any exception that you want to retry and skip the message deletion, you need to let the MessageListener not bump its `lastMessageId` since it will always try to pull the latest message that it haven't process before.

src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
ericsciple: good catch.... had a bunch of query params early when testing. In the process of switching to message ID as the only parameter now. Will switch to a value in the path. I like that.

src/Runner.Listener/MessageRef.cs
ericsciple: I like that. This is specific to job requests and not for example cancel messages

src/Runner.Common/RunServer.cs
ericsciple: will look more, i think i copied this :)

ericsciple:
moving to draft... working out some e2e communication patterns...

PR #1910, opened 2022-05-24
Increase maximal issue count from 10 to 100

j-liedtke:
In my opinion, the [problem matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) is an underrated feature (e.g. not documented [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions)) that is currently limited by the number of issues that are reported. This change removes this limitation and will thus greatly improve the problem matcher feature.

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -114,7 +114,7 @@ namespace GitHub.Runner.Worker
 
     public sealed class ExecutionContext : RunnerService, IExecutionContext
     {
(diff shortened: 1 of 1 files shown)


cbandera:
@actions/actions-runtime could you approve the workflow run?
I also believe that this PR is very helpful and would approve our developer experience in PRs!

mikebowyer (approved):

LisaBurke40 (approved):

cbandera:
Ping

PR #1913, opened 2022-05-25
Created env var for forcing node12 actions to run on node16

### LINKED ISSUE #1911: Create an env variable to force `node12` actions to run on node16
We should have a way for a self hosted runner admin to force `node12` actions to run on `node16`

```
ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION=16
```

While this could be a breaking change, it's very unlikely to impact most actions. The goal is to allow runner admins more control over what node version is used, particularly in scenarios where a node version is out of date or deprecated.

ruvceskistefan:
Closes #1911


src/Runner.Worker/Handlers/StepHost.cs
```diff
@@ -128,7 +128,14 @@ public async Task<string> DetermineNodeRuntimeVersion(IExecutionContext executio
 
             var output = new List<string>();
             var execExitCode = await dockerManager.DockerExec(executionContext, Container.ContainerId, string.Empty, osReleaseIdCmd, output);
-            string nodeExternal;
+            
+            string nodeExternal = preferredVersion;
+            if (executionContext.Global.EnvironmentVariables != null)
```
TingluoHuang, line 133: should we do it here or in ActionsManger.cs where we load the action.yaml?

thboop, line 133: We should get variables from the runner machine (host context), here, not the[ job level envs](https://docs.github.com/en/actions/learn-github-actions/environment-variables)

thboop: I think we can do something like

```
bool.TryParse(Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion), out forceNode16);
```

thboop: To test this, what you can do is create an `.env` file at the root of the runner layout folder, and set this value, then validate you see it works as expected

src/Runner.Worker/ActionManager.cs
```diff
@@ -399,6 +399,13 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                     else if (definition.Data.Execution.ExecutionType == ActionExecutionType.NodeJS)
                     {
                         var nodeAction = definition.Data.Execution as NodeJSActionExecutionData;
+                        
+                        string nodeVersionOut = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
```
TingluoHuang, line 403: sorry... my bad...
i was suggesting `NodeScriptActionHandler` which we already have some code to force to use node16 for osx-arm64.

src/Runner.Common/Constants.cs
```diff
@@ -238,6 +238,7 @@ public static class Agent
 
                 // Set this env var to "node12" to downgrade the node version for internal functions (e.g hashfiles). This does NOT affect the version of node actions.
                 public static readonly string ForcedInternalNodeVersion = "ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION"; 
+                public static readonly string ForcedActionsNodeVersion = "ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION";
```
TingluoHuang, line 241: If we suggest customer to delete `node12` from the runner, then this is not only for Actions, we might need to update the `NodeUtil.GetInternalNodeVersion` to also respect this.

might need to consider the collision case with `ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION` set to `node12`

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -15,8 +15,14 @@ public static class NodeUtil
 
         public static string GetInternalNodeVersion()
         {
-            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
-            return !string.IsNullOrEmpty(forcedNodeVersion) && BuiltInNodeVersions.Contains(forcedNodeVersion) ? forcedNodeVersion : _defaultNodeVersion;
+            var forcedInternalNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
+            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
+
+            return !string.IsNullOrEmpty(forcedNodeVersion) && BuiltInNodeVersions.Contains(forcedNodeVersion)
```
fhammerl, line 21: NIT: Can you extract some of these conditions into variables so the return statement is a bit easier to understand?

src/Runner.Worker/ActionManager.cs
```diff
@@ -399,6 +399,7 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                     else if (definition.Data.Execution.ExecutionType == ActionExecutionType.NodeJS)
                     {
                         var nodeAction = definition.Data.Execution as NodeJSActionExecutionData;
+
```
fhammerl, line 402:
```suggestion
```

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -15,8 +15,21 @@ public static class NodeUtil
 
         public static string GetInternalNodeVersion()
         {
-            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
-            return !string.IsNullOrEmpty(forcedNodeVersion) && BuiltInNodeVersions.Contains(forcedNodeVersion) ? forcedNodeVersion : _defaultNodeVersion;
+            var forcedInternalNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
+            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
```
fhammerl, line 19: This method is for determining the node version for **internal** JS processes, like hashfiles.js or hook execution. This has no effect on how a nodeJS github action is executed.

Why do you consider the forced **action** node version here?

ruvceskistefan: I think that you're right, now it makes more sense. I assume I misunderstood Thomas' request, thank you for your explanation.

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -102,6 +102,11 @@ public async Task RunAsync(ActionRunStage stage)
                 Data.NodeVersion = "node16";
             }
 #endif
+            string nodeVersionOut = System.Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
+            if(nodeVersionOut == "16" && Data.NodeVersion != "node16")
```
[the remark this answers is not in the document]
ruvceskistefan: Makes sense, I agree with you on this one

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -13,10 +13,21 @@ public static class NodeUtil
         public static readonly ReadOnlyCollection<string> BuiltInNodeVersions = new(new[] { "node12", "node16" });
 #endif
 
+        public static string GetNodeVersion()
+        {
+            return System.Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
```
TingluoHuang, line 18: do we need a method that only has a single line?

src/Test/L0/Worker/StepHostL0.cs
```diff
@@ -55,7 +55,7 @@ public async Task DetermineNodeRuntimeVersionInContainerAsync()
                 Assert.Equal("node12", nodeVersion);
             }
         }
-
+
```
fhammerl, line 58: It looks like the whitespace is still there

src/Runner.Common/Util/NodeUtil.cs
fhammerl: NIT: Let's remove the method and do it inline

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -102,6 +103,11 @@ public async Task RunAsync(ActionRunStage stage)
                 Data.NodeVersion = "node16";
             }
 #endif
+            string nodeVersionOut = NodeUtil.GetNodeVersion();
```
fhammerl, line 106:
```suggestion
            var forcedNodeVersion = System.Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion);
```

Let's remove the method above above

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -15,8 +15,14 @@ public static class NodeUtil
 
         public static string GetInternalNodeVersion()
         {
-            var forcedNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
-            return !string.IsNullOrEmpty(forcedNodeVersion) && BuiltInNodeVersions.Contains(forcedNodeVersion) ? forcedNodeVersion : _defaultNodeVersion;
+            var forcedInternalNodeVersion = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion);
```
thboop, line 18: Is this just a refactor? It doesn't appear to do anything if i'm reading it correctly. Should we just revert these changes?

ruvceskistefan: @fhammerl asked to refactor this one with meaningful var names in [this](https://github.com/actions/runner/pull/1913#discussion_r886521138) comment

thboop (dismissed):
lgtm

thboop (approved):
LGTM

PR #1930, opened 2022-06-08
add job_name to github context

IndraGunawan:
this is for the context https://github.community/t/get-current-job-name-within-the-workflow/255214

`message.JobDisplayName` will contain `jobs.<job_id>.name` if present otherwise `jobs.<job_id>`


IndraGunawan:
@TingluoHuang any chances to look at this?

PR #1932, opened 2022-06-09
Revert bash and sh `-e` filePath escape

fhammerl:
It generated invalid arguments for `System.Diagnostic.Process` when the `bash` command itself was an argument as well, for example:

```
_proc.StartInfo.FileName = "/usr/bin/docker";
_proc.StartInfo.Arguments = "exec -i --workdir /__w/container-hook-e2e/container-hook-e2e 47105c66144d8809d9fa2bce9a58ea0564cd14def0ae7952cd6231fba3576db1 sh -e '/__w/_temp/fd086560-cb92-4f3b-a99c-35a6b7b1bbdb.sh'";
```

`sh` here would look for a literal path containing a single quote, which doesn't exist.

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs
@@ -13,8 +13,8 @@ namespace GitHub.Runner.Worker.Handlers
             ["cmd"] = "/D /E:ON /V:OFF /S /C \"CALL \"{0}\"\"",
             ["pwsh"] = "-command \". '{0}'\"",
             ["powershell"] = "-command \". '{0}'\"",
-            ["bash"] = "--noprofile --norc -e -o pipefail '{0}'",
-            ["sh"] = "-e '{0}'",
+            ["bash"] = "--noprofile --norc -e -o pipefail {0}",
+            ["sh"] = "-e {0}",
             ["python"] = "{0}"
         };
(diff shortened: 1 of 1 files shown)


nikola-jokic:
Just to put more context on this end
If we invoke bash with arguments, bash goes through parsing the arguments and quote removal at the end
If we invoke docker as a process, it does not have the quote removal, and it will parse arguments as is, having the quotes around the script (here, we could not even use the multi-word argument).
You would normally invoke docker from a shell (bash, zsh...) which does this command line argument parsing when it invokes the docker.
We should make sure that bash is our `entrypoint` for the `ProcessInvoker` if we want to add quotes to allow for arguments with spaces in them.

ChristopherHX:
> We should make sure that bash is our `entrypoint` for the `ProcessInvoker` if we want to add quotes to allow for arguments with spaces in them.


I'm not shure you understand the problem here
- Dotnet parses your commandline on linux, macOS etc. and only supports double quotes in cmdline, then execv is called with the parsed argv array.
  - The caller program here dotnet is responsible for cmdline parsing
  - see dotnet sourcecode, no single quote parsing
- Dotnet calls windows api with cmdline.
  - The called program is responsible for cmdline parsing
  - single quotes can work on windows, if the called program has a custom cmdline interpreter
  - By default windows cmdline parser only supports double quotes.

PR #1934, opened 2022-06-09
Handle message from broker

ericsciple:


src/Runner.Listener/Runner.cs
```diff
@@ -469,14 +501,6 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                                     Trace.Info($"Skip message deletion for cancellation message '{message.MessageId}'.");
                                 }
                             }
-                            else if (string.Equals(message.MessageType, Pipelines.HostedRunnerShutdownMessage.MessageType, StringComparison.OrdinalIgnoreCase))
```
ericsciple, line 472: revert

src/Runner.Listener/Runner.cs
```diff
@@ -510,18 +534,15 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                         await jobDispatcher.ShutdownAsync();
                     }
 
-                    if (!skipSessionDeletion)
```
ericsciple, line 513: revert

src/Runner.Listener/Runner.cs
```diff
@@ -1,18 +1,22 @@
...
 using GitHub.Runner.Sdk;
-using GitHub.Services.WebApi;
-using Pipelines = GitHub.DistributedTask.Pipelines;
+using System.Linq;
+using GitHub.Runner.Listener.Check;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
```
ericsciple, line 19: fix

src/Runner.Listener/Runner.cs
```diff
@@ -458,6 +458,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
+                                {
+                                    var messageRef = StringUtil.ConvertFromJson<RunnerJobRequestRef>(message.Body);
+
+                                    // Create connection
+                                    var credMgr = HostContext.GetService<ICredentialManager>();
+                                    var creds = credMgr.LoadCredentials();
+
+                                    // todo: add retries https://github.com/github/actions-broker/issues/49
+                                    var runServer = HostContext.CreateService<IRunServer>();
+                                    await runServer.ConnectAsync(new Uri(settings.ServerUrl), creds);
```
TingluoHuang, line 479: we are creating a new connection on every message.

ericsciple: Well, we won't when we switch over to the run service 😄

TingluoHuang: I am fine with the change now, but long term we might want to figure out all the URLs we need and where we get them. ex: broker URL, admin URL, runService URL.

TingluoHuang (approved):

PR #1939, opened 2022-06-10
Add retry logic around getting job messages from broker

takost:
Add retry logic for Broker flow aroung getting messages.
Retries are executed for ~5 minuts and then abandoned.
Issue: https://github.com/github/actions-broker/issues/49

### DIFF
--- a/src/Runner.Common/RunServer.cs
+++ b/src/Runner.Common/RunServer.cs
@@ -98,7 +98,7 @@ namespace GitHub.Runner.Common
                     if (currentRetryAttempt < maxRetryAttempts)
                     {
                         var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(15));
-                        Trace.Warning($"Back off {backOff.TotalSeconds} seconds before retry.");
+                        Trace.Warning($"Back off {backOff.TotalSeconds} seconds before next retry. {maxRetryAttempts - currentRetryAttempt} attempt left.");
                         currentRetryAttempt++;
                         await Task.Delay(backOff, cancellationToken);
                     }
--- a/src/Runner.Common/RunServer.cs
+++ b/src/Runner.Common/RunServer.cs
@@ -17,7 +17,7 @@ namespace GitHub.Runner.Common
     {
         Task ConnectAsync(Uri serverUrl, VssCredentials credentials);
 
-        Task<AgentJobRequestMessage> GetJobMessageAsync(string id);
+        Task<AgentJobRequestMessage> GetJobMessageAsync(string id, CancellationToken token);
     }
 
     public sealed class RunServer : RunnerService, IRunServer
@@ -67,23 +67,25 @@ namespace GitHub.Runner.Common
             }
         }
 
-        public Task<AgentJobRequestMessage> GetJobMessageAsync(string id)
+        public Task<AgentJobRequestMessage> GetJobMessageAsync(string id, CancellationToken cancellationToken)
         {
             CheckConnection();
-            var jobMessage = RetryRequestWithTimeout<AgentJobRequestMessage>(async () =>
+            var jobMessage = RetryRequest<AgentJobRequestMessage>(async () =>
                                                     {
-                                                        return await _taskAgentClient.GetJobMessageAsync(id);
-                                                    });
+                                                        return await _taskAgentClient.GetJobMessageAsync(id, cancellationToken);
+                                                    }, cancellationToken);
             return jobMessage;
         }
 
-        private async Task<T> RetryRequestWithTimeout<T>(Func<Task<T>> func,
-            int maxTimeoutMinutes = 5)
+        private async Task<T> RetryRequest<T>(Func<Task<T>> func,
+                                                        CancellationToken cancellationToken,
+                                                        int maxRetryAttempts = 5
+                                                        )
         {
-            var currentRetryTime = TimeSpan.Zero;
-            var maxRetryTime = TimeSpan.FromMinutes(maxTimeoutMinutes);
+            var currentRetryAttempt = 1;
             while (true)
             {
+                cancellationToken.ThrowIfCancellationRequested();
                 try
                 {
                     return await func();
@@ -93,16 +95,16 @@ namespace GitHub.Runner.Common
                     Trace.Error("Catch exception during get full job message");
                     Trace.Error(ex);
 
-                    if (currentRetryTime < maxRetryTime)
+                    if (currentRetryAttempt < maxRetryAttempts)
                     {
                         var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(15));
                         Trace.Warning($"Back off {backOff.TotalSeconds} seconds before retry.");
-                        currentRetryTime += backOff;
-                        await Task.Delay(backOff);
+                        currentRetryAttempt++;
+                        await Task.Delay(backOff, cancellationToken);
                     }
                     else
                     {
-                        Trace.Info("Retry time limit exceeded. Abandoning getting message");
+                        Trace.Info("Retry attempt limit exceeded. Abandoning getting message");
                         throw;
                     }
                 }
--- a/src/Runner.Common/RunServer.cs
+++ b/src/Runner.Common/RunServer.cs
@@ -70,28 +70,38 @@ namespace GitHub.Runner.Common
         public Task<AgentJobRequestMessage> GetJobMessageAsync(string id)
         {
             CheckConnection();
-            return RequestJobMessageAsync(id);
+            var jobMessage = RetryRequestWithTimeout(async () =>
+                                                    {
+                                                        return await _taskAgentClient.GetJobMessageAsync(id);
+                                                    },
+                                                    TimeSpan.FromSeconds(5),
+                                                    TimeSpan.FromSeconds(15),
+                                                    TimeSpan.FromMinutes(5));
+            return jobMessage;
         }
 
-        private async Task<AgentJobRequestMessage> RequestJobMessageAsync(string id)
+        private async Task<AgentJobRequestMessage> RetryRequestWithTimeout(Func<Task<AgentJobRequestMessage>> func,
+            TimeSpan minBackoff,
+            TimeSpan maxBackoff,
+            TimeSpan maxTimeoutMinutes)
         {
-            var remainingRetryLimitTime = TimeSpan.FromMinutes(5);
+            var currentRetryTime = TimeSpan.Zero;
             while (true)
             {
                 try
                 {
-                    return await _taskAgentClient.GetJobMessageAsync(id);;
+                    return await func();
                 }
                 catch (Exception ex)
                 {
                     Trace.Error("Catch exception during get full job message");
                     Trace.Error(ex);
 
-                    if (remainingRetryLimitTime > TimeSpan.Zero)
+                    if (currentRetryTime < maxTimeoutMinutes)
                     {
-                        var backOff = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(15));
+                        var backOff = BackoffTimerHelper.GetRandomBackoff(minBackoff, maxBackoff);
                         Trace.Warning($"Back off {backOff.TotalSeconds} seconds before retry.");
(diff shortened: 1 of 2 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -477,7 +478,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                                     // todo: add retries https://github.com/github/actions-broker/issues/49
```
fhammerl, line 478: NIT: Todo looks done

src/Runner.Listener/Runner.cs
```diff
@@ -477,7 +478,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
                                     var runServer = HostContext.CreateService<IRunServer>();
                                     await runServer.ConnectAsync(new Uri(settings.ServerUrl), creds);
-                                    var jobMessage = await runServer.GetJobMessageAsync(messageRef.RunnerRequestId);
+                                    Pipelines.AgentJobRequestMessage jobMessage = null;
+
+                                    var remainingRetryLimitTime = TimeSpan.FromMinutes(5);
```
fhammerl, line 483: NIT: just curious, why a 5 minute window for retrying instead of a number of retries?

src/Runner.Listener/Runner.cs
```diff
@@ -477,7 +478,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
+                                        {
+                                            jobMessage = await runServer.GetJobMessageAsync(messageRef.RunnerRequestId);
+                                            break;
+                                        }
+                                        catch (Exception ex)
+                                        {
+                                            Trace.Error("Catch exception during get full job message");
+                                            Trace.Error(ex);
+
+                                            if (remainingRetryLimitTime > TimeSpan.Zero)
```
fhammerl, line 496: Do we always want to retry? Should we break on 4xx response codes?

src/Runner.Listener/Runner.cs
```diff
@@ -477,7 +478,35 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
                                     await runServer.ConnectAsync(new Uri(settings.ServerUrl), creds);
-                                    var jobMessage = await runServer.GetJobMessageAsync(messageRef.RunnerRequestId);
+                                    Pipelines.AgentJobRequestMessage jobMessage = null;
+
+                                    var remainingRetryLimitTime = TimeSpan.FromMinutes(5);
+                                    while (true)
+                                    {
+                                        try
+                                        {
+                                            jobMessage = await runServer.GetJobMessageAsync(messageRef.RunnerRequestId);
```
fhammerl, line 488: Info: What happens now in case `runServer.GetJobMessageAsync` throws?
Doesn't the loop at `while (!HostContext.RunnerShutdownToken.IsCancellationRequested)` at Line 337 handle retries?

takost: The loop on 337 always tries to retrieve latest message from MessageListener.
It is hard and is not worth to make it not retrieve latest message if we want to retry only retrieving message from broker.

> Info: What happens now in case `runServer.GetJobMessageAsync` throws?

Do you mean before the change? Right now it will be skipped. Current message from messagelistener will be deleted, and in the new iteration we will work already with other message

src/Runner.Listener/Runner.cs
takost: I discussed it with @ericsciple, right now it is fine to retry all errors, but I keep thinking about non-retriable exceptions

src/Runner.Listener/Runner.cs
takost: I disscussed limits with @ericsciple , so I just based on his thoughts. 
But I think that for errors that can be caused by servers\services instability it is good idea to rely on timeouts and retry several times during some timeframe, instead of rely on number of retries itself

src/Runner.Listener/Runner.cs
```diff
@@ -474,10 +475,37 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
```
TingluoHuang, line 487: Why do we need a retry at this layer? 
- the VSSHttpClient already has retry built-in.
- I don't think we want to retry on all exceptions
- If we really want to retry, I would suggest retrying within the `runServer.GetJobMessageAsync` to keep the `Runner.cs` simple.

ericsciple: @TingluoHuang that is interesting. Keep in mind VSSF is temporary. Eventually will switch over to Run svc.

We could avoid retrying some HTTP codes (400/401/403) but due to eventual consistency edge cases, i think we want to retry 404.

With all ☝️ in mind, do you think we need to add any extra retry logic at this time? I suppose we could simply wait until we switch to Run svc.

takost: @TingluoHuang @ericsciple I rewrite a bit.

BTW, I had some other draft solution: https://github.com/actions/runner/pull/1938

src/Runner.Common/RunServer.cs
```diff
@@ -70,7 +70,43 @@ private void CheckConnection()
         public Task<AgentJobRequestMessage> GetJobMessageAsync(string id)
```
ericsciple, line 70: do we need to pass the cancellation token from the caller?

TingluoHuang: +1

src/Runner.Common/RunServer.cs
```diff
@@ -70,7 +70,43 @@ private void CheckConnection()
...
         {
             CheckConnection();
-            return _taskAgentClient.GetJobMessageAsync(id);
+            var jobMessage = RetryRequestWithTimeout<AgentJobRequestMessage>(async () =>
```
ericsciple, line 73: plan
- will leave the retries in this file
- pass the cancellation token (in case runner is shutting down; `messageQueueLoopTokenSource.Token` from runner.cs)
- let's switch to count-based retries (5 retries; keep the 5-15s backoff). Argument is purely about consistency. Same algorithm for reader bouncing between files. Apologies, this is my fault for the original time-based proposal :(
- will not move to a retry util for now, since would require caller to provide retriable. Today we're using taskAgentClient but tomorrow switching to ___.

takost: I have reworked retries to  count-based, and started to pass cancellation token

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,48 @@ private void CheckConnection()
...
+                                                    }, cancellationToken);
+            return jobMessage;
+        }
+
+        private async Task<T> RetryRequest<T>(Func<Task<T>> func,
+                                                        CancellationToken cancellationToken,
+                                                        int maxRetryAttempts = 5
+                                                        )
+        {
+            var currentRetryAttempt = 1;
```
fhammerl, line 85: NIT: consider renaming `currentRetryAttempt` to `retryCount`, it is the terminology most often used across the runner code.

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,48 @@ private void CheckConnection()
...
+                try
+                {
+                    return await func();
+                }
+                catch (Exception ex)
+                {
+                    Trace.Error("Catch exception during get full job message");
+                    Trace.Error(ex);
+
+                    if (currentRetryAttempt < maxRetryAttempts)
```
fhammerl, line 98: NIT: I tried doing the math: with `currentRetryAttempt=1` and `maxRetryAttempts=5`, this code executes a maximum of 5 times. Technically, that's 1 try and 4 retries.

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,48 @@ private void CheckConnection()
...
+        {
+            var currentRetryAttempt = 1;
+            while (true)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                try
+                {
+                    return await func();
+                }
+                catch (Exception ex)
```
fhammerl, line 93: We could catch if `retryCount / currentRetryAttempt` is not too high, otherwise let it bubble through.

`catch(Exception ex) when (retryCount < maxRetryAttempt)`

src/Runner.Common/RunServer.cs
takost: Yeeeaaah 😅 , actually my goal was to run code 5 times, so it is 1 try + 4 retries. But I might miss the point and we should have 1 try + 5 retries. What do you think?

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,42 @@ private void CheckConnection()
...
+            var retryCount = 0;
+            while (retryCount < maxRetryAttemptsCount)
+            {
+                retryCount++;
+                cancellationToken.ThrowIfCancellationRequested();
+                try
+                {
+                    return await func();
+                }
+                catch (Exception ex) when (retryCount < maxRetryAttemptsCount)
```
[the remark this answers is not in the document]
ericsciple: We don't lock on this today. Checking if we have a backlog item to resolve conflicting assignment...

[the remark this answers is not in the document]
ericsciple: Yikes, we were missing a tracking item. [Created now](https://github.com/github/actions-broker/issues/122), and added to the backlog of all potential items for "Next quarter"

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,48 @@ private void CheckConnection()
...
+            return jobMessage;
+        }
+
+        private async Task<T> RetryRequest<T>(Func<Task<T>> func,
+                                                        CancellationToken cancellationToken,
+                                                        int maxRetryAttempts = 5
+                                                        )
+        {
+            var currentRetryAttempt = 1;
+            while (true)
```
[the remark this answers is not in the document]
takost: I am not sure that it makes sense to duplicate the same condition twice:
1) `catch (Exception ex) when (retryCount < maxRetryAttemptsCount)`
2) `while (retryCount < maxRetryAttemptsCount)`

Actually, condition in `while` will never return false because the exception in `catch ... when ...` will be thrown earlier and having condition in two places can confuse so I prefer to keep the `catch ... when ...` approach + `while (true)`.
I have looked through some retries implementations in different places and actions-dotnet and it seems that `while (true)` approach is widely used.

Also if we put condition to `while`, C# compiler thinks that code after loop can be reached and force to write some `return` after loop. it makes code less readable because it is useless code lines that will never be reached.

src/Runner.Common/RunServer.cs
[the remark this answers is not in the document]
ericsciple: .sparkle @patrickcarnahan good insight wrt this type of non-retriable exception 😄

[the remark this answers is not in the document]
ericsciple: 401 and 403 are other non-retriable exceptions. When we switch from Actions Service to Run service, we will switch the underlying HTTP client and should handle those non-retriable here as well.

src/Runner.Common/RunServer.cs
```diff
@@ -67,10 +67,39 @@ private void CheckConnection()
...
+            return jobMessage;
+        }
+
+        private async Task<T> RetryRequest<T>(Func<Task<T>> func,
+                                              CancellationToken cancellationToken,
+                                              int maxRetryAttemptsCount = 5
+                                             )
+        {
+            var retryCount = 0;
+            while (true)
```
TingluoHuang, line 86:
```suggestion
            while (!cancellationToken.IsCancellationRequested)
```
?

takost: I thought about it, if we put condition to while, C# compiler thinks that code after loop can be reached and force to write some `return` or `throw` there. 
So in case when cancellation is requested before start of the loop, probably, we will have to either:
1) throw `OperationCanceledException`or any other exception manually
or
2) return some "empty" value and create additional condition for empty value in Runner function.

I'd prefer not to write the condition and create extra logic after the loop.

@TingluoHuang What do you think?

takost: Let's leave as it is ❤️

src/Runner.Common/RunServer.cs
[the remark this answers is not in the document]
takost: Added comment for future improvement to add handling of non-retriable exceptions

ericsciple (approved):

PR #1944, opened 2022-06-13
Added to lower on action_status to be consistent with the job.status

### LINKED ISSUE #1841: Composite action status is using mixed case not lower case only
**Describe the bug**

Using the composite action status in a string comparison, as documented [here](https://docs.github.com/en/actions/learn-github-actions/expressions#evaluate-status-explicitly), does not work as expected because it returns mixed case `Failure` instead of `failure`.

**Expected behavior**

Status should be consistent across both job and composite actions and should be using all lowercase, as documented, to avoid needing further string manipulation like with `toLowerCase()`.

Version of your runner? GHES 3.4.0; Runner 2.287.1

OS of the machine running the runner? Linux

## What's not working?

```js
${{ github.action_status == 'failure' }} # returns 'Failure'
```

nikola-jokic:
If we echo `job.status`, it will output `failure`.
But if we run echo `github.action_status` in a composite action, it would echo `Failure`.

This PR is just for consistency.

Fixes #1841


fhammerl (approved):
LGTM 
No change in behaviour ( checks like `{{ github.action_status == 'Failure'}}` were already equivalent `{{ github.action_status == 'failure'}}`

But printing with `echo ${{ github.action_status }}` will now print 'success' or 'failure', consistent with job status (`github.status`)

PR #1982, opened 2022-07-03
Fix URL typo

kichik:


kichik:
PR #1973 already fixes this.

Issue #1929, opened 2022-06-08
Ephemeral runner doing unnecessary cleanup


zhiwei-arbo:
**Describe the enhancement**
We are self-hosting Github runners, running them in ephemeral mode. This means the whole pod gets cleaned up after the job. However, we've noticed that the runner does a [CleanupTempDir](https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Runner.Worker/TempDirectoryManager.cs#L48) after the jobs, which takes significant amount of time for us (due to large number of files and size in our Bazel build) and seems to be redundant

Skipping the cleanup for Ephemeral steps might be possible here

**Code Snippet**
N/A

**Additional information**
N/A

nikola-jokic:
Hi @zhiwei-arbo,

Thank you for submitting this enhancement. I added it to the board, and it will be considered in the future :relaxed:

akanieski:
#2016  should address this if the team feels it fits.

aktech:
*Disclaimer: This doesn't answers the actual question, but suggests an alternative:*

You can achieve this easily with https://cirun.io/ It creates on demand **ephemeral** runners for GitHub Actions on your cloud and manages the complete lifecycle. You simply connect your cloud provider and define what runners you need in a simple yaml file and that's it.

See https://docs.cirun.io/reference/examples.html#aws for example.

PR #2016, opened 2022-07-22
Added configurable SkipTempDirectoryCleanup to improve ephemeral runner cleanup performance

akanieski:
As discussed in #1929, scenarios when runner is ephemeral I added `SkipTempDirectoryCleanup` to `RunnerSettings` to allow operators to avoid having to delete the temp directory, leaving the cleanup to the underlying host after ephemeral runner tears down (as is often the case with containerized runners).


surmabck:
@akanieski  Any update on this?

PR #2022, opened 2022-07-25
POC: Windows arm64 runner build

thboop:


src/Runner.Service/Windows/App.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <configuration>
-    <startup> 
-        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
+    <startup>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8" />
```
thboop, line 4: Needed for a supported arm version

src/dev.sh
```diff
@@ -43,6 +43,9 @@ if [[ "$CURRENT_PLATFORM" == 'windows' ]]; then
     if [[ "$PROCESSOR_ARCHITECTURE" == 'x86' ]]; then
         RUNTIME_ID='win-x86'
     fi
+    if [[ "$PROCESSOR_ARCHITECTURE" == 'ARM64' ]] || [[ "$PROCESSOR_ARCHITECTURE" == 'AMD64' ]]; then
```
thboop, line 46: AMD64 is emulating x64 on an AMD system. I'm happy to remove this if we want to be more explicit about building for arm64, but given git bash on windows 11 arm is currently x64 we would have to manually specify win-arm64 builds when building

src/dev.sh
```diff
@@ -309,7 +312,7 @@ if [[ (! -d "${DOTNETSDK_INSTALLDIR}") || (! -e "${DOTNETSDK_INSTALLDIR}/.${DOTN
         echo "Convert ${DOTNETSDK_INSTALLDIR} to Windows style path"
         sdkinstallwindow_path=${DOTNETSDK_INSTALLDIR:1}
         sdkinstallwindow_path=${sdkinstallwindow_path:0:1}:${sdkinstallwindow_path:1}
-        powershell -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "& \"./Misc/dotnet-install.ps1\" -Version ${DOTNETSDK_VERSION} -InstallDir \"${sdkinstallwindow_path}\" -NoPath; exit \$LastExitCode;" || checkRC dotnet-install.ps1
+        pwsh -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "& \"./Misc/dotnet-install.ps1\" -Version ${DOTNETSDK_VERSION} -InstallDir \"${sdkinstallwindow_path}\" -NoPath; exit \$LastExitCode;" || checkRC dotnet-install.ps1
```
thboop, line 315: todo, we need to fallback here or only do this if its windows-arm. We need to use a native arm powershell to build win-arm builds, other dotnet-install doesn't run as expected. 

We could also look into specifying the architecture rather than changing the shell, I think there's a supported configuration option for that.

src/Runner.Service/Windows/App.config
TingluoHuang, line 4: bump this means we are dropping support on the old windows version.

releaseNote.md
```diff
@@ -16,6 +16,22 @@ Add-Type -AssemblyName System.IO.Compression.FileSystem ;
 [System.IO.Compression.ZipFile]::ExtractToDirectory("$PWD\actions-runner-win-x64-<RUNNER_VERSION>.zip", "$PWD")
 ```
 
+## Windows arm64
```
TingluoHuang, line 19:
```suggestion
## [Pre-release] Windows arm64
```

src/Runner.Service/Windows/RunnerService.csproj
```diff
@@ -11,10 +11,15 @@
...
     <SignAssembly>false</SignAssembly>
     <DelaySign>false</DelaySign>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(PackageRuntime)' == 'win-arm64' ">
+      <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
+  </PropertyGroup>
+    <PropertyGroup Condition=" '$(PackageRuntime)' != 'win-arm64' ">
```
TingluoHuang, line 20:
```suggestion
  <PropertyGroup Condition=" '$(PackageRuntime)' != 'win-arm64' ">
```

src/Test/L0/Listener/SelfUpdaterL0.cs
```diff
@@ -1,4 +1,4 @@
-﻿#if !(OS_OSX && ARM64)
+﻿#if !((OS_OSX || OS_WINDOWS) && ARM64)
```
TingluoHuang, line 1:
```suggestion
﻿#if !(OS_WINDOWS && ARM64)
```

I think we were commenting out `OS_OSX and ARM64` in the initial release since it can't do any upgrade testing.

src/Test/Test.csproj
```diff
@@ -15,7 +15,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
```
TingluoHuang, line 18: is this a required change?

src/dev.sh
TingluoHuang, line 315: is `pwsh` installed by default on all Windows versions?
Do we need to `where pwsh` and fallback to `powershell`?

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -97,11 +97,11 @@ public async Task RunAsync(ActionRunStage stage)
                 workingDirectory = HostContext.GetDirectory(WellKnownDirectory.Work);
             }
 
-#if OS_OSX
+#if OS_OSX || OS_WINDOWS
             if (string.Equals(Data.NodeVersion, "node12", StringComparison.OrdinalIgnoreCase) &&
                 Constants.Runner.PlatformArchitecture.Equals(Constants.Architecture.Arm64))
             {
-                ExecutionContext.Output($"The node12 is not supported on macOS ARM64 platform. Use node16 instead.");
+                ExecutionContext.Output($"The node12 is not supported on the macOS or windows ARM64 platforms. Use node16 instead.");
```
TingluoHuang, line 104: should we have 2 messages for each platform?
it might be weird to see a message about windows when I am on a macOS device.

TingluoHuang, line 104:
```suggestion
#if OS_OSX
                ExecutionContext.Output($"The node12 is not supported on macOS ARM64 platform. Use node16 instead.");
#elif OS_WINDOWS
                ExecutionContext.Output($"The node12 is not supported on windows ARM64 platform. Use node16 instead.");
#endif
```

src/dev.sh
thboop: Done! Luckily Which is included in the bash utils so we can use it across linux/win/mac

src/Test/Test.csproj
thboop: Yeah, otherwise it doesn't have the appropriate win-arm files to load. My understanding is this isn't a breaking change.

If you revert the change and run test on a win-arm machine you get

```  
Starting test execution, please wait...
  A total of 1 test files matched the specified pattern.
EXEC : Testhost process exited with error : Unhandled exception. System.BadImageFormatException: Could not load file or
 assembly 'C:\runner\src\Test\bin\Release\net6.0\testhost.dll'. An attempt was made to load a prog
ram with an incorrect format. 
```

src/Runner.Listener/Program.cs
```diff
@@ -60,6 +60,17 @@ private async static Task<int> MainAsync(IHostContext context, string[] args)
                         terminal.WriteLine("This runner version is built for Windows. Please install a correct build for your OS.");
                         return Constants.Runner.ReturnCode.TerminatedError;
                     }
+                    #if ARM64
+                        // A little hacky, but windows gives no way to differentiate between windows 10 and 11.
+                        // By default only 11 supports native x64 app emulation on arm, so we only want to support windows 11
+                        // https://docs.microsoft.com/en-us/windows/arm/overview#build-windows-apps-that-run-on-arm
+                        // Windows 10 and 11 share a MajorVersion, so we also check the build version. Minor for both is 0, so doing < 0 doesn't really make a lot of sense.
+                        if (Environment.OSVersion.Version.Build < Constants.OperatingSystem.Windows11BuildVersion || Environment.OSVersion.Version.Major < Constants.OperatingSystem.Windows11MajorVersion )
```
TingluoHuang, line 68:
```suggestion
                        if (Environment.OSVersion.Version.Major < Constants.OperatingSystem.Windows11MajorVersion || 
                            Environment.OSVersion.Version.Build < Constants.OperatingSystem.Windows11BuildVersion)
```

src/dev.sh
```diff
@@ -309,7 +318,8 @@ if [[ (! -d "${DOTNETSDK_INSTALLDIR}") || (! -e "${DOTNETSDK_INSTALLDIR}/.${DOTN
         echo "Convert ${DOTNETSDK_INSTALLDIR} to Windows style path"
         sdkinstallwindow_path=${DOTNETSDK_INSTALLDIR:1}
         sdkinstallwindow_path=${sdkinstallwindow_path:0:1}:${sdkinstallwindow_path:1}
-        powershell -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "& \"./Misc/dotnet-install.ps1\" -Version ${DOTNETSDK_VERSION} -InstallDir \"${sdkinstallwindow_path}\" -NoPath; exit \$LastExitCode;" || checkRC dotnet-install.ps1
+        which pwsh
```
TingluoHuang, line 321:
```suggestion
```

src/dev.sh
```diff
@@ -28,6 +28,12 @@ RUNNER_VERSION=$(cat runnerversion)
 
 pushd "$SCRIPT_DIR"
 
+if which pwsh; then
+    POWERSHELL_VERSION="pwsh"
```
TingluoHuang, line 32:
```suggestion
    POWERSHELL="pwsh"
```

what about just calling it `POWERSHELL`?


`POWERSHELL_VERSION` sounds like a version `10.1.1`, etc.

TingluoHuang (approved):

PR #266, opened 2020-01-12
allow container to be null/empty

ericsciple:
fixes https://github.com/actions/runner/issues/265

Issue #265, opened 2020-01-12
Allow an empty value to be specified for jobs.<id>.container


ericsciple:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to build a matrix workflow where some jobs run in a container and some run directly on the VM.  I may have a cross-platform tool where I have strange dependencies that I provide via a container, or I want to run on multiple different Linux distributions, but also run on Windows and macOS.  So I would like a way to specify that I run on _no_ container.  This would allow me to create a workflow like:

```yaml
name: CI

on: [push]

jobs:    
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        image: ['ubuntu:latest', 'centos:latest', '']
```

A more realistic scenario would be a matrix of `runs-on` that includes linux, windows and macos, and then `include`ing the container value on linux hosts.

**Describe the solution you'd like**
That some value exists for `container` that means "none".  Perhaps: an empty string (`''`), a dash (`'-'`), or a dot (`'.'`), all of which are not legal image names.

ericsciple:
@ethomson fyi i moved the issue here

ericsciple:
Fixed on the runner (expression scenario). Still need to fix on the server (non-expression)

mnishamk-freiheit:
could this be done for jobs.<id>.container.image as well?

PR #2025, opened 2022-07-27
Allow container.image to be null/empty

moonblade:
For issue https://github.com/actions/runner/issues/265, A pull request https://github.com/actions/runner/pull/266/files was created which allows for container to be an empty string.

Since container (of type string) is an alias for container.image. It makes sense to allow container.image to have empty values as well.

Similar to the original pull request which was to allow jobs that run on public container, or no container within the same matrix, this would allow jobs that run on private container, or no container within the same matrix.

### DIFF
--- a/src/Sdk/DTPipelines/workflow-v1.0.json
+++ b/src/Sdk/DTPipelines/workflow-v1.0.json
@@ -369,7 +369,7 @@
     "container-mapping": {
       "mapping": {
         "properties": {
-          "image": "non-empty-string",
+          "image": "string",
           "options": "non-empty-string",
           "env": "container-env",
           "ports": "sequence-of-non-empty-string",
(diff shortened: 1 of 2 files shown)


webknjaz:
It'd be nice to have the same for services.

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -401,10 +401,23 @@
       ],
       "one-of": [
         "non-empty-string",
-        "container-mapping"
+        "services-container-mapping"
```
JoannaaKL, line 404:
```suggestion
        "container-mapping"
```

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -401,10 +401,23 @@
...
+          "image": "non-empty-string",
+          "options": "non-empty-string",
+          "env": "container-env",
+          "ports": "sequence-of-non-empty-string",
+          "volumes": "sequence-of-non-empty-string",
+          "credentials": "container-registry-credentials"
+        }
+      }
+    },
+
```
JoannaaKL, line 420: We don't need the section 'services-container-mapping', could you please remove it? :)

JoannaaKL (changes requested):

PR #2026, opened 2022-07-27
Include current runner status while getting messages

lokesh755:
**Context**

Current PR is to send an additional parameter that indicates the current runner status while getting messages. This will help to update the runner status in the broker.

When runner status is changed from idle to busy or busy to idle, it cancels the current request to get messages and creates a new request with the corresponding runner status.

Current changes are behind FF by using the environment variable `USE_BROKER_FLOW`

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -27,6 +27,7 @@ namespace GitHub.Runner.Listener
         bool Cancel(JobCancelMessage message);
         Task WaitAsync(CancellationToken token);
         Task ShutdownAsync();
+        event EventHandler<JobStatusEventArgs> JobStatus;
     }
 
     // This implementation of IJobDispatcher is not thread safe.
@@ -54,6 +55,8 @@ namespace GitHub.Runner.Listener
         private TimeSpan _channelTimeout;
 
         private TaskCompletionSource<bool> _runOnceJobCompleted = new TaskCompletionSource<bool>();
+
+        public event EventHandler<JobStatusEventArgs> JobStatus;
 
         public override void Initialize(IHostContext hostContext)
         {
@@ -335,6 +338,11 @@ namespace GitHub.Runner.Listener
             Busy = true;
             try
             {
+                if (JobStatus != null)
+                {
+                    JobStatus(this, new JobStatusEventArgs(TaskAgentStatus.Busy));
+                }
+
                 if (previousJobDispatch != null)
                 {
                     Trace.Verbose($"Make sure the previous job request {previousJobDispatch.JobId} has successfully finished on worker.");
@@ -650,6 +658,11 @@ namespace GitHub.Runner.Listener
             finally
             {
                 Busy = false;
+                
+                if (JobStatus != null)
+                {
+                    JobStatus(this, new JobStatusEventArgs(TaskAgentStatus.Online));
+                }
             }
         }
 
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -23,6 +23,7 @@ namespace GitHub.Runner.Listener
         Task DeleteSessionAsync();
         Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token);
         Task DeleteMessageAsync(TaskAgentMessage message);
+        void OnJobStatus(object sender, JobStatusEventArgs e);
     }
 
     public sealed class MessageListener : RunnerService, IMessageListener
@@ -38,6 +39,8 @@ namespace GitHub.Runner.Listener
         private readonly TimeSpan _sessionConflictRetryLimit = TimeSpan.FromMinutes(4);
         private readonly TimeSpan _clockSkewRetryLimit = TimeSpan.FromMinutes(30);
         private readonly Dictionary<string, int> _sessionCreationExceptionTracker = new Dictionary<string, int>();
+        private TaskAgentStatus runnerStatus = TaskAgentStatus.Online;
+        private CancellationTokenSource _getMessagesTokenSource;
 
         public override void Initialize(IHostContext hostContext)
         {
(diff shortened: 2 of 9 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -360,6 +360,9 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                     bool runOnceJobReceived = false;
                     jobDispatcher = HostContext.CreateService<IJobDispatcher>();
 
+                    jobDispatcher.JobStarted += _listener.OnJobStarted;
```
TingluoHuang, line 363: need to unsubscribe the event handler at some point. 😄

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -55,6 +57,11 @@ public sealed class JobDispatcher : RunnerService, IJobDispatcher
 
         private TaskCompletionSource<bool> _runOnceJobCompleted = new TaskCompletionSource<bool>();
 
+        public event EventHandler<EventArgs> JobStarted;
+        public event EventHandler<EventArgs> JobCompleted;
+
+        private ITerminal _term;
```
TingluoHuang, line 63: you might want to update line 363 to use the new `_term` you added, it was getting the terminal inline.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -55,6 +57,11 @@ public sealed class JobDispatcher : RunnerService, IJobDispatcher
 
         private TaskCompletionSource<bool> _runOnceJobCompleted = new TaskCompletionSource<bool>();
 
+        public event EventHandler<EventArgs> JobStarted;
```
TingluoHuang, line 60: What about we only have one `EventHandler<JobStatusEventArgs> JobStatus` and use the `EventArgs` to distinct JobStart vs JobComplete?

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -335,6 +344,12 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
             Busy = true;
             try
             {
+                //invoke JobStarted Event
+                if (JobStarted != null)
+                {
+                    JobStarted(this, new EventArgs());
```
TingluoHuang, line 350: do we want to feed in some job info for tracing? ex jobId/name etc.

src/Runner.Listener/MessageListener.cs
```diff
@@ -183,27 +199,52 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
 
-                    // Decrypt the message body if the session is using encryption
-                    message = DecryptMessage(message);
-
-                    if (message != null)
+                    if (completedTask == getMessageTask)
                     {
-                        _lastMessageId = message.MessageId;
-                    }
+                        if (getMessageTask.IsCanceled)
```
TingluoHuang, line 218:
```C#
if(!getMessageTask.IsCompletedSuccessfully)
{
    await getMessageTask;
}
```

TingluoHuang: or should you just do
```
if (completedTask == getMessageTask)
{
    message = await getMessageTask;
}
```

src/Runner.Listener/MessageListener.cs
```diff
@@ -183,27 +199,52 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
+                            encounteringError = false;
+                            continuousError = 0;
+                        }
+                    }
+                    else 
                     {
-                        _term.WriteLine($"{DateTime.UtcNow:u}: Runner reconnected.");
-                        encounteringError = false;
-                        continuousError = 0;
+                        getMessagesTokenSource.Cancel();
```
TingluoHuang, line 246: you might want to still `await getMessageTask;` and catch/trace all exceptions from it.

src/Runner.Listener/MessageListener.cs
```diff
@@ -170,6 +174,18 @@ public async Task DeleteSessionAsync()
             }
         }
 
+        public void OnJobStarted(object sender, EventArgs e)
+        {
+            runnerStatus = TaskAgentStatus.Busy;
```
TingluoHuang, line 179: you might want to add some tracing so we know that the status changed at what time.

src/Runner.Listener/MessageListener.cs
```diff
@@ -183,27 +199,52 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
             while (true)
             {
                 token.ThrowIfCancellationRequested();
+                jobStatusNotification = new TaskCompletionSource<bool>();
```
TingluoHuang, line 202: Do we need to have both `jobStatusNotification` and `getMessagesTokenSource`?
will only have a global `_getMessagesTokenSource` work? let the OnJobStarted() to call `_getMessagesTokenSource?.cancel();`

src/Runner.Listener/MessageListener.cs
```diff
@@ -183,27 +199,52 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
                 TaskAgentMessage message = null;
+                CancellationTokenSource getMessagesTokenSource = CancellationTokenSource.CreateLinkedTokenSource(token);
```
TingluoHuang, line 204: You might want to dispose every `CancellationTokenSource` at some point.

src/Runner.Listener/JobDispatcher.cs
lokesh755: Was going to remove that.. just added for my debugging.

bryanmacfarlane:
There's no description on the PR.  What issue is this fixing?

src/Runner.Listener/JobDispatcher.cs
lokesh755: I think it won't be necessary.

src/Runner.Sdk/Util/VssUtil.cs
```diff
@@ -57,7 +57,7 @@ public static VssConnection CreateConnection(Uri serverUri, VssCredentials crede
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
 
-            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_ALLOW_REDIRECT")))
```
TingluoHuang, line 60: why we need to change this?

src/Runner.Listener/MessageListener.cs
```diff
@@ -206,6 +228,11 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                         continuousError = 0;
                     }
                 }
+                catch (OperationCanceledException) when (_getMessagesTokenSource.Token.IsCancellationRequested)
```
TingluoHuang, line 231: `_getMessagesTokenSource.Token.IsCancellationRequested` will always be true when `token` is cancelled

I think you need to do
`_getMessagesTokenSource.Token.IsCancellationRequested && !token.IsCancellationRequested`

lokesh755: updated pr

TingluoHuang (approved):

PR #2029, opened 2022-07-28
Stefan ruvceski/bugfix/update script

ruvceskistefan:
Proof of work:

- update file: we are getting new log for creation succeed
Screenshot 2022-07-28 at 11 21 17

- run-helper: we are getting new console output after file is detected 
Screenshot 2022-07-28 at 11 20 35

### DIFF
--- a/src/Misc/layoutbin/update.sh.template
+++ b/src/Misc/layoutbin/update.sh.template
@@ -181,6 +181,8 @@ fi
 date "+[%F %T-%4N] Update succeed" >> "$logfile"
 
 touch update.finished
+date "+[%F %T-%4N] update.finished file creation succeed" >> "$logfile"
+
 # rename the update log file with %logfile%.succeed/.failed/succeedneedrestart
 # runner service host can base on the log file name determin the result of the runner update
 date "+[%F %T-%4N] Rename $logfile to be $logfile.succeed" >> "$logfile" 2>&1
--- a/src/Misc/layoutbin/update.sh.template
+++ b/src/Misc/layoutbin/update.sh.template
@@ -180,6 +180,7 @@ fi
 
 date "+[%F %T-%4N] Update succeed" >> "$logfile"
 
+touch update.finished
 # rename the update log file with %logfile%.succeed/.failed/succeedneedrestart
 # runner service host can base on the log file name determin the result of the runner update
 date "+[%F %T-%4N] Rename $logfile to be $logfile.succeed" >> "$logfile" 2>&1
--- a/src/Misc/layoutroot/run-helper.sh.template
+++ b/src/Misc/layoutroot/run-helper.sh.template
@@ -38,6 +38,7 @@ elif [[ $returnCode == 3 ]]; then
     for i in {0..30} do
         if test -f "$FILE"; then
             echo "Update finished successfully."
+            rm "$FILE"
             break
         fi
         "$DIR"/safe_sleep.sh 1
--- a/src/Misc/layoutroot/run-helper.sh.template
+++ b/src/Misc/layoutroot/run-helper.sh.template
@@ -31,9 +31,17 @@ elif [[ $returnCode == 2 ]]; then
     "$DIR"/safe_sleep.sh 5
     exit 2
 elif [[ $returnCode == 3 ]]; then
+    FILE=update.finished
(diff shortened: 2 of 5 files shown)


src/Misc/layoutroot/run-helper.sh.template
```diff
@@ -31,9 +31,18 @@ elif [[ $returnCode == 2 ]]; then
     "$DIR"/safe_sleep.sh 5
     exit 2
 elif [[ $returnCode == 3 ]]; then
-    # Sleep 5 seconds to wait for the runner update process finish
-    echo "Runner listener exit because of updating, re-launch runner in 5 seconds"
-    "$DIR"/safe_sleep.sh 5
+    # Wait for 30 seconds or for flag file to exists for the runner update process finish
+    echo "Runner listener exit because of updating, re-launch runner after successful update."
```
fhammerl, line 35:
``` suggestion
    echo "Runner listener exit because of updating, re-launch runner after successful update"
```

NIT: Remove '.'

src/Misc/layoutbin/update.sh.template
```diff
@@ -180,6 +180,9 @@ fi
 
 date "+[%F %T-%4N] Update succeed" >> "$logfile"
 
+touch update.finished
```
thboop, line 183: Should we make sure to best effort cleanup this file if it already exists?

fhammerl: `run-helper.sh` will always wait for `Runner.Listener` to finish before attempting a restart.

We should probably clean up update.finished in `Runner.Listener` right before it starts a process with the generated `_update.sh`. This way `update.finished` is cleaned up before `_update.sh` starts, so no way for the shell script to generate the update file, and before `Runner.Listener` stops so `run-helper.sh` for sure doesn't attempt a restart.

src/Runner.Listener/SelfUpdater.cs
```diff
@@ -131,6 +131,11 @@ public async Task<bool> SelfUpdate(AgentRefreshMessage updateMessage, IJobDispat
                 // For L0, we will skip execute update script.
                 if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("_GITHUB_ACTION_EXECUTE_UPDATE_SCRIPT")))
                 {
+                    string flagFile = "update.finished";
+                    if (File.Exists(flagFile))
```
fhammerl, line 135: `IOUtil.Delete` should be a little safer and you don't have to check with `File.Exists`

src/Misc/layoutroot/run-helper.sh.template
```diff
@@ -31,9 +31,18 @@ elif [[ $returnCode == 2 ]]; then
     "$DIR"/safe_sleep.sh 5
     exit 2
 elif [[ $returnCode == 3 ]]; then
```
TingluoHuang, line 33: do we need to do something to the `.cmd` version?

fhammerl: We considered it, but this issue is unlikely to come up with windows machines as it usually happens on containers with unreliable IO, and these containers are overwhelmingly not running windows.

src/Misc/layoutbin/update.sh.template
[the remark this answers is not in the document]
fhammerl: We considered it, but this issue is unlikely to come up with windows machines as it usually happens on containers with unreliable IO, and these containers are overwhelmingly not running windows.

src/Misc/layoutroot/run-helper.sh.template
```diff
@@ -31,9 +31,18 @@ elif [[ $returnCode == 2 ]]; then
...
-    "$DIR"/safe_sleep.sh 5
+    # Wait for 30 seconds or for flag file to exists for the runner update process finish
+    echo "Runner listener exit because of updating, re-launch runner after successful update"
+
+    FILE=update.finished
+    for i in {0..30} do
+        if test -f "$FILE"; then
+            echo "Update finished successfully."
+            rm "$FILE"
+            break
```
fhammerl, line 42:
```suggestion
            exit 2
```
NIT: ☝️

fhammerl:
Fixed in https://github.com/actions/runner/pull/2044

PR #2031, opened 2022-07-29
prepend ErrorView='NormalView' to pwsh and powershell content

### LINKED ISSUE #382: The default shell for Windows (Powershell) does not show the stderr/stdout on exit error
**Describe the bug**
For example, this build my which used the default shell for windows (powershell), did not showed the stdout or stderr of the error. It just says the command exited with some error code. Doing so, you have no ideia from which line the error comes from or what was happening: https://github.com/evandroforks/anki/runs/522520124?check_suite_focus=true#step:5:10
image

This problem does not happens if I change the shell to `cmd`: https://github.com/evandroforks/anki/runs/522428852?check_suite_focus=true#step:5:26
image

**Question, Bug, or Feature?**: 
"Bug"

**Virtual environments affected**
- [ ] macOS 10.15
- [ ] Ubuntu 16.04 LTS
- [ ] Ubuntu 18.04 LTS
- [x] Windows Server 2016 R2
- [x] Windows Server 2019

nikola-jokic:
Fixes #382 

Example workflow:
```yaml
name: PWSH err

on:
  workflow_dispatch:

jobs:
  test:
    runs-on: self-hosted
    steps:
    - name: fail here
      run: Get-ChildItem C:\NoDirectory
      shell: powershell
```
After execution displays:
image


fhammerl:
I ran this 3 times with different errorviews, and the errors look like this:

```pwsh
$ErrorView = "CategoryView" #NormalView or ConciseView
echo $ErrorView
Get-ChildItem -Path C:\nofile.txt
```

image

fhammerl:
We have decided against modifying the `powershell` default behaviour here. Alternative behaviours are straightforward to enerate by setting the `ErrorView` ENV to `NormalView` or `ConciseView`.

PR #2049, opened 2022-08-09
Display full job name and nested workflow details in log

nicholasbergesen:
This forms part of the work being done for the nested workflow feature. Jobs that are nested more than 1 level in depth have their job name truncated in the UI. The change displays the full job name as well as the calling workflow ref and inputs that were passed in, in the UI and raw logs 

Complete job name:
image

Calling workflow ref with inputs:
image
Expanded:
image

Raw logs:
image

actions-dotnet change: https://github.com/github/actions-dotnet/pull/12910

Closes: https://github.com/github/c2c-actions-policy/issues/580
Epic: https://github.com/github/c2c-actions-policy/issues/399

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -318,12 +318,13 @@ namespace GitHub.Runner.Worker
 
                     if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
                     {
-                        context.Output($"uses: {workflowFileFullPath.Value}");
-                        var inputs = message.ContextData["inputs"].AssertDictionary("inputs");
-                        if (inputs.Any()) {
-                            context.Output("with:");
-                            foreach(var input in inputs) {
-                                context.Output($"  {input.Key}: {input.Value}");
+                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData)) {
+                            var inputs = pipelineContextData.AssertDictionary("inputs");
+                            if (inputs.Any()) {
+                                context.Output("with:");
+                                foreach(var input in inputs) {
+                                    context.Output($"  {input.Key}: {input.Value}");
+                                }
                             }
                         }
                         
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -324,7 +324,6 @@ namespace GitHub.Runner.Worker
                             var inputs = pipelineContextData.AssertDictionary("inputs");
                             if (inputs.Any()) {
                                 context.Output($"##[group] inputs");
-                                context.Output("with:");
                                 foreach(var input in inputs) {
                                     context.Output($"  {input.Key}: {input.Value}");
                                 }
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -316,8 +316,9 @@ namespace GitHub.Runner.Worker
                         }
                     }
 
-                    if (!string.IsNullOrWhiteSpace(message.DataWorkflowFilePathRaw)) {
-                        context.Output($"uses: {message.DataWorkflowFilePathRaw}");
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
                         var inputs = message.ContextData["inputs"].AssertDictionary("inputs");
                         if (inputs.Any()) {
                             context.Output("with:");
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -322,16 +322,19 @@ namespace GitHub.Runner.Worker
                         if (message.ContextData.TryGetValue("inputs", out var pipelineContextData))
                         {
                             var inputs = pipelineContextData.AssertDictionary("inputs");
-                            if (inputs.Any()) {
+                            if (inputs.Any()) 
+                            {
                                 context.Output($"##[group] inputs");
-                                foreach(var input in inputs) {
+                                foreach(var input in inputs) 
+                                {
                                     context.Output($"  {input.Key}: {input.Value}");
                                 }
                                 context.Output("##[endgroup]");
                             }
                         }
 
-                        if (!string.IsNullOrWhiteSpace(message.JobDisplayName)) {
+                        if (!string.IsNullOrWhiteSpace(message.JobDisplayName)) 
+                        {
                             context.Output($"Complete job name: {message.JobDisplayName}");
                         }
                     }
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -326,10 +326,10 @@ namespace GitHub.Runner.Worker
                                 context.Output($"  {input.Key}: {input.Value}");
                             }
                         }
-                    }
-
-                    if (!string.IsNullOrWhiteSpace(message.JobDisplayName)) {
-                        context.Output($"Complete job name: {message.JobDisplayName}");
+                        
+                        if (!string.IsNullOrWhiteSpace(message.JobDisplayName)) {
+                            context.Output($"Complete job name: {message.JobDisplayName}");
+                        }
                     }
 
                     var intraActionStates = new Dictionary<Guid, Dictionary<string, string>>();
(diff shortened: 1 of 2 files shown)


src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -184,6 +184,20 @@ public WorkspaceOptions Workspace
             private set;
         }
 
+        [DataMember(EmitDefaultValue = false)]
```
yujincat, line 187: Maybe we need to update the initialization method `AgentJobRequestMessage` as well? I think this should match with the definition in Actions Service, so maybe it is not manually updated but somehow updated with dependency. I am not sure what is the best practice here.

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -184,6 +184,20 @@ public WorkspaceOptions Workspace
...
+        public string JobWorkflowRef
```
yujincat, line 188: We do not use JobWorkfloRef in your PR

yujincat:
In the example snapshot, I believe we used `./.github/workflows/end.yml` but it is adding the nwo. 
What does it look like when we refer the workflow with nwo and version? Does it show the version as well?

nicholasbergesen (commented):
remove unused JobWorkflowRef property

nicholasbergesen (commented):
add mapping to dataWorkflowFilePathRaw in constructor

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
nicholasbergesen: It's not necessary since this class is only initialized through deserialization but I've added the constructor mapping for completeness.

nicholasbergesen:
> In the example snapshot, I believe we used `./.github/workflows/end.yml` but it is adding the nwo. What does it look like when we refer the workflow with nwo and version? Does it show the version as well?

@yujincat if I use `github/public-server/.github/workflows/end.yml@main` in the workflow file it displays `github/public-server/.github/workflows/end.yml` in the log, with the nwo but without the version

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
nicholasbergesen: I've removed it since it changes the constructor signature and causes some tests to break as the constructor signature no longer matches.

TingluoHuang:
Why do we need to log this info to the `set up job` step?
We might need to have PM sign off on this kind of change since these are informational message that is not really scoped for how the runner set up the job.

yujincat:
@TingluoHuang We already have discussed this with @chrispat before, hence we are working on this. The reason we want to log this is we want to provide more context of workflow execution in the log so that the users can understand it; mostly reusable/nested workflows perspective.

Users may want to know what inputs they are given. I thin it makes sense to have that in the "set up job" section because the inputs apply per job. Also we want to show the text given from user in the `uses` field in the yaml file to provide more context of using reusable workflows for this job.

Providing the full job name is also what we want, as the job name using nested workflows will be truncated in the UI. 

@nicholasbergesen I think we should check if we want to make this change for all workflows or reusable case only.

nicholasbergesen:
I've moved the display of complete job name inside the check on `system.workflowFileFullPath` so it will only display in the log for reusable workflows.

nicholasbergesen:
@TingluoHuang Do you think it would be better to create a new section? Something along the lines of `Job Trigger` or `Job Source` for displaying details relating to how the job was called?

chrispat:
I think this is good. We already display a bunch of information in that area that has nothing to do with the runner such as secret source and token permissions.

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,22 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
+                        var inputs = message.ContextData["inputs"].AssertDictionary("inputs");
```
yujincat, line 322: I guess it works without exception even though there is no entry with key "data" in the ContextData. Maybe cause I am c# newbie? Can you tell me how this can work?

nicholasbergesen: Good catch, it appears the inputs key is being populated as a result of `PopulateWorkflowPhaseInputs`. I'll add a null check in.

yujincat (approved):

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,23 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData)) {
```
TingluoHuang, line 321: do we want to put these into a `##[group]` so it can collapse in case there are lots of inputs?

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,26 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
                     }
 
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
+                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData)) {
+                            var inputs = pipelineContextData.AssertDictionary("inputs");
+                            if (inputs.Any()) {
+                                context.Output($"##[group] inputs");
+                                context.Output("with:");
```
ajaykn, line 326:
```suggestion
```
in case of workflow dispatch we have inputs from dispatch inputs too.

So i think its better we just mention inputs and remove `with`, so that it wont be confusing for customers ?

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,26 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
+                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData)) {
```
ajaykn, line 322:
```suggestion
                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData))
                        {
```

ajaykn: please check other if's as well

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,29 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
+                        if (message.ContextData.TryGetValue("inputs", out var pipelineContextData))
+                        {
+                            var inputs = pipelineContextData.AssertDictionary("inputs");
+                            if (inputs.Any()) 
+                            {
+                                context.Output($"##[group] inputs");
+                                foreach(var input in inputs)
```
ajaykn, line 328:
```suggestion
                                foreach (var input in inputs) 
```

ajaykn (approved):
Looks good to me 🚀

ajaykn (dismissed):

src/Runner.Worker/JobExtension.cs
```diff
@@ -316,6 +316,29 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                         }
                     }
 
+                    if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
+                    {
+                        context.Output($"uses: {workflowFileFullPath.Value}");
```
TingluoHuang, line 321:
```suggestion
                        context.Output($"Uses: {workflowFileFullPath.Value}");
```

Since all other lines are upper case first letters.

TingluoHuang (dismissed):

TingluoHuang (approved):

PR #2050, opened 2022-08-09
Fix broken run-helper update syntax and run-helper not updating on restart

fhammerl:


src/Misc/layoutroot/run-helper.sh.template
```diff
@@ -35,7 +35,7 @@ elif [[ $returnCode == 2 ]]; then
 elif [[ $returnCode == 3 ]]; then
     # Wait for 30 seconds or for flag file to exists for the runner update process finish
     echo "Runner listener exit because of updating, re-launch runner after successful update"
-    for i in {0..30} do
+    for i in {0..30}; do
```
fhammerl, line 38: ^ this change was only made in `_layout/run-helper.sh`, not in the source so it wasn't checked in

src/Misc/layoutroot/run.sh
```diff
@@ -9,10 +9,10 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
     [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-cp -f "$DIR"/run-helper.sh.template "$DIR"/run-helper.sh
 # run the helper process which keep the listener alive
 while :;
 do
+    cp -f "$DIR"/run-helper.sh.template "$DIR"/run-helper.sh
```
fhammerl, line 15: This is why the canary tests weren't signalling even though run-helper had the above error: run.sh would use the old version of run-helper.sh until `run.sh` itself was restarted, because the copying of the new `run-helper.sh` was outside the runner restart loop.

When the tests were set up, `run.sh` itself would get restarted once so this went undetected.

This is bash only, the .cmd loop is correct.

TingluoHuang (approved):

PR #2069, opened 2022-08-19
Include step context name and start/finish time in step telemetry

TingluoHuang:
We want to have a human-readable string to distinct steps using the same action within a job.


thboop (approved):
lgtm

densto88 (approved):

PR #2093, opened 2022-09-01
Add warning for job stalls

vanZeben:
This solves a problem where stalled jobs end up timing out after a certain period of time and there is no useful information shown in the logs as to what happened. This wont solve whatever process is actually stalled, but will let the user know that something in their workflow (at that specific step) is stalled out and we havent received any output from the process.

Example inline within job output:
Example inline within job output

Summary annotation:
Screen Shot 2022-09-01 at 12 45 40 AM

This can be enabled using the feature flag `DistributedTask.AllowRunnerStallDetect`

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -321,13 +321,15 @@ namespace GitHub.Runner.Worker.Handlers
             ExecutionContext.Debug($"{fileName} {arguments}");
 
             Inputs.TryGetValue("standardInInput", out var standardInInput);
-            using (var stdoutManager = new OutputManager(ExecutionContext, ActionCommandManager))
-            using (var stderrManager = new OutputManager(ExecutionContext, ActionCommandManager))
+            StallManager stallManager = FeatureManager.IsStallDetectEnabled(ExecutionContext.Global.Variables) ? new StallManager(ExecutionContext) : null;
+            using (OutputManager stdoutManager = new OutputManager(ExecutionContext, ActionCommandManager, null, stallManager),
+                                 stderrManager = new OutputManager(ExecutionContext, ActionCommandManager, null, stallManager))
             {
                 StepHost.OutputDataReceived += stdoutManager.OnDataReceived;
                 StepHost.ErrorDataReceived += stderrManager.OnDataReceived;
 
                 // Execute
+                stallManager?.Initialize();
                 int exitCode = await StepHost.ExecuteAsync(ExecutionContext,
                                             workingDirectory: StepHost.ResolvePathForStepHost(ExecutionContext, workingDirectory),
                                             fileName: fileName,
--- a/src/Runner.Worker/Handlers/StallManager.cs
+++ b/src/Runner.Worker/Handlers/StallManager.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Timers;
+using GitHub.Runner.Common;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Worker.Handlers
+{
+    [ServiceLocator(Default = typeof(TimerAdapter))]
+    public interface ITimer
+    {
+        void Start();
+        void Stop();
+        double Interval { get; set; }
+        event ElapsedEventHandler Elapsed;
+        bool AutoReset { get; set; }
+        void Dispose();
+    }
+
+    public class TimerAdapter : Timer, ITimer { }
+
+    public sealed class StallManager : IDisposable
+    {
+        public static TimeSpan DefaultStallInterval = TimeSpan.FromMinutes(30);
+
+        private readonly IExecutionContext _executionContext;
+        private readonly double _interval;
+
+        private ITimer _timer { get; set; }
+        private int _intervalsElapsedWhileStalled = 0;
+
+        public StallManager(IExecutionContext executionContext, double interval, ITimer timer)
+        {
+            _executionContext = executionContext;
+            _interval = interval;
+            _timer = timer;
+
+            _timer.Interval = _interval;
+            _timer.Elapsed += TriggerWarning;
+        }
+        public StallManager(IExecutionContext executionContext, double interval) : this(executionContext, interval, new TimerAdapter()) { }
+        public StallManager(IExecutionContext executionContext) : this(executionContext, StallManager.DefaultStallInterval.TotalMilliseconds) { }
+
+        public void Initialize()
+        {
+            this.OnDataReceived(null, null);
+        }
+
+        public void Dispose()
+        {
+            try
+            {
+                _timer.Dispose();
+            }
+            catch { }
+        }
+
+        public void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
+        {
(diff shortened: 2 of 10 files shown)


src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -321,12 +321,19 @@ public async Task RunAsync(ActionRunStage stage)
             ExecutionContext.Debug($"{fileName} {arguments}");
 
             Inputs.TryGetValue("standardInInput", out var standardInInput);
-            using (var stdoutManager = new OutputManager(ExecutionContext, ActionCommandManager))
```
TingluoHuang, line 324: I think we need to do this for all action invocation, not only for `run` script.

src/Runner.Worker/Handlers/ScriptStalledManager.cs
```diff
@@ -0,0 +1,64 @@
...
+            IntervalsElapsedWhileStalled = 0;
+            _timer.Stop();
+            _timer.Start();
+        }
+
+        private void TriggerWarning(object source, ElapsedEventArgs e)
+        {
+            IntervalsElapsedWhileStalled++;
+            string friendlyDuration = StringUtil.TimespanToHumanReadableString(TimeSpan.FromMilliseconds(IntervalsElapsedWhileStalled * _interval));
+            _executionContext.Warning($"No output has been detected in the last {friendlyDuration} and the process has not yet exited. This step may have stalled and might require some investigation.");
```
TingluoHuang, line 61: we might want to consult with PM to see what other information we should/can provide to make this more actionable to our customers.

src/Runner.Worker/Handlers/ScriptHandler.cs
vanZeben: Do all step invocations go through the `OutputManager`? Perhaps a cleaner approach for that would be to tie into that class (since it already does the event registration) with the `StalledManager` as a singleton instance. What are your thoughts on that?

thboop: I believe they do, but we can check the three types
- nodejs actions
- docker actions
- run script steps

src/Runner.Sdk/Util/StringUtil.cs
```diff
@@ -127,5 +127,23 @@ public static string SubstringPrefix(string value, int count)
         {
             return value?.Substring(0, Math.Min(value.Length, count));
         }
+
+        public static string TimespanToHumanReadableString(TimeSpan duration)
+        {
+            if (duration.Duration().TotalSeconds == 0) {
```
thboop, line 133: nit: I like the idea of this function, but I'm really not sure how often we will use it and really for this pr its only being used to say `last 10 minutes`, should we consider a constant here instead and revisit this if there are more places we want to use it?

src/Runner.Worker/Handlers/ScriptHandler.cs
```diff
@@ -321,12 +321,19 @@ public async Task RunAsync(ActionRunStage stage)
...
-            using (var stderrManager = new OutputManager(ExecutionContext, ActionCommandManager))
+            using (ScriptStalledManager stalledManager = new ScriptStalledManager(ExecutionContext, TimeSpan.FromMinutes(10).TotalMilliseconds))
```
thboop, line 324: nit: IMO double this to 20 or 30. Jobs timeout by default after 6 hours for hosted machines, we should optimize to limit false positives, not responsiveness for users as most users will just see the annotations/build logs after the fact and won't be monitoring in real time.

src/Runner.Sdk/Util/StringUtil.cs
vanZeben: Thanks for the feedback! I can understand that and was kinda thinking the same as I was writing it. I opted to include it as it's testable and gives a much cleaner user experience since you can pass any `TimeSpan` to this, will automatically handle different intervals, so the warning message doesn't just have to say `last x minutes` and can break up into `last 2 hours, 40 minutes` or whatever. 

I can totally understand the rational though and it's easy enough to add back in if we decide to use something like that in the future so Ill get rid of it for now!

src/Runner.Worker/Handlers/ScriptHandler.cs
vanZeben: It seems they all do. Ill rejig it a bit

TingluoHuang: Do we want to start with some doc or ADR to help drive the design conversation?

src/Runner.Worker/Handlers/ScriptHandler.cs
vanZeben: Upped it to 30 minutes in [cda3bfc](https://github.com/actions/runner/pull/2093/commits/cda3bfcd90cdd1863e726e2fc23b94542da36a56) :)

src/Runner.Sdk/Util/StringUtil.cs
vanZeben: Got rid of the function and the helper function in favour or a more simple timespan. Still need to sync with product about the wording

src/Runner.Worker/Handlers/StallManager.cs
```diff
@@ -0,0 +1,70 @@
...
+        void Stop();
+        double Interval { get; set; }
+        event ElapsedEventHandler Elapsed;
+        bool AutoReset { get; set; }
+        void Dispose();
+    }
+
+    public class TimerAdapter : Timer, ITimer {}
+
+    public sealed class StallManager: IDisposable
```
TingluoHuang, line 23: Should we push all this logic into the processInvoker at the bottom layer, and any caller just needs to say `processInvoker.ExecuateAsync(x,x,x,x, stallDetect: true)`

src/Runner.Worker/Handlers/StallManager.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            _intervalsElapsedWhileStalled = 0;
+            _timer.Stop();
+            _timer.Start();
+        }
+
+        private void TriggerWarning(object source, ElapsedEventArgs e)
+        {
+            _intervalsElapsedWhileStalled++;
+            _executionContext.Warning($"No output has been detected in the last {TimeSpan.FromMilliseconds(_intervalsElapsedWhileStalled * _interval).TotalMinutes} minutes and the process has not yet exited. This step may have stalled and might require some investigation.");
```
TingluoHuang, line 67: do we want to add a bunch of warnings or regular output would be enough?

vanZeben: This adds a bunch of output every configured interval. This is beneficial as the issue is that the process hangs for unknown reasons without any output, so letting them know multiple times that the runner is still trying to process something, but the end process isnt responding is helpful to help them narrow down the issue

src/Runner.Worker/Handlers/StallManager.cs
vanZeben: I think this is valuable to do, but would like to do it in another PR as this has been stale for awhile and it's a bunch of reconfiguring

PR #2096, opened 2022-09-02
Adding a new vars context for non-secret variables

tauhid621:
Adding a new `vars` context for non-secret action configuration variables.

This has been added to support plain-text variables in GitHub Actions.

Configuration variables will be similar to secrets and will be used to store non-confidential data.

Actions service will add the new `vars` context as part of the `contextData` of `AgentJobRequestMessage`

More details on the [ADR](https://github.com/github/c2c-actions/pull/5372)

### DIFF
--- a/src/Runner.Worker/Variables.cs
+++ b/src/Runner.Worker/Variables.cs
@@ -171,9 +171,11 @@ namespace GitHub.Runner.Worker
             var result = new DictionaryContextData();
             foreach (var variable in _variables.Values)
             {
-                if (!variable.Secret)
+                // Only add the configuration variables to the vars context.
+                if (!variable.Secret && variable.Name.StartsWith(Constants.Variables.ConfigurationVariablePrefix, StringComparison.OrdinalIgnoreCase))
                 {
-                    result[variable.Name] = new StringContextData(variable.Value);
+                    var name = variable.Name.Replace(Constants.Variables.ConfigurationVariablePrefix, string.Empty);
+                    result[name] = new StringContextData(variable.Value);
                 }
             }
             return result;
--- a/src/Runner.Worker/Variables.cs
+++ b/src/Runner.Worker/Variables.cs
@@ -165,6 +165,19 @@ namespace GitHub.Runner.Worker
             }
             return result;
         }
+
+        public DictionaryContextData ToVarsContext()
+        {
+            var result = new DictionaryContextData();
+            foreach (var variable in _variables.Values)
+            {
+                if (!variable.Secret)
+                {
+                    result[variable.Name] = new StringContextData(variable.Value);
+                }
+            }
+            return result;
+        }
     }
 
     public sealed class Variable
--- a/src/Runner.Worker/Variables.cs
+++ b/src/Runner.Worker/Variables.cs
@@ -165,21 +165,6 @@ namespace GitHub.Runner.Worker
             }
             return result;
         }
-
-        public DictionaryContextData ToVarsContext()
-        {
-            var result = new DictionaryContextData();
-            foreach (var variable in _variables.Values)
-            {
-                // Only add the configuration variables to the vars context.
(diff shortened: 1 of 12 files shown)


TingluoHuang (commented):
Can you run a workflow and dump the whole `vars` context so we can see what it looks like?

tauhid621:
@TingluoHuang 
**Initial dump** : All the non-secret variables are added to `vars` context.
Screenshot 2022-09-15 at 12 44 54 PM

Made changes in local to only add variables with the prefix `CONFIGURATION_VARIABLE.` Actions Service will add this prefix and runner will add those variables to `vars` context.
Screenshot 2022-09-15 at 1 43 11 PM

Action's service [PR](https://github.com/github/actions-dotnet/pull/13170) will be updated to add the prefix.

src/Runner.Worker/Variables.cs
```diff
@@ -165,6 +165,21 @@ public DictionaryContextData ToSecretsContext()
...
+
+        public DictionaryContextData ToVarsContext()
+        {
+            var result = new DictionaryContextData();
+            foreach (var variable in _variables.Values)
+            {
+                // Only add the configuration variables to the vars context.
+                if (!variable.Secret && variable.Name.StartsWith(Constants.Variables.ConfigurationVariablePrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    var name = variable.Name.Replace(Constants.Variables.ConfigurationVariablePrefix, string.Empty);
```
nikola-jokic, line 177:
```suggestion
                    var name = variable.Name.Substring(Constants.Variables.ConfigurationVariablePrefix.Length);
```
This might be safer since Replace will replace all occurrences, and we really want to trim the prefix, right?

src/Runner.Worker/Variables.cs
```diff
@@ -165,6 +165,21 @@ public DictionaryContextData ToSecretsContext()
             }
             return result;
         }
+
+        public DictionaryContextData ToVarsContext()
+        {
+            var result = new DictionaryContextData();
+            foreach (var variable in _variables.Values)
```
TingluoHuang, line 172: I am wondering whether we should not build this new feature on top of the `Variables` object.
The `Variables` is a legacy thing only used for sending secrets and system variables.

tauhid621: We used the `Variables` object as it used to handle both secrets and plain-text variables in case of AzureDevops. 

I am not sure what it would men to introduce a new field like `ConfigVariables` in runner and the Actions Service. I believe the variables are right now stored as part of the plan.

tauhid621: I tried to pass `vars` as part of the context object. The context was passed and not explicit handling was required.

src/Runner.Worker/Variables.cs
tauhid621: Removed the prefix logic. We will be passing the variables as part of the context object.

src/Runner.Worker/Variables.cs
tauhid621: @TingluoHuang Both Action's service and runner PRs are updated to use the context to pass `vars`.
Could you please review again.

src/Runner.Worker/action_yaml.json
```diff
@@ -180,6 +181,7 @@
                 "matrix",
                 "job",
                 "runner",
+                "vars",
```
TingluoHuang, line 184: why we need `vars` here?

TingluoHuang:
I will approve this after approving the server-side PR. 🙇

src/Runner.Worker/action_yaml.json
ajaykn: +1 we dont need in actions yaml file, since we support `vars` only in workflow file.

tauhid621: Thanks for mentioning this. I'll remove the changes from this file.

tauhid621:
Given that the server side [PR](https://github.com/github/actions-dotnet/pull/13170) is merged.
@TingluoHuang Could you review this again?

TingluoHuang (approved):

Issue #341, opened 2020-02-21
Show docker logs when service fails to start


dakale:
If a service fails to start (healthcheck becomes `unhealthy`) we could print the `docker logs`. Without being able to see logs, the job just fails and it is hard to debug

cc @jclem

TrejGun:
must feature

JoannaaKL:
Hi @dakale is this ticket about adding more explicit error message in the 'Initialise containers' step or the case where logs are not visible at all? 
We noticed that all docker logs are being printed in 'Stop containers' step (under 1 and 2 points):

![image](https://user-images.githubusercontent.com/67866556/188593131-c9a01dbd-c71e-4671-a7d8-4986cd6e2014.png)

Ps. This screenshot is from our work-in-progress branch but 'Stop containers' section was not modified. :)

PR #2110, opened 2022-09-07
Avastancu/joannaakl/service container error log

JoannaaKL:
Coauthor @AvaStancu 
Fixing issue [Issue 341](https://github.com/actions/runner/issues/341).
Adding subsection in 'Start Containers' for each docker service failure. Services that fail to start will not be logged in 'Stop Containers' section so we don't have duplicated logs. 

If the service doesn't define the healthcheck method, we will check the container's exit code. The user will be able to see the exit code in the console, which might also add some additional context:
image

### DIFF
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -16,6 +16,7 @@ namespace GitHub.Runner.Worker.Container
     {
         string DockerPath { get; }
         string DockerInstanceLabel { get; }
+        IList<ContainerInfo> UnhealthyContainers {get; set; }
         Task<DockerVersion> DockerVersion(IExecutionContext context);
         Task<int> DockerPull(IExecutionContext context, string image);
         Task<int> DockerPull(IExecutionContext context, string image, string configFileDirectory);
@@ -41,6 +42,8 @@ namespace GitHub.Runner.Worker.Container
         public string DockerPath { get; private set; }
 
         public string DockerInstanceLabel { get; private set; }
+
+        public IList<ContainerInfo> UnhealthyContainers {get; set; }
 
         public override void Initialize(IHostContext hostContext)
         {
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -351,7 +351,6 @@ namespace GitHub.Runner.Worker.Container
         {
             return await ExecuteDockerCommandAsync(context, "inspect", $"{options} {dockerObject}");
         }
-        
 
         public async Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId)
         {
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -32,6 +32,7 @@ namespace GitHub.Runner.Worker.Container
         Task<int> DockerExec(IExecutionContext context, string containerId, string options, string command);
         Task<int> DockerExec(IExecutionContext context, string containerId, string options, string command, List<string> outputs);
         Task<List<string>> DockerInspect(IExecutionContext context, string dockerObject, string options);
+        Task<List<string>> DockerInspectLogs(IExecutionContext context, string dockerContainerId);
         Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId);
         Task<int> DockerLogin(IExecutionContext context, string configFileDirectory, string registry, string username, string password);
     }
@@ -350,6 +351,11 @@ namespace GitHub.Runner.Worker.Container
         public async Task<List<string>> DockerInspect(IExecutionContext context, string dockerObject, string options)
         {
             return await ExecuteDockerCommandAsync(context, "inspect", $"{options} {dockerObject}");
+        }
+
+        public async Task<List<string>> DockerInspectLogs(IExecutionContext context, string dockerContainerId)
+        {
+            return await ExecuteDockerCommandAsync(context, "logs", $"{dockerContainerId}");
         }
 
         public async Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId)
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -32,7 +32,6 @@ namespace GitHub.Runner.Worker.Container
         Task<int> DockerExec(IExecutionContext context, string containerId, string options, string command);
         Task<int> DockerExec(IExecutionContext context, string containerId, string options, string command, List<string> outputs);
         Task<List<string>> DockerInspect(IExecutionContext context, string dockerObject, string options);
-        Task<List<string>> DockerInspectLogs(IExecutionContext context, string dockerContainerId);
         Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId);
         Task<int> DockerLogin(IExecutionContext context, string configFileDirectory, string registry, string username, string password);
     }
@@ -352,11 +351,7 @@ namespace GitHub.Runner.Worker.Container
         {
             return await ExecuteDockerCommandAsync(context, "inspect", $"{options} {dockerObject}");
         }
-
-        public async Task<List<string>> DockerInspectLogs(IExecutionContext context, string dockerContainerId)
-        {
-            return await ExecuteDockerCommandAsync(context, "logs", $"{dockerContainerId}");
-        }
+        
 
         public async Task<List<PortMapping>> DockerPort(IExecutionContext context, string containerId)
         {
--- a/src/Runner.Worker/Container/DockerCommandManager.cs
+++ b/src/Runner.Worker/Container/DockerCommandManager.cs
@@ -16,7 +16,6 @@ namespace GitHub.Runner.Worker.Container
     {
         string DockerPath { get; }
         string DockerInstanceLabel { get; }
-        IList<ContainerInfo> UnhealthyContainers {get; set; }
         Task<DockerVersion> DockerVersion(IExecutionContext context);
         Task<int> DockerPull(IExecutionContext context, string image);
         Task<int> DockerPull(IExecutionContext context, string image, string configFileDirectory);
@@ -42,8 +41,6 @@ namespace GitHub.Runner.Worker.Container
         public string DockerPath { get; private set; }
 
         public string DockerInstanceLabel { get; private set; }
-
-        public IList<ContainerInfo> UnhealthyContainers {get; set; }
 
         public override void Initialize(IHostContext hostContext)
         {
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -425,7 +425,7 @@ namespace GitHub.Runner.Worker
             }
             else
             {
-                executionContext.Output($"##[group]Docker logs for container id {container.ContainerId}");
+                executionContext.Output($"##[group] Container {container.ContainerImage} failed healthchecks, printing logs:");
                 await _dockerManager.DockerInspectLogs(context: executionContext, dockerContainerId: container.ContainerId);
                 executionContext.Output("##[endgroup]");
             }
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -108,7 +108,7 @@ namespace GitHub.Runner.Worker
             var unhealthyContainers = new List<ContainerInfo>();
             foreach (var container in containers.Where(c => !c.IsJobContainer))
             {
-                var healthcheck = await Healthcheck(executionContext, container);
+                var healthcheck = await ContainerHealthcheck(executionContext, container);
 
                 if (!string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
                 {
@@ -328,7 +328,7 @@ namespace GitHub.Runner.Worker
             {
                 if (!container.IsJobContainer && !container.FailedInitialization)
                 {
-                    var healthcheck = await Healthcheck(executionContext, container);
+                    var healthcheck = await ContainerHealthcheck(executionContext, container);
                     if (string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
                     {
                         // Print logs for service container jobs (not the "action" job itself b/c that's already logged).
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -426,8 +426,7 @@ namespace GitHub.Runner.Worker
             else
             {
                 executionContext.Output($"##[group]Docker logs for container id {container.ContainerId}");
-                List<string> dockerLogs = await _dockerManager.DockerInspectLogs(context: executionContext, dockerContainerId: container.ContainerId);
-                dockerLogs.ForEach(log => executionContext.Output(log));
+                await _dockerManager.DockerInspectLogs(context: executionContext, dockerContainerId: container.ContainerId);
                 executionContext.Output("##[endgroup]");
             }
         }
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -330,11 +330,6 @@ namespace GitHub.Runner.Worker
             {
                 if (!container.IsJobContainer && !container.FailedInitialization)
                 {
-                    var healthcheck = await ContainerHealthcheck(executionContext, container);
-                    if (string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
-                    {
-                        // Print logs for service container jobs (not the "action" job itself b/c that's already logged).
-                        // Print them only if the service was healthy, else they were already logged in the initialize containers section.
                         executionContext.Output($"Print service container logs: {container.ContainerDisplayName}");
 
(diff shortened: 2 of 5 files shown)


src/Runner.Worker/Container/DockerCommandManager.cs
```diff
@@ -351,6 +351,7 @@ public async Task<List<string>> DockerInspect(IExecutionContext context, string
         {
             return await ExecuteDockerCommandAsync(context, "inspect", $"{options} {dockerObject}");
         }
+
```
TingluoHuang, line 354:
```suggestion
```

src/Test/L0/TestHostContext.cs
```diff
@@ -12,7 +12,7 @@
 using GitHub.DistributedTask.Logging;
 using System.Net.Http.Headers;
 using GitHub.Runner.Sdk;
-
+using GitHub.Runner.Worker.Container;
```
TingluoHuang, line 15: do we need to have this?

src/Test/L0/TestHostContext.cs
```diff
@@ -26,6 +26,7 @@ public sealed class TestHostContext : IHostContext, IDisposable
         private string _suiteName;
         private string _testName;
         private Tracing _trace;
+
```
TingluoHuang, line 29:
```suggestion
```

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -98,12 +98,12 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
                 await StartContainerAsync(executionContext, container);
             }
 
-            executionContext.Output("##[group]Waiting for all services to be ready");
```
TingluoHuang, line 101: Without this `##[group]`, the log might be flooded with a bunch of `xxxx service is healthy.` in the health check pass scenario.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -301,14 +301,18 @@ private async Task StopContainerAsync(IExecutionContext executionContext, Contai
...
-                    // Print logs for service container jobs (not the "action" job itself b/c that's already logged).
-                    executionContext.Output($"Print service container logs: {container.ContainerDisplayName}");
-
-                    int logsExitCode = await _dockerManager.DockerLogs(executionContext, container.ContainerId);
-                    if (logsExitCode != 0)
-                    {
-                        executionContext.Warning($"Docker logs fail with exit code {logsExitCode}");
-                    }
+                    var healthcheck = await Healthcheck(executionContext, container);
+                    if (string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase)){
```
TingluoHuang, line 305: What if the service container becomes unhealthy in the middle of the job? I think we still want to get the log for them.

src/Test/L0/TestHostContext.cs
JoannaaKL: fixed

thboop (commented):
some thoughts!

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -99,11 +100,41 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
...
+                {
+                    executionContext.Output($"##[group]Service container {container.ContainerNetworkAlias} failed.");
+                    await ContainerErrorLogs(executionContext, container);
+                    executionContext.Output("##[endgroup]"); 
+                }
+                throw new InvalidOperationException("One or more containers failed to start.");
+            }         
+        }
+
+        public void printHello()
```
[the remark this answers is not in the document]
JoannaaKL: Me + Copilot ....

JoannaaKL:
Thank you @thboop for your feedback! I have applied your suggestions and refactored a whole block of code performing the containers healthcheck & logging into separate method. This way unit testing is easier and we don't have to change the visibility of other private methods.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -89,6 +89,7 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
             executionContext.Output("##[group]Create local container network");
             var containerNetwork = $"github_network_{Guid.NewGuid().ToString("N")}";
             await CreateContainerNetworkAsync(executionContext, containerNetwork);
+
```
TingluoHuang, line 92:
```suggestion
```

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -413,14 +446,14 @@ private async Task ContainerHealthcheck(IExecutionContext executionContext, Cont
...
-                executionContext.Output($"{container.ContainerNetworkAlias} service is healthy.");
-            }
-            else
-            {
-                throw new InvalidOperationException($"Failed to initialize, {container.ContainerNetworkAlias} service is {serviceHealth}.");
-            }
+            return serviceHealth;
+        }
+
+        private async Task ContainerErrorLogs(IExecutionContext executionContext, ContainerInfo container)
```
TingluoHuang, line 452: do we need a method for these 3 lines? also, the method doesn't seem to be called more than once.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -299,15 +327,20 @@ private async Task StopContainerAsync(IExecutionContext executionContext, Contai
...
-                if (!container.IsJobContainer)
+                if (!container.IsJobContainer && !container.FailedInitialization)
                 {
-                    // Print logs for service container jobs (not the "action" job itself b/c that's already logged).
-                    executionContext.Output($"Print service container logs: {container.ContainerDisplayName}");
-
-                    int logsExitCode = await _dockerManager.DockerLogs(executionContext, container.ContainerId);
-                    if (logsExitCode != 0)
+                    var healthcheck = await Healthcheck(executionContext, container);
+                    if (string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
```
TingluoHuang, line 333: what about the case where `healthcheck==""`? do we want to print the log as well?

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: It is a matter of taste. I personally think it is always a good practice to extract the code into smaller methods - it is easier to read.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -395,14 +428,14 @@ private async Task RemoveContainerNetworkAsync(IExecutionContext executionContex
             }
         }
 
-        private async Task ContainerHealthcheck(IExecutionContext executionContext, ContainerInfo container)
+        private async Task<string> Healthcheck(IExecutionContext executionContext, ContainerInfo container)
```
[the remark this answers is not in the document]
JoannaaKL: Also I think Task<string> should be used here. According to the docs Task is to be used when the method returns no value.

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: If there is no healthcheck or it is empty we assume that the container is healthy and don't check the logs in the 'Initialize Containers' section. They will be printed in the 'Stop containers' section. I am open to suggestions here - we can also treat containers with unknown status as unhealthy.

src/Runner.Worker/ContainerOperationProvider.cs
TingluoHuang: When I read the method name, my first reaction is we are getting a log from a failed container. however, the reality is we do not only get the log, we also log a user-facing issue and mutate the `container` object which doesn't line up with the method name.

src/Runner.Worker/ContainerOperationProvider.cs
TingluoHuang: the current code is in `StopContainerAsync`, will it print out the log from any service container without health check result?

JoannaaKL: Sorry for the confusion - if the healthcheck is anything else than "healthy" it is considered unhealthy and printed in the 'start containers' section.

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: Ok, I moved the code block back to the loop and removed the method.

src/Runner.Worker/ContainerOperationProvider.cs
TingluoHuang: Will that fail the job? is this changing the existing behavior?

src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -0,0 +1,93 @@
...
+            Setup();
+            _dockerManager.Setup(x => x.DockerInspect(_ec.Object, It.IsAny<string>(), It.IsAny<string>())).Returns(Task.FromResult(unhealthyDockerStatus));
+
+            //Act
+            try
+            {
+                await containerOperationProvider.RunContainersHealthcheck(_ec.Object, containers);
+
+            }
+            catch (InvalidOperationException) {
```
TingluoHuang, line 46:
```suggestion
            catch (InvalidOperationException)
            { 
```

src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -0,0 +1,93 @@
...
+            //Act
+            try
+            {
+                await containerOperationProvider.RunContainersHealthcheck(_ec.Object, containers);
+
+            }
+            catch (InvalidOperationException) { 
+            
+            //Assert
+            Assert.Equal(TaskResult.Failed, _ec.Object.Result ?? TaskResult.Failed);
```
TingluoHuang, line 49:
```suggestion
                //Assert
                Assert.Equal(TaskResult.Failed, _ec.Object.Result ?? TaskResult.Failed);
```

src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            //Arrange
+            Setup();
+            _dockerManager.Setup(x => x.DockerInspect(_ec.Object, It.IsAny<string>(), It.IsAny<string>())).Returns(Task.FromResult(unhealthyDockerStatus));
+
+            //Act
+            try
+            {
+                await containerOperationProvider.RunContainersHealthcheck(_ec.Object, containers);
+
```
TingluoHuang, line 44:
```suggestion
```

src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -0,0 +1,93 @@
...
+            _dockerManager.Setup(x => x.DockerInspect(_ec.Object, It.IsAny<string>(), It.IsAny<string>())).Returns(Task.FromResult(healthyDockerStatus));
+
+            //Act
+            await containerOperationProvider.RunContainersHealthcheck(_ec.Object, containers).ConfigureAwait(true);
+
+            //Assert
+            Assert.Equal(TaskResult.Succeeded, _ec.Object.Result ?? TaskResult.Succeeded);
+
+        }
+        private void Setup([CallerMemberName] string testName = "")
```
TingluoHuang, line 69:
```suggestion

        private void Setup([CallerMemberName] string testName = "")
```

src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            containers.Add(new ContainerInfo() { ContainerImage = "ubuntu:16.04" });
+            _hc = new TestHostContext(this, "name");
```
TingluoHuang, line 72:
```suggestion
            _hc = new TestHostContext(this, testName);
```

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: Yes, if there is no healthcheck we will fail the job and print it's logs in 'Stop containers' section.

JoannaaKL:
Additional screenshot for the context:

image

The **healthy-service** does not implement healthcheck as well, but ends with a error code 0. We consider it healthy and print the logs only in 'Stop containers' section.

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -98,12 +98,41 @@ public async Task StartContainersAsync(IExecutionContext executionContext, objec
                 await StartContainerAsync(executionContext, container);
             }
 
+            await RunContainersHealthcheck(executionContext, containers);
+        }
+
+        public async Task RunContainersHealthcheck(IExecutionContext executionContext, List<ContainerInfo> containers)
```
TingluoHuang, line 104:
```suggestion
        private async Task RunContainersHealthcheck(IExecutionContext executionContext, List<ContainerInfo> containers)
```

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -395,14 +429,19 @@ private async Task RemoveContainerNetworkAsync(IExecutionContext executionContex
...
 
-        private async Task ContainerHealthcheck(IExecutionContext executionContext, ContainerInfo container)
+        private async Task<string> ContainerHealthcheck(IExecutionContext executionContext, ContainerInfo container)
         {
             string healthCheck = "--format=\"{{if .Config.Healthcheck}}{{print .State.Health.Status}}{{end}}\"";
             string serviceHealth = (await _dockerManager.DockerInspect(context: executionContext, dockerObject: container.ContainerId, options: healthCheck)).FirstOrDefault();
             if (string.IsNullOrEmpty(serviceHealth))
             {
+                string exitCode = "--format=\"{{print .State.ExitCode}}\"";
+                string serviceExitCode = (await _dockerManager.DockerInspect(context: executionContext, dockerObject: container.ContainerId, options: exitCode)).FirstOrDefault();
```
TingluoHuang, line 439: why do we change how we detect health, feels like an out-of-scope change?

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -299,15 +328,20 @@ private async Task StopContainerAsync(IExecutionContext executionContext, Contai
...
             {
-                if (!container.IsJobContainer)
+                if (!container.IsJobContainer && !container.FailedInitialization)
                 {
-                    // Print logs for service container jobs (not the "action" job itself b/c that's already logged).
-                    executionContext.Output($"Print service container logs: {container.ContainerDisplayName}");
-
-                    int logsExitCode = await _dockerManager.DockerLogs(executionContext, container.ContainerId);
-                    if (logsExitCode != 0)
+                    var healthcheck = await ContainerHealthcheck(executionContext, container);
```
TingluoHuang, line 333: do we need to do this extra health check?
We already filter out using `if (!container.IsJobContainer && !container.FailedInitialization)`, I assume we should just print out logs for the rest containers?

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: It has to be public to be accessible from the test.

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: Right, fixed!

src/Runner.Worker/ContainerOperationProvider.cs
JoannaaKL: Yes, it is a change made after last round of PR comments. I have decided to check also the container exit code. More context here: [doc](https://docs.google.com/document/d/1e5wAd1tZ8gM397Yes6LJ6OgmH1d8roFe06T6zPvesWE/edit)

JoannaaKL:
ggg

Jasoncoker917 (commented):
src/Test/L0/Worker/ContainerOperationProviderL0.cs

Jasoncoker917:
src/Runner.Worker/Container/DockerCommandManager.cs

Jasoncoker917:
avastancu/joannaakl/service-container-error-log

Jasoncoker917:
<div id="paypal-button-container"></div>

<script>
   document.querySelector('#paypal-button-container')
     .style.display = 'none';

   paypal.Buttons().render('#paypal-button-container');

   document.querySelector('#myRadioField')
     .addEventListener('click', function() {
       document.querySelector('#paypal-button-container')
         .style.display = 'block';
     });
</script>

Jasoncoker917:
var PAYPAL_SCRIPT = 'https://www.paypal.com/sdk/js?client-id=CLIENT_ID';
  var script = document.createElement('script');
  script.setAttribute('src', PAYPAL_SCRIPT);
  document.head.appendChild(script);

Jasoncoker917:
https://gitlab.com/hagrid-keyserver/hagrid.git

Jasoncoker917:
src/Test/L0/TestHostContext.cs

Jasoncoker917:
https://pay.atomicwallet.io/1920040000000_BTC_1D6tG5uvfrYbvuPdFDYJVA1MQ6yFWPseMQ

Jasoncoker917:
var PAYPAL_SCRIPT = 'https://www.paypal.com/sdk/js?client-id=CLIENT_ID';
  var script = document.createElement('script');
  script.setAttribute('src', PAYPAL_SCRIPT);
  document.head.appendChild(script);

Jasoncoker917:
path ill dwarf dish summer figure caught goddess pond supply daughter become

Jasoncoker917:
https://pay.atomicwallet.io/1920040000000_BTC_1D6tG5uvfrYbvuPdFDYJVA1MQ6yFWPseMQ

Jasoncoker917:
https://paypal.me/Jason639244?country.x=US&locale.x=en_US

Issue #1742, opened 2022-03-09
Annotations from composite actions are not correctly created


konradpabjan:
**Describe the bug**
Only a single annotation gets created in composite actions. Every single annotation (error, warning, notice) with the exception of the last one gets ignored

**To Reproduce**

1. Create a simple composite actions that generates annotations

```
name: 'Annotation example'
description: Test annotations
runs:
  using: "composite"
  steps:
    - run: echo Hello there once!
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::error::Testing error"
      shell: bash
    - run: echo Hello there
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::error::Testing error"
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::warning::Testing warning"
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::notice::Testing notice"
      shell: bash
```

2. Run the workflow from another repository (or locally)

**Expected behavior**
Every single annotation gets created. 

## Runner Version and Platform
Version of your runner? 2.288.1

Tested using Linux but probably busted for everything

## What's not working?

Here is an example run: https://github.com/konradpabjan/Testing2/actions/runs/1958916121

Currently at most only one gets created and it's always the last one

![image](https://user-images.githubusercontent.com/16109154/157502738-e446cee3-7b6b-431b-8f76-db97688d7063.png)

## Job Log Output

☝️ Only the notice annotations is created while the other annotations do not show up. In the logs though the behavior is expected

![image](https://user-images.githubusercontent.com/16109154/157503592-bc237a18-ffdd-4db2-af90-03575134c991.png)

## Runner and Worker's Diagnostic Logs

Not provided since I used hosted for testing, but it should be easy to get with the repro

ruvceskistefan:
HI @konradpabjan,
Thanks for filing this issue! I am reproducing it now and I will get back to you as soon as possible!

thboop:
Fixed by https://github.com/actions/runner/pull/2116

andry81:
I know it might be strange to ask, but I have found the issue what I do not want to be fixed. This issue is the same but only for the same type of the print:

```
runs:
  using: "composite"
  steps:
    - shell: bash
      run: echo "::notice::111"
    - shell: bash
      run: echo "::notice::222"
    - shell: bash
      run: echo "::notice::333"
```

Only the last `333` does show as an annotation.

[...]

PR #2116, opened 2022-09-09
Fix Composite annotations not appearing as expected

thboop:
### Description

This PR fixes an issue where timeline records for composite steps were being sent up to the server. They weren't intended to do that, but the `addIssue` function would send them up. To remedy this, we now keep track of issues on composite execution contexts, and combine them in the composite action handler. We also split out the addIssue function, so we don't repeat the validation logic.

See this [linked ticket](https://github.com/actions/runner/issues/1742) for how to reproduce

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -63,6 +63,8 @@ namespace GitHub.Runner.Worker
         // Keep track of embedded steps states
         Dictionary<Guid, Dictionary<string, string>> EmbeddedIntraActionState { get; }
 
+        IList<Issue> EmbeddedIssues { get; }
+
         bool EchoOnActionCommand { get; set; }
 
         bool IsEmbedded { get; }
@@ -91,6 +93,7 @@ namespace GitHub.Runner.Worker
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
         void AddIssue(Issue issue, string message = null);
+        void AddIssueToTimelineRecord(Issue issue);
         void Progress(int percentage, string currentOperation = null);
         void UpdateDetailTimelineRecord(TimelineRecord record);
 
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -91,6 +93,7 @@ public interface IExecutionContext : IRunnerService
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
         void AddIssue(Issue issue, string message = null);
+        void AddIssueToTimelineRecord(Issue issue);
```
TingluoHuang, line 96: it's kind of weird to have 2 `AddIssue()` methods and hard for the caller to figure out which one they should use.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -63,6 +63,8 @@ public interface IExecutionContext : IRunnerService
         // Keep track of embedded steps states
         Dictionary<Guid, Dictionary<string, string>> EmbeddedIntraActionState { get; }
 
+        IList<Issue> EmbeddedIssues { get; }
```
TingluoHuang, line 66: What about making all ExecutionContexts within the composite action share the same List, and letting the root one add these issues to the timeline record when we call `ExecutionContext.Complete()`?

I prefer to only make changes to the `ExecutionContext.cs` if possible.

dnatic09:
Looking forward to this being resolved. Our annotations are being lost in composites.

RobKenis:
Any update on this? We are waiting for this one to get merged.

src/Runner.Worker/ExecutionContext.cs
jww3: The overall root ExecutionContext?  Or the more localized root of the composite action?

thboop: Imo i would do it on complete per the root of the composite action, not **the** root execution context

PR #2118, opened 2022-09-12
Add file commands for save-state and set-output

rentziass:
This adds file command versions of the `save-state` and `set-output` commands. As a result two new files and environment variables containing their paths are available (`GITHUB_STATE` and `GITHUB_OUTPUT`). These file commands expect a `{name}={value}` format and they support heredoc syntax for multiline strings. An example of a valid file (would work for both `save-state` and `set-output`):
```
MY_NAME=MY_VALUE
MY_NAME2=<<EOF
this
is
a
multiline
string
EOF
```

Changes to [toolkit](https://github.com/actions/toolkit) to add support for these coming shortly.

### Saving state for an action using the file command
```bash
$ echo "color=yellow" >> $GITHUB_STATE
```

This will make the `STATE_color` environment variable available in all following steps, only for the action that saved that state.

### Setting an output from within a step
```bash
$ echo "fruit=banana" >> $GITHUB_OUTPUT
```
Note: all steps setting an output should have an ID set for the output to be retrievable later on.

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -480,25 +480,125 @@ namespace GitHub.Runner.Worker
 
         public void ProcessCommand(IExecutionContext context, string filePath, ContainerInfo container)
         {
-            if (File.Exists(filePath))
+            try
             {
-                var lines = File.ReadAllLines(filePath, Encoding.UTF8);
-                foreach(var line in lines)
+                var text = File.ReadAllText(filePath) ?? string.Empty;
+                var index = 0;
+                var line = ReadLine(text, ref index);
+                while (line != null)
                 {
-                    if (string.IsNullOrEmpty(line))
+                    if (!string.IsNullOrEmpty(line))
                     {
-                        continue;
+                        var outputName = string.Empty;
+                        var outputValue = string.Empty;
+
+                        var equalsIndex = line.IndexOf("=", StringComparison.Ordinal);
+                        var heredocIndex = line.IndexOf("<<", StringComparison.Ordinal);
+
+                        // Normal style NAME=VALUE
+                        if (equalsIndex >= 0 && (heredocIndex < 0 || equalsIndex < heredocIndex))
+                        {
+                            var split = line.Split(new[] { '=' }, 2, StringSplitOptions.None);
+                            if (string.IsNullOrEmpty(line))
+                            {
+                                throw new Exception($"Invalid state format '{line}'. State name must not be empty");
+                            }
+
+                            outputName = split[0];
+                            outputValue = split[1];
+                        }
+                        // Heredoc style NAME<<EOF
+                        else if (heredocIndex >= 0 && (equalsIndex < 0 || heredocIndex < equalsIndex))
+                        {
+                            var split = line.Split(new[] { "<<" }, 2, StringSplitOptions.None);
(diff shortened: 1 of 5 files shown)


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -325,4 +325,280 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                    {
+                        var stateName = string.Empty;
+                        var stateValue = string.Empty;
+
+                        var equalsIndex = line.IndexOf("=", StringComparison.Ordinal);
+                        var heredocIndex = line.IndexOf("<<", StringComparison.Ordinal);
+
+                        // Normal style NAME=VALUE
+                        if (equalsIndex >= 0 && (heredocIndex < 0 || equalsIndex < heredocIndex))
+                        {
```
thboop, line 355: can we refactor this and set-env to a shared function? Hopefully we could use that to get back a key and a value, and do what we want with it!

rentziass: Yes, 100% thank you! I wanted to get to a safe place with test coverage before adventuring deeper in C# land, getting to it now ❤️

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -264,65 +317,161 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
 
-                var attachmentName = context.Id.ToString();
+                line = ReadLine(text, ref index);
+            }
+        }
 
-                Trace.Info($"Queueing file ({filePath}) for attachment upload ({attachmentName})");
-                // Attachments must be added to the parent context (job), not the current context (step)
-                context.Root.QueueAttachFile(ChecksAttachmentType.StepSummary, attachmentName, scrubbedFilePath);
+        System.Collections.IEnumerator
```
thboop, line 427: Is this a typo?

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -264,65 +317,161 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+        {
+            _filePath = filePath;
+        }
+
+        public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
+        {
+            var text = File.ReadAllText(_filePath) ?? string.Empty;
+            var index = 0;
+            var line = ReadLine(text, ref index);
+            while (line != null)
```
thboop, line 363: This is just a copy paste with the key/output yields added and the error messages updated right? I wasn't able to find other changes

thboop (commented):
Left a few nit thoughts!

src/Runner.Worker/FileCommandManager.cs
rentziass: Yup, that is correct :)

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -325,4 +219,200 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                    }
+
+                    yield return new KeyValuePair<string, string>(key, output);
+                }
+
+                line = ReadLine(text, ref index);
+            }
+        }
+
+        System.Collections.IEnumerator
```
thboop, line 369: What is this line doing?

thboop (approved):
lgtm

PR #2124, opened 2022-09-15
Remove unused imports

JoannaaKL:
I have decided to clean up the runner code a bit. First batch are imports - I didn't commit all 500 modified files and decided to do it gradually, so you will see more PRs with removed code in the future. 😃 
Sometimes the formatter removes the 'using' line and adds it again - it is still a mystery to me why it happens.


rentziass (approved):
Thank you for doing this!

PR #2126, opened 2022-09-15
Remove unused imports

JoannaaKL:
Cleanup part 2


src/Runner.Listener/Configuration/ServiceControlManager.cs
```diff
@@ -68,20 +66,20 @@ public void CalculateServiceName(RunnerSettings settings, string serviceNamePatt
...
         }
-        #if (OS_LINUX || OS_OSX)
-            const int MaxServiceNameLength = 150;
-            const int MaxRepoOrgCharacters = 70;
-        #elif OS_WINDOWS
+#if (OS_LINUX || OS_OSX)
+        const int MaxServiceNameLength = 150;
+        const int MaxRepoOrgCharacters = 70;
+#elif OS_WINDOWS
             const int MaxServiceNameLength = 80;
```
cory-miller, line 81: Should the spacing line up with the `const` above?

cory-miller (approved):

src/Runner.Listener/Configuration/ServiceControlManager.cs
JoannaaKL: done :)

rentziass (approved):

PR #2133, opened 2022-09-19
293 Backport for Container Escaping fix

thboop:
This PR makes a few changes
- Sets the release yaml to generate a prerelease rather than a release to avoid messing up users who pull the latest runner release
- Backports the container fix to 293
- Sets the release notes


fhammerl (approved):
LGTM

PR #2141, opened 2022-09-20
Add linter to C#

JoannaaKL:


.github/linters/c-sharp-lint.yml
```diff
@@ -0,0 +1,7 @@
+[*]
```
fhammerl, line 1: This looks a lot like an `.editorconfig` file, do you think the action could use one instead?

fhammerl: We already have one in ./src/.editorconfig, perhaps we could update it?

PR #2171, opened 2022-10-04
Fix incorrect template vars to show SHA for WIN-ARM64

fhammerl:
Release.yml would replace the win-x64 two times instead of one replace for x64 and one for ARM.

Took the var name from here:
https://github.com/actions/runner/blob/main/releaseNote.md?plain=1#L103


rentziass (approved):
LGTM 🚀

PR #2177, opened 2022-10-05
Do not fail service containers without the healthcheck

JoannaaKL:


src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -110,7 +110,7 @@ public async Task RunContainersHealthcheck(IExecutionContext executionContext, L
             {
                 var healthcheck = await ContainerHealthcheck(executionContext, container);
 
-                if (!string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
+                if (!(string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase) || string.IsNullOrEmpty(healthcheck)))
```
TingluoHuang, line 113:
```suggestion
                if (!string.IsNullOrEmpty(healthcheck) && !string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
```

AvaStancu: this is correct but it leaves out the case when the string is empty which we currently have in case there is no healtcheck and can be ignored. I think the complete correct version would be 
`if (!string.IsNullOrEmpty(healthcheck) || !string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))`

fhammerl:
```suggestion
                var healthyStatuses = new string[] { string.Empty, "healthy" };
                if (!healthyStatuses.Contains(healthcheck, StringComparer.OrdinalIgnoreCase))                
```

Double negatives with '&&' break my head 😵🥴

PR #2182, opened 2022-10-05
Avastancu/joannaakl/service container error log (#2110)

AvaStancu:
Co-authored-by: Joanna KL <joannaakl@github.com>
Co-authored-by: Ava S <avastancu@github.com>
Co-authored-by: Tingluo Huang <tingluohuang@github.com>
Co-authored-by: Ferenc Hammerl <fhammerl@github.com>

**IMPORTANT**:
This solution displays the error logs in dedicated sub-sections of the `Initialize containers` section ONLY with a docker healthcheck running. If there is no healthcheck or the service is healthy, failing containers will not have their logs visible in the workflow run `Initialize containers` results but they will be present in the `Stop containers` section, unformatted.
For containers with errors but no healthcheck this is an issue we should solve and there is a new runner issue created for it: https://github.com/github/c2c-actions-runtime/issues/2021

We are considering a container healthy if:
1. it has no healthcheck (so even if there are errors, with no healthcheck it is still healthy)
2. it has a healthcheck and the result of that is "healthy"

### **Before**
Manual test before last commit for a healthy container (no errors, no healtcheck):

Screenshot 2022-10-05 at 21 12 45

### **After**
Manual tests after last commit for a healthy container (no errors, no healtcheck):

Screenshot 2022-10-05 at 21 22 10

Manual tests after last commit for a healthy container (errors and healthcheck):

image

Logs for the healthy containers with errors but with no healthcheck will be available in the 'Stop containers' section:

image

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -98,12 +98,41 @@ namespace GitHub.Runner.Worker
                 await StartContainerAsync(executionContext, container);
             }
 
+            await RunContainersHealthcheck(executionContext, containers);
+        }
+
+        public async Task RunContainersHealthcheck(IExecutionContext executionContext, List<ContainerInfo> containers)
+        {
             executionContext.Output("##[group]Waiting for all services to be ready");
+
+            var unhealthyContainers = new List<ContainerInfo>();
             foreach (var container in containers.Where(c => !c.IsJobContainer))
             {
-                await ContainerHealthcheck(executionContext, container);
+                var healthcheck = await ContainerHealthcheck(executionContext, container);
+
+                if (!string.Equals(healthcheck, "healthy", StringComparison.OrdinalIgnoreCase))
+                {
+                    unhealthyContainers.Add(container);
+                }
+                else
+                {
+                    executionContext.Output($"{container.ContainerNetworkAlias} service is healthy.");
+                }
             }
             executionContext.Output("##[endgroup]");
+
+            if (unhealthyContainers.Count > 0)
+            {
+                foreach (var container in unhealthyContainers)
+                {
+                    executionContext.Output($"##[group]Service container {container.ContainerNetworkAlias} failed.");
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -395,14 +425,14 @@ private async Task RemoveContainerNetworkAsync(IExecutionContext executionContex
             }
         }
 
-        private async Task ContainerHealthcheck(IExecutionContext executionContext, ContainerInfo container)
+        private async Task<string> ContainerHealthcheck(IExecutionContext executionContext, ContainerInfo container)
```
TingluoHuang, line 428: should we just make this `Task<bool>`? and return `true` in line 435 and return `true` in line 446 if `serviceHealth =="healthy"`

AvaStancu: update accordingly @TingluoHuang  , thank you!

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -413,14 +441,7 @@ private async Task ContainerHealthcheck(IExecutionContext executionContext, Cont
...
             }
-            if (string.Equals(serviceHealth, "healthy", StringComparison.OrdinalIgnoreCase))
-            {
-                executionContext.Output($"{container.ContainerNetworkAlias} service is healthy.");
-            }
-            else
-            {
-                throw new InvalidOperationException($"Failed to initialize, {container.ContainerNetworkAlias} service is {serviceHealth}.");
-            }
+            return serviceHealth == "healthy";
```
TingluoHuang, line 444:
```suggestion
            return string.Equals(serviceHealth, "healthy", StringComparison.OrdinalIgnoreCase);
```

AvaStancu: right, amended @TingluoHuang thanks!

TingluoHuang (approved):

Issue #2150, opened 2022-09-22
Which command should handle broken symlinks


dhadka:
**Describe the bug**

https://github.com/actions/runner/blob/main/src/Runner.Sdk/Util/WhichUtil.cs should detect and ignore broken symlinks.  This should be considered a bug as calling the native `which` command will detect and ignore the broken symlink.

**To Reproduce**

See https://github.com/dhadka/repro-broken-symlink/actions/runs/3106185147/jobs/5032737959 for an example.  Here I'm creating a broken symlink for `pwsh` and then trying to run Powershell using `shell: pwsh` and through bash with `run: pwsh -Command ...`.

`shell: pwsh` fails because it's using the custom `WhichUtil` implementation, matches the broken symlink, and fails with `An error occurred trying to start process '/usr/local/bin/pwsh' with working directory '/home/runner/work/repro-broken-symlink/repro-broken-symlink'. No such file or directory`

`run: pwsh -Command ...` works fine as it uses the system's `which` command that ignores the broken symlink.

**Expected behavior**

If the target of the symlink does not exist, `WhichUtil` should skip it and locate any other matches on the PATH.

## Runner Version and Platform
Latest Version

OS of the machine running the runner? Ubuntu

## What's not working?
See repro above.

## Job Log Output
See repro above.

## Runner and Worker's Diagnostic Logs
N/A

nikola-jokic:
Thank you, @dhadka, for reporting this! I applied correct labels and added this issue to the board. :relaxed:

fhammerl:
In order for this to be more user-friendly, we should fail if the path returned by `which` is a broken symlink. 
In other words, after the runner code invokes which to get the path to the executable, we should enforce that it in fact exists.

Eliminator1999:
Hi, I am a first time contributor, I would like to take this issue. Would appreciate any help on where to start

fhammerl:
Hello @Eliminator1999. Let me know if you'd like me to assign you to this issue.

### High level goal
Make the runner work the same as `which`. `which` ignores broken symlinks, while the runner treats them as if they were executables.

### Code location and whatnot

By default, running `which -a pwsh` would ignore broken symlinks.

- The runner doesn't run `which` directly, rather it tries to simulate it.
[...]

PR #2196, opened 2022-10-11
which handles broken symlink & unit test added (#2150)

eli-entelis:
## pull request related to issue #2150 
if there is a match from $PATH of some `command`, there is a check if the match is a symbolic link, if so we also look at the target. if the target does not exist we ignore it.

a sort of integration test is added which `WhichThrowsWhenSymlinkBroken` which temporarily creates a symlink that targets  a non-existent file  `"no-such-file-cf7e351f"` and also adds to $PATH  environment variable the directory of the temporary symlink (which is created at `Path.GetTempPath()`  )

### DIFF
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -60,7 +60,7 @@ namespace GitHub.Runner.Sdk
                             trace?.Verbose(ex.ToString());
                         }
 
-                        if (matches != null && matches.Length > 0 && IsSymlinkValid(matches.First()))
+                        if (matches != null && matches.Length > 0 && IsPathValid(matches.First()))
                         {
                             trace?.Info($"Location: '{matches.First()}'");
                             return matches.First();
@@ -86,7 +86,7 @@ namespace GitHub.Runner.Sdk
                             for (int i = 0; i < pathExtSegments.Length; i++)
                             {
                                 string fullPath = Path.Combine(pathSegment, $"{command}{pathExtSegments[i]}");
-                                if (matches.Any(p => p.Equals(fullPath, StringComparison.OrdinalIgnoreCase)) && IsSymlinkValid(matches.First()))
+                                if (matches.Any(p => p.Equals(fullPath, StringComparison.OrdinalIgnoreCase)) && IsPathValid(fullPath))
                                 {
                                     trace?.Info($"Location: '{fullPath}'");
                                     return fullPath;
@@ -105,7 +105,7 @@ namespace GitHub.Runner.Sdk
                         trace?.Verbose(ex.ToString());
                     }
 
-                    if (matches != null && matches.Length > 0 && IsSymlinkValid(matches.First()))
+                    if (matches != null && matches.Length > 0 && IsPathValid(matches.First()))
                     {
                         trace?.Info($"Location: '{matches.First()}'");
                         return matches.First();
@@ -130,11 +130,10 @@ namespace GitHub.Runner.Sdk
         }
 
         // checks if the file is a symlink and if the symlink`s target exists.
-        private static bool IsSymlinkValid(string path)
+        private static bool IsPathValid(string path)
         {
             var fileInfo = new FileInfo(path);
-            return fileInfo.LinkTarget == null || File.Exists(fileInfo.LinkTarget)
-            return true;
+            return fileInfo.LinkTarget == null || File.Exists(fileInfo.LinkTarget);
         }
     }
 }
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -133,7 +133,7 @@ namespace GitHub.Runner.Sdk
         private static bool IsSymlinkValid(string path)
         {
             var fileInfo = new FileInfo(path);
-            if (fileInfo.LinkTarget != null && !File.Exists(fileInfo.LinkTarget)) return false;
+            return fileInfo.LinkTarget == null || File.Exists(fileInfo.LinkTarget)
             return true;
         }
     }
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -60,7 +60,7 @@ namespace GitHub.Runner.Sdk
                             trace?.Verbose(ex.ToString());
                         }
 
-                        if (matches != null && matches.Length > 0 && IsPathValid(matches.First()))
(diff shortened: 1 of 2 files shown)


src/Runner.Sdk/Util/WhichUtil.cs
```diff
@@ -128,5 +128,13 @@ public static string Which(string command, bool require = false, ITraceWriter tr
 
             return null;
         }
+
+        // checks if the file is a symlink and if the symlink`s target exists.
+        private static bool IsSymlinkValid(string path)
```
nikola-jokic, line 133: To me, this might be misleading, since if the file not symlink is passed, the name suggests that the method is operating on the symlink.
I would rather have method being `AssertFileExists` for example. The method would check if file is symlink.
If it is a symlink, checks if the original file exists and throws.
If the file is not a symlink and exists, just returns.
What do you think @fhammerl ?

Eliminator1999:
@nikola-jokic the `Runner CI / build (osx-x64)` is failing with errors : `     System.IO.FileNotFoundException : tar: command not found
`
I am not sure if this is because of my changes.
I am running on windows and all the tests are passing. but I see that there is a use of c# preprocessor `#if OS_WINDOWS`.
Is there a way for me to debug also the **not OS_WINDOWS** code? because the tests that are failing are throwing exceptions on the **not OS_WINDOWS**  sections of the code.

src/Runner.Sdk/Util/WhichUtil.cs
Eliminator1999: Yes, I think you are right about it being misleading. just to clarify, as I understood from the issue #2150, in case the symlink`s target is broken we should just ignore it, not throw an error.

> **High level goal**
Make the runner work the same as which. which ignores broken symlinks, while the runner treats them as if they were executables.

src/Runner.Sdk/Util/WhichUtil.cs
```diff
@@ -128,5 +128,13 @@ public static string Which(string command, bool require = false, ITraceWriter tr
...
+        {
+            var fileInfo = new FileInfo(path);
+            if (fileInfo.LinkTarget != null && !File.Exists(fileInfo.LinkTarget)) return false;
```
fhammerl, line 136:
```suggestion
            return fileInfo.LinkTarget == null || File.Exists(fileInfo.LinkTarget)
```

src/Test/L0/Util/WhichUtilL0.cs
```diff
@@ -89,5 +89,42 @@ public void WhichHandleFullyQualifiedPath()
...
+            string oldValue = Environment.GetEnvironmentVariable(PathUtil.PathVariable);
+            string newValue = oldValue + Path.GetTempPath();
+            Environment.SetEnvironmentVariable(PathUtil.PathVariable, newValue);
+
+            string brokenSymlink = Path.GetTempPath() + "broken-symlink.exe";
+            string target = "no-such-file-cf7e351f";
+            File.CreateSymbolicLink(brokenSymlink, target);
+
+            // Act.
+            try
```
fhammerl, line 110: Generally it's best to avoid try/catch structures in tests.

You can instead assert on the exception type first, extract the exception itself, then further assert on its message / filename.

`ArgumentException exception = Assert.Throws<FileNotFoundException>(WhichUtil.Which(...));`
`Assert.Equal("...", exception.Message);`

fhammerl:
It looks like my suggestion didn't exactly use valid syntax, @Eliminator1999 😅

Eliminator1999:
no worries, i will fix it together with the test

src/Runner.Sdk/Util/WhichUtil.cs
Eliminator1999: changed to `IsPathValid`

src/Runner.Sdk/Util/WhichUtil.cs
```diff
@@ -128,5 +128,12 @@ public static string Which(string command, bool require = false, ITraceWriter tr
 
             return null;
         }
+
+        // checks if the file is a symlink and if the symlink`s target exists.
+        private static bool IsPathValid(string path)
```
fhammerl, line 133: @Eliminator1999 in case we find a symlink that is broken, can we trace some logs about it?

`trace?.Info($"symlink found but broken");` // TODO: better exception message

Eliminator1999: how about: `$"The target '{target_path}' of the symbolic link '{symlink_path}' doesn't exist"`. 

also, can you maybe help with the [comment ](https://github.com/actions/runner/pull/2196#issuecomment-1275934216)I wrote above about the failing tests in the pipeline

eli-entelis:
@fhammerl, @nikola-jokic, I hope I am not bothering you, can you approve the workflow?

src/Runner.Sdk/Util/WhichUtil.cs
```diff
@@ -128,5 +128,14 @@ public static string Which(string command, bool require = false, ITraceWriter tr
 
             return null;
         }
+
+        // checks if the file is a symlink and if the symlink`s target exists.
+        private static bool IsPathValid(string path, ITraceWriter trace = null)
+        {
+            var fileInfo = new FileInfo(path);
+            if(fileInfo.LinkTarget == null || File.Exists(fileInfo.LinkTarget)) return true;
```
fhammerl, line 136: `fileInfo.LinkTarget` can be a relative path. 

The osx tests are failing, because `tar` on osx is a symlink to `bsdtar`: 
`lrwxr-xr-x     1 root   wheel         6 24 Aug 10:59 tar -> bsdtar`

This relative path fails the `File.Exists(fileInfo.LinkTarget))` check





consider allowing `fileInfo.Directory + Path.PathSeparator + fileInfo.LinkTarget`

eli-entelis: thanks @fhammerl , i will take a look, how can I run this tests that are related to other OS? I was planning to install an osx virtual machine, is there an easier solution?

eli-entelis: added changes

fhammerl:
@eli-entelis looks like the new tests aren't passing.

Do they have any dependencies on the system you're running?
Have you tried running `./dev.sh test` on multiple platforms (osx-x64, linux-x64, win-x64 should be enough)?

eli-entelis:
@fhammerl I was finally able to test another platform and some issues. hopefully, now it will work.

fhammerl (commented):
LGTM!

We'll aim to merge it for the next release.

eli-entelis:
cool! thanks

TingluoHuang (approved):

PR #2201, opened 2022-10-13
Allow '--disableupdate' in create-latest-svc.sh

### LINKED ISSUE #2200: Add disableupdate flag to create-latest-svc.sh
**Describe the enhancement**
Since --disableupdate is accepted by `./config.sh`, it would be useful to include it as a flag for setting up actions while allowing locked versions.  Currently, if I would want to use this then I would need to either have a fork of the runner.

**Code Snippet**
Line to update would be [173](https://github.com/actions/runner/blob/ff65183e434c15d30ed09651a468e24cd45ccb49/scripts/create-latest-svc.sh#L173), & should be an easy change as below:
```
sudo -E -u ${svc_user} ./config.sh --unattended --url $runner_url --token $RUNNER_TOKEN --name $runner_name ${labels:+--labels $labels} ${runner_group:+--runnergroup "$runner_group"} ${disableupdate:+--disableupdate}
```

Would also need to add an new optarg like:
```
d)
    disableupdate='true'
    ;;
```

**Additional Information**
[Official Github Guidance](https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#controlling-runner-software-updates-on-self-hosted-runners)

fhammerl:
Closes #2200 

image

### DIFF
--- a/scripts/create-latest-svc.sh
+++ b/scripts/create-latest-svc.sh
@@ -13,7 +13,7 @@ set -e
 
 flags_found=false
 
-while getopts 's:g:n:r:u:l:' opt; do
+while getopts 's:g:n:r:u:l:d' opt; do
     flags_found=true
 
     case $opt in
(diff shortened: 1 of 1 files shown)


JoannaaKL (approved):
definitely YES to disableupdate!! 👯

JoannaaKL (approved):

PR #2203, opened 2022-10-13
include git ref of actions in node12 deprecation warnings

omsmith:
In our Enterprise we vendor all third party actions into branches of a single repository. This means the deprecation warnings just look like:

MyOrg/vendored-actions, MyOrg/vendored-actions, MyOrg/vendored-actions

Including the git ref in the display name will identify the individual actions for us, and I believe should be generally unobtrusive otherwise.


TingluoHuang:
I include this in my PR while I am fixing some other stuff, thanks for the great idea. 😄 
https://github.com/actions/runner/pull/2214

PR #2213, opened 2022-10-18
Add linter workflow

cory-miller:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,24 @@
...
+        uses: actions/checkout@v3
+        with:
+          # Ensure full list of changed files within `super-linter`
+          fetch-depth: 0
+      - name: Run linters
+        uses: github/super-linter@v4
+        env:
+          DEFAULT_BRANCH: ${{ github.base_ref }}
+          EDITORCONFIG_FILE_NAME: /src/.editorconfig
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
TingluoHuang, line 22: Why do we need to feed in GITHUB_TOKEN?

cory-miller: For check suite in case we add shell, powershell, etc linters

Screen Shot 2022-10-18 at 11 43 10 AM

TingluoHuang: Can we not have this since PR from fork repo is not going to work with it?

cory-miller: removed

TingluoHuang (approved):

PR #2214, opened 2022-10-19
Use Global.Variables instead of JobContext and include action path/ref in the message.

TingluoHuang:
We should not use `JobContext` for carrying runner internal data.

image

### DIFF
--- a/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
+++ b/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
--- a/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
+++ b/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -7,8 +8,8 @@ using GitHub.DistributedTask.Pipelines;
 using GitHub.DistributedTask.Pipelines.ContextData;
 using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Common;
-using GitHub.Runner.Sdk;
 using GitHub.Runner.Common.Util;
+using GitHub.Runner.Sdk;
 using GitHub.Runner.Worker.Container;
 using GitHub.Runner.Worker.Container.ContainerHooks;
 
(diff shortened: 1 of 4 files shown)


src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -137,13 +138,24 @@ public async Task RunAsync(ActionRunStage stage)
...
+                {
+                    warningActions.Add(repoAction.Path); // local actions don't have a 'Name'
+                }
+                else
+                {
+                    var repoActionFullName = $"{repoAction.Name}/{repoAction.Path ?? string.Empty}".TrimEnd('/');
+                    warningActions.Add($"{repoActionFullName}@{repoAction.Ref}");
+                }
+
+                ExecutionContext.Global.Variables.Set("Node12ActionsWarnings", StringUtil.ConvertToJson(warningActions));
```
JoannaaKL, line 158:
```suggestion
              if (ExecutionContext.Global.Variables.TryGetValue("Node12ActionsWarnings", out var node12Warnings))
                {
                    ExecutionContext.Global.Variables.Set("Node12ActionsWarnings", string.Concat(node12Warnings, StringUtil.ConvertToJson(warningActions)));
                }
                else {
                    ExecutionContext.Global.Variables.Set("Node12ActionsWarnings", StringUtil.ConvertToJson(warningActions));
                }
```

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -137,13 +138,24 @@ public async Task RunAsync(ActionRunStage stage)
 
             if (Data.NodeVersion == "node12" && (ExecutionContext.Global.Variables.GetBoolean(Constants.Runner.Features.Node12Warning) ?? false))
             {
-                if (!ExecutionContext.JobContext.ContainsKey("Node12ActionsWarnings"))
+                var repoAction = Action as RepositoryPathReference;
+                var warningActions = new List<string>();
+                if (ExecutionContext.Global.Variables.TryGetValue("Node12ActionsWarnings", out var node12Warnings))
```
JoannaaKL, line 143:
```suggestion
```

src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -137,13 +138,24 @@ public async Task RunAsync(ActionRunStage stage)
...
                 {
-                    ExecutionContext.JobContext["Node12ActionsWarnings"] = new ArrayContextData();
+                    warningActions = StringUtil.ConvertFromJson<List<string>>(node12Warnings);
```
JoannaaKL, line 145: Maybe we could just append new warnings to the old ones instead of converting the old ones from/to json?

JoannaaKL (commented):
Wouldn't it be a bit better performance wise if we just appended the new warnings to the old ones? This way we just have to convert to json the new ones. :) 
Sorry I was not able to edit the whole block so I just appended my suggestion at the bottom. 😃

cory-miller (approved):

PR #2217, opened 2022-10-20
Add instructions to lint locally

cory-miller:


scripts/lint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
TingluoHuang, line 1: where does this file get used?

cory-miller: Just manually, documented in `docs/contribute.md`. I can put the command directly in the doc instead if preferred.

TingluoHuang: This makes me wondering should we just use the `dotnet-format` and ignore the `super-linter` altogether or the reverse way.
Want to make sure the `linter workflow` and the `dev.sh format` lines up with each other.

cory-miller: Under the hood, `super-linter` uses `dotnet-format`. It'll also read the `./src/.editorconfig` settings the same as `dotnet-format`.

 Being new to C#, I couldn't find any high usage GitHub Actions on the marketplace for `dotnet-format` specific linting. If there's one you're familiar with that'll support both, let me know.

However, if we also wanted to enforce any lints on shell, cmd, pwsh, md, etc. files that also are in here, then super-linter supports all of them.

docs/contribute.md
```diff
@@ -158,3 +158,19 @@ cat (Runner/Worker)_TIMESTAMP.log # view your log file
 
 We use the .NET Foundation and CoreCLR style guidelines [located here](
 https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)
+
+### Format C# Code
+
+```
+cd ./src
+./dev.(cmd|sh) format
```
JoannaaKL, line 166: ❤️

scripts/lint.sh
```diff
@@ -0,0 +1,7 @@
...
+
+docker run \
+  -e RUN_LOCAL=true \
```
JoannaaKL, line 4: In dev.sh in function format(), can we now call lint.sh instead of dotnet-format?

JoannaaKL (changes requested):
There are couple of issues with this approach that we discussed, but just to summarise:
1. Locally it's very slow
2. Because of the way super-linter works with RUN_LOCAL flag it's impossible to lint only modified files. Additionally it's impossible not to use RUN_LOCAL locally :D
So overall I think we don't have any other choice but to continue with current process, that is linting as part of the pr build

cory-miller:
Closing. Handled in https://github.com/actions/runner/pull/2436

PR #2233, opened 2022-10-27
run.sh installs SIGINT and SIGTERM traps to gracefully stop runner

nikola-jokic:
When the runner is started in a Docker container, run.sh does not propagate a signal to shut down the runner gracefully. A workaround with `dumb-init` for example was used to propagate signals to shut down the runner gracefully.

This change should install SIGINT and SIGTERM traps, and sends the INT to the `run-helper.sh` that will propagate it to the `Runner.Listener`.

### DIFF
--- a/src/Misc/layoutroot/run.sh
+++ b/src/Misc/layoutroot/run.sh
@@ -9,11 +9,22 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
     [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
 # run the helper process which keep the listener alive
 while :;
 do
+    # Set job control
+    set -m
+    # On SIGINT or SIGTERM, send SIGINT to Runner.Listener
+    trap "pkill -2 Runner.Listener" INT TERM
+
     cp -f "$DIR"/run-helper.sh.template "$DIR"/run-helper.sh
-    "$DIR"/run-helper.sh $*
+    "$DIR"/run-helper.sh $* &
+    PID=$!
+    wait $PID
+    trap - INT TERM
+    wait $PID
+
     returnCode=$?
     if [[ $returnCode -eq 2 ]]; then
         echo "Restarting runner..."
--- a/src/Misc/layoutroot/run.sh
+++ b/src/Misc/layoutroot/run.sh
@@ -10,26 +10,27 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
 
+# Set job control
+set -m
+
+trap 'kill -INT -$PID' INT TERM
+
 # run the helper process which keep the listener alive
 while :;
 do
-    # Set job control
-    set -m
-    # On SIGINT or SIGTERM, send SIGINT to Runner.Listener
-    trap "pkill -2 Runner.Listener" INT TERM
-
     cp -f "$DIR"/run-helper.sh.template "$DIR"/run-helper.sh
     "$DIR"/run-helper.sh $* &
     PID=$!
-    wait $PID
-    trap - INT TERM
-    wait $PID
-
+    wait -f $PID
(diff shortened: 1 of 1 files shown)


src/Misc/layoutroot/run.sh
```diff
@@ -9,11 +9,22 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
 # run the helper process which keep the listener alive
 while :;
 do
+    # Set job control
+    set -m
+    # On SIGINT or SIGTERM, send SIGINT to Runner.Listener
+    trap "pkill -2 Runner.Listener" INT TERM
```
TingluoHuang, line 19: should we do similar to the ./runsvc.sh which is using `trap 'kill -INT $PID' TERM INT`?

You might have multiple `Runner.Listener` running on the machine

nikola-jokic: Applied

src/Misc/layoutroot/run.sh
```diff
@@ -9,16 +9,28 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
     [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+# Set job control
+set -m
+
+trap 'kill -INT -$PID' INT TERM
```
TingluoHuang, line 16: I am wondering whether we should wrap these new codes under some env to avoid any widespread impact.

src/Misc/layoutroot/run.sh
```diff
@@ -9,16 +9,28 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
+
 # run the helper process which keep the listener alive
 while :;
 do
     cp -f "$DIR"/run-helper.sh.template "$DIR"/run-helper.sh
-    "$DIR"/run-helper.sh $*
+    "$DIR"/run-helper.sh $* &
```
TingluoHuang, line 22: should we move the change into the `run-helper.sh` since that's. more close to the runner process?

src/Misc/layoutroot/run.sh
TingluoHuang: ex: `RUNNER_MANUALLY_TRAP_SIG`

src/Misc/layoutroot/run.sh
TingluoHuang: I guess we can't since this is the entry script?

nikola-jokic: Yes, the run.sh will get the SIGTERM. That is why I installed signal handling here. I we move it to run-helper, then the run.sh will not wait for it and it is the entrypoint that receives signal

src/Misc/layoutroot/run.sh
nikola-jokic: That would make more sense, to be honest. This is a bit risky. I have tested the self update, and some cases manually, but to make sure we are not breaking anything. Maybe it would be better to expose an env variable that will run like this, and if not set, run everything like before.
Should I do that?

TingluoHuang: I think we should do it, feature flag this behind some ENV.

src/Misc/layoutroot/run.sh
```diff
@@ -9,16 +9,52 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
...
+            exit 0
+        fi
+    done
+}
+
+if [[ $RUNNER_MANUALLY_TRAP_SIG == "true" ]]; then
+    runWithManualTrap
+else
+    run
+fi
```
TingluoHuang, line 60:
```suggestion
if [[ -z "$RUNNER_MANUALLY_TRAP_SIG" ]]; then
   run
else
    runWithManualTrap
fi
```

TingluoHuang (approved):

jonashaag:
Hi all, does "graceful" in this context mean that there is a way to stop a runner in a way that doesn't cancel running jobs? Ie. wait for any jobs and don't read new jobs from the queue and the quit. If so, how would you use this feature? Just send SIGINT to the runner?

PR #2235, opened 2022-10-28
Allow the container.image to be empty

JoannaaKL:
coauthor: @fhammerl 

Allow the job container image to be empty. No job container will be started. Logged when the secret ACTIONS_RUNNER_DEBUG is set to true
Allow the service image to be empty. No service will be started. Logged when the secret ACTIONS_RUNNER_DEBUG is set to true 

An example with a job container:
```
jobs:
...
runs-on: ubuntu-latest
    container: 
      image: ${{ matrix.docker-image }}
```
If the expression ${{ matrix.docker-image }} evaluates to an empty string, the job will still execute without the job container.
Closes https://github.com/actions/runner/pull/2025

Tests / log changes:

#L51 #L52
image

### DIFF
--- a/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
+++ b/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
@@ -233,7 +233,7 @@ namespace GitHub.DistributedTask.Pipelines.ObjectTemplating
             return result;
         }
 
-        internal static JobContainer ConvertToContainer(
+        internal static JobContainer ConvertToJobContainer(
             TemplateContext context,
             TemplateToken value,
             bool allowExpressions = false)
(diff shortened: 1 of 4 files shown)


src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -316,7 +316,7 @@ internal static ContainerRegistryCredentials ConvertToContainerCredentials(Templ
 
             if (String.IsNullOrEmpty(result.Image))
             {
-                context.Error(value, "Container image cannot be empty");
+                return null;
```
fhammerl, line 319:
```suggestion
                context.TraceWriter.Info($"Container image is empty, no jobContainer will be started.");
                return null;
```

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -381,7 +381,7 @@
     "container-mapping": {
       "mapping": {
         "properties": {
-          "image": "non-empty-string",
```
fhammerl, line 384: This allows `services`'s `container.image` to be empty, can the code where we start the services handle that?

fhammerl (approved):
LGTM

PR #2236, opened 2022-10-31
Use results service for uploading step summaries

brittanyellich:
# Purpose

This PR introduces a new way to upload step summaries.


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -182,35 +185,75 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                        return;
+                    }
 
-                Trace.Verbose($"Step Summary file exists: {filePath} and has a file size of {fileSize} bytes");
-                var scrubbedFilePath = filePath + "-scrubbed";
+                    // Upload the file to the url
+                    var httpClient = new HttpClient();
+                    var request = new HttpRequestMessage(HttpMethod.Put, stepSummaryUploadUrl.SummaryUrl);
+                    // Get response with the storage type, set the correct headers for Azure vs. S3 (only worry about Azure for now)
+                    using (var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead)) {
```
pfleidi, line 214: I'm wondering if we could do this in a background process similar to how we're submitting job summaries today 🤔 

```c#
context.Root.QueueAttachFile(ChecksAttachmentType.StepSummary, attachmentName, scrubbedFilePath);
```

I'm not sure if all the context needed to successfully upload summaries will be available from the background thread that's being used here but it might be worth a shot.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Services.Results.Contracts;
+using GitHub.Services.WebApi.Utilities.Internal;
+using System.Net.Http.Formatting;
+
+namespace GitHub.Services.Results.Client
+{
+    public class ResultsHttpClient
```
TingluoHuang, line 12: please use `RawHttpClientBase` as the base class and extend it with anything you need.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -182,35 +185,75 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                            context.Error(String.Format(Constants.Runner.SummaryUploadError, response.StatusCode));
+                            Trace.Info($"Step Summary file ({filePath}) failed to upload with status code {response.StatusCode} and message {response.Content}; skipping attachment upload");
+                            return;
+                        }
+                    }
+                    
+                    // Make upload complete call to url
+                    var stepSummaryUploadCompleteResponse = await resultsClient.StepSummaryUploadCompleteAsync(jobId, planId, stepId, fileSize, new CancellationToken());
+                    Trace.Info($"Step Summary file ({filePath}) upload marked as complete, response ok: {stepSummaryUploadCompleteResponse.Ok}");
+                } else {
```
TingluoHuang, line 225:
```suggestion
                }
                else
                {
```

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -182,35 +185,75 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                    Trace.Info("Using new Actions Results Job Summaries upload");
+
+                    var ResultsReceiverEndpointAddr = context.Global.Variables.Get("ResultsReceiverEndpoint");
+
+                    var token = context.Global.Variables.Get("System.AccessToken");
+                    var resultsClient = new ResultsHttpClient(token, ResultsReceiverEndpointAddr);
+                    var planId = context.Global.Variables.Get("System.PlanId");
+                    var jobId = context.Global.Variables.Get("System.JobId");
+                    var stepId = context.Global.Variables.Get("System.StepId");
+                    var stepSummaryUploadUrl = await resultsClient.GetStepSummaryUploadUrlAsync(jobId, planId, stepId, new CancellationToken());
```
TingluoHuang, line 199: You should not make any sync IO call (ex: HTTP request) when `processCommand` as it will block STDOUT/ERR from the process and slow down the customer's workflow job, follow the existing pattern to ask the `JobServerQueue` to process these files in the background.

src/Runner.Worker/FileCommandManager.cs
TingluoHuang: Please do the background way. 🙇

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -182,35 +185,75 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
                     return;
                 }
 
-                if (fileSize > AttachmentSizeLimit)
-                {
-                    context.Error(String.Format(Constants.Runner.UnsupportedSummarySize, AttachmentSizeLimit / 1024, fileSize / 1024));
-                    Trace.Info($"Step Summary file ({filePath}) is too large ({fileSize} bytes); skipping attachment upload");
+                Trace.Verbose($"Step Summary file exists: {filePath} and has a file size of {fileSize} bytes");
```
TingluoHuang, line 188: you can keep the change in your fork repo and not create PR to the main repo if the changes are not ready for review. 😄

src/Runner.Worker/FileCommandManager.cs
brittanyellich: @TingluoHuang This is great info, thank you! Since this will be uploading to a cloud service provider I'm guessing I should be creating a new ConcurrentQueue in the JobServerQueue to handle this type of file upload? Just wanted to verify 😄

src/Runner.Worker/FileCommandManager.cs
brittanyellich: Ah great, I will close this PR for now, then 😄

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
brittanyellich: Great, thank you! It isn't super clear how to go about getting the token from VssOAuthCredential to be able to make calls to other services. Do you have any insight on where that would be?

src/Runner.Worker/FileCommandManager.cs
TingluoHuang: sure, or fork the logic when right before upload files, to log/to attachment/to your new thing.

PR #2237, opened 2022-11-01
Expose GITHUB_JOB_DISPLAY_NAME in the environment

kylekthompson:
Hello 👋 I've been working a bit with GitHub Actions and the GitHub REST API and was finding it, in some cases, challenging to align the jobs that are running with the jobs that are returned in the API. One such reason is because there's a mismatch where the API returns the job's database identifier and display name, but the runner only has the YAML key for the job. By exposing the job display name, this will help get _closer_ to being able to line up a job running via the Actions runner to a job returned in the API (albeit not perfectly, as job display names need not be unique; separately, I've opened an issue with GitHub support to see if it'd be possible to return the YAML key in the API).

I'm happy to make changes/improvements/consider alternatives here, but this was a first pass (following similar work done in https://github.com/actions/runner/pull/366) that I was hoping could spark some conversation. I'd really _prefer_ to have the job ID in the environment so I could make this API call from an action, but I suspect there's an architectural reason precluding that:

```sh
gh api repos/{owner}/{repo}/actions/jobs/{job_id}
```

**Some screenshots of this working**:

Screen Shot 2022-11-01 at 12 48 35 PM
Screen Shot 2022-11-01 at 12 48 24 PM

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -749,6 +749,10 @@ namespace GitHub.Runner.Worker
             {
                 githubContext["job"] = new StringContextData(githubJob);
             }
+            if (!string.IsNullOrEmpty(message.JobDisplayName))
+            {
+                githubContext["job_display_name"] = new StringContextData(message.JobDisplayName);
+            }
             var githubDictionary = ExpressionValues["github"].AssertDictionary("github");
             foreach (var pair in githubDictionary)
             {
--- a/src/Runner.Worker/GitHubContext.cs
+++ b/src/Runner.Worker/GitHubContext.cs
@@ -21,6 +21,7 @@ namespace GitHub.Runner.Worker
             "graphql_url",
             "head_ref",
             "job",
+            "job_display_name",
             "output",
             "path",
             "ref_name",


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -749,6 +749,10 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             {
                 githubContext["job"] = new StringContextData(githubJob);
             }
+            if (!string.IsNullOrEmpty(message.JobDisplayName))
+            {
+                githubContext["job_display_name"] = new StringContextData(message.JobDisplayName);
+            }
```
kylekthompson, line 755: I did try to write a test for this, but was having a hard time finding the minimal amount of setup required to initialize an execution context and populate this. Before investing more time, I figured I'd open this and see if there was interest in the addition.

fhammerl: Thanks @kylekthompson , there's a relatively quick internal process for adding new context vars other than the runner change (update docs, auditing, consider future supportability etc). I'll get the ball rolling and will get back to you. 

Once it's green lit, as you suggested,
 we can look at some unit-tests for it.

fhammerl: I'd also recommend taking the feedback you had regarding the API and opening a discussion about it on the [community](https://github.com/orgs/community/discussions/) forum / discussions to get some eyes on it :)

kylekthompson: Much appreciated! I'll go ahead and open a discussion there as well, thanks!

kylekthompson: For posterity, here’s the community discussion post: https://github.com/orgs/community/discussions/38277

fhammerl:
@kylekthompson We've decided that this is something we're interested in, but instead of patching it in the runner, it is something we'd like to take care of server side.

I'll close this PR with that in mind and will ping you here once the change is ready.

kylekthompson:
Sure thing, thanks for looking into it further and for circling back!

mhumeSF:
@fhammerl What came of this? Is this available/exposed anywhere?

PR #2240, opened 2022-11-02
Forward parameters into run() func in run.sh.

TingluoHuang:
Regression introduced in #2233


src/Misc/layoutroot/run.sh
```diff
@@ -54,7 +54,7 @@ runWithManualTrap() {
 }
 
 if [[ -z "$RUNNER_MANUALLY_TRAP_SIG" ]]; then
-   run
+    run $*
```
TingluoHuang, line 57: We need to forward all args since we introduce the new method.

cory-miller (approved):

PR #2280, opened 2022-11-24
set env in ProcessInvoker sanitized

ruvceskistefan:


src/Runner.Sdk/ProcessInvoker.cs
```diff
@@ -264,7 +264,19 @@ public ProcessInvoker(ITraceWriter trace)
             {
                 foreach (KeyValuePair<string, string> kvp in environment)
                 {
+#if OS_WINDOWS
+                     if(!String.IsNullOrWhiteSpace(kvp.Key) && !String.IsNullOrWhiteSpace(kvp.Value))
```
thboop, line 268: The value could be null correct? IIRC we can set a key to a null value which would just unset the env. 

We may need to do something like

string tempKey = String.IsNullOrWhitespace(kvp.key) ? kvp.key : kvp.Key.Split("\0")[0];
string tempValue // same as above
then set it in the _proc.StartInfo

src/Runner.Sdk/ProcessInvoker.cs
```diff
@@ -264,7 +264,19 @@ public ProcessInvoker(ITraceWriter trace)
...
+                    {
+                        string tempKey = kvp.Key.Split("\0")[0];
+                        if(!String.IsNullOrWhiteSpace(tempKey))
+                        {
+                             _proc.StartInfo.Environment[tempKey.Trim()] = kvp.Value.Split("\0")[0].Trim();
```
thboop, line 273: What does the trim do here?

src/Runner.Sdk/ProcessInvoker.cs
thboop: What happens if the initial character is '\0'?

ruvceskistefan: If Key is `\0` we will skip it, if value is `\0` we will store key:empty

pje:
Thanks for the contribution, @ruvceskistefan! 

Can you add a small test for the new behavior?

https://github.com/actions/runner/blob/main/src/Test/L0/ProcessInvokerL0.cs

JoannaaKL (approved):
looks good

Issue #2165, opened 2022-09-30
Liquorix kernel breaks self-hosted install


Jire:
![unknown](https://user-images.githubusercontent.com/707385/193164007-c27be9a8-5674-4368-a248-99a436dfff06.png)

When trying to use `config.sh`

fhammerl:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

We use dotnet's [OSDescription](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osdescription?view=net-7.0) to get a string that we then include in our UserAgent header as a `ProductInfoHeaderValue`.

The constructor of `ProductInfoHeaderValue` validates if a string is compliant with UserAgent formatting. There aren't many requirements, but one of them is matching brackets -> `()` 
Brackets represent a `comment` in a UserAgent.

Liquorix has an OS description that contains mismatched brackets, so the constructor of `ProductInfoHeaderValue` throws.

I'd probably go with dropping all `()` to prevent nested UserAgent `comments`

PR #2288, opened 2022-11-30
Replace '(' and ')' with '[' and '] from OS.Description so it doesn't fail User-Agent header validation 

fhammerl:
Closes https://github.com/actions/runner/issues/2165
Uses behaviour from [UserAgentUtility.cs](https://github.com/actions/runner/blob/d0a4a41a632072512aee5aedc9c4430d50c06341/src/Sdk/WebApi/WebApi/Utilities/UserAgentUtility.cs#L39)?

## Bug

Linux Liquorix has an OS description that contains mismatched brackets, so the constructor of `ProductInfoHeaderValue` throws.

## Solution
Drop all `()` to prevent nested UserAgent `comments`
I don't know of other chars that would need to be sanitized.

## Context

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

We use dotnet's [OSDescription](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osdescription?view=net-7.0) to get a string that we then include in our UserAgent header as a `ProductInfoHeaderValue`.

The constructor of `ProductInfoHeaderValue` validates if a string is compliant with UserAgent formatting. There aren't many requirements, but one of them is matching brackets -> `()` 
Brackets represent a `comment` in a UserAgent.


nikola-jokic (dismissed):
LGTM

AvaStancu (approved):

src/Runner.Sdk/ActionPlugin.cs
```diff
@@ -73,7 +73,7 @@ public VssConnection InitializeVssConnection()
         {
             var headerValues = new List<ProductInfoHeaderValue>();
             headerValues.Add(new ProductInfoHeaderValue($"GitHubActionsRunner-Plugin", BuildConstants.RunnerPackage.Version));
-            headerValues.Add(new ProductInfoHeaderValue($"({RuntimeInformation.OSDescription.Trim()})"));
+            headerValues.Add(new ProductInfoHeaderValue($"({RuntimeInformation.OSDescription.Trim().Replace("(", "").Replace(")", "")})"));
```
fhammerl, line 76: Consider only dropping `(` and `)` if they are mismatched?

src/Runner.Sdk/Util/StringUtil.cs
```diff
@@ -123,5 +123,46 @@ public static string SubstringPrefix(string value, int count)
...
+                }
+            }
+            // We have 'openingBracketCount' number of unpaired '(' left
+            return openingBracketCount == 0;
+        }
+
+        // Fixes format violations e.g. https://github.com/actions/runner/issues/2165
+        public static string SanitizeUserAgentHeader(string header)
+        {
+            header = header.Trim();
```
jww3, line 159: nit:  since this `Trim` behavior was folded-in to this helper method, we should probably include some leading/trailing whitespace permutations in the new unit test.

src/Runner.Sdk/ActionPlugin.cs
```diff
@@ -73,7 +73,7 @@ public VssConnection InitializeVssConnection()
         {
             var headerValues = new List<ProductInfoHeaderValue>();
             headerValues.Add(new ProductInfoHeaderValue($"GitHubActionsRunner-Plugin", BuildConstants.RunnerPackage.Version));
-            headerValues.Add(new ProductInfoHeaderValue($"({RuntimeInformation.OSDescription.Trim()})"));
+            headerValues.Add(new ProductInfoHeaderValue($"({StringUtil.SanitizeUserAgentHeader(RuntimeInformation.OSDescription)})"));
```
jww3, line 76: Looks like we're already solving this a different way [elsewhere](https://github.com/actions/runner/blob/d0a4a41a632072512aee5aedc9c4430d50c06341/src/Sdk/WebApi/WebApi/Utilities/UserAgentUtility.cs#L39)?

src/Test/L0/Util/StringUtilL0.cs
```diff
@@ -186,5 +185,33 @@ public void ConvertStringToBool()
...
+        [Theory]
+        [InlineData("", "")]
+        [InlineData("(())", "[[]]")]
+        [InlineData("()()", "[][]")]
+        [InlineData("  Liquorix kernel OS Description is poorly formatted (linux version ", "Liquorix kernel OS Description is poorly formatted [linux version")]
+        [InlineData("Liquorix kernel OS Description is poorly formatted (linux version", "Liquorix kernel OS Description is poorly formatted [linux version")]
+        [InlineData("()((.", "[][[.")]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public void AreBracketsBalanced(string input, string expected)
```
jww3, line 198: This test should be renamed.

src/Runner.Sdk/Util/StringUtil.cs
```diff
@@ -123,5 +123,12 @@ public static string SubstringPrefix(string value, int count)
         {
             return value?.Substring(0, Math.Min(value.Length, count));
         }
+
+        // Fixes format violations e.g. https://github.com/actions/runner/issues/2165
+        public static string SanitizeUserAgentHeader(string header)
+        {
+            return header.Replace("(", "[").Replace(")", "]").Trim();
```
jww3, line 130: Can UserAgentUtility leverage this helper method as well?  or vice-versa?

jww3 (approved):

PR #2291, opened 2022-12-01
Listener stdout logging

AvaStancu:
Solves the Listener part needed for https://github.com/github/c2c-actions-runtime/issues/2065
Worker part to follow


src/Runner.Common/HostTraceListener.cs
```diff
@@ -89,10 +89,15 @@ public override void WriteLine(string message)
             }
 
             Flush();
+
+            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout)))
+            {
+                Console.WriteLine(message);
```
TingluoHuang, line 95: I am wondering whether we should add a `ConsoleTraceListener` base on the setting instead of writing to the console within the `TextWriterTraceListener`.

src/Runner.Common/Terminal.cs
```diff
@@ -27,14 +28,15 @@ public interface ITerminal : IRunnerService, IDisposable
 
     public sealed class Terminal : RunnerService, ITerminal
     {
-        public bool Silent { get; set; }
```
fhammerl, line 30: Repurposing the unused `Silent` prop as `PrintLogToStdout` to avoid double printing

src/Runner.Common/HostTraceListener.cs
AvaStancu: Thanks @TingluoHuang , changes have been added

src/Runner.Common/StdoutTraceListener.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace GitHub.Runner.Common
+{
+    public sealed class StdoutTraceListener : RunnerTraceListener
+    {
+        // Console.Out is the stdout stream
+        public StdoutTraceListener() : base(Console.Out) { }
```
TingluoHuang, line 11: this is exactly [ConsoleTraceListener](https://source.dot.net/System.Diagnostics.TextWriterTraceListener/R/d34b6cbcf7aebc48.html)

src/Runner.Common/Terminal.cs
TingluoHuang: `Silent` was used by testing, I think. do we want to keep that?

src/Runner.Common/Terminal.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
```
TingluoHuang, line 1: I do not quite get why we need to change this file.

src/Runner.Common/Tracing.cs
```diff
@@ -27,6 +27,11 @@ public Tracing(string name, ISecretMasker secretMasker, SourceSwitch sourceSwitc
             }
 
             _traceSource.Listeners.Add(traceListener);
+
+            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout)))
+            {
+                _traceSource.Listeners.Add(new StdoutTraceListener());
```
TingluoHuang, line 33: can we make the caller decide whether we need to pass a `StdoutTraceListener` to `Tracing()`?
ex: 
`public Tracing(string name, ISecretMasker secretMasker, SourceSwitch sourceSwitch, HostTraceListener traceListener, StdoutTraceListener stdoutListener = null)`

src/Runner.Common/HostTraceListener.cs
fhammerl: ConsoleTraceListener didn't format it, we implemented it in an `StdoutTraceListener` class instead

src/Runner.Common/StdoutTraceListener.cs
fhammerl: Yes, I was inspired by https://github.com/microsoft/referencesource/blob/master/System/compmod/system/diagnostics/ConsoleTraceListener.cs

TingluoHuang: what about leave `HostTraceListener.cs` unchanged and create
```
using System;
using System.Diagnostics;
using System.Globalization;
using GitHub.Runner.Sdk;

namespace GitHub.Runner.Common
{
    public sealed class RunnerStdoutTraceListener : ConsoleTraceListener
    {
        private readonly string _hostType;

        public RunnerStdoutTraceListener(string hostType)
        {
            _hostType = hostType;
        }

        // Copied and modified slightly from .Net Core source code. Modification was required to make it compile.
        // There must be some TraceFilter extension class that is missing in this source code.
        public override void TraceEvent(TraceEventCache eventCache, string source, TraceEventType eventType, int id, string message)
        {
            if (Filter != null && !Filter.ShouldTrace(eventCache, source, eventType, id, message, null, null, null))
            {
                return;
            }

            WriteHeader(source, eventType, id);
            WriteLine(message);
            WriteFooter(eventCache);
        }

        internal bool IsEnabled(TraceOptions opts)
        {
            return (opts & TraceOutputOptions) != 0;
        }

        // Altered from the original .Net Core implementation.
        private void WriteHeader(string source, TraceEventType eventType, int id)
        {
            string type = null;
            switch (eventType)
            {
                case TraceEventType.Critical:
                    type = "CRIT";
                    break;
                case TraceEventType.Error:
                    type = "ERR ";
                    break;
                case TraceEventType.Warning:
                    type = "WARN";
                    break;
                case TraceEventType.Information:
                    type = "INFO";
                    break;
                case TraceEventType.Verbose:
                    type = "VERB";
                    break;
                default:
                    type = eventType.ToString();
                    break;
            }

            Write(StringUtil.Format("[{0} {1:u} {2} {3}] ", _hostType, DateTime.UtcNow, type, source));
        }

        // Copied and modified slightly from .Net Core source code to make it compile. The original code
        // accesses a private indentLevel field. In this code it has been modified to use the getter/setter.
        private void WriteFooter(TraceEventCache eventCache)
        {
            if (eventCache == null)
                return;

            IndentLevel++;
            if (IsEnabled(TraceOptions.ProcessId))
                WriteLine("ProcessId=" + eventCache.ProcessId);

            if (IsEnabled(TraceOptions.ThreadId))
                WriteLine("ThreadId=" + eventCache.ThreadId);

            if (IsEnabled(TraceOptions.DateTime))
                WriteLine("DateTime=" + eventCache.DateTime.ToString("o", CultureInfo.InvariantCulture));

            if (IsEnabled(TraceOptions.Timestamp))
                WriteLine("Timestamp=" + eventCache.Timestamp);

            IndentLevel--;
        }
    }
}

```

src/Runner.Common/Terminal.cs
fhammerl: Undid the renaming, makes more sense to leave it `Silent`

fhammerl: `Silent = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout));`

This is needed to avoid double printing, should we set it somewhere else?

src/Runner.Common/Terminal.cs
```diff
@@ -168,6 +170,10 @@ public void WriteSection(string message)
                 Console.WriteLine($"# {message}");
                 Console.WriteLine();
             }
+            else
+            {
+                Trace.Info($"# {message}");
```
fhammerl, line 175: So that we still print to stdout through the `ConsoleListener` even if Terminal.cs is silent

src/Runner.Common/Tracing.cs
fhammerl: 👍

src/Runner.Common/HostContext.cs
```diff
@@ -94,6 +94,13 @@ public HostContext(string hostType, string logFile = null)
             this.SecretMasker.AddValueEncoder(ValueEncoders.PowerShellPreAmpersandEscape);
             this.SecretMasker.AddValueEncoder(ValueEncoders.PowerShellPostAmpersandEscape);
 
+            // Create StdoutTraceListener if ENV is set
+            StdoutTraceListener stdoutTraceListener = null;
+            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout)))
+            {
+                stdoutTraceListener = new StdoutTraceListener();
```
TingluoHuang, line 101: you might want to pass hostType to `StdoutTraceListener`, ex: `new StdoutTraceListener(hostType)`

src/Runner.Common/Terminal.cs
```diff
@@ -168,6 +170,10 @@ public void WriteSection(string message)
                 Console.WriteLine($"# {message}");
                 Console.WriteLine();
             }
+            else
```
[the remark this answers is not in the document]
TingluoHuang: might be wired to change those errors get printed to STDOUT instead STDERR.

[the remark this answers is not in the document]
fhammerl: > do we need to update all the Write*() functions in the file to have the else?
only the `Write*()` that don't have a `Trace.*` Instead of an `else`, should we actually Trace these always?


We could also do `Trace.Info($"WRITE SECTION # {message}");` and `Trace.Info($"WRITE SUCCESS √ {message}");` to keep it consistent with other traces in the other `Write*()` functions?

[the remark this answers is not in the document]
TingluoHuang: I was more focused on the `WriteError()` parts. 😄

TingluoHuang (approved):

PR #2301, opened 2022-12-05
Use results for uploading step summaries

brittanyellich:
This PR introduces a new way to queue and upload step summaries.


yacaovsnc (approved):
LGTM!

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            var request = new StepSummaryMetadataCreate()
+            {
+                workflow_job_run_backend_id = jobId,
+                workflow_run_backend_id = planId,
+                step_backend_id = stepId,
+                size = size,
+                uploaded_at = timestamp
+            };
+
+            var stepSummaryUploadCompleteRequest = m_resultsServiceUrl+"twirp/results.services.receiver.Receiver/CreateStepSummaryMetadata";
```
robherley, line 61: we have some helpers to join uri paths: https://github.com/actions/runner/blob/852a80fcbd21fa12d6b53fc6ab79029cc0f9b90a/src/Sdk/Common/Common/Utility/UriUtility.cs#L258

that way we don't need to worry about trailing or leading `/`

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            HttpContent content = new ObjectContent<GetSignedStepSummaryURLRequest>(request, m_formatter);
+            requestMessage.Content = content;
+
+            var response = await SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationToken: cancellationToken);
+            return await ReadJsonContentAsync<GetSignedStepSummaryURLResponse>(response, cancellationToken);
+        }
+
+        public async Task StepSummaryUploadCompleteAsync(string jobId, string planId, string stepId, long size, CancellationToken cancellationToken)
+        {
+            var timestamp = DateTime.UtcNow.ToString("yyyy-MM-dd'T'HH:mm:ss.fffK");
```
robherley, line 51: we talked about this in dm's, but our proto uses timestamp which is seconds and fractions of seconds at utc epoch

```json
"uploaded_at": {
   "seconds": 1670271570,
   "nanos": 110206903
}
```

it'd be great to use that for consistency instead of relying on string parsing + rfc3339

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Services.Results.Contracts;
+using System.Net.Http.Formatting;
+using Sdk.WebApi.WebApi;
+
+namespace GitHub.Services.Results.Client
+{
+    public class ResultsHttpClient : RawHttpClientBase
```
TingluoHuang, line 13: one thing to check if you plan to use `ResultsHttpClient` in production for customers, does the RawHttpClientBase support all the network settings for the runner, ex: proxy, SSL cert, etc?

TingluoHuang: it should, but I didn't review the initial PR and am not sure whether we implement all required pieces, it was added to testing RunService only in the dev environment.

yacaovsnc (dismissed):

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
yacaovsnc: From what I can tell, `RawHttpMessagehandler.cs` handles [timeouts](https://github.com/actions/runner/blob/main/src/Sdk/Common/Common/RawHttpMessageHandler.cs#L129), [retries](https://github.com/actions/runner/blob/main/src/Sdk/Common/Common/RawHttpMessageHandler.cs#L145) and has [partial support for proxy](https://github.com/actions/runner/blob/main/src/Sdk/Common/Common/RawHttpMessageHandler.cs#L83) - it doesn't apply proxy authentication.  

It also doesn't support certs which is okay in Results case since we don't need to support self signed certificate yet.  

I will try to add the proxy authentication support in the other issue created.

src/Runner.Common/Constants.cs
```diff
@@ -164,7 +164,8 @@ public static class Features
...
-            public static readonly string Node12DetectedAfterEndOfLife = "Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: {0}. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.";
+            public static readonly string SummaryUploadError = "$GITHUB_STEP_SUMMARY upload aborted, an error occurred when uploading the summary. For more information see: https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions#adding-a-markdown-summary";
+            public static readonly string Node12DetectedAfterEndOfLife = "Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: {0}";
```
TingluoHuang, line 168: do you have a bad merge/rebase? the `Node12DetectedAfterEndOfLife` looks a little bit different.

src/Runner.Common/JobServer.cs
```diff
@@ -143,6 +148,11 @@ public void InitializeWebsocketClient(ServiceEndpoint serviceEndpoint)
             InitializeWebsocketClient(TimeSpan.Zero);
         }
 
+        public void InitializeResultsClient(Uri uri, string token) {
+            var httpMessageHandler = HostContext.CreateHttpClientHandler();
+            this._resultsClient = new ResultsHttpClient(uri, httpMessageHandler, token, true);
```
TingluoHuang, line 153:
```suggestion
            this._resultsClient = new ResultsHttpClient(uri, httpMessageHandler, token, disposeHandler: true);
```

src/Runner.Common/JobServer.cs
```diff
@@ -305,6 +315,17 @@ public Task<TaskAttachment> CreateAttachmentAsync(Guid scopeIdentifier, string h
...
         }
 
+        public Task CreateResultsStepSymmaryAsync(string planId, string jobId, string stepId, string file, CancellationToken cancellationToken)
+        {
+            if (_resultsClient != null) 
+            {
+                return _resultsClient.UploadStepSummaryAsync(jobId, planId, stepId, file, cancellationToken: cancellationToken);
+            }
+            Trace.Info("Results client is not initialized. Skipping the step summary upload.");
+            return Task.CompletedTask;
```
TingluoHuang, line 325: do we need to skip or throw?

src/Runner.Common/JobServerQueue.cs
```diff
@@ -20,6 +20,7 @@ public interface IJobServerQueue : IRunnerService, IThrottlingReporter
         void Start(Pipelines.AgentJobRequestMessage jobRequest);
         void QueueWebConsoleLine(Guid stepRecordId, string line, long? lineNumber = null);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
+        void QueueResultsSummaryUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, string stepId, bool deleteSource);
```
TingluoHuang, line 23: I would suggest renaming to `QueueSummaryUpload`.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -30,6 +31,7 @@ public sealed class JobServerQueue : RunnerService, IJobServerQueue
         private static readonly TimeSpan _delayForWebConsoleLineDequeue = TimeSpan.FromMilliseconds(500);
         private static readonly TimeSpan _delayForTimelineUpdateDequeue = TimeSpan.FromMilliseconds(500);
         private static readonly TimeSpan _delayForFileUploadDequeue = TimeSpan.FromMilliseconds(1000);
+        private static readonly TimeSpan _delayForResultsSummaryUploadDequeue = TimeSpan.FromMilliseconds(1000);
```
TingluoHuang, line 34: I would rename everything in this file to not include `Result`, `IJobServerQueue` doesn't need to know it's uploading stuff to ResultService.

src/Runner.Common/JobServer.cs
```diff
@@ -22,11 +24,13 @@ public interface IJobServer : IRunnerService, IAsyncDisposable
         Task ConnectAsync(VssConnection jobConnection);
 
         void InitializeWebsocketClient(ServiceEndpoint serviceEndpoint);
+        void InitializeResultsClient(Uri uri, string token);
 
         // logging and console
         Task<TaskLog> AppendLogContentAsync(Guid scopeIdentifier, string hubName, Guid planId, int logId, Stream uploadStream, CancellationToken cancellationToken);
         Task AppendTimelineRecordFeedAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, Guid timelineRecordId, Guid stepId, IList<string> lines, long? startLine, CancellationToken cancellationToken);
         Task<TaskAttachment> CreateAttachmentAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, Guid timelineRecordId, String type, String name, Stream uploadStream, CancellationToken cancellationToken);
+        Task CreateResultsStepSymmaryAsync(string planId, string jobId, string stepId, string file, CancellationToken cancellationToken);
```
TingluoHuang, line 33:
```suggestion
        Task CreateStepSymmaryAsync(string planId, string jobId, string stepId, string file, CancellationToken cancellationToken);
```

TingluoHuang: we are creating a summary, and it happens to upload to the ResultService.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -93,6 +98,20 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest)
 
             _jobServer.InitializeWebsocketClient(serviceEndPoint);
 
+            // This code is usually wrapped by an instance of IExecutionContext which isn't available here.
+            jobRequest.Variables.TryGetValue("system.github.results_endpoint", out VariableValue resultsEndpointVariable);
```
TingluoHuang, line 102: should we use the SystemConnection to pass the URL? we have the same pattern for passing `CacheServerUrl` and `GenerateIdTokenUrl` from the service to the runner.

src/Runner.Common/Constants.cs
yacaovsnc: If the PR looks okay overall, do you mind if we merge this and then I will address all the comments in a separate PR.  Brittany is out and I have no permission to update her branch.  😭

src/Runner.Common/JobServerQueue.cs
```diff
@@ -665,6 +768,42 @@ private async Task UploadFile(UploadFileInfo file)
                 }
             }
         }
+
+    private async Task UploadResultsSummaryFile(ResultsSummaryUploadFileInfo file)
```
TingluoHuang, line 772:
```suggestion
        private async Task UploadResultsSummaryFile(ResultsSummaryUploadFileInfo file)
```

src/Runner.Common/JobServerQueue.cs
```diff
@@ -665,6 +768,42 @@ private async Task UploadFile(UploadFileInfo file)
...
+        {
+            bool uploadSucceed = false;
+            try
+            {
+                Trace.Info($"Starting to upload summary file to results service {file.Type}, {file.Name}, {file.Path}");
+
```
TingluoHuang, line 778:
```suggestion
```

src/Runner.Common/JobServerQueue.cs
```diff
@@ -665,6 +768,42 @@ private async Task UploadFile(UploadFileInfo file)
...
+    private async Task UploadResultsSummaryFile(ResultsSummaryUploadFileInfo file)
+        {
+            bool uploadSucceed = false;
+            try
+            {
+                Trace.Info($"Starting to upload summary file to results service {file.Type}, {file.Name}, {file.Path}");
+                
+
+                if (String.Equals(file.Type, ChecksAttachmentType.ResultsStepSummary, StringComparison.OrdinalIgnoreCase))
```
TingluoHuang, line 780: will there every be other types of files?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: What's the difference and the advantage of using `SystemConnection`?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Not planned yet.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -204,6 +230,29 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
...
+        public void QueueResultsSummaryUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, string stepId, bool deleteSource)
+        {
+            ArgUtil.NotEmpty(timelineId, nameof(timelineId));
+            ArgUtil.NotEmpty(timelineRecordId, nameof(timelineRecordId));
+
+            // all parameter not null, file path exist.
+            var newFile = new ResultsSummaryUploadFileInfo()
+            {
+                TimelineId = timelineId,
+                TimelineRecordId = timelineRecordId,
```
TingluoHuang, line 242: do we need these 2 fields? `TimelineId` and `TimelineRecordId`

src/Runner.Common/JobServerQueue.cs
```diff
@@ -204,6 +230,29 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
             _fileUploadQueue.Enqueue(newFile);
         }
 
+        public void QueueResultsSummaryUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, string stepId, bool deleteSource)
```
TingluoHuang, line 233: do we need to pass `Guid timelineId, Guid timelineRecordId` anymore?

src/Runner.Common/JobServerQueue.cs
```diff
@@ -683,6 +822,20 @@ internal class UploadFileInfo
         public bool DeleteSource { get; set; }
     }
 
+    internal class ResultsSummaryUploadFileInfo
```
TingluoHuang, line 825: make sure to remove all fields that you don't need.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -749,6 +750,7 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             {
                 githubContext["job"] = new StringContextData(githubJob);
             }
+
```
TingluoHuang, line 753:
```suggestion
```

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Yeah we still need `planId`, `jobId` and `stepId` in Results.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -846,6 +848,20 @@ public void QueueAttachFile(string type, string name, string filePath)
             _jobServerQueue.QueueFileUpload(_mainTimelineId, _record.Id, type, name, filePath, deleteSource: false);
         }
 
+        public void QueueResultsSummaryFile(string type, string name, string filePath, string stepId)
```
TingluoHuang, line 851: do we need to pass type any more since it's always a summery type?

src/Sdk/DTWebApi/WebApi/TaskAttachment.cs
```diff
@@ -107,5 +107,6 @@ public class CoreAttachmentType
     public class ChecksAttachmentType
     {
         public static readonly String StepSummary = "Checks.Step.Summary";
+        public static readonly String ResultsStepSummary = "Checks.Results.Step.Summary";
```
TingluoHuang, line 110: why do you need a different type?
I assume the object will send to a different service endpoint.

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
```diff
@@ -68,6 +68,20 @@ public class RawHttpClientBase: IDisposable
...
+        protected RawHttpClientBase(
+            Uri baseUrl,
+            HttpMessageHandler pipeline,
+            string token,
+            bool disposeHandler)
+        {
+            m_client = new HttpClient(pipeline, disposeHandler);
+
+            // Disable their timeout since we handle it ourselves
+            m_client.Timeout = TimeSpan.FromMilliseconds(-1.0);
```
TingluoHuang, line 80: Don't do this, play keep the timeout.
What if the client can't event reach your service, do we want it to sit there forever?

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
```diff
@@ -68,6 +68,20 @@ public class RawHttpClientBase: IDisposable
             m_formatter = new VssJsonMediaTypeFormatter();
         }
 
+        protected RawHttpClientBase(
+            Uri baseUrl,
+            HttpMessageHandler pipeline,
+            string token,
```
TingluoHuang, line 74: where is this `token` get used in this method?

src/Sdk/WebApi/WebApi/Results/Contracts.cs
```diff
@@ -0,0 +1,36 @@
+namespace GitHub.Services.Results.Contracts
+{
+    public class GetSignedStepSummaryURLRequest
+    {
+        public string workflow_job_run_backend_id;
+        public string workflow_run_backend_id;
+        public string step_backend_id;
+    }
+
```
TingluoHuang, line 9:
```suggestion
```

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
yacaovsnc: I believe this is for a different purpose.  We will be uploading directly to Blob Storage, so maybe the intention was to allow an upload for longer?

However I agree we shouldn't change this timeout value.  I will check, but I don't believe today when uploading to Actions Service we have a different timeout value.

src/Runner.Worker/ExecutionContext.cs
yacaovsnc: Yeah we should be able to get rid of those types.  The function should only care about summaries.

src/Sdk/WebApi/WebApi/Results/Contracts.cs
```diff
@@ -0,0 +1,36 @@
+namespace GitHub.Services.Results.Contracts
```
TingluoHuang, line 1: move this file up in the folder structure.
`src/Sdk/Results/WebApi/Contract.cs`.
same for the `ResultsHttpClient.cs`

src/Sdk/WebApi/WebApi/Results/Contracts.cs
```diff
@@ -0,0 +1,36 @@
...
+        public long size;
+        public string uploaded_at;
+    }
+
+    public class CreateStepSummaryMetadataResponse
+    {
+        public bool ok;
+    }
+
+    public static class BlobStorageTypes {
```
TingluoHuang, line 32:
```suggestion
    public static class BlobStorageTypes
    {
```

src/Sdk/WebApi/WebApi/Results/Contracts.cs
```diff
@@ -0,0 +1,36 @@
+namespace GitHub.Services.Results.Contracts
+{
+    public class GetSignedStepSummaryURLRequest
```
TingluoHuang, line 3: are these `DataContract`?
if so, you might want to follow the C# conversion in other file and adding `[DataContract]` and `[DataMember]` to all the class/property.

TingluoHuang: Also, you need to name it `WorkflowJobRunBackendId` for styling.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            var request = new GetSignedStepSummaryURLRequest()
+            {
+                workflow_job_run_backend_id = jobId,
+                workflow_run_backend_id = planId,
+                step_backend_id = stepId
+            };
+
+            var stepSummaryUploadRequest = new Uri(m_resultsServiceUrl, "twirp/results.services.receiver.Receiver/GetStepSummarySignedBlobURL");
+
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Post, stepSummaryUploadRequest);
```
TingluoHuang, line 38: you need to dispose this with `using`

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", m_token);
+            requestMessage.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json"));
+
+            HttpContent content = new ObjectContent<GetSignedStepSummaryURLRequest>(request, m_formatter);
```
TingluoHuang, line 42: also need to dispose `HttpContent`

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            requestMessage.Content = content;
+
+            var response = await SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationToken: cancellationToken);
```
TingluoHuang, line 45: same for the response, need to dispose.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            return await ReadJsonContentAsync<GetSignedStepSummaryURLResponse>(response, cancellationToken);
+        }
+
+        public async Task StepSummaryUploadCompleteAsync(string jobId, string planId, string stepId, long size, CancellationToken cancellationToken)
```
TingluoHuang, line 49: public or private?

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            string token,
+            bool disposeHandler)
+            : base(baseUrl, pipeline, disposeHandler)
+        {
+            m_token = token;
+            m_resultsServiceUrl = baseUrl;
+            m_formatter = new JsonMediaTypeFormatter();
+        }
+
+        public async Task<GetSignedStepSummaryURLResponse> GetStepSummaryUploadUrlAsync(string jobId, string planId, string stepId, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
TingluoHuang: make the parameters in the right order, `planId, jobId, stepId`

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
TingluoHuang: make the parameters in the right order, `planId, jobId, stepId`

src/Sdk/WebApi/WebApi/Results/Contracts.cs
yacaovsnc: Those are Twirp service requests.  However I think they are just json anyway.  I will add the contract attributes.  

It's `WorkflowJobRun*`. 🤷🏼

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                workflow_job_run_backend_id = jobId,
+                workflow_run_backend_id = planId,
+                step_backend_id = stepId,
+                size = size,
+                uploaded_at = timestamp
+            };
+
+            var stepSummaryUploadCompleteRequest = new Uri(m_resultsServiceUrl, "twirp/results.services.receiver.Receiver/CreateStepSummaryMetadata");
+
+            HttpRequestMessage requestMessage = new HttpRequestMessage(HttpMethod.Post, stepSummaryUploadCompleteRequest);
```
TingluoHuang, line 63: same comments for all the dispose.

src/Runner.Common/Constants.cs
yacaovsnc: Okay I am going to fork her branch and create a new PR.  Looks like we have sufficient comments that should be addressed first.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            requestMessage.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse("application/json"));
+
+            HttpContent content = new ObjectContent<StepSummaryMetadataCreate>(request, m_formatter);
+            requestMessage.Content = content;
+
+            var response = await SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationToken: cancellationToken);
+            var jsonResponse = await ReadJsonContentAsync<CreateStepSummaryMetadataResponse>(response, cancellationToken);
+            if (!jsonResponse.ok)
+            {
+                throw new Exception($"Failed to mark step summary upload as complete, status code: {response.StatusCode}, ok: {jsonResponse.ok}, jobId: {jobId}, planId: {planId}, stepId: {stepId}, size: {size}, timestamp: {timestamp}");
```
TingluoHuang, line 74: will this exception be on the user's face? if so, we might need to only keep the high-level info, I don't think users want to see all the GUID about those planId/JobId/StepId.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        public static async Task<HttpResponseMessage> UploadFileAsync(string url, string blobStorageType, FileStream file, CancellationToken cancellationToken) {
+            // Upload the file to the url
+            var httpClient = new HttpClient();
```
TingluoHuang, line 80: why do we create a new HttpClient if we are using the RawHttpClientBase? should we use the `RawHttpClientBase.SendAsync`?

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            var request = new HttpRequestMessage(HttpMethod.Put, url)
+            {
+                Content = new StreamContent(file)
+            };
+
+            if (blobStorageType == BlobStorageTypes.AzureBlobStorage) {
```
TingluoHuang, line 86:
```suggestion
            if (blobStorageType == BlobStorageTypes.AzureBlobStorage)
            {
```

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            if (!jsonResponse.ok)
+            {
+                throw new Exception($"Failed to mark step summary upload as complete, status code: {response.StatusCode}, ok: {jsonResponse.ok}, jobId: {jobId}, planId: {planId}, stepId: {stepId}, size: {size}, timestamp: {timestamp}");
+            }
+        }
+
+        public static async Task<HttpResponseMessage> UploadFileAsync(string url, string blobStorageType, FileStream file, CancellationToken cancellationToken) {
+            // Upload the file to the url
+            var httpClient = new HttpClient();
+            var request = new HttpRequestMessage(HttpMethod.Put, url)
```
TingluoHuang, line 81: also same about dispose.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                Content = new StreamContent(file)
+            };
+
+            if (blobStorageType == BlobStorageTypes.AzureBlobStorage) {
+                request.Content.Headers.Add("x-ms-blob-type", "BlockBlob");
+            }
+
+            using (var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)) {
+                if (!response.IsSuccessStatusCode) {
```
TingluoHuang, line 91:
```suggestion
            using (var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken))
            {
                if (!response.IsSuccessStatusCode)
                {
```

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                    throw new Exception($"Failed to upload file, status code: {response.StatusCode}, reason: {response.ReasonPhrase}, blobStorageType: {blobStorageType}");
```
TingluoHuang, line 92: same about whether this will be on user's face.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            var response = await SendAsync(requestMessage, HttpCompletionOption.ResponseContentRead, cancellationToken: cancellationToken);
+            var jsonResponse = await ReadJsonContentAsync<CreateStepSummaryMetadataResponse>(response, cancellationToken);
+            if (!jsonResponse.ok)
+            {
+                throw new Exception($"Failed to mark step summary upload as complete, status code: {response.StatusCode}, ok: {jsonResponse.ok}, jobId: {jobId}, planId: {planId}, stepId: {stepId}, size: {size}, timestamp: {timestamp}");
+            }
+        }
+
+        public static async Task<HttpResponseMessage> UploadFileAsync(string url, string blobStorageType, FileStream file, CancellationToken cancellationToken) {
```
TingluoHuang, line 78: public or private?

TingluoHuang, line 78:
```suggestion
        public static async Task<HttpResponseMessage> UploadFileAsync(string url, string blobStorageType, FileStream file, CancellationToken cancellationToken)
        {
```

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                    throw new Exception($"Failed to upload file, status code: {response.StatusCode}, reason: {response.ReasonPhrase}, blobStorageType: {blobStorageType}");
+                }
+                return response;
+            }
+        }
+
+        // Handle file upload for step summary
+        public async Task UploadStepSummaryAsync(string jobId, string planId, string stepId, string file, CancellationToken cancellationToken)
+        {
+            try {
```
TingluoHuang, line 101:
```suggestion
            try
            {
```

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+            using (var response = await httpClient.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken)) {
+                if (!response.IsSuccessStatusCode) {
+                    throw new Exception($"Failed to upload file, status code: {response.StatusCode}, reason: {response.ReasonPhrase}, blobStorageType: {blobStorageType}");
+                }
+                return response;
+            }
+        }
+
+        // Handle file upload for step summary
+        public async Task UploadStepSummaryAsync(string jobId, string planId, string stepId, string file, CancellationToken cancellationToken)
```
TingluoHuang, line 99: order the parameters.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
yacaovsnc: I believe the intention is this upload goes to Blob, not Results - that said, I think the `Raw` http client should still work.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            try {
+                // Get the upload url
+                var uploadUrlResponse = await GetStepSummaryUploadUrlAsync(jobId, planId, stepId, cancellationToken);
+
+                // Do we want to throw an exception here or should we just be uploading/truncating the data
+                var fileSize = new FileInfo(file).Length;
+                if (fileSize > uploadUrlResponse.soft_size_limit)
+                {
+                    throw new Exception($"File size is larger than the upload url allows, file size: {fileSize}, upload url size: {uploadUrlResponse.soft_size_limit}, blobType: {uploadUrlResponse.blob_storage_type}");
```
TingluoHuang, line 109: does the user care about the storage type?

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                // Upload the file
+                using (var fileStream = new FileStream(file, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, true))
+                {
+                    var response = await UploadFileAsync(uploadUrlResponse.summary_url, uploadUrlResponse.blob_storage_type, fileStream, cancellationToken);
+                }
+
+                // Send step summary upload complete message
+                await StepSummaryUploadCompleteAsync(jobId, planId, stepId, fileSize, cancellationToken);
+            }
+            catch (Exception ex) {
```
TingluoHuang, line 121:
```suggestion
            catch (Exception ex)
            {
```

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
```diff
@@ -0,0 +1,130 @@
...
+                throw new Exception("Failed to upload step summary to results", ex);
```
TingluoHuang, line 122: will this print out a nest error message with tons of stack about the inner exception?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Going to keep is as a variable for now.  The `SystemConnction` has authorization parameters and for Results we are just sticking with the system access token.

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
yacaovsnc: Hmmm, I see this was already there in another block.  We don't need to introduce a new constructor so I removed it.

src/Sdk/WebApi/WebApi/Results/ResultsHttpClient.cs
yacaovsnc: This is caught at `JobServerQueue` eventually and we do a trace.  

I can remove the ids.

src/Runner.Common/JobServerQueue.cs
TingluoHuang: the auth in the `SystemConnection` is exactly the system access token.
the advantage is aline with the existing pattern, the `SystemConnection` was holding some connection info back to the service, I think the result service URL is just one of them.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -683,6 +814,19 @@ internal class UploadFileInfo
         public bool DeleteSource { get; set; }
     }
 
+    internal class SummaryUploadFileInfo
+    {
+        public Guid TimelineId { get; set; }
+        public Guid TimelineRecordId { get; set; }
```
TingluoHuang, line 820: I still don't get why we need these fields.
```
await _jobServer.CreateStepSymmaryAsync(file.PlanId, file.JobId, file.StepId, file.Path, cancellationTokenSource.Token);
```

Do we only pass PlanId/JobId/StepId/Path to the `_jobServer`?

PR #2307, opened 2022-12-07
Made worker logs available to stdout

AvaStancu:
Solves the Worker part needed for https://github.com/github/c2c-actions-runtime/issues/2065

Console sample
Screenshot 2022-12-08 at 19 15 17

Runner logs file sample
Screenshot 2022-12-08 at 19 15 54

Worker logs file sample
Screenshot 2022-12-08 at 19 17 14

### DIFF
--- a/src/Runner.Common/Terminal.cs
+++ b/src/Runner.Common/Terminal.cs
@@ -18,7 +18,7 @@ namespace GitHub.Runner.Common
         string ReadSecret();
         void Write(string message, ConsoleColor? colorCode = null);
         void WriteLine();
-        void WriteLine(string line, ConsoleColor? colorCode = null);
+        void WriteLine(string line, ConsoleColor? colorCode = null, bool skipTracing = false);
         void WriteError(Exception ex);
         void WriteError(string line);
         void WriteSection(string message);
@@ -116,9 +116,12 @@ namespace GitHub.Runner.Common
 
         // Do not add a format string overload. Terminal messages are user facing and therefore
         // should be localized. Use the Loc method in the StringUtil class.
-        public void WriteLine(string line, ConsoleColor? colorCode = null)
+        public void WriteLine(string line, ConsoleColor? colorCode = null, bool skipTracing = false)
         {
-            Trace.Info($"WRITE LINE: {line}");
+            if (!skipTracing)
+            {
+                Trace.Info($"WRITE LINE: {line}");
+            }
             if (!Silent)
             {
                 if (colorCode != null)
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -512,14 +512,14 @@ namespace GitHub.Runner.Listener
                             var completedTask = await Task.WhenAny(renewJobRequest, workerProcessTask, Task.Delay(-1, jobRequestCancellationToken));
                             if (completedTask == workerProcessTask)
                             {
+                                var detailInfo = string.Join(Environment.NewLine, workerOutput);
+                                Trace.Info(detailInfo);
                                 // worker finished successfully, complete job request with result, attach unhandled exception reported by worker, stop renew lock, job has finished.
                                 int returnCode = await workerProcessTask;
                                 Trace.Info($"Worker finished for job {message.JobId}. Code: " + returnCode);
 
-                                string detailInfo = null;
                                 if (!TaskResultUtil.IsValidReturnCode(returnCode))
                                 {
-                                    detailInfo = string.Join(Environment.NewLine, workerOutput);
                                     Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
 
                                     var jobServer = HostContext.GetService<IJobServer>();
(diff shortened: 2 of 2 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -512,6 +512,7 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                             var completedTask = await Task.WhenAny(renewJobRequest, workerProcessTask, Task.Delay(-1, jobRequestCancellationToken));
                             if (completedTask == workerProcessTask)
                             {
+                                Trace.Info(string.Join(Environment.NewLine, workerOutput));
```
JoannaaKL, line 515: I'd propose to extract it to the variable`detailInfo = string.Join(Environment.NewLine, workerOutput);` and reuse here and in lines 531 and 534 :)

AvaStancu: SGTM! :) Amended, thanks!

JoannaaKL (approved):

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -512,14 +512,14 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                             var completedTask = await Task.WhenAny(renewJobRequest, workerProcessTask, Task.Delay(-1, jobRequestCancellationToken));
                             if (completedTask == workerProcessTask)
                             {
+                                var detailInfo = string.Join(Environment.NewLine, workerOutput);
+                                Trace.Info(detailInfo);
```
fhammerl, line 516: `Trace.Info` formats the input still - the first line would look like

`[RUNNER YYYY-MM-DD:HH:mm:ss TYPE] [WORKER YYYY`

Would not mislead a human, but it can trip up some automation

fhammerl, line 516: This will dump the entire collection of Worker logs in a single string - it can easily be thousands of lines long.
Should we consider trying to print worker logs asap?

fhammerl: In case the runner loses connection / its pod is destroyed, waiting until the end of the job risks losing the logs alltogether

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -422,6 +422,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                                             lock (_outputLock)
                                             {
                                                 workerOutput.Add(stdout.Data);
+                                                if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout)))
```
TingluoHuang, line 425: store the result of the `StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.PrintLogToStdout))` in to a variable outside of the `using`, so we don't need to read env on every line we get from the worker.

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -422,6 +422,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+                                                {
+                                                    Console.WriteLine(stdout.Data);
```
TingluoHuang, line 427: Can we use `ITerminal` here to writeline?

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -422,6 +422,10 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
                                             lock (_outputLock)
                                             {
                                                 workerOutput.Add(stdout.Data);
```
TingluoHuang, line 424: We might need to figure out how to handle the existing logic about stdout from worker.
Today, if the worker printout anything to STDOUT, we will create an issue on the service for it, since we assume anything output from the worker means something busted...

TingluoHuang: you can check the code about how we use `workerOutput` later in the file.

TingluoHuang, line 424:
```
if(!stdout.Data.StartsWith("[WORKER "))
{
    workerOutput.Add(stdout.Data);
}
```

src/Runner.Common/Terminal.cs
```diff
@@ -16,6 +16,7 @@ public interface ITerminal : IRunnerService, IDisposable
         bool Silent { get; set; }
         string ReadLine();
         string ReadSecret();
+        void StdOutWriteLine(string line);
```
TingluoHuang, line 19: Let's add an extra-parameter to `void WriteLine(string line, ConsoleColor? colorCode = null, bool skipTracing = false);`


```
public void WriteLine(string line, ConsoleColor? colorCode = null, bool skipTracing = false)
        {
            if (!skipTracing)
            {
                Trace.Info($"WRITE LINE: {line}");
            }
            if (!Silent)
            {
                if (colorCode != null)
                {
                    Console.ForegroundColor = colorCode.Value;
                    Console.WriteLine(line);
                    Console.ResetColor();
                }
                else
                {
                    Console.WriteLine(line);
                }
            }
        }
```

TingluoHuang (approved):

PR #2311, opened 2022-12-09
[1742] Ensure multiple composite annoations are correctly written.

jww3:
This implementation uses a collector pattern to allow embedded `ExecutionContext's` to stash `Issue` objects for later processing by a non-embedded ancestor `ExecutionContext`.

### Context
Proposed fix for https://github.com/actions/runner/issues/1742

### Also:
 - Provide explicit constructor implementations for `ExecutionContext`
 - Leverage explicit constructors to solidify immutability of several `ExecutionContext` class members.
 - Use a consistent timestamp for `FinishTime` in `ExecutionContext::Complete`
 - Fixed erroneous call to `ExecutionContext::Complete` in `CompositeActionHandler.cs`
 - Addressed serialization inconsistencies in `TimelineRecord`

### Demonstration of Problem and Fix

image

source: https://github.com/bbq-beets/jww3-2022a/actions/runs/4445968310 

#### Workflow files

```
name: Issue 1742 Test
on:
  workflow_dispatch:
jobs:   
  demonstrate-bug:
    runs-on: ubuntu-latest
    steps:
      - name: Demonstrate the bug on a GitHub-hosted runner.
        run:  echo "Prior to the fix being released, this should demonstrate the bug."

      - name: Composite action
        uses: konradpabjan/composite-action-test@main

  verify-fix:
    runs-on: self-hosted
    steps:
      - name: Verification of the fix on a Custom Runner Build
        run:  echo "Now, we'll try and verify the fix."

      - name: Composite action
        uses: konradpabjan/composite-action-test@main
```
source:  https://github.com/bbq-beets/jww3-2022a/blob/main/.github/workflows/verify-fix-1742.yml

```
name: 'Hello World'
description: Greeting
runs:
  using: "composite"
  steps:
    - run: echo Hello there once!
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::error::Testing error"
      shell: bash
    - run: echo Hello there
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::error::Testing error"
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::warning::Testing warning"
      shell: bash
    - name: Testing the Error Workflow Command
      run: echo "::notice::Testing notice"
      shell: bash
```
source:  https://github.com/konradpabjan/composite-action-test/blob/main/action.yml


jww3:
The idea here is similar to @thboop's [proposed fix](https://github.com/actions/runner/pull/2116)

src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -294,7 +294,7 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps, ActionRunStage stage
                     // Evaluation error
                     Trace.Info("Caught exception from expression for embedded step.env");
                     step.ExecutionContext.Error(ex);
-                    step.ExecutionContext.Complete(TaskResult.Failed);
+                    SetStepConclusion(step, TaskResult.Failed);
```
jww3, line 297: @thboop called this out as a bug, noting that `::Complete` shouldn't be called on a Composite step.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -411,21 +432,29 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
                 this.Warning($"The job has experienced {TimeSpan.FromMilliseconds(_totalThrottlingDelayInMilliseconds).TotalSeconds} seconds total delay caused by server throttling.");
             }
 
+            DateTime now = DateTime.UtcNow;
```
jww3, line 435: fetch the current time once-and-only-once in this method.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -411,21 +432,29 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
...
+            _record.FinishTime = now;
             _record.PercentComplete = 100;
             _record.Result = _record.Result ?? TaskResult.Succeeded;
             _record.State = TimelineRecordState.Completed;
 
+            foreach (var issue in _embeddedIssueCollector)
+            {
+                // Before our main timeline's final QueueTimelineRecordUpdate,
+                //    inject any issues collected by embedded ExecutionContexts.
+                AddIssue(issue);
```
jww3, line 447: Is there a better point in time to do this work?  This was the best spot I could find -- right after other (more fundamental) `_record` updates and (more importantly) before the final call to `QueueTimelineRecordUpdate`

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -365,11 +388,9 @@ public void RegisterPostJobStep(IStep step)
                 child._logger.Setup(_mainTimelineId, recordId);
             }
 
-            child.IsEmbedded = isEmbedded;
             child.StepTelemetry.StepId = recordId;
             child.StepTelemetry.Stage = stage.ToString();
-            child.StepTelemetry.IsEmbedded = isEmbedded;
```
jww3, line 371: `child.IsEmbedded` and `child.StepTelemetry.IsEmbedded` are now initialized in the workhorse constructor.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -365,11 +388,9 @@ public void RegisterPostJobStep(IStep step)
...
-            child.StepTelemetry.StepContextName = child.GetFullyQualifiedContextName(); ;
```
jww3, line 372: only change to this line was to get rid of the dangling `;`

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -147,6 +148,29 @@ public sealed class ExecutionContext : RunnerService, IExecutionContext
...
+
+            _parentExecutionContext = parent;
+            this.IsEmbedded = embedded;
+            this.StepTelemetry = new ActionsStepTelemetry
+            {
+                IsEmbedded = embedded
+            };
+
+            //Embedded Execution Contexts pseudo-inherit their parent's embeddedIssueCollector.
+            _embeddedIssueCollector = embedded ? parent._embeddedIssueCollector : new();
```
jww3, line 171: Declaring these properties as immutable and assigning them within an explicit constructor makes it easier to guarantee that they're initialized properly and consistently.

jww3:
It's not clear what `lock` [this comment](https://github.com/actions/runner/pull/2311/files#diff-4085092f16aac893358021b20c7a92b02a504a1a99830bc88454ae0380638c45R577) is referring to:

```
// This is not thread safe, the caller need to take lock before calling issue()
```


None of the callers of `ExecutionContext::AddIssue` make any attempt to acquire a lock.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -124,16 +124,17 @@ public sealed class ExecutionContext : RunnerService, IExecutionContext
 
         private readonly TimelineRecord _record = new();
         private readonly Dictionary<Guid, TimelineRecord> _detailRecords = new();
+        private readonly List<Issue> _embeddedIssueCollector;
         private readonly object _loggerLock = new();
         private readonly object _matchersLock = new();
+        private readonly ExecutionContext _parentExecutionContext;
```
jww3, line 130: now declared as immutable.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -411,21 +432,32 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
...
             _record.Result = _record.Result ?? TaskResult.Succeeded;
             _record.State = TimelineRecordState.Completed;
 
+            // Before our main timeline's final QueueTimelineRecordUpdate,
+            //    inject any issues collected by embedded ExecutionContexts.
+            if (!this.IsEmbedded)
+            {
+                foreach (var issue in _embeddedIssueCollector)
+                {
+                    AddIssue(issue);
```
thboop, line 449: The AddIssue function has a side effect of writing a line to the job logs. Does this change make those changes write twice?

As a part of this refactor, we may want to remove the side effect from the AddIssue function, have one function that
- masks any secrets
- writes to a log
- returns a hydrated timelinerecordissue object
And a second that
- Adds that timelinerecordissue to the timeline record and queues an update.

src/Runner.Worker/ExecutionContext.cs
thboop: I'm not sure we care about the fundamental record updates on _record for composite steps (embedded).

for embedded steps it may just be okay to mask secrets and output to logs, and when we bubble fully up to the parent, then we can update only the parent's timeline record.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -411,21 +432,32 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
...
+
             _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
 
             // complete all detail timeline records.
             if (_detailTimelineId != Guid.Empty && _detailRecords.Count > 0)
             {
                 foreach (var record in _detailRecords)
                 {
-                    record.Value.FinishTime = record.Value.FinishTime ?? DateTime.UtcNow;
+                    record.Value.FinishTime = record.Value.FinishTime ?? now;
```
TingluoHuang, line 460: technically, we can remove all the code that deal with `_detailRecords` in a separate PR., those were from ADO and I am 90.999% sure we don't use them now.

src/Runner.Worker/ExecutionContext.cs
jww3: It doesn't appear that anything will get written to the job log as a result of this call.  All the write logic is gated on a non-null `logMessage` being passed as the second (optional) parameter.

see https://github.com/actions/runner/blob/7036627d4717dadc479a2917b6f79859681fd606/src/Runner.Worker/ExecutionContext.cs#L581

jww3: I created a one-off exploration of something along those lines [here](https://github.com/actions/runner/pull/2378)

jww3: This should be fixed, now.  I decided not to tease apart the logging calls from `AddIssue` since all callers except one desire/expect that behavior.  Instead, I replaced the log message override string parameter with a small struct that allows the caller to clearly communicate their intent.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -744,7 +744,7 @@ private List<TimelineRecord> MergeTimelineRecords(List<TimelineRecord> timelineR
             foreach (var record in mergedRecords)
             {
                 Trace.Verbose($"    Record: t={record.RecordType}, n={record.Name}, s={record.State}, st={record.StartTime}, {record.PercentComplete}%, ft={record.FinishTime}, r={record.Result}: {record.CurrentOperation}");
-                if (record.Issues != null && record.Issues.Count > 0)
+                if (record.Issues != null)
                 {
                     foreach (var issue in record.Issues)
```
jww3, line 749: `foreach` will just be a no-op if the `Count` is 0.  No need to check it explicitly.

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -1,4 +1,4 @@
-﻿using GitHub.DistributedTask.Pipelines.ContextData;
```
jww3, line 1: some of these files had a UTF-8 byte-order-mark (BOM).  The linter complained about the BOM, so I changed the file encoding to be BOMless.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -92,7 +98,7 @@ public interface IExecutionContext : IRunnerService
         void SetGitHubContext(string name, string value);
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
-        void AddIssue(Issue issue, string message = null);
+        void AddIssue(Issue issue, ExecutionContextLogOptions logOptions);
```
jww3, line 101: give callers a mechanism to opt-out of logging entirely for a particular call to `AddIssue`

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -413,21 +440,32 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
...
             _record.Result = _record.Result ?? TaskResult.Succeeded;
             _record.State = TimelineRecordState.Completed;
 
+            // Before our main timeline's final QueueTimelineRecordUpdate,
+            //    inject any issues collected by embedded ExecutionContexts.
+            if (!this.IsEmbedded)
+            {
+                foreach (var issue in _embeddedIssueCollector)
+                {
+                    AddIssue(issue, ExecutionContextLogOptions.None);
```
jww3, line 457: Suppress logging when embedded issues bubble-up.  (They were already logged when they were initially added.)

src/Runner.Worker/ExecutionContext.cs
jww3: 90.999% or 99.999% ?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -995,8 +1041,7 @@ public void PublishStepTelemetry()
                         StepTelemetry.FinishTime = _record.FinishTime;
                     }
 
-                    if (!IsEmbedded &&
-                        _record.Issues.Count > 0)
```
jww3, line 999: the only thing going on in this conditional is a `foreach`.  `foreach` is a no-op for an empty list.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -703,17 +703,17 @@ private List<TimelineRecord> MergeTimelineRecords(List<TimelineRecord> timelineR
                     timelineRecord.State = rec.State ?? timelineRecord.State;
                     timelineRecord.WorkerName = rec.WorkerName ?? timelineRecord.WorkerName;
 
-                    if (rec.ErrorCount != null && rec.ErrorCount > 0)
+                    if (rec.ErrorCount > 0)
                     {
                         timelineRecord.ErrorCount = rec.ErrorCount;
                     }
```
jww3, line 709: `ErrorCount`, `WarningCount`, and `NoticeCount` are no longer nullable (previously `Int32?`, now `Int32`)

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -1,78 +1,86 @@
...
-                this.PreviousAttempts.AddRange(recordToBeCloned.PreviousAttempts);
-            }
+                if (recordToBeCloned.m_previousAttempts?.Count > 0)
+                {
+                    this.m_previousAttempts.AddRange(recordToBeCloned.m_previousAttempts);
+                }
 
-            if (recordToBeCloned.m_variables?.Count > 0)
-            {
-                this.m_variables = recordToBeCloned.Variables.ToDictionary(k => k.Key, v => v.Value.Clone());
```
jww3, line 75: This was a bug.  The original might have had a case-insensitive Variables Dictionary, but the clone's would be case-sensitive.

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -209,35 +217,46 @@ public TimelineReference Details
...
-            set;
+            get
+            {
+                return m_errorCount.GetValueOrDefault(0);
+            }
+            set
+            {
+                m_errorCount = value;
+            }
         }
```
jww3, line 230: I'm not sure why these were nullable in the first place.  I'm guessing it was for backwards compatibility with older serialized versions that lacked these fields?

Making them non-nullable has some compelling advantages and eliminates subtle bugs like not having to worry about the impact null might have on everyday operators like `++`, `<`, `>`, etc.

see https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#lifted-operators

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -209,35 +217,46 @@ public TimelineReference Details
...
 
         public List<Issue> Issues
         {
             get
             {
-                if (m_issues == null)
-                {
-                    m_issues = new List<Issue>();
-                }
```
jww3, line 240: `m_issues`, `m_previousAttempts`, and `m_variables` are now instantiated once and only once within `EnsureInitialized`.

I wanted to declare them as `readonly` but had to stop short of that because of the quirks of `DataContractSerializer`

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -299,13 +310,53 @@ public TimelineRecord Clone()
...
+            m_warningCount = m_warningCount ?? 0;
+            m_noticeCount = m_noticeCount ?? 0;
+
+            m_issues = m_issues ?? new List<Issue>();
+            m_previousAttempts = m_previousAttempts ?? new List<TimelineAttempt>();
+            this.Attempt = Math.Max(this.Attempt, 1);
+
+            // Ensure whatever content may have been deserialized for m_values is backed by a case-insensitive Dictionary.
+            var empty = Enumerable.Empty<KeyValuePair<string, VariableValue>>();
+            m_variables = new Dictionary<string, VariableValue>(m_variables ?? empty, StringComparer.OrdinalIgnoreCase);
```
jww3, line 341: `DataContractSerializer` has no way of knowing that what we really want is a case-insensitive Dictionary.  We have to take whatever data may have been deserialized and re-bundle it into a new (case-insensitive) Dictionary.

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -299,13 +310,53 @@ public TimelineRecord Clone()
...
+        }
+
+        [DataMember(Name = nameof(ErrorCount), Order = 40)]
+        private Int32? m_errorCount;
+
+        [DataMember(Name = nameof(WarningCount), Order = 50)]
+        private Int32? m_warningCount;
+
+        [DataMember(Name = nameof(NoticeCount), Order = 55)]
+        private Int32? m_noticeCount;
```
jww3, line 351: We need to preserve nullability on these backing fields for as long as older versions of this type are floating around in serialized form.  Once they've all become irrelevant, we can do away with the backing fields and just have `get; set;` on the non-nullable properties.

vmjoseph:
> It's not clear what lock [this comment](https://github.com/actions/runner/pull/2311/files#diff-4085092f16aac893358021b20c7a92b02a504a1a99830bc88454ae0380638c45R577) is referring to:
> 
> // This is not thread safe, the caller need to take lock before calling issue()
> None of the callers of ExecutionContext::AddIssue make any attempt to acquire a lock.
> 
> 
Hm, I think that comment may be referring to `Write()`'s lock statement here:
https://github.com/actions/runner/blob/bb7b1e825962e2d6b2041782341df851ac2eefa2/src/Runner.Worker/ExecutionContext.cs#L828-L836

src/Test/L0/Worker/OutputManagerL0.cs
```diff
@@ -984,10 +984,15 @@ public void CaptureTelemetryForGitUnsafeRepository()
...
-            _executionContext.Setup(x => x.AddIssue(It.IsAny<DTWebApi.Issue>(), It.IsAny<string>()))
-                .Callback((DTWebApi.Issue issue, string logMessage) =>
+            _executionContext.Setup(x => x.AddIssue(It.IsAny<DTWebApi.Issue>(), It.IsAny<ExecutionContextLogOptions>()))
+                .Callback((DTWebApi.Issue issue, ExecutionContextLogOptions logOptions) =>
                 {
-                    _issues.Add(new Tuple<DTWebApi.Issue, string>(issue, logMessage));
+                    var resolvedMessage = issue.Message;
+                    if (logOptions.WriteToLog && !string.IsNullOrEmpty(logOptions.LogMessageOverride))
+                    {
+                        resolvedMessage = logOptions.LogMessageOverride;
```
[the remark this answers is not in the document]
jww3: will do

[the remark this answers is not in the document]
jww3: done

vmjoseph (approved):
Looks good to me!

src/Runner.Common/JobServerQueue.cs
TingluoHuang: it's no-op... but you are wasting CPU cycles to do an extra `GetEnumerator()` and `MoveNext()`. 😄

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -559,14 +597,10 @@ public void Progress(int percentage, string currentOperation = null)
         }
 
         // This is not thread safe, the caller need to take lock before calling issue()
-        public void AddIssue(Issue issue, string logMessage = null)
+        public void AddIssue(Issue issue, ExecutionContextLogOptions logOptions)
```
TingluoHuang, line 600: Can we make `logOptions` defaults to `ExecutionContextLogOptions.Default` so you don't need to change all existing callers?
`public void AddIssue(Issue issue, ExecutionContextLogOptions logOptions = ExecutionContextLogOptions.Default)`

src/Test/L0/Worker/StepsRunnerL0.cs
```diff
@@ -590,7 +590,7 @@ private Mock<IActionRunner> CreateStep(TestHostContext hc, TaskResult result, st
             step.Setup(x => x.Condition).Returns(condition);
             step.Setup(x => x.ContinueOnError).Returns(new BooleanToken(null, null, null, continueOnError));
             step.Setup(x => x.Action)
-                .Returns(new DistributedTask.Pipelines.ActionStep()
+                .Returns(new GitHub.DistributedTask.Pipelines.ActionStep()
```
TingluoHuang, line 593: why do we need to change this?

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -1,78 +1,86 @@
-﻿using System;
+using System;
```
TingluoHuang, line 1: is this file changed in the actions service?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -149,6 +156,29 @@ public sealed class ExecutionContext : RunnerService, IExecutionContext
         private long _totalThrottlingDelayInMilliseconds = 0;
         private bool _stepTelemetryPublished = false;
 
+        public ExecutionContext()
+            : this(null, false)
```
TingluoHuang, line 160:
```suggestion
            : this(parent: null, embedded: false)
```
maybe ☝️ for better readability?

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: 99.9999999999% 😄

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: i would still keep it to avoid waste CPU.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -581,53 +615,59 @@ public void AddIssue(Issue issue, string logMessage = null)
...
+                    previousCountForIssueType = _record.NoticeCount++;
+                    break;
+            }
 
-                if (_record.ErrorCount < _maxIssueCount)
+            if (!string.IsNullOrEmpty(wellKnownTag))
+            {
+                if (previousCountForIssueType < _maxCountPerIssueType)
                 {
                     _record.Issues.Add(issue);
```
TingluoHuang, line 640: do we need to not add to the `_record.Issues` when it's embedded?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -92,7 +98,7 @@ public interface IExecutionContext : IRunnerService
         void SetGitHubContext(string name, string value);
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
-        void AddIssue(Issue issue, string message = null);
```
TingluoHuang, line 95: `void AddIssue(Issue issue, string message = null, bool writeToLog = true);`
☝️  will this help reduce the amount of code changes since we don't need to change all existing callers?

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: ignore this, for now, check the other comment at the top. 😄

src/Runner.Common/JobServerQueue.cs
jww3: which is probably about equal to a `get_Issues`, a `get_Count` and a `jmp` evaluation ?

src/Runner.Worker/ExecutionContext.cs
jww3: There are only 11 callers outside of L0 tests.  My usual rule-of-thumb for resorting to an optional param is around 200 pre-existing callers for an unexported method.

src/Test/L0/DistributedTask/WebApi/TimelineRecordL0.cs
```diff
@@ -0,0 +1,270 @@
+#nullable enable
```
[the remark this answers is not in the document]
jww3: I've included it there as well.  The intent of this test was to prove that I didn't break deserialization.

The definition of `TimelineRecord` in the service and the definition of `TimelineRecord` in `actions/runner` have diverged somewhat, so I'd feel more comfortable having these assurances in both places.

Longer term, I'd love to migrate these classes to a shared library that can be consumed by both client and server.

src/Test/L0/Worker/StepsRunnerL0.cs
jww3: The compiler was complaining about namespace ambiguity.

Without the qualification, it assumes that `DistributedTask` refers to new namespace `GitHub.Runner.Common.Tests.DistributedTask` (introduced in new test file, `TimelineRecordL0.cs`)

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
jww3: I ported these changes to the actions service in a separate PR.  There were already significant divergences between the two copies.  As I mentioned above, longer-term I'd love to migrate all the replicated code into a library that can be shared between client and server.

It's less of a refactor than it is a baselining.  It wasn't strictly necessary, but it makes `TimelineRecord` much easier to reason about and maintain moving forward.

There's no longer any ambiguity about whether members are instantiated (non-null) or not.

For example, `record.ErrorCount++` is guaranteed to work correctly now.

Previously, the only way to guarantee a non-null outcome was to do something like:

```
record.ErrorCount = (record.ErrorCount ?? 0) + 1;
```

src/Runner.Worker/ExecutionContext.cs
jww3: Good question.

`PublishStepTelemetry` (line 1038 in the new version) retains its long-standing guarantee that only issues on top-level (non-embedded) ExecutionContexts are exfiltrated.

That leaves `JobServerQueue.MergeTimelineRecords` as the only other place where `TimelineRecord.Issues` is accessed in a read operation.

I can't say for sure what the ramifications are of allowing issues from embedded `ExecutionContexts` to remain in "merged" `TimelineRecords`, but historically (it seems), the system has tolerated it?

src/Runner.Common/JobServerQueue.cs
TingluoHuang: I prefer to reduce changes in PR to only the required things to keep the review focus (especially if the change contains more files)

I am fine with the refactor if you like. 😄

src/Runner.Worker/ExecutionContext.cs
TingluoHuang: 10+ is a lot for the runner since it's a tiny codebase 😄

src/Runner.Worker/ExecutionContext.cs
thboop: I don't think embedded timeline records should be sent up to the server, so it likely doesn't matter. Only root step level timeline records are sent up. 

That being said, whether or not we should add them to the timeline records for embedded is really up to the implementation, it shouldn't matter either way because it just serves as a vehicle to send them up to the server. For confusion sake, it may make sense to not add them unless there is a specific reason we are adding them.

jww3: Fixed.  Note that the latest commit also avoids updating the record's Error/Warning/NoticeCount when the ExecutionContext is embedded.

jww3: ### Screenshot of verification run after this fix:

image

source:  https://github.com/bbq-beets/jww3-2022a/actions/runs/4914104452

TingluoHuang (approved):

Issue #2309, opened 2022-12-08
Display Name is not fully expanded for some contexts on skipped steps


thboop:
**Describe the bug**
We try to evaluate the name of a step many times during the course of a job. Whether or not that is successful depends on if we have access to all the correct [context values](https://docs.github.com/en/actions/learn-github-actions/contexts) at the time we are trying to evaluate. The very last evaluation is done just as we start a step, and is intended as a catch all to evaluate the step name if it hasn't been evaluated yet.

However, this catch all doesn't work if the step doesn't actually run. You can see in the workflow below skipped steps may or may not fully evaluate the name based on what contexts are used in the name.

Rather than doing this last evaluation as the step starts, maybe we should do it right before we evaluate the condition, so that we can ensure even for skipped steps this value gets updated.

One tricky note, we should ensure that this evaluation cannot fail (skipped steps should never turn to failed and cause a job to fail that would have otherwise succeeded). I'm not sure if this is something the tryEvaluateStepName function can do, but we will want to double check with incorrectly formatted syntax inside the `{{ }}` brackets. In the case where a name cannot be evaluated, it would be okay to fallback to the unevaluated name.

**To Reproduce**
Run the workflow
```
name: Skipping Steps
on:
  workflow_dispatch:
  push:
jobs:   
  skippedstep:
    env:
      NPM_PACKAGE_ORG: "random-name"
    strategy:
      matrix:
        os: ["windows","linux"]
    runs-on: ubuntu-latest
    steps:
      - name: ${{ github.actor }}
        run: echo "this is expanded"
        if: false
      - name: ${{ env.NPM_PACKAGE_ORG }}
        run: echo "I'll be expanded"
      - name: ${{ env.NPM_PACKAGE_ORG }}
        if: false
        run: echo "I'll not be expanded"
      - name: ${{ matrix.os }} 
        run: echo "am expanded as well"
        if: false
```

**Expected behavior**
We should evaluate the step name even if we skip the step.

### Resources
[Here](https://github.com/actions/runner/blob/088981a372fc962a3d86cb0bfcbe5173a01d0d2e/src/Runner.Worker/ActionRunner.cs#LL288C30-L288C30) is where we evaluate the step name, we do it multiple times, the last being as soon as we run a step.

[Here](https://github.com/actions/runner/blob/088981a372fc962a3d86cb0bfcbe5173a01d0d2e/src/Runner.Worker/StepsRunner.cs#L262) is the last place we evaluate step names (just in time)

[...]

PR #2313, opened 2022-12-13
Defer evaluation of a step's DisplayName until its condition is evaluated.

jww3:
Proposed fix for https://github.com/actions/runner/issues/2309

This fix relocates the logic for expanding a step's display name.

In the previous implementation, the display name was evaluated as part of step execution.  It was therefore possible for display names to remain unevaluated for steps that ended up being skipped.

With this change, we now evaluate the display name right before we evaluate the step's condition.

In this way, the display name gets updated even if the condition check results in the step being skipped.

### Workflow used to Test

```
name: Issue 2309 Test
on:
  workflow_dispatch:
jobs:   
  skippedstep:
    env:
      NPM_PACKAGE_ORG: "random-name"
    strategy:
      matrix:
        variation: ["aaa","bbb"]
    runs-on: self-hosted
    steps:
      - name: Step 1 -- Actual = ${{ github.actor }}; Expected = jww3 
        run: echo "this is expanded"
        if: false
      - name: Step 2 -- Actual = ${{ env.NPM_PACKAGE_ORG }}; Expected = random-name
        run: echo "I'll be expanded"
      - name: Step 3 (fix) -- Actual = ${{ env.NPM_PACKAGE_ORG }}; Expected = random-name
        if: false
        run: echo "I'll not be expanded"
      - name: Step 4 -- Actual = ${{ matrix.variation }}; Expected = aaa or bbb
        run: echo "am expanded as well"
        if: false
```
source:  https://github.com/bbq-beets/jww3-2022a/blob/main/.github/workflows/verify-fix-2309.yml

### Before

image

### After

image


AvaStancu:
Congrats on one of your very first PRs as part of the runtime team 👏 
Would it be possible to extract your implementation to a separate method and add some unit tests for it?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -195,6 +195,20 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                            {
+                                if (step is IActionRunner actionRunner && actionRunner.Stage == ActionRunStage.Main &&
+                                    actionRunner.TryEvaluateDisplayName(step.ExecutionContext.ExpressionValues, step.ExecutionContext))
+                                {
+                                    step.ExecutionContext.UpdateTimelineRecordDisplayName(actionRunner.DisplayName);
+                                }
+                            }
+                            catch (Exception ex)
+                            {
+                                Trace.Info("Caught exception while attempting to evaulate/update the step's DisplayName.  Exception Details:  {0}", ex);
```
AvaStancu, line 209: if I got this C# code right, this `Exception` is the 'parent' of SystemException and C#'s docs say `Because SystemException serves as the base class of a variety of exception types, your code should not throw a SystemException exception, nor should it attempt to handle a SystemException exception unless you intend to re-throw the original exception.`. Would at least using `Trace.Error` in this case be better or not catching this generic exception at all? 🤔

flobernd: > or not catching this generic exception at all

+1 for not using a `try..catch` block here at all, **unless** exceptions during the display name evaluation must explicitly be ignored by design/acceptance-criteria.

If there is no concrete exception type thrown by the wrapped code **and** any possible exceptions should not be ignored, it's usually best letting the exception bubbling up to a higher / the topmost handler (which usually does some kind of logging).

jww3: I considered this.

Evaluation of the condition is paramount.  We have to ensure that that happens no-matter-what.  In other words, we can't allow a failure to update the display name (which would be benign in the grand scheme of things) to derail evaluation of the condition.

jww3:
> Congrats on one of your very first PRs as part of the runtime team 👏 Very clear issue descriptions also 👏 👏 👏 Would it be possible to extract your implementation to a separate method and add some unit tests for it?

I spent a couple hours trying to devise a unit test for this given our unit testing framework and had to give up.  Thomas and I both looked at it.  As best as I can tell, there's no great way in our unit testing framework to exercise this code path without extensive mocking -- that is, mocking to the point where you're not really proving anything useful.

src/Runner.Worker/StepsRunner.cs
fhammerl: @flobernd with the correct *unless* - from the issue / reqs doc:

> we should ensure that this evaluation cannot fail (skipped steps should never turn to failed and cause a job to fail that would have otherwise succeeded).

We really _do_ want to make this a best / last effort display name calculation.
Let's make sure however that we don't use `catch (Exception ex)` for flow handling, like catching ArgUtil.NotNull exceptions. For example, could `actionRunner.DisplayName` be `null` at the time of calling `step.ExecutionContext.UpdateTimelineRecordDisplayName`?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -195,6 +195,26 @@ public async Task RunAsync(IExecutionContext jobContext)
                         }
                         else
                         {
+                            // This is our last, best chance to expand the display name.  (At this point, all the requirements for successful expansion should be met.)
```
fhammerl, line 198: As per @AvaStancu's recommendation:
> Would it be possible to extract your implementation to a separate method and add some unit tests for it?

I agree with you @jww3 on your comments below regarding the difficulty and usefulness of unit testing here 
What do you think about a `TrySetDisplayname()` function?

src/Runner.Worker/ActionRunner.cs
```diff
@@ -285,16 +285,20 @@ public async Task RunAsync()
...
             ArgUtil.NotNull(Action, nameof(Action));
 
-            // If we have already expanded the display name, there is no need to expand it again
-            // TODO: Remove the ShouldEvaluateDisplayName check and field post m158 deploy, we should do it by default once the server is updated
+            updated = false;
+            // If we have already expanded the display name, don't bother attempting [re-]expansion.
             if (_didFullyEvaluateDisplayName || !string.IsNullOrEmpty(Action.DisplayName))
             {
-                return false;
+                return true;
```
jww3, line 301: This is a subtle change in behavior, but fortunately there weren't many existing callers (mainly just unit tests) and even fewer existing callers that even bothered checking the return value

src/Test/L0/Worker/ActionRunnerL0.cs
```diff
@@ -183,13 +178,51 @@ public void EvaluateExpansionOfDisplayNameToken()
...
+
+            // Arrange
+            Setup();
+            var actionId = Guid.NewGuid();
+            var action = new Pipelines.ActionStep()
+            {
+                Name = "action",
+                Id = actionId,
+                DisplayName = explicitDisplayName,
+                DisplayNameToken = new BasicExpressionToken(null, null, null, "matrix.node"),
```
jww3, line 204: in this unit test, we're setting both `DisplayName` and `DisplayNameToken`.

AvaStancu (dismissed):

src/Runner.Worker/ActionRunner.cs
fhammerl: The return value is `IsDisplayNameSet`, and the out var says `IsUpdated[JustNow]`.

Why do we need both? 🤔 Is it to avoid setting it again in `StepsRunner.TryUpdateDisplayName()`?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -195,6 +195,12 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                            // That being said, evaluating the display name should still be considered as a "best effort" exercise.  (It's not critical or paramount.)
+                            // For that reason, we call a safe "Try..." wrapper method to ensure that any potential problems we encounter in evaluating the display name
+                            // don't interfere with our ultimate goal within this code block:  evaluation of the condition.
+                            TryUpdateDisplayName(step as IActionRunner, out _);
```
fhammerl, line 202: Do we need the `out _` var here / in the function signature?

jww3: yeah, I think so, for consistency.

There are effectively three possible outcomes.

Call succeeded as the result of a successful update to the display name.
Call succeeded because the display name was already up-to-date.
Call failed

We've seen that there are some codepaths (one level down) that need to distinguish between those two flavors of success, so for consistency, I'm preserving that level of detail.

If it's not appealing, the next best alternative would be to drop the out parameter and instead return a bitwise enum that encodes the various flavors of success.  I considered that, but it felt a bit over the top.

src/Runner.Worker/ActionRunner.cs
jww3: Consider code that invokes `EvaluateDisplayName` multiple times.

Example:
```
if (EvaluateDisplayName(this.ExecutionContext.ExpressionValues, this.ExecutionContext)) 
{
   this.ExecutionContext.UpdateTimelineRecordDisplayName(this.DisplayName);
}
   .
   .
   .
if (EvaluateDisplayName(this.ExecutionContext.ExpressionValues, this.ExecutionContext)) 
{
   this.ExecutionContext.UpdateTimelineRecordDisplayName(this.DisplayName);
}
```

Which TimelineRecordUpdate is the correct one?

src/Runner.Worker/ActionRunner.cs
```diff
@@ -285,25 +284,59 @@ public async Task RunAsync()
...
+        /// </summary>
+        public bool TryUpdateDisplayName(out bool updated)
+        {
+            updated = false;
+
+            // REVIEW:  This try/catch can be removed if some future implementation of EvaluateDisplayName and UpdateTimelineRecordDisplayName
+            // can make reasonable guarantees that they won't throw an exception.
+            try
+            {
+                if (this.Stage == ActionRunStage.Main && EvaluateDisplayName(this.ExecutionContext.ExpressionValues, this.ExecutionContext, out updated))
```
jww3, line 300: Add a comment to clarify why we only attempt this on `Stage == ActionRunStage.Main`

thboop: Its been a while but I believe a job consists of the following stages

- Job started
  - We try to evaluate all display names here
- PreJobSteps
   - No contexts have changed from the above, so no need to retry if we couldn't fully expand
- Main Steps
    - Context may now have changed, lets update the display names before we start the steps  
- Post Job Steps
    - Name should have already been fully expanded, no need to try again.
- Job complete
    - Nothing to do, job is done don't evaluate display names

jww3: Thank you!

fhammerl (dismissed):
LGTM

jww3:
Fix verification with latest changes:

image

fhammerl (approved):
LGTM

PR #2317, opened 2022-12-16
Initialize container manager based on whether the ContainerHooksPath is  set

nikola-jokic:
When the operation provider is called, it initializes both the dockerManager and the containerHookManager.
The dockerManager will call `which` on docker, which can cause initialization to fail. 

The fix should relax the docker cli dependency if the runner is going to use container hooks.

Related to the issue in runner-container-hooks: https://github.com/actions/runner-container-hooks/issues/30

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -33,8 +33,14 @@ namespace GitHub.Runner.Worker
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
-            _dockerManager = HostContext.GetService<IDockerCommandManager>();
-            _containerHookManager = HostContext.GetService<IContainerHookManager>();
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.Hooks.ContainerHooksPath)))
+            {
+                _dockerManager = HostContext.GetService<IDockerCommandManager>();
+            }
+            else
+            {
+                _containerHookManager = HostContext.GetService<IContainerHookManager>();
+            }
         }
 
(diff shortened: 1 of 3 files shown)


src/Test/L0/Worker/ContainerOperationProviderL0.cs
```diff
@@ -111,7 +151,6 @@ private void Setup([CallerMemberName] string testName = "")
             _containerHookManager = new Mock<IContainerHookManager>();
             containerOperationProvider = new ContainerOperationProvider();
 
-            _hc.SetSingleton<IDockerCommandManager>(_dockerManager.Object);
```
fhammerl, line 114: Do you know why this was set in the first place if never used?

src/Runner.Worker/ContainerOperationProvider.cs
```diff
@@ -33,8 +33,14 @@ public class ContainerOperationProvider : RunnerService, IContainerOperationProv
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
-            _dockerManager = HostContext.GetService<IDockerCommandManager>();
-            _containerHookManager = HostContext.GetService<IContainerHookManager>();
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.Hooks.ContainerHooksPath)))
```
fhammerl, line 36: Should we do `same` in the `ContainerActionHandler`?

https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ContainerActionHandler.cs#L41

nikola-jokic: Great catch! Thanks!

fhammerl (approved):
LGTM

PR #2321, opened 2022-12-16
Re-add file size check prior to reading file

bethanyj28:
The job hangs when uploading a large summary file since we had moved a file size check to another location. We suspect it has to do with `MaskSecrets` reading the file, but we will re-add the check to fix the issue for the time being.

Prior to the change:
SCR-20221216-hue

After the change:
SCR-20221216-hul

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -182,6 +182,14 @@ namespace GitHub.Runner.Worker
                     return;
                 }
 
+                if (fileSize > AttachmentSizeLimit)
+                {
+                    context.Error(String.Format(Constants.Runner.UnsupportedSummarySize, AttachmentSizeLimit / 1024, fileSize / 1024));
+                    Trace.Info($"Step Summary file ({filePath}) is too large ({fileSize} bytes); skipping attachment upload");
+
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -182,6 +182,14 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
                     return;
                 }
 
+                if (fileSize > AttachmentSizeLimit)
```
brittanyellich, line 185: We should probably also remove this block where it is added in the if/else block below

bethanyj28: Removed in [674f15b](https://github.com/actions/runner/pull/2321/commits/674f15bb436e7ed6c46d805e8b2d57ce3ca6a100)!

Antonioftl: Beautiful!!

TingluoHuang:
@bethanyj28 Please cherry-pick the fix back to the main branch.

bethanyj28:
@TingluoHuang PR open: https://github.com/actions/runner/pull/2330

Thanks!

PR #2328, opened 2022-12-19
Using VssHttpClient instead of RawHttpClient

yacaovsnc:
In order to take advantage of the proxy and ssl cert logic, we will use `VssHttpClientBase` instead of `RawHttpClientBase` for the time being since the vss client is production ready.


TingluoHuang:
we should revert this since it's not landed in the main branch.

PR #2334, opened 2022-12-20
Always upload to avoid issues

yacaovsnc:
Uploading summaries and gather telemetry.  Also make some style changes to be consistent.

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -20,7 +20,7 @@ namespace GitHub.Runner.Common
         void Start(Pipelines.AgentJobRequestMessage jobRequest);
         void QueueWebConsoleLine(Guid stepRecordId, string line, long? lineNumber = null);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
-        void QueueSummaryUpload(Guid timelineId, Guid timelineRecordId, string stepId, string name, string path, bool deleteSource);
+        void QueueSummaryUpload(Guid stepRecordId, string name, string path, bool deleteSource);
         void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRecord);
     }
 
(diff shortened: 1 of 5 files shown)


src/Runner.Common/JobServerQueue.cs
```diff
@@ -476,8 +471,27 @@ private async Task ProcessSummaryUploadQueueAsync(bool runOnce = false)
                         }
                         catch (Exception ex)
                         {
-                            Trace.Info("Catch exception during summary file upload to results, keep going since the process is best effort.");
-                            Trace.Error(ex);
+                            try
+                            {
+
+                                var issue = new Issue() { Type = IssueType.Warning, Message = $"Caught exception during summary file upload to results. {ex.Message}" };
```
TingluoHuang, line 477: Do we want to make this warning shows up on the user's face?
You might need to do some server work to hide it if we don't want.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -476,8 +471,27 @@ private async Task ProcessSummaryUploadQueueAsync(bool runOnce = false)
                         }
                         catch (Exception ex)
                         {
-                            Trace.Info("Catch exception during summary file upload to results, keep going since the process is best effort.");
-                            Trace.Error(ex);
+                            try
```
TingluoHuang, line 474: I would suggest removing `try-catch` since none of the code should throw within here.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Definitely.  Server side PR on separate thread.

src/Runner.Common/JobServerQueue.cs
TingluoHuang: if we have a bug, then we would like to make the runner crash and find the bug via e2e test and canary runs, we use `try-catch` only in places where we know it's going to throw an exception.

yacaovsnc: Sure, removed.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -476,8 +471,21 @@ private async Task ProcessSummaryUploadQueueAsync(bool runOnce = false)
                         }
                         catch (Exception ex)
                         {
+                            var issue = new Issue() { Type = IssueType.Warning, Message = $"Caught exception during summary file upload to results. {ex.Message}" };
+                            issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.ResultsUploadFailure;
+
+                            var telemetryRecord = new TimelineRecord()
```
TingluoHuang, line 477: 😭 I realize a problem with this approach... let's sync in working days.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -476,8 +471,21 @@ private async Task ProcessSummaryUploadQueueAsync(bool runOnce = false)
...
+                            {
+                                Id = Constants.Runner.TelemetryRecordId,
+                            };
+                            telemetryRecord.Issues.Add(issue);
```
TingluoHuang, line 481:
```suggestion
                            var telemetryRecord = new TimelineRecord()
                            {
                                Id = Constants.Runner.TelemetryRecordId,
                            };
                            telemetryRecord.Issues.Add(new Issue() { Type = IssueType.Warning, Message = $"Caught exception during summary file upload to results. {ex.Message}" });
```

yacaovsnc: I still want `data` field so I can trace and filter on it.  If we use this to pass up telemetry in the future, I don't want to see other failiures.  

Otherwise I have to rely on a sentence, I would much prefer to filter on a key instead.

TingluoHuang (approved):

PR #2335, opened 2022-12-21
Log jitconfig value as secret

nikola-jokic:


AvaStancu (dismissed):

TingluoHuang (changes requested):
I will create a different PR for fixing this.

TingluoHuang (approved):

Issue #2342, opened 2022-12-26
config cannot handle symlinks


IceflowRE:
**Describe the bug**
Iam inside a docker container based on this [image](https://github.com/actions/runner/blob/main/images/Dockerfile).
OS: `Linux/Ubuntu`
Version: `2.300.2`
Arch: `arm64`

Iam using symlinks for
- `.credentials`
- `.credentials_rsaparams`
- `.env`
- `.path`
- `.runner`

```
RUN mkdir /actions-runner-config/ \
&& touch /actions-runner-config/.credentials \
&& ln -s /actions-runner-config/.credentials .credentials \
&& touch /actions-runner-config/.credentials_rsaparams \
&& ln -s /actions-runner-config/.credentials_rsaparams .credentials_rsaparams \
&& touch /actions-runner-config/.env \
&& ln -s /actions-runner-config/.env .env \
&& touch /actions-runner-config/.path \
&& ln -s /actions-runner-config/.path .path \
&& touch /actions-runner-config/.runner \
&& ln -s /actions-runner-config/.runner .runner
```

Now i run
```
root@d12f802e0d14:/actions-runner# ./config.sh --url https://github.com/actions --token ?????????????????????? --unattended --disableupdate
```

```
[RUNNER 2022-12-26 20:46:30Z INFO HostContext] No proxy settings were found based on environmental variables (http_proxy/https_proxy/HTTP_PROXY/HTTPS_PROXY)
[RUNNER 2022-12-26 20:46:30Z INFO HostContext] Well known directory 'Bin': '/actions-runner/bin'
[RUNNER 2022-12-26 20:46:30Z INFO HostContext] Well known directory 'Root': '/actions-runner'
[RUNNER 2022-12-26 20:46:30Z INFO HostContext] Well known config file 'Credentials': '/actions-runner/.credentials'
[RUNNER 2022-12-26 20:46:31Z INFO HostContext] Well known directory 'Bin': '/actions-runner/bin'
[RUNNER 2022-12-26 20:46:31Z INFO HostContext] Well known directory 'Root': '/actions-runner'
[RUNNER 2022-12-26 20:46:31Z INFO HostContext] Well known config file 'Runner': '/actions-runner/.runner'
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at GitHub.Runner.Common.HostContext..ctor(String hostType, String logFile)
   at GitHub.Runner.Listener.Program.Main(String[] args)
[...]

PR #2343, opened 2022-12-27
Guard against NullReference while creating HostContext

nikola-jokic:
Fixes null pointer exception creating `HostContext` if the `.runner` file exists, but is empty.
Null pointer exception is described in issue https://github.com/actions/runner/issues/2342, but this does not solve the issue itself.

It just lets the HostContext proceed and the error message will be displayed as the runner is already configured.


AvaStancu (approved):

src/Runner.Common/HostContext.cs
```diff
@@ -221,8 +221,11 @@ public HostContext(string hostType, string logFile = null)
             if (File.Exists(runnerFile))
             {
                 var runnerSettings = IOUtil.LoadObject<RunnerSettings>(runnerFile);
-                _userAgents.Add(new ProductInfoHeaderValue("RunnerId", runnerSettings.AgentId.ToString(CultureInfo.InvariantCulture)));
-                _userAgents.Add(new ProductInfoHeaderValue("GroupId", runnerSettings.PoolId.ToString(CultureInfo.InvariantCulture)));
+                if (runnerSettings != null)
```
TingluoHuang, line 224: should we just throw and crash the runner if `runnerSettings` is NULL?

nikola-jokic: Thanks!

src/Runner.Sdk/Util/IOUtil.cs
```diff
@@ -46,6 +46,21 @@ public static T LoadObject<T>(string path)
             return StringUtil.ConvertFromJson<T>(json);
         }
 
+        public static T LoadObject<T>(string path, bool required)
+        {
+            string json = File.ReadAllText(path, Encoding.UTF8);
+            if (string.IsNullOrEmpty(json))
+            {
+                throw new InvalidOperationException($"File {path} is empty");
```
nikola-jokic, line 54: @TingluoHuang any suggestions for exception that we should throw :relaxed: ?

TingluoHuang: do we want to merge `public static T LoadObject<T>(string path, bool required)` merge with `public static T LoadObject<T>(string path)` to be something like `public static T LoadObject<T>(string path, bool required = false)`?

TingluoHuang: `ArgumentNullException` or `ArgumentException` both seems fine to me.

nikola-jokic: Applied `ArgumentNullException` if file is empty or null, `ArgumentException` if convert from json returns null

TingluoHuang (dismissed):

TingluoHuang (approved):

TingluoHuang (approved):

PR #2373, opened 2023-01-16
Change directories first

cinderblock:
If you don't change to the correct directory first, it partially defeats the purpose of doing so.

Fixes bugs like:
```log
ldd: ./bin/libcoreclr.so: No such file or directory
ldd: ./bin/libSystem.Security.Cryptography.Native.OpenSsl.so: No such file or directory
ldd: ./bin/libSystem.IO.Compression.Native.so: No such file or directory
```


cinderblock:
:(

PR #2374, opened 2023-01-16
Formalize `TryUpdateDisplayName` and `EvaluateDisplayName` as members of interface `IStep`

jww3:
based on a conversation with @fhammerl regarding https://github.com/actions/runner/pull/2313

### DIFF
--- a/src/Runner.Worker/StepsRunner.cs
+++ b/src/Runner.Worker/StepsRunner.cs
@@ -1,12 +1,9 @@
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using GitHub.DistributedTask.Expressions2;
 using GitHub.DistributedTask.ObjectTemplating.Tokens;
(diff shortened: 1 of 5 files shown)


src/Runner.Worker/StepsRunner.cs
```diff
@@ -26,6 +21,8 @@ public interface IStep
         string DisplayName { get; set; }
         IExecutionContext ExecutionContext { get; set; }
         TemplateToken Timeout { get; }
+        bool TryUpdateDisplayName(out bool updated);
+        bool EvaluateDisplayName(DictionaryContextData contextData, IExecutionContext context, out bool updated);
```
jww3, line 25: I co-located these in the `IStep` interface because I strongly suspect the latter can be eliminated in a second pass refactor.  Having them side-by-side should make that refactor easier to reason about.

fhammerl (approved):
LGTM, let's merge it in the original PR then let's go over the refactoring you mentioned with `EvaluateDisplayName`

PR #2377, opened 2023-01-17
Show more information in the runner log

yujincat:
To provide more information in the run log, add
 - the reference used when calling reusable workflows
 - the job name for both non-reusable/reusable workflow jobs

Actions service side PR: https://github.com/github/actions-dotnet/pull/14151

Related to https://github.com/github/actions-enablers-policy/issues/704

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -321,7 +321,13 @@ namespace GitHub.Runner.Worker
 
                     if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
                     {
-                        context.Output($"Uses: {workflowFileFullPath.Value}");
+                        var usesOutput = $"Uses: {workflowFileFullPath.Value}";
+                        if (message.Variables.TryGetValue("system.workflowFileRef", out VariableValue workflowFileRef))
+                        {
+                            usesOutput += $"@{workflowFileRef.Value}";
+                        }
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -321,7 +321,21 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
 
                     if (message.Variables.TryGetValue("system.workflowFileFullPath", out VariableValue workflowFileFullPath))
                     {
-                        context.Output($"Uses: {workflowFileFullPath.Value}");
+                        var usesOutput = $"Uses: {workflowFileFullPath.Value}";
+                        if (message.Variables.TryGetValue("system.workflowFileRef", out VariableValue workflowFileRef)
+                            && message.Variables.TryGetValue("system.workflowFileSha", out VariableValue workflowFileSha))
```
TingluoHuang, line 326:
```suggestion
                            || message.Variables.TryGetValue("system.workflowFileSha", out VariableValue workflowFileSha))
```

should it be `OR`?

yujincat: oh yea you are right :-) 
Just to note, we are sending both values but in case the user is using reusable workflow using sha then `workflowFileRef` will be empty. Still that will return `true` for this `TryGetValue`.

TingluoHuang (dismissed):

TingluoHuang (approved):

PR #2378, opened 2023-01-17
Pivot toward immutable `DistributedTask.WebApi::Issue` instances.

jww3:
The proper way to instantiate a `DistributedTask.WebApi::Issue` is now via the factory method on an `ExecutionContext` (`IExecutionContext::CreateIssue`)

This variation addresses feedback in https://github.com/actions/runner/pull/2311

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -299,7 +299,7 @@ namespace GitHub.Runner.Common
                         {
                             try
                             {
-                                // Give at most 60s for each request. 
+                                // Give at most 60s for each request.
                                 using (var timeoutTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(60)))
                                 {
                                     await _jobServer.AppendTimelineRecordFeedAsync(_scopeIdentifier, _hubName, _planId, _jobTimelineId, _jobTimelineRecordId, stepRecordId, batch.Select(logLine => logLine.Line).ToList(), batch[0].LineNumber, timeoutTokenSource.Token);
@@ -600,8 +600,7 @@ namespace GitHub.Runner.Common
                 {
                     foreach (var issue in record.Issues)
                     {
-                        String source;
-                        issue.Data.TryGetValue("sourcepath", out source);
+                        string source = issue["sourcepath"];
                         Trace.Verbose($"        Issue: c={issue.Category}, t={issue.Type}, s={source ?? string.Empty}, m={issue.Message}");
                     }
                 }
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -648,7 +648,7 @@ namespace GitHub.Runner.Worker
             var filteredDictionaryEntries = command.Properties
                                                    .Where(kvp => !string.Equals(kvp.Key, keyToExclude, StringComparison.OrdinalIgnoreCase));
 
-            var metadata = new IssueMetadata(issueCategory, false, filteredDictionaryEntries);
+            var metadata = new IssueMetadata(issueCategory, false, null, filteredDictionaryEntries);
             var issue = context.CreateIssue(this.Type, command.Data, metadata, true);
             context.AddIssue(issue);
         }
--- a/src/Runner.Worker/ActionCommandManager.cs
+++ b/src/Runner.Worker/ActionCommandManager.cs
@@ -270,13 +270,10 @@ namespace GitHub.Runner.Worker
                 if (string.Equals(blocked, envName, StringComparison.OrdinalIgnoreCase))
                 {
                     // Log Telemetry and let user know they shouldn't do this
-                    var issue = new Issue()
-                    {
-                        Type = IssueType.Error,
-                        Message = $"Can't update {blocked} environment variable using ::set-env:: command."
-                    };
-                    issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = $"{Constants.Runner.UnsupportedCommand}_{envName}";
-                    context.AddIssue(issue, writeToLog: true);
+                    var message = $"Can't update {blocked} environment variable using ::set-env:: command.";
+                    var metadata = new IssueMetadata(Constants.Runner.InternalTelemetryIssueDataKey, $"{Constants.Runner.UnsupportedCommand}_{envName}");
+                    var issue = context.CreateIssue(IssueType.Error, message, metadata, true);
+                    context.AddIssue(issue);
 
                     return;
                 }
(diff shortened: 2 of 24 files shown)


src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -270,12 +270,9 @@ public void ProcessCommand(IExecutionContext context, string line, ActionCommand
...
                 {
                     // Log Telemetry and let user know they shouldn't do this
-                    var issue = new Issue()
-                    {
-                        Type = IssueType.Error,
-                        Message = $"Can't update {blocked} environment variable using ::set-env:: command."
-                    };
-                    issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = $"{Constants.Runner.UnsupportedCommand}_{envName}";
+                    var message = $"Can't update {blocked} environment variable using ::set-env:: command.";
+                    var metadata = new IssueMetadata(Constants.Runner.InternalTelemetryIssueDataKey, $"{Constants.Runner.UnsupportedCommand}_{envName}");
```
jww3, line 274: I defined `IssueMetadata` a data bucket to pass to `CreateIssue` It holds the seldom-used fields of an `Issue`.  It was easier than polluting the `CreateIssue` function signature with those seldom-used fields.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -577,74 +579,66 @@ public void Progress(int percentage, string currentOperation = null)
...
+
+            if (!string.IsNullOrEmpty(wellKnownTag))
             {
-                if (!string.IsNullOrEmpty(logMessage))
+                if (writeToLog && !string.IsNullOrEmpty(refinedMessage))
                 {
-                    long logLineNumber = Write(WellKnownTags.Warning, logMessage);
-                    issue.Data["logFileLineNumber"] = logLineNumber.ToString();
+                    long logLineNumber = Write(wellKnownTag, refinedMessage);
+                    result.Data["logFileLineNumber"] = logLineNumber.ToString();
```
jww3, line 626: This is the last remaining side-effect.

src/Runner.Worker/Handlers/OutputManager.cs
```diff
@@ -139,12 +139,10 @@ public void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
 
                         // Convert to issue
                         var issue = ConvertToIssue(match);
-
                         if (issue != null)
                         {
                             // Log issue
-                            _executionContext.AddIssue(issue, stripped);
```
jww3, line 146: This was the only place we were passing a message override to `AddIssue`

What's interesting is that, as far as I can tell, it was unnecessary.

Have a look here.

src/Runner.Worker/Handlers/OutputManager.cs
```diff
@@ -107,16 +108,15 @@ public void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
                         // Match
                         try
                         {
-                            match = matcher.Match(stripped);
-
+                            match = matcher.Match(refinedLine);
```
jww3, line 111: The `IssueMatch` returned by `Match` now includes a field that preserves the original string passed to it.

src/Runner.Worker/Handlers/OutputManager.cs
```diff
@@ -139,12 +139,10 @@ public void OnDataReceived(object sender, ProcessDataReceivedEventArgs e)
 
                         // Convert to issue
                         var issue = ConvertToIssue(match);
-
                         if (issue != null)
                         {
                             // Log issue
-                            _executionContext.AddIssue(issue, stripped);
-
+                            _executionContext.AddIssue(issue);
```
jww3, line 145: `ConvertToIssue` now leverages the original text stored within `match` so we no longer need to pass a log message override to `AddIssue`

src/Runner.Worker/Handlers/OutputManager.cs
```diff
@@ -327,9 +321,11 @@ private DTWebApi.Issue ConvertToIssue(IssueMatch match)
             }
             catch (Exception ex)
             {
-                _executionContext.Debug($"Dropping file value '{match.File}' and fromPath value '{match.FromPath}'. Exception during validation: {ex.ToString()}");
+                _executionContext.Debug($"Dropping file value '{match.File}' and fromPath value '{match.FromPath}'. Exception during validation: {ex}");
             }
 
+            var metadata = new IssueMetadata(null, false, match.SourceText, issueData);
+            var issue = _executionContext.CreateIssue(issueType, match.Message, metadata, true);
```
jww3, line 328: note how `match.SourceText` is stored in the metadata passed to `CreateIssue`.  This is how we achieve log message overrides, now.

src/Runner.Worker/IssueMatcher.cs
```diff
@@ -8,6 +8,28 @@ namespace GitHub.Runner.Worker
...
+        {
+            this.Category = category;
+            this.IsInfrastructureIssue = infrastructureIssue;
+            this.LogMessageOverride = logMessageOverride;
+            this.Data = data;
+        }
+
+        public readonly string Category;
+        public readonly bool IsInfrastructureIssue;
+        public readonly string LogMessageOverride;
```
jww3, line 29: `LogMessageOverride` is transient data.  We won't persist it into any `Issue` objects we create.

src/Runner.Worker/IssueMatcher.cs
```diff
@@ -184,8 +206,9 @@ public IssuePattern(IssuePatternConfig config, TimeSpan timeout)
 
     public sealed class IssueMatch
     {
-        public IssueMatch(IssueMatch runningMatch, IssuePattern pattern, GroupCollection groups, string defaultSeverity = null)
+        public IssueMatch(string sourceText, IssueMatch runningMatch, IssuePattern pattern, GroupCollection groups, string defaultSeverity = null)
         {
+            SourceText = sourceText;
```
jww3, line 211: This allows the original text to "survive" long enough for us to pass it along to the mechanism that logs (`ExecutionContext::Write`)

src/Runner.Worker/ActionCommandManager.cs
```diff
@@ -658,14 +644,12 @@ public void ProcessCommand(IExecutionContext context, string inputLine, ActionCo
...
-            foreach (var property in command.Properties)
-            {
-                if (!string.Equals(property.Key, Constants.Runner.InternalTelemetryIssueDataKey, StringComparison.OrdinalIgnoreCase))
-                {
-                    issue.Data[property.Key] = property.Value;
-                }
-            }
+            string keyToExclude = Constants.Runner.InternalTelemetryIssueDataKey;
+            var filteredDictionaryEntries = command.Properties
+                                                   .Where(kvp => !string.Equals(kvp.Key, keyToExclude, StringComparison.OrdinalIgnoreCase));
```
jww3, line 649: I probably need to close over this `IEnumerable` since there's no guarantee that `command` won't get modified by the time the `IssueMetadata` is consumed.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -600,8 +600,7 @@ private List<TimelineRecord> MergeTimelineRecords(List<TimelineRecord> timelineR
                 {
                     foreach (var issue in record.Issues)
                     {
-                        String source;
-                        issue.Data.TryGetValue("sourcepath", out source);
+                        string source = issue["sourcepath"];
```
KetchupOnMyKetchup, line 603: Is there a risk/issue if `issue["sourcepath"]` is null?

src/Runner.Worker/ActionCommandManager.cs
jww3: fixed

jww3:
Decided to go a different (less aggressive) route.  There are still some fixes in this PR worth considering.

PR #2383, opened 2023-01-19
Revert "split by regex"

TingluoHuang:
Reverts actions/runner#2333

```
uses: actions/github-script@v6
with:
  script: |
    if (context.workflow != "foo") {
      console.log("hello")
    }
```

failed with `Unhandled error: SyntaxError: Unexpected token ')'`


thboop (approved):
lgtm

PR #2395, opened 2023-01-25
Bypass all proxies for all hosts if `no_proxy='*'` is set

fhammerl:
The behaviour of `no_proxy` is in no way standardised, and it is handled differently by the runner app vs in github actions (see actions/toolkit PR below). The behaviour in this PR is consistent with 

`no_proxy='*'` will now bypass any proxy for any host.

### DIFF
--- a/src/Runner.Sdk/RunnerWebProxy.cs
+++ b/src/Runner.Sdk/RunnerWebProxy.cs
@@ -164,13 +164,6 @@ namespace GitHub.Runner.Sdk
                         {
                             continue;
                         }
-
-                        // Strip leading '.' fron noproxy host
-                        if (noProxyInfo.Host.StartsWith("."))
-                        {
-                            noProxyInfo.Host = noProxyInfo.Host[1..];
-                        }
-
                         _noProxyList.Add(noProxyInfo);
                     }
                 }
@@ -231,7 +224,14 @@ namespace GitHub.Runner.Sdk
                     matchPort = string.Equals(noProxy.Port, input.Port.ToString());
                 }
 
-                matchHost = string.Equals(input.Host, noProxy.Host, StringComparison.OrdinalIgnoreCase) || input.Host.EndsWith($".{noProxy.Host}", StringComparison.OrdinalIgnoreCase);
+                if (noProxy.Host.StartsWith('.'))
+                {
+                    matchHost = input.Host.EndsWith(noProxy.Host, StringComparison.OrdinalIgnoreCase);
+                }
+                else
+                {
+                    matchHost = string.Equals(input.Host, noProxy.Host, StringComparison.OrdinalIgnoreCase) || input.Host.EndsWith($".{noProxy.Host}", StringComparison.OrdinalIgnoreCase);
+                }
 
                 if (matchHost && matchPort)
                 {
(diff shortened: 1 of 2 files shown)


src/Runner.Sdk/RunnerWebProxy.cs
```diff
@@ -165,6 +165,12 @@ public RunnerWebProxy()
                             continue;
                         }
 
+                        // Strip leading '.' fron noproxy host
+                        if (noProxyInfo.Host.StartsWith("."))
+                        {
+                            noProxyInfo.Host = noProxyInfo.Host[1..];
```
JoannaaKL, line 171: What by mistake someone set the host to sth like "cs."?

src/Test/L0/RunnerWebProxyL0.cs
```diff
@@ -351,12 +351,51 @@ public void WebProxyFromEnvironmentVariablesNoProxy()
                 Assert.False(proxy.IsBypassed(new Uri("https://actions.com")));
                 Assert.False(proxy.IsBypassed(new Uri("https://ggithub.com")));
                 Assert.False(proxy.IsBypassed(new Uri("https://github.comm")));
-                Assert.False(proxy.IsBypassed(new Uri("https://google.com")));
                 Assert.False(proxy.IsBypassed(new Uri("https://example.com")));
                 Assert.False(proxy.IsBypassed(new Uri("http://example.com:333")));
                 Assert.False(proxy.IsBypassed(new Uri("http://192.168.0.123:123")));
                 Assert.False(proxy.IsBypassed(new Uri("http://192.168.1.123/home")));
 
+                Assert.True(proxy.IsBypassed(new Uri("https://google.com")));
```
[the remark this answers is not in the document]
JoannaaKL: Ahh right, I should've checked the whole file. Thank you! 🍫

src/Runner.Sdk/RunnerWebProxy.cs
fhammerl: Like `no_proxy="example.com, cs."`? 
I don't think any valid Hostnames end with `cs.`, end with a `.`. So, `cs.` would match nothing.

JoannaaKL: Aha this is a range operation, so if the host will be equal to `cs.` this logic will set noProxyInfo.Host to just `.` 
It's definitely not a valid host but it could be added by mistake. :)

fhammerl:
Closing this to keep current behaviour which is consistent with `wget`

src/Test/L0/RunnerWebProxyL0.cs
```diff
@@ -351,7 +351,7 @@ public void WebProxyFromEnvironmentVariablesNoProxy()
                 Assert.False(proxy.IsBypassed(new Uri("https://actions.com")));
                 Assert.False(proxy.IsBypassed(new Uri("https://ggithub.com")));
                 Assert.False(proxy.IsBypassed(new Uri("https://github.comm")));
-                Assert.False(proxy.IsBypassed(new Uri("https://google.com")));
+                Assert.False(proxy.IsBypassed(new Uri("https://google.com"))); // no_proxy has '.google.com', specifying only subdomains bypass
```
fhammerl, line 354: Not a change in behaviour in this PR. It's an additional test clarifying existing behaviour.

rentziass (approved):
LGTM! 🚀

rentziass (approved):

src/Test/L0/RunnerWebProxyL0.cs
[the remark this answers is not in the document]
Jerrusjrmeny44: porque no revisalo

PR #2396, opened 2023-01-25
Run service: fix small bug

yaananth:
URL schema is missing and honestly not sure why we have it `new Uri(serverUri.Authority)`

Ideally, we need to have tests around this, but things are still in flux w.r.t run service so skipping that for now, I hope it's okay


TingluoHuang (approved):

HTC37:
gh pr checkout 2396

PR #2412, opened 2023-02-01
start calling run service for job completion

yaananth:
Contributes to https://github.com/github/actions-service-team/issues/3446

This enables runner to call run service directly to complete the job.

There is more work left to send right data for job completion which will be in follow ups. We also need to start streaming events, upload logs etc., , which are all out of scope of this PR.


src/Runner.Common/RunnerService.cs
```diff
@@ -68,6 +68,19 @@ protected async Task<VssConnection> EstablishVssConnection(Uri serverUrl, VssCre
             throw new InvalidOperationException(nameof(EstablishVssConnection));
         }
 
+        protected async Task RetryRequest(Func<Task> func,
```
yaananth, line 71: This is added so that this can be called for Tasks that don't have any return type. To achieve that (since we can't do `Task<void>`), it's a general pattern to use something like `Unit` that represents absent value. I added `Unit` class and used it here.

src/Runner.Common/Unit.cs
```diff
@@ -0,0 +1,8 @@
+// Represents absence of value.
+namespace GitHub.Runner.Common
```
yaananth, line 2: Used in `RunnerService`, see there

src/Runner.Worker/JobRunner.cs
```diff
@@ -40,21 +39,34 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
             Trace.Info("Job ID {0}", message.JobId);
 
             DateTime jobStartTimeUtc = DateTime.UtcNow;
+            IJobServer jobServer = null;
+            IRunServer runServer = null;
 
             ServiceEndpoint systemConnection = message.Resources.Endpoints.Single(x => string.Equals(x.Name, WellKnownServiceEndpointNames.SystemVssConnection, StringComparison.OrdinalIgnoreCase));
+            if (string.Equals(message.MessageType, JobRequestMessageTypes.RunnerJobRequest, StringComparison.OrdinalIgnoreCase))
```
yaananth, line 46: we already have test cases that would have caught any null refs with existing logic of using `jobServer`. I added one for `runServer`

src/Runner.Worker/JobRunner.cs
```diff
@@ -40,21 +39,34 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+            {
+                runServer = HostContext.GetService<IRunServer>();
```
yaananth, line 48: yes we are not starting `_jobServerQueue` here, that's the one that drives sending updates, that's all out of scope and will come in later.

src/Runner.Worker/JobRunner.cs
```diff
@@ -206,6 +218,59 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+            {
+                return await CompleteJobAsync(runServer, jobContext, message, taskResult);
+            }
+            else
+            {
+                return await CompleteJobAsync(jobServer, jobContext, message, taskResult);
+            }
+        }
+
+        private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecutionContext jobContext, Pipelines.AgentJobRequestMessage message, TaskResult? taskResult = null)
```
yaananth, line 233: copy of the other method, with out stuff that run service doesn't care.

src/Sdk/DTWebApi/WebApi/RunServiceHttpClient.cs
```diff
@@ -62,14 +62,38 @@ public class RunServiceHttpClient : RawHttpClientBase
...
         }
+
+        public Task CompleteJobAsync(
+            Uri requestUri,
+            Guid planId,
+            Guid jobId,
+            CancellationToken cancellationToken = default)
+        {
+            HttpMethod httpMethod = new HttpMethod("POST");
+            var payload = new {
```
yaananth, line 83: more needed, but that's for later

src/Runner.Common/RunServer.cs
```diff
@@ -55,5 +56,12 @@ public Task<AgentJobRequestMessage> GetJobMessageAsync(string id, CancellationTo
             return jobMessage;
         }
 
+        public Task CompleteJobAsync(Guid planId, Guid jobId, CancellationToken cancellationToken)
+        {
+            CheckConnection();
+            return RetryRequest(
+                async () => await _runServiceHttpClient.CompleteJobAsync(requestUri, planId, jobId, cancellationToken), cancellationToken);
+        }
+
```
TingluoHuang, line 65:
```suggestion
```

src/Runner.Worker/JobRunner.cs
```diff
@@ -40,21 +39,34 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
+                Uri jobServerUrl = systemConnection.Url;
+
+                Trace.Info($"Creating job server with URL: {jobServerUrl}");
+                // jobServerQueue is the throttling reporter.
+                _jobServerQueue = HostContext.GetService<IJobServerQueue>();
+                VssConnection jobConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, new DelegatingHandler[] { new ThrottlingReportHandler(_jobServerQueue) });
+                await jobServer.ConnectAsync(jobConnection);
+
+                // yash: kicks start jobServerQueue
```
yaananth, line 65: LOL

src/Runner.Common/Unit.cs
```diff
@@ -0,0 +1,8 @@
...
+{
+    public readonly struct Unit
```
TingluoHuang, line 4: do we have to have this struct?

yaananth: It's used in `RunnerService`, to have a way to get data "empty" data for Task. I think it's nice to have this, can be used for any later uses. [Dotnet code has it as well](https://github.com/github/actions-dotnet/blob/b5850133b3c04d4293f029eb911e7fdcb3bd2987/Launch/Sdk/Server/Unit.cs)

src/Runner.Worker/JobRunner.cs
```diff
@@ -206,6 +217,59 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
             }
         }
 
+        private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IJobServer jobServer, IExecutionContext jobContext, Pipelines.AgentJobRequestMessage message, TaskResult? taskResult = null)
```
thboop, line 220: Should we just pass in a single IRunnerService for the server we intend to return here, and provide overloads for it?

Its kind of weird to pass in two variables when one is expected to be null

yaananth: they both don't share common props  :( `IJobServer` and `IRunServer` are different)...
Any alternatives to make it clean? I did have two methods for their specific calls, but I need a some way of abstracting that, so I created that common method that takes both, else I have to have `if else` every where we call this, which is bit disturbing to see...

src/Runner.Worker/JobRunner.cs
```diff
@@ -206,6 +217,59 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+                jobContext.Warning(string.Format(Constants.Runner.Node12DetectedAfterEndOfLife, actions));
+            }
+
+            // Load any upgrade telemetry
+            LoadFromTelemetryFile(jobContext.Global.JobTelemetry);
+
+            // Make sure we don't submit secrets as telemetry
+            MaskTelemetrySecrets(jobContext.Global.JobTelemetry);
+
+            Trace.Info($"Raising job completed");
```
thboop, line 248:
```suggestion
            Trace.Info($"Raising job completed against run service");
```

Just for my own sanity if we could 🥇

src/Runner.Worker/JobRunner.cs
yaananth: Like we call `await CompleteJobAsync` 5 times and they are common to both flows, so 🤷‍♂️

thboop: Definitely a super nit suggestion so feel free to ignore

But they both inherit from IJobServer iirc.

Base case could throw, overloads of IJobServer and IRunServer do as expected, and when you call it c# should automatically call the more detailed overload if one exists. Or you could just pass in a single IRunService and check the type and fail if its not an expected type.

src/Runner.Worker/JobRunner.cs
```diff
@@ -206,6 +217,59 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+            }
+            else
+            {
+                return await CompleteJobAsync(jobServer, jobContext, message, taskResult);
+            }
+        }
+
+        private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecutionContext jobContext, Pipelines.AgentJobRequestMessage message, TaskResult? taskResult = null)
+        {
+            jobContext.Debug($"Finishing: {message.JobDisplayName}");
```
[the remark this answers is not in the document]
yaananth: `CleanupTempDirectory` is not intentional, thanks for catching it.

yeah the rest is intentional

src/Runner.Worker/JobRunner.cs
yaananth: addressed!

src/Runner.Worker/JobRunner.cs
```diff
@@ -206,6 +218,66 @@ public async Task<TaskResult> RunAsync(Pipelines.AgentJobRequestMessage message,
...
+            TaskResult result = jobContext.Complete(taskResult);
+            if (jobContext.Global.Variables.TryGetValue("Node12ActionsWarnings", out var node12Warnings))
+            {
+                var actions = string.Join(", ", StringUtil.ConvertFromJson<HashSet<string>>(node12Warnings));
+                jobContext.Warning(string.Format(Constants.Runner.Node12DetectedAfterEndOfLife, actions));
+            }
+
+            // Clean TEMP after finish process jobserverqueue, since there might be a pending fileupload still use the TEMP dir.
```
thboop, line 247:
```suggestion
            // Make sure to clean temp after file upload since they may be pending fileupload still use the TEMP dir.
```

thboop (approved):
lgtm

PR #2426, opened 2023-02-09
Runner service exit after consecutive re-try exits

### LINKED ISSUE #879: Improve error handling
**Describe the enhancement**
So-called "retryable" errors (for example, error code 2: Not configured) cause the runner not to exit. It is assumed, that the underlying cause will be fixed within the next attempts. Every 5 seconds, the runner tries to start again: 

There should be a maximum number of attempts - self-hosted runners are often deployed in an unattended manner and this might cause indefinite loops.

**Code Snippet**
https://github.com/actions/runner/blob/main/src/Misc/layoutbin/RunnerService.js#L53

nikola-jokic:
Fixes #879

Note: If you think 10 is not a good number, let me know :relaxed:

### DIFF
--- a/src/Misc/layoutbin/RunnerService.js
+++ b/src/Misc/layoutbin/RunnerService.js
@@ -24,8 +24,8 @@ if (exitServiceAfterNFailures <= 0) {
   exitServiceAfterNFailures = NaN;
 }
 
-var consecutiveFailureCount = 0;
-var retryRestarts = 0;
+var unknownFailureRetryCount = 0;
+var retriableFailureRetryCount = 0;
 
 var gracefulShutdown = function () {
   console.log("Shutting down runner listener");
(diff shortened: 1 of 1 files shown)


src/Misc/layoutbin/RunnerService.js
```diff
@@ -25,6 +25,7 @@ if (exitServiceAfterNFailures <= 0) {
 }
 
 var consecutiveFailureCount = 0;
+var retryRestarts = 0;
```
fhammerl, line 28: var unknownFailureRetryCount = 0;
var retriableFailureRetryCount = 0;

fhammerl (approved):
LGTM

pranaliwalunj306-ai:
ca78e773f560cfb32daea3ee52f63428

PR #2430, opened 2023-02-12
Fix color code detection in OutputManager

### LINKED ISSUE #2341: Problem matchers do not work on colored output
**Describe the bug**
Color codes appear to affect regex matching of problem matchers 

**To Reproduce**
1. Create a problem matcher that matches plaintext, e.g. `"regexp": "Hello"`
1. Run an action that produces color output, for example: `bash -c "printf '\e[31mHello!\e(B\e[m\n'"`
1. Observe that no match appears in the actions output or inline on a PR "files" tab.

As a more real-life example, I have a PR with colored output and none of the expected annotations:
https://github.com/corewa-rs/corewars/actions/runs/3777317784/jobs/6421196314

**Expected behavior**
Most (all?) are written against plain (un-colored) text, but it would be nice to have colorized output in the action logs while still supporting colored output.

[This document](https://github.com/jsoref/actions-runner/blob/f9c2bf1dd72541bf039c3c5fa4129814181ca261/docs/adrs/0276-problem-matchers.md?plain=1#L113) indicates that color codes should be stripped before applying matchers.

Version of your runner?

## What's not working?
Please include error messages and screenshots.

## Job Log Output
If applicable, include the relevant part of the job / step log output here.  All sensitive information should already be masked out, but please double-check before pasting here.

If applicable, add relevant diagnostic log information.  Logs are located in the runner's `_diag` folder. The runner logs are prefixed with `Runner_` and the worker logs are prefixed with `Worker_`. Each job run correlates to a worker log.  All sensitive information should already be masked out, but please double-check before pasting here.

ian-h-chamberlain:
Fixes #2341 

This was a simple case of a mistaken translation from shell escape codes (which support octal `\NNN`) to C# (which only supports hex `\xNN` and `\uNNNN` in string literals, but does support octal `\NNN` in `Regex`). The test was passing because it also used incorrect escape codes, which matched the incorrect prefix.

Edit: manual testing against the original repro repo: https://github.com/ian-h-chamberlain/actions-color-repro/actions/runs/4157866967

### DIFF
--- a/src/Runner.Worker/Handlers/OutputManager.cs
+++ b/src/Runner.Worker/Handlers/OutputManager.cs
@@ -13,10 +13,10 @@ namespace GitHub.Runner.Worker.Handlers
 {
     public sealed class OutputManager : IDisposable
     {
-        private const string _colorCodePrefix = "\033[";
+        private const string _colorCodePrefix = "\x1b[";
         private const int _maxAttempts = 3;
         private const string _timeoutKey = "GITHUB_ACTIONS_RUNNER_ISSUE_MATCHER_TIMEOUT";
-        private static readonly Regex _colorCodeRegex = new(@"\x0033\[[0-9;]*m?", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+        private static readonly Regex _colorCodeRegex = new(@"\x1b\[[0-9;]*m?", RegexOptions.Compiled | RegexOptions.CultureInvariant);
         private readonly IActionCommandManager _commandManager;
         private readonly ContainerInfo _container;
         private readonly IExecutionContext _executionContext;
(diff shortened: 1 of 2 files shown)


ian-h-chamberlain:
This was marked as stale, but as far as I know is still needed and the fix is still valid. Not sure if there's any way to unstale or get a review / CI besides just commenting here...

nchammas:
Since this PR was created, the main README has been updated with the following note:

https://github.com/actions/runner/blob/dde968bf5724d5b9ba7a759f9a114a305e2c76e7/README.md#L25

Given that, plus the silence from maintainers on this PR, I guess we are not going to see this merged.

I am hitting the same issue with problem matchers that you are trying to fix here. It's frustrating that we have to choose between contorting matcher regexes to handle color codes, or stripping color entirely from log output.

PR #2439, opened 2023-02-15
Properly guard upload

yacaovsnc:
Fix 2.302 upload log issue

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -237,6 +237,18 @@ namespace GitHub.Runner.Common
             if (!_resultsClientInitiated)
             {
                 Trace.Verbose("Skipping results upload");
+                try
+                {
+                    if (deleteSource)
+                    {
+                        File.Delete(path);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Trace.Info("Catch exception during delete skipped results upload file.");
+                    Trace.Error(ex);
+                }
                 return;
             }
 
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -20,7 +20,7 @@ namespace GitHub.Runner.Common
         void Start(Pipelines.AgentJobRequestMessage jobRequest);
         void QueueWebConsoleLine(Guid stepRecordId, string line, long? lineNumber = null);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
-        void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines = 0);
+        void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines);
         void QueueTimelineRecordUpdate(Guid timelineId, TimelineRecord timelineRecord);
     }
 
@@ -84,6 +84,8 @@ namespace GitHub.Runner.Common
         private bool _webConsoleLineAggressiveDequeue = true;
         private bool _firstConsoleOutputs = true;
 
+        private bool _resultsClientInitiated = false;
+
         public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
@@ -109,8 +111,8 @@ namespace GitHub.Runner.Common
             {
                 Trace.Info("Initializing results client");
                 _jobServer.InitializeResultsClient(new Uri(resultsReceiverEndpoint), accessToken);
+                _resultsClientInitiated = true;
             }
-
 
             if (_queueInProcess)
             {
@@ -230,11 +232,17 @@ namespace GitHub.Runner.Common
             _fileUploadQueue.Enqueue(newFile);
         }
 
-        public void QueueResultsUpload(Guid recordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines)
+        public void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines)
         {
-            if (recordId == _jobTimelineRecordId) 
+            if (!_resultsClientInitiated)
             {
-                Trace.Verbose("Skipping job log {0} for record {1}", path, recordId);
+                Trace.Verbose("Skipping results upload");
+                return;
+            }
+
+            if (timelineRecordId == _jobTimelineRecordId && String.Equals(type, CoreAttachmentType.ResultsLog, StringComparison.Ordinal))
+            {
+                Trace.Verbose("Skipping job log {0} for record {1}", path, timelineRecordId);
                 return;
             }
 
@@ -246,14 +254,14 @@ namespace GitHub.Runner.Common
                 Type = type,
                 PlanId = _planId.ToString(),
                 JobId = _jobTimelineRecordId.ToString(),
-                RecordId = recordId,
+                RecordId = timelineRecordId,
                 DeleteSource = deleteSource,
                 Finalize = finalize,
                 FirstBlock = firstBlock,
                 TotalLines = totalLines,
             };
 
-            Trace.Verbose("Enqueue results file upload queue: file '{0}' attach to job {1} step {2}", newFile.Path, _jobTimelineRecordId, recordId);
+            Trace.Verbose("Enqueue results file upload queue: file '{0}' attach to job {1} step {2}", newFile.Path, _jobTimelineRecordId, timelineRecordId);
             _resultsFileUploadQueue.Enqueue(newFile);
         }
 
@@ -483,11 +491,8 @@ namespace GitHub.Runner.Common
                             }
                             else if (String.Equals(file.Type, CoreAttachmentType.ResultsLog, StringComparison.OrdinalIgnoreCase))
                             {
-                                if (file.RecordId != _jobTimelineRecordId)
-                                {
-                                    Trace.Info($"Got a step log file to send to results service.");
-                                    await UploadResultsStepLogFile(file);
-                                }
+                                Trace.Info($"Got a step log file to send to results service.");
+                                await UploadResultsStepLogFile(file);
                             }
                         }
                         catch (Exception ex)
@@ -801,8 +806,7 @@ namespace GitHub.Runner.Common
             {
                 // Upload the step summary
                 Trace.Info($"Starting to upload summary file to results service {file.Name}, {file.Path}");
-                var cancellationTokenSource = new CancellationTokenSource();
-                await _jobServer.CreateStepSummaryAsync(file.PlanId, file.JobId, file.RecordId, file.Path, cancellationTokenSource.Token);
+                await _jobServer.CreateStepSummaryAsync(file.PlanId, file.JobId, file.RecordId, file.Path, CancellationToken.None);
 
                 uploadSucceed = true;
             }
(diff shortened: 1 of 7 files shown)


src/Runner.Common/JobServerQueue.cs
```diff
@@ -230,21 +232,31 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
             _fileUploadQueue.Enqueue(newFile);
         }
 
-        public void QueueSummaryUpload(Guid stepRecordId, string name, string path, bool deleteSource)
+        public void QueueResultsUpload(Guid recordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines)
```
TingluoHuang, line 235: `recordId` or `timelineRecordId`, i think the name better matches what you have in the interface.

TingluoHuang: same for `long totalLines` => `long totalLines = 0`

src/Runner.Common/JobServerQueue.cs
```diff
@@ -140,12 +142,12 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest)
             _fileUploadDequeueTask = ProcessFilesUploadQueueAsync();
 
             Trace.Info("Start results file upload queue.");
-            _summaryUploadDequeueTask = ProcessSummaryUploadQueueAsync();
+            _resultsUploadDequeueTask = ProcessResultsUploadQueueAsync();
```
TingluoHuang, line 145: should we skip this if `_resultsClientInitiated == false`?

src/Runner.Common/JobServerQueue.cs
```diff
@@ -230,21 +232,31 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
...
 
-        public void QueueSummaryUpload(Guid stepRecordId, string name, string path, bool deleteSource)
+        public void QueueResultsUpload(Guid recordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines)
         {
+            if (recordId == _jobTimelineRecordId) 
+            {
+                Trace.Verbose("Skipping job log {0} for record {1}", path, recordId);
+                return;
+            }
+
```
TingluoHuang, line 242: should we also return early if `_resultsClientInitiated == false`?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Default parameters are a source of bugs because it's possible the default values on the interface and the implementation are different.  Then depending on how the type is initiated, there could be different behaviors.  I believe the best practice is to only have the value defined at the interface level?  This way if someone mistakenly instantiate the concrete type, then they must explicitly pass in the value for this parameter.  

I will update the `recordId` name.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: If we don't enqueue then I think we can skip process the queue.  I can guard at enqueue too then we can skip this.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Sure, doesn't hurt to guard here as well.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: hmmm, it looks pretty ugly to skip this as this is also called in `Shutdown`.

Do you think it's enough just to:
1. Skip enqueue any file
2. Make sure any file on the queue is dropped with no-op

If we have to remove processing this queue, the code looks pretty ugly. 😓

TingluoHuang: Skip enqueue should be good enough.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -20,7 +20,7 @@ public interface IJobServerQueue : IRunnerService, IThrottlingReporter
         void Start(Pipelines.AgentJobRequestMessage jobRequest);
         void QueueWebConsoleLine(Guid stepRecordId, string line, long? lineNumber = null);
         void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type, string name, string path, bool deleteSource);
-        void QueueSummaryUpload(Guid stepRecordId, string name, string path, bool deleteSource);
+        void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines = 0);
```
TingluoHuang, line 23:
```suggestion
        void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines);
```

src/Runner.Common/JobServerQueue.cs
```diff
@@ -437,18 +455,21 @@ private async Task ProcessFilesUploadQueueAsync(bool runOnce = false)
...
-                List<SummaryUploadFileInfo> filesToUpload = new();
-                SummaryUploadFileInfo dequeueFile;
-                while (_summaryFileUploadQueue.TryDequeue(out dequeueFile))
+                List<ResultsUploadFileInfo> filesToUpload = new();
+                ResultsUploadFileInfo dequeueFile;
+                while (_resultsFileUploadQueue.TryDequeue(out dequeueFile))
                 {
-                    filesToUpload.Add(dequeueFile);
-                    // process at most 10 file upload.
+                    if (_resultsClientInitiated)
```
TingluoHuang, line 468: do you still need this if we skip it in the enqueue side already?

src/Runner.Common/Logging.cs
```diff
@@ -32,15 +32,30 @@ public class PagingLogger : RunnerService, IPagingLogger
         private string _pagesFolder;
         private IJobServerQueue _jobServerQueue;
 
+        // For Results
+        public static string BlocksFolder = "blocks";
+
+        // 2 MB
+        public const int BlockSize = 2 * 1024 * 1024;
```
TingluoHuang, line 39: move these public variables up to where other public variables live.

TingluoHuang: any reason not keep using 8MB?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -872,7 +872,7 @@ public void QueueSummaryFile(string name, string filePath, Guid stepRecordId)
                 throw new FileNotFoundException($"Can't upload (name:{name}) file: {filePath}. File does not exist.");
             }
 
-            _jobServerQueue.QueueSummaryUpload(stepRecordId, name, filePath, deleteSource: false);
+            _jobServerQueue.QueueResultsUpload(stepRecordId, name, filePath, ChecksAttachmentType.StepSummary, deleteSource: false, finalize: false, firstBlock: false);
```
TingluoHuang, line 875: the `finalize: false, firstBlock: false` feels odd, since you are done with uploading the only block for summary.
should it be `finalize: true, firstBlock: true`?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: hmm, not really, I can remove this.  I just feel it's safer to check the condition right before we act on it, but since the queue is a private member variable, I guess it's okay to just guard it at enqueue time.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -804,6 +833,34 @@ private async Task UploadSummaryFile(SummaryUploadFileInfo file)
...
             }
         }
+
+        private async Task UploadResultsStepLogFile(ResultsUploadFileInfo file)
+        {
+            bool uploadSucceed = false;
+            try
+            {
+                Trace.Info($"Starting upload of step log file to results service {file.Name}, {file.Path}");
+                var cancellationTokenSource = new CancellationTokenSource();
```
TingluoHuang, line 843: why create a cancellationTokenSouce here?
i didn't find anything that calls `Cancel()` on it.

yacaovsnc: I think this was created because `RawHttpClientBase` expected one.

I am making a lot of changes in the http client in a separate branch and will submit a new PR very soon, and I will remove those `cancellationTokenSource` there.

src/Runner.Common/Logging.cs
yacaovsnc: Yeah Azure Append Blob supports 4MB blocks only - and since this is really minimum 8MB, we need to reduce the size to be safe.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -459,19 +480,30 @@ private async Task ProcessSummaryUploadQueueAsync(bool runOnce = false)
...
                         try
                         {
-                            await UploadSummaryFile(file);
+                            if (String.Equals(file.Type, ChecksAttachmentType.StepSummary, StringComparison.OrdinalIgnoreCase))
+                            {
+                                await UploadSummaryFile(file);
+                            }
+                            else if (String.Equals(file.Type, CoreAttachmentType.ResultsLog, StringComparison.OrdinalIgnoreCase))
+                            {
+                                if (file.RecordId != _jobTimelineRecordId)
```
TingluoHuang, line 497: do we need this filter? i thought you already filter it out in the enqueue part.

src/Runner.Common/JobServerQueue.cs
TingluoHuang: you can just pass `CancellationToken.None` to the `await _jobServer.CreateResultsStepLogAsync()` if you don't want to have cancellation for now.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Well, this is written in mind for the next PR.  I am removing the filter up there already.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -871,8 +871,8 @@ public void QueueSummaryFile(string name, string filePath, Guid stepRecordId)
             {
                 throw new FileNotFoundException($"Can't upload (name:{name}) file: {filePath}. File does not exist.");
             }
-
-            _jobServerQueue.QueueSummaryUpload(stepRecordId, name, filePath, deleteSource: false);
+
```
TingluoHuang, line 874:
```suggestion

```

TingluoHuang (approved):

src/Runner.Common/JobServerQueue.cs
```diff
@@ -230,21 +232,37 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
             _fileUploadQueue.Enqueue(newFile);
         }
 
-        public void QueueSummaryUpload(Guid stepRecordId, string name, string path, bool deleteSource)
+        public void QueueResultsUpload(Guid timelineRecordId, string name, string path, string type, bool deleteSource, bool finalize, bool firstBlock, long totalLines)
         {
+            if (!_resultsClientInitiated)
+            {
+                Trace.Verbose("Skipping results upload");
```
TingluoHuang, line 239: should we also delete the file base on `deleteSource==true` when we skip?

src/Runner.Common/JobServerQueue.cs
```diff
@@ -230,21 +232,37 @@ public void QueueFileUpload(Guid timelineId, Guid timelineRecordId, string type,
...
+                return;
+            }
+
+            if (timelineRecordId == _jobTimelineRecordId && String.Equals(type, CoreAttachmentType.ResultsLog, StringComparison.Ordinal))
+            {
+                Trace.Verbose("Skipping job log {0} for record {1}", path, timelineRecordId);
+                return;
```
TingluoHuang, line 246: same here, do we need to delete the file?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Good call, deleted file here.

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Not going to delete here since I am going to remove this check now in my next PR.  I already have the code mostly ready, just need to test and submit a PR.  

The general ff case is handled by the first delete above.

TingluoHuang (approved):

PR #2441, opened 2023-02-16
Remove latest tag from runner image publishing workflow

Link-:
### Context

The `latest` tag creates a big maintenance and security burden. Staying aligned with https://github.com/actions/actions-runner-controller/issues/2056 I think it's best to remove the latest tag from this runner image as well.

Users will need to pin a specific runner version and create more controlled mechanisms to switch to the latest version.


Link-:
Pending discussions

PR #2443, opened 2023-02-16
Add --no-default-labels config option to self-hosted runners

gabriel-samfira:
My apologies for the large wall of text bellow :smile: .

This change allows us to remove the default labels (`self-hosted`, `linux`, `x64` for example) attached to runners. In a lot of cases, workflow authors only include the `self-hosted` label to target runners. In an environment where we manage pools of ephemeral self-hosted runners, this scenario negates the usefulness of custom labels, causing a random runner to pick up a job.

For example, if you have a set of runners with custom labels like:

  * gpu
  * high-memory

And another with:

  * fpga
  * medium-memory

And the workflow author defines just ```self-hosted``` (because it's easy and the documentation encourages authors to use it to target self hosted runners), you may get a runner with a `gpu` when in fact you required one with `fpga`.

Yes, teams could be encouraged to target runners using only custom labels, but when you have many teams, this doesn't always work out, and we end up consuming runner types we should not, potentially generating unnecessary cost (gpu instances may be more expensive than general purpose ones).

It's easier to deal with support tickets where a team asks why the `self-hosted` label is not working, than having the ops team responsible for maintaining the runners, ping each team asking them to change their workflows so as not to accidentally consume expensive runners when they actually should be using general purpose runners.

Runner groups are only available to enterprise users. Removing the default labels would be useful even for single repos with more collaborators and free tier orgs that a lot of open source projects use. It's also simpler to use labels to target runners if we don't have strict ACL requirements  (that groups enable) and we simply want a simple way to target specific runners.

Allowing us to remove the default labels will give us the ability to define unique label sets and thus schedule jobs more efficiently. It also allows us to better react to `queued` workflow webhook and pick the right runner type to spin up if we have automation tools that allow us to define multiple pools with different characteristics (like detailed above). This saves on cost as we can correctly target runner types we can choose to spin up only on-demand as opposed to having them linger around.

Hoping this makes sense smile :smile:.

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -132,6 +132,7 @@ namespace GitHub.Runner.Common
                     public static readonly string GenerateServiceConfig = "generateServiceConfig";
                     public static readonly string Help = "help";
                     public static readonly string Local = "local";
+                    public static readonly string NoDefaultLabels = "no-default-labels";
                     public static readonly string Replace = "replace";
                     public static readonly string DisableUpdate = "disableupdate";
                     public static readonly string Once = "once"; // Keep this around since customers still relies on it
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace GitHub.Runner.Common
 {
--- a/src/Runner.Listener/CommandSettings.cs
+++ b/src/Runner.Listener/CommandSettings.cs
@@ -29,8 +29,8 @@ namespace GitHub.Runner.Listener
         private readonly Dictionary<string, string[]> validOptions = new()
         {
             // Valid configure flags and args
-            [Constants.Runner.CommandLine.Commands.Configure] = 
-                new string[] 
+            [Constants.Runner.CommandLine.Commands.Configure] =
+                new string[]
                 {
                     Constants.Runner.CommandLine.Flags.DisableUpdate,
                     Constants.Runner.CommandLine.Flags.Ephemeral,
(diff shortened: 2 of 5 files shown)


gabriel-samfira:
Hi folks!

Any chance to get some feedback on this PR? Thanks!

alexsanderp:
Any chance to get some feedback on this PR? Thanks!

pavlovic-ivan:
I would just like to express support for this issue and PR by adding this comment. Hopefully @actions/actions-runtime will pick this up soon.

sameeraksc:
we also need this feature badly

TingluoHuang (approved):

gabriel-samfira:
Hi @TingluoHuang

Some of the linting failures seem to be at lines I didn't change. Would you like me to fix those as well? There is also a file encoding error thrown.

gabriel-samfira:
Ohh! Thanks for making the changes and for merging! :smile:

shankarRaman:
@gabriel-samfira does this mean that the default labels will no longer be added when registering runners? We use https://github.com/actions/actions-runner-controller for creating our runners.

gabriel-samfira:
@shankarRaman This PR adds the ability to opt out of the default labels when configuring the runner using `Runner.Listener configure`. But this is something that the automation you're using has to opt into. I don't think `action-runner-controller` uses `config.sh` to configure the runners, so this might not help.

If you're using `action-runner-controller` and need to configure your runners without the default labels, you should probably open an issue or ask on that projects page. You should get a much more informed answer than I am able to give.

nkomm00:
In which version of actions runner "--no-default-labels" is available?

Rayshard:
> In which version of actions runner "--no-default-labels" is available?

2.305.0+ https://github.com/actions/runner/releases/tag/v2.305.0

PR #2447, opened 2023-02-17
Add job log upload support

yacaovsnc:
Upload job logs to results and some refactoring to reduce duplicated code.

### DIFF
--- a/src/Runner.Common/JobServer.cs
+++ b/src/Runner.Common/JobServer.cs
@@ -32,6 +32,7 @@ namespace GitHub.Runner.Common
         Task<TaskAttachment> CreateAttachmentAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, Guid timelineRecordId, String type, String name, Stream uploadStream, CancellationToken cancellationToken);
         Task CreateStepSummaryAsync(string planId, string jobId, Guid stepId, string file, CancellationToken cancellationToken);
         Task CreateResultsStepLogAsync(string planId, string jobId, Guid stepId, string file, bool finalize, bool firstBlock, long lineCount, CancellationToken cancellationToken);
+        Task CreateResultsJobLogAsync(string planId, string jobId, string file, bool finalize, bool firstBlock, long lineCount, CancellationToken cancellationToken);
         Task<TaskLog> CreateLogAsync(Guid scopeIdentifier, string hubName, Guid planId, TaskLog log, CancellationToken cancellationToken);
         Task<Timeline> CreateTimelineAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, CancellationToken cancellationToken);
         Task<List<TimelineRecord>> UpdateTimelineRecordsAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, IEnumerable<TimelineRecord> records, CancellationToken cancellationToken);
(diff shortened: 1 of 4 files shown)


src/Runner.Common/JobServer.cs
```diff
@@ -335,6 +336,14 @@ public Task CreateResultsStepLogAsync(string planId, string jobId, Guid stepId,
             throw new InvalidOperationException("Results client is not initialized.");
         }
 
+        public Task CreateResultsJobLogAsync(string planId, string jobId, string file, bool finalize, bool firstBlock, long lineCount, CancellationToken cancellationToken)
+        {
+            if (_resultsClient != null)
+            {
+                return _resultsClient.UploadResultsJobLogAsync(planId, jobId, file, finalize, firstBlock, lineCount, cancellationToken: cancellationToken);
+            }
+            throw new InvalidOperationException("Results client is not initialized.");
```
robherley, line 345: Double checking this time to make sure the exception will be caught 😄

brittanyellich: Since we're dual-uploading, do we even need this exception? I wonder if we could just add some tracing here instead.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -272,12 +310,32 @@ public async Task UploadResultsStepLogAsync(string planId, string jobId, Guid st
...
             }
         }
 
         private MediaTypeFormatter m_formatter;
         private Uri m_resultsServiceUrl;
         private string m_token;
     }
+
+    // Constants specific to results
+    public static class Constants
```
robherley, line 323: ❤️

src/Runner.Common/JobServer.cs
yacaovsnc: We actually caught the exception before as well - the issue was in the handling part. 😓 

https://github.com/actions/runner/blob/main/src/Runner.Common/JobServerQueue.cs#L520

Here we wanted to log a telemetry for all the failures, but on GHES server doesn't understand this record and errored out.

robherley (dismissed):

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -60,27 +66,51 @@ public async Task<GetSignedStepLogsURLResponse> GetStepLogUploadUrlAsync(string
...
+        {
+            var request = new GetSignedJobLogsURLRequest()
+            {
+                WorkflowJobRunBackendId = jobId,
+                WorkflowRunBackendId = planId,
+            };
+
+            var getJobLogsSignedBlobURLEndpoint = new Uri(m_resultsServiceUrl, Constants.GetJobLogsSignedBlobURL);
+
+            return await GetResultsSasURLResponse<GetSignedJobLogsURLRequest, GetSignedJobLogsURLResponse>(getJobLogsSignedBlobURLEndpoint, cancellationToken, request);
```
brittanyellich, line 84: Can we use `SignedURL` instead of `SasURL` for consistency?

src/Runner.Common/JobServer.cs
yacaovsnc: Now we are properly guarded and if the feature flag is not on, nothing will be put on the Results processing queue.

I also tested these changes against a GHES instance.

brittanyellich (approved):

src/Runner.Common/JobServer.cs
yacaovsnc: @brittanyellich we shouldn't get to this point if we aren't going to upload to results, so I feel the exception is still appropriate.

lokesh755 (dismissed):

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
yacaovsnc: Renamed!

lokesh755 (approved):

Issue #403, opened 2020-04-02
./config.sh remove - Failed: Removing service


nicolas-marcq:
**Describe the bug**
Cannot remove the runner

**To Reproduce**
Run:
```
./config.sh remove
```

**Expected behavior**
The runner is removed.

## Runner Version and Platform
2.165.2
Linux Ubuntu 18.04

## Job Log Output
```
 ./config.sh remove

# Runner removal

Removing service
Failed: Removing service
Unconfigure service first
```

I tried by manually removing the service from systemd. Still not ok.

TingluoHuang:
try ‘sudo ./svc.sh uninstall’ first

TingluoHuang:
close now, please reopen if suggested doesn't help.

ffittschen:
@TingluoHuang  We still see the same error on macOS. When calling `sudo ./svc.sh uninstall`, the script just responds with _Must not run with sudo_.

m-hilgendorf:
This is failing for me as well. `svc.sh` fails with:

```
./svc.sh uninstall
uninstalling actions.runner.jitbot-test-runner.MacCloud
stopping actions.runner.jitbot-test-runner.MacCloud
Could not find domain for port (Aqua)
Failed: failed to unload /Users/user198536/Library/LaunchAgents/actions.runner.<OMITTED>.plist
```

I've looked at #349 for a workaround but my user doesn't have sudoers permissions on this machine. 

Is there a better way to reconfigure the runner than uninstalling it?

jgwinner:
I ran into this issue.

The fix for me was the  `sudo ./svc.sh uninstall ` command, which is not mentioned in the GitHub control panel when you try to remove the service.

I think "unconfigure service first' is slightly misleading, as the command is actually uninstall at the O/S level, and "remove" at the GitHub level.

Secondly, if you clone a VM with a GitLab runner, the two are tied at the hip; removal of one Runner screws up the other one, but I will open a second ticket for that.

         == John ==

jgwinner:
To be clear: this should be reopened, and `sudo ./svc.sh` uninstall command, should be mentioned in the GitHub control panel when you try to remove the service.

SeriousM:
> To be clear: this should be reopened, and `sudo ./svc.sh` uninstall command, should be mentioned in the GitHub control panel when you try to remove the service.

and should be mentioned in the error message after `./config.sh remove` fails.

```
[...]

Issue #1959, opened 2022-06-21
Github action runner svc start not working on MacOS Monterey


c3-kaspesi:
**Describe the bug**

Trying to set up a self-hosted github actions runner as a service so it launches on startup.  Currently am getting 
```
starting actions.runner.X
Load failed: 5: Input/output error
Try running `launchctl bootstrap` as root for richer errors.
status actions.runner.X:
/Users/ec2-user/Library/LaunchAgents/actions.runner.X.plist
Stopped
```
**To Reproduce**

1. Configure action runner with ./config.sh
2. Install the service with ./svc.sh install (Successful) 
3. Start the service with ./svc.sh start 

**Expected behavior**
Service starts and can be seen running:
`launchctl list | grep X`

## Runner Version and Platform
Version of your runner?
Latest 2.293

OS of the machine running the runner? OSX/Windows/Linux/...
MacOS Monterey

## What's not working?
Service Starting

## Workaround:
I can get the runner started by running:
sudo chown root:wheel /Users/ec2-user/Library/LaunchAgents/actions.runner.X
sudo launchctl bootstrap system ~/Library/LaunchAgents/actions.runner.c3-e-c3vsce.GAR-5.plist
Though this is installing the runner in the system domain instead of the user, and the service does not start on restart.

asenyaev:
I've faced the same issue today on my 2 different devices with macOS (x86_64 and ARM64). 

In addition, if to run the command which starts a service in svc.sh:
```
launchctl load -w /Users/<user>/Library/LaunchAgents/actions.runner.<organization>.<runner_name>.plist

echo $?
134
```
it doesn't show anything, but the exit code is 134.

I use 2.294.0 version of Actions Runner.

asenyaev:
Found the source of this issue and a solution - launchd can't load user's services until it logs in via GUI, that's why need to login under the certain user via GUI. In addition, auto login feature can help in that case.

F-WRunTime:
Does Mac have a method to remote log into the GUI? We do not have easy access to the hardware. Any recommendations? 
I am having this same issue. 
Notably our M1 regularly cycles between online and offline when connected to github. Is there a networking configuration change needed in order to fix connectivity and stay online for extended period of time?

F-WRunTime:
> You need to remote login via vnc and then enable `Automatic Login` to the account you are using in system preferences, under `User & Groups` -> `Login Options`. Now next time the machine turns on GUI should be enabled on that account.

[...]

PR #2449, opened 2023-02-18
fix #349 #403 #947 - fix headless setup on Mac OSX using ssh

### LINKED ISSUE #349: OSX Runner service fails to start

tcptps:
This should fix #349 #403 #947 because it fixes the `Could not find domain for port (Aqua)` error, by moving the agent to the Background context where it belongs. 

**Why it works**  

By default (without the LimitLoadToSessionType key in the plist file) launchd tries to load the agent into the Aqua context but this context doesnt exist when you are only logged in using ssh. The same applies to launchctl as it also by default tries to load the agent into the Aqua context unless you specify a different context (using the `-S` option in this case `-S Background` ). The Background context on the other hand exists when you are logged in using ssh and when you are logged in using the GUI, or to be more precises almost always when you are logged in somehow.

**Potential Problems**  
 If the runner (or more specificly the sh file I changed) is updated to my changed version while the service is loaded, you won't be able to unload the currently loaded agent as it is not loaded in the Background context but instead int the Aqua context. 
I dont know how you want to deal with this problem, but I have created seperate pull request #2450  that fixes this potential problem as well, because I don't know If my solution is acceptable. If someone wants to implement their own solution see the further explanation below.
<details>
  <summary>further explanation of the potential problem</summary>

In the beginning the agent was loaded using this line ` launchctl load -w "${PLIST_PATH}" || failed "failed to load ${PLIST_PATH}"` since no context is specified the agent is now loaded into the Aqua context on the system because it is the default one.   
Now the runner is updated in a way that my code changes are applied without unloading the service before updating, or restarting the system (because a restart will also unload all services), meaning the runner is updated while the agent is loaded, threfore the following line will be used to unload the service after the update `launchctl unload -S Background  "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"` which will fail because the agent is (as mentioned above) still loaded into the Aqua context and because of that cant be unloaded from the Background context but instead must be unloaded from the Aqua context using this line `launchctl unload "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"` (yes `launchctl unload -S Aqua  "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"` would also work.) when unloading for the first time after the update.   
After this first initial unload after the update the sevice can be unloaded from the Background context without any problems using this line `launchctl unload -S Background  "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"` because it was (breforehand) loaded in the Background context using `launchctl load -w -S Background "${PLIST_PATH}" || failed "failed to load ${PLIST_PATH}"`
No you cant load the same service twice into different contexts. 

 **TLDR**. 
The problem occurs with liveupdates only.
To mitigate the problem, the agent when unloaded for the first time after the update must be unloaded using`launchctl unload "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"`

</details> 

**Background**  
  I am currently working on a small complete admins guide to launchd and therefore doing extensive research.
  I came across this [Issue](https://github.com/buildkite/docs/issues/353) and saw it mentioned here multiple times.

### DIFF
--- a/src/Misc/layoutbin/actions.runner.plist.template
+++ b/src/Misc/layoutbin/actions.runner.plist.template
@@ -27,5 +27,12 @@
     <string>Interactive</string>
     <key>SessionCreate</key>
     <true/>
+    <key>LimitLoadToSessionType</key>
+    <array>
+      <string>Aqua</string>
+      <string>LoginWindow</string>
+      <string>Background</string>
+      <string>StandardIO</string>
+    </array>
   </dict>
 </plist>
--- a/src/Misc/layoutbin/actions.runner.plist.template
+++ b/src/Misc/layoutbin/actions.runner.plist.template
@@ -33,6 +33,7 @@
       <string>LoginWindow</string>
       <string>Background</string>
       <string>StandardIO</string>
+      <string>System</string>
     </array>
   </dict>
 </plist>
--- a/src/Misc/layoutbin/darwin.svc.sh.template
+++ b/src/Misc/layoutbin/darwin.svc.sh.template
@@ -80,14 +80,14 @@ function install()
 function start()
 {
     echo "starting ${SVC_NAME}"
-    launchctl load -w "${PLIST_PATH}" || failed "failed to load ${PLIST_PATH}"
+    launchctl load -w -S Background "${PLIST_PATH}" || failed "failed to load ${PLIST_PATH}"
     status
 }
 
 function stop()
 {
     echo "stopping ${SVC_NAME}"
-    launchctl unload "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"
+    launchctl unload -S Background  "${PLIST_PATH}" || failed "failed to unload ${PLIST_PATH}"
     status
 }
 


tcptps:
#1959 #1056 might be fixed with this as well

PR #2450, opened 2023-02-18
Fix live update as mentioned in #2449

tcptps:
see 👉  #2449 
really go take look at it this just a companion PR not to break things because of the update.

### DIFF
--- a/src/Misc/layoutbin/actions.runner.plist.template
+++ b/src/Misc/layoutbin/actions.runner.plist.template
@@ -27,5 +27,12 @@
     <string>Interactive</string>
     <key>SessionCreate</key>
     <true/>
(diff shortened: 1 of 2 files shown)


tcptps:
I only fixed what I broke, so the script is now back to its previous state in terms of error handling (none zero exit code by launchctl produces failure), except for the one error this whole PR is about (agent is unloaded regardless of the context [Aqua or Background] it is loaded in).
If you are looking for a fix to correctly check if launchctl did not produce any errors, I would reccomend to check if stderr is empty and non zero error code as well.
But this PR is only supposed to prevent problems with live updates caused by and mentioned in #2449 and that is exactly what it now does.

tcptps:
#1959 #1056 might be fixed with this as well

PR #2471, opened 2023-03-03
Add update certificates to ./run.sh if RUNNER_UPDATE_CA_CERTS env is set

nikola-jokic:
When certificates are injected, the image should call update certificates before starting the runner.

If env `RUNNER_UPDATE_CA_CERTS` is set, the `run.sh` will try updating certificates before starting the runner


rentziass (approved):
LGTM, thank you!

images/Dockerfile
```diff
@@ -46,4 +46,8 @@ COPY --chown=runner:docker --from=build /actions-runner .
 
 RUN install -o root -g root -m 755 docker/* /usr/bin/ && rm -rf docker
 
-USER runner
\ No newline at end of file
+USER runner
+
+COPY entrypoint.sh .
+
+ENTRYPOINT [ "./entrypoint.sh" ]
```
TingluoHuang, line 53: I guess we have to create an entry point to reload the cert, how should we change the helm chart?

nikola-jokic: You don't need to provide command now by default :relaxed: I tried running it with env JIT and the JIT flag, works fine
So however you want to pass the JIT, it will work
And when you copy assets for example with init containers, entrypoint is in extracted runner directory and it goes with all other files

TingluoHuang: Do we need to add trap sigint sigterm to the entrypoint?

nikola-jokic: No, because we use exec which will replace the entrypoint with the run.sh that already traps the signal :smile:

images/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# update ca certificates if they are injected with a volume mount
+sudo update-ca-certificates
```
TingluoHuang, line 4: do we need to do this behind some ENV so we don't have to refresh the cert store every time?

nikola-jokic: Not sure, it can delay startup slightly, but I think it won't be noticeable because the operations is pretty fast.
Inside the container, the operation took 0.360s so I guess it should be fine executing it each time.
What I did is:
`docker run --entrypoint bash -it {my image}`
Then executed `time sudo update-ca-certificates`

nikola-jokic: @rentziass @Link- any thoughts? I have no idea to be honest. I think if someone wants to build the image, they would probably check to see this `Dockerfile` and the `entrypoint.sh`, but I have no idea tbh...

TingluoHuang: another way is to run `update-ca-certificates (ubuntu/debian)` or `update-ca-trust(redhat/centos)` in the `run.sh` based on an environment variable passed in from ARC.
similar to `RUNNER_WAIT_FOR_DOCKER_IN_SECONDS` in the `run.sh`

nikola-jokic: Wrote helper function to check for tools if env is exported. Please let me know what do you think about this approach :smile:

images/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+    if [[ ! -z "$(which update-ca-certificates)" ]]; then
+        echo "Running sudo update-ca-certificates"
+        sudo update-ca-certificates
+    elif [[ ! -z "$(which update-ca-trust)" ]]; then
+        echo "Running sudo update-ca-trust"
+        sudo update-ca-trust
+    fi
+}
+
+if [[ ! -z $RUNNER_UPDATE_CA ]]; then
```
Link-, line 14: nit: `$RUNNER_UPDATE_TRUST_STORE` might be more descriptive than `$RUNNER_UPDATE_CA`

images/entrypoint.sh
Link-: @TingluoHuang @nikola-jokic - I definitely like the idea of triggering the update store based on the status of a flag. However, I'm failing to see how hiding that flag in `run.sh` would solve the problem of using other distributions? What if we add a couple of conditions to identify the base image use and decide (in addition to the flag) whether we should run this or not?

images/entrypoint.sh
nikola-jokic: Updated :relaxed: :+1:

images/entrypoint.sh
nikola-jokic: To avoid extensive checks for the base image, I tried just seeing if cli utility is present in the image. That was the easiest way I could think of to avoid different formats for different distributions (I tried to see different formats of `/etc/os-release`, so those checks would be more difficult)

My understanding is that this env is just a signal to the `entrypoint.sh` to trigger ca update before the run.sh is called.
So if you don't have any mounts with certs, it would not be called

On the other hand, I would prefer to call it in every case, since anyone can mount custom certs as config maps for example, and the controller will be completely unaware of it.

The call is not too expensive, it executed in less than 1s when I tried it inside the image. But I don't know if that can become expensive on some nodes...

images/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# update ca certificates in case they are injected with a volume mount
+function updateCerts() {
```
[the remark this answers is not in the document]
nikola-jokic: No I think that is a great point. The only reason I can think of is because the entrypoint is really tied to the docker image where update certs make sense if you mount them as volumes.
The entrypoint can also provide a place where we can include extra behaviour only for containers. For example, a nice addition here might be to export env registering signal handling on `run.sh` by default, since we know we are going to need it. That can eliminate the code on ARC controller keeping track of those envs.

I don't hold a strong opinion. Right now, the `entrypoint.sh` is just a tiny wrapper.

images/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+    if [[ ! -z "$(which update-ca-certificates)" ]]; then
```
TingluoHuang, line 5: `which` might be not installed.

images/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+        echo "Running sudo update-ca-certificates"
+        sudo update-ca-certificates
```
TingluoHuang, line 7: `sudo` might not be installed.

images/entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+    elif [[ ! -z "$(which update-ca-trust)" ]]; then
+        echo "Running sudo update-ca-trust"
+        sudo update-ca-trust
+    fi
+}
+
+if [[ ! -z $RUNNER_UPDATE_TRUST_STORE ]]; then
```
TingluoHuang, line 14: `RUNNER_UPDATE_CA_CERTS`?

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
...
 }
 
+# update ca certificates in case they are injected with a volume mount
+function updateCerts() {
+    local sudo_prefix=""
+
+    local user_id=`id -u`
+    if ! [ $user_id -eq 0 ]; then
+        if [[ ! -x "$(command -v sudo)" ]]; then
+            echo "Warning: failed to update certificate store: sudo not found"
```
Link-, line 63:
```suggestion
            echo "Warning: failed to update certificate store, sudo is required but not found"
```
nit

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
...
+            sudo_prefix="sudo"
+        fi
+    fi
+
+    if [[ -x "$(command -v update-ca-certificates)" ]]; then
+        eval $sudo_prefix "update-ca-certificates"
+    elif [[ -x "$(command -v update-ca-trust)" ]]; then
+        eval $sudo_prefix "update-ca-trust"
+    else
+        echo "Warning: failed to update certificate store: update-ca-certificates or update-ca-trust not found"
```
Link-, line 75:
```suggestion
        echo "Warning: failed to update certificate store, update-ca-certificates or update-ca-trust not found. This can happen if you're using a different runner base image."
```
nit

src/Misc/layoutroot/run.sh
rentziass: ultranit: `{level}: {issue}: {reason}` feels more Unix-like?

Link- (approved):
LGTM 🚀

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
...
+        eval $sudo_prefix "update-ca-certificates"
+    elif [[ -x "$(command -v update-ca-trust)" ]]; then
+        eval $sudo_prefix "update-ca-trust"
+    else
+        echo "Warning: failed to update certificate store: update-ca-certificates or update-ca-trust not found. This can happen if you're using a different runner base image."
+        return 1
+    fi
+}
+
+if [[ ! -z $RUNNER_UPDATE_CA_CERTS ]]; then
```
TingluoHuang, line 80:
```suggestion
if [[ ! -z "$RUNNER_UPDATE_CA_CERTS" ]]; then
```

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
     done
 }
 
+# update ca certificates in case they are injected with a volume mount
+function updateCerts() {
+    local sudo_prefix=""
+    local user_id=`id -u`
+
+    if ! [ $user_id -eq 0 ]; then
```
TingluoHuang, line 61:
```suggestion
    if [ $user_id -ne 0 ]; then
```

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
...
+        if [[ ! -x "$(command -v sudo)" ]]; then
+            echo "Warning: failed to update certificate store: sudo is required but not found"
+            return 1
```
TingluoHuang, line 64: `return 1` or just `return`?

nikola-jokic: I felt like the function should report an error with an exit code even if we ignore it
It should signal the failure anyway, and it is up to the caller to ignore it or use it to terminate the execution

src/Misc/layoutroot/run.sh
```diff
@@ -53,6 +53,34 @@ runWithManualTrap() {
     done
 }
 
+# update ca certificates in case they are injected with a volume mount
```
TingluoHuang, line 56: update the comment to be just 
```
update ca certificates via `update-ca-certificates` or `update-ca-trust`
```
?

nikola-jokic: Maybe I should delete the comment?
I don't want to tie it up with implementation in case we end up using another tool if needed, and the name should be enough... Does not have to be a volume mount
What do you think?

TingluoHuang: sure, removing also seems fine to me. 😄

TingluoHuang (approved):

PR #2474, opened 2023-03-06
Allow ExecutionContext in hashfiles

fhammerl:


src/Runner.Worker/ActionRunner.cs
```diff
@@ -181,7 +181,7 @@ public async Task RunAsync()
             else
             {
                 var templateEvaluator = ExecutionContext.ToPipelineTemplateEvaluator();
-                inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, ExecutionContext.ExpressionValues, ExecutionContext.ExpressionFunctions);
```
fhammerl, line 184: The context of HashFilesFunction is decided based on the inputs here

fhammerl: Passing down ExecutionContext (like we do for StepIf, like `success()` makes it available in the function later)

src/Runner.Worker/ActionRunner.cs
```diff
@@ -355,7 +355,7 @@ public bool EvaluateDisplayName(DictionaryContextData contextData, IExecutionCon
         {
             DictionaryContextData expressionValues = ExecutionContext.GetExpressionValues(stepHost);
             var templateEvaluator = ExecutionContext.ToPipelineTemplateEvaluator();
-            var inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, expressionValues, ExecutionContext.ExpressionFunctions);
+            var inputs = templateEvaluator.EvaluateStepInputs(Action.Inputs, expressionValues, ExecutionContext.ExpressionFunctions, ExecutionContext.ToExpressionState());
```
fhammerl, line 358: Passing down ExecutionContext (like we do for StepIf, like `success()` makes it available in the function later)

src/Runner.Worker/Expressions/HashFilesFunction.cs
```diff
@@ -29,6 +29,9 @@ public sealed class HashFilesFunction : Function
             githubContext.TryGetValue(PipelineTemplateConstants.Workspace, out var workspace);
             var workspaceData = workspace as StringContextData;
             ArgUtil.NotNull(workspaceData, nameof(workspaceData));
+            var executionContext = templateContext.State[nameof(IExecutionContext)] as IExecutionContext;
```
fhammerl, line 32: Is there now, because `CreateContext` has expressionState

PR #2488, opened 2023-03-13
WIP Composite Action Incidents

vmjoseph:
Fix for https://github.com/github/c2c-actions-runtime/issues/2182


src/Runner.Worker/ActionManager.cs
```diff
@@ -682,12 +682,16 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
                         // * Repo or tag doesn't exist, or isn't public
                         // * Policy validation failed
                         if (ex is WebApi.UnresolvableActionDownloadInfoException)
-                        {
+                        {   // Log the error and fail the JobExtension Initialization.
+                            Trace.Error($"Caught exception from JobExtenion Initialization: {ex}");
+                            executionContext.InfrastructureError(ex.Message);
```
thboop, line 687: I'm not convinced this should be considered infrastructure error. If the repo doesn't or they try to use a repo they don't have access to, should that be user error instead?

src/Runner.Worker/ActionManager.cs
```diff
@@ -682,12 +682,16 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
...
                             throw;
                         }
                         else
                         {
                             // This exception will be traced as an infrastructure failure
+                            Trace.Error($"Caught exception from JobExtenion Initialization: {ex}");
```
thboop, line 693: We also catch this and throw [here](https://github.com/actions/runner/blob/9958fc0374820841ec0c131fa1259a442593fb1b/src/Runner.Worker/JobExtension.cs#L434). This will result in us double printing in that scenario right?



We could
- Delete the lines in that file and handle it centrally here
- Add this trace.error and infrastructure inside a try catch [here](https://github.com/actions/runner/blob/9958fc0374820841ec0c131fa1259a442593fb1b/src/Runner.Worker/ActionRunner.cs#L92)

src/Runner.Worker/ActionManager.cs
```diff
@@ -100,43 +100,56 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
...
+                    {
+                        executionContext.Output("Container action is only supported on Linux, skip pull and build docker images.");
+                        containerSetupSteps.Clear();
+                    }
 #endif
+                }
+                return new PrepareResult(containerSetupSteps, result.PreStepTracker);
+
+            }
+            catch (Exception)
```
thboop, line 146: Should we remove this try catch? It isn't doing anything besides catching and rethrowing right?

src/Runner.Worker/ActionManager.cs
```diff
@@ -174,7 +187,26 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
             if (repositoryActions.Count > 0)
             {
                 // Get the download info
-                var downloadInfos = await GetDownloadInfoAsync(executionContext, repositoryActions);
+                var downloadInfos = new Dictionary<string, WebApi.ActionDownloadInfo>
+                {
+                    ["patterns"] = new WebApi.ActionDownloadInfo()
+                };
+
+                try
```
thboop, line 195: Same feedback as above

src/Runner.Worker/ActionManager.cs
jww3: +1

thboop: I think we should use the `try...catch` patterns in scenarios where we need to do something with the exception.

Examples of that could be
- Logging
- Bubbling up this as an infra error (most relevant in this case)
- Checking if the exception is of a certain type, and if so ignoring it, otherwise re-throwing it

I'm not sure what value a catch block with `throw` as the only command buys it, I think

```
throw new Exception("test")
```

and 

``` 
try {
  throw new Exception("test")
} catch {
throw;
} 
```

functionally do the same exact thing, but one is less lines of code (and maybe more readable?)

src/Runner.Worker/ActionRunner.cs
```diff
@@ -89,7 +91,18 @@ public async Task RunAsync()
...
+                {
+                    prepareResult = await actionManager.PrepareActionsAsync(ExecutionContext, compositeHandlerData.Steps, ExecutionContext.Id);
+                }
+                catch (Exception ex)
+                {
+                    if (ex is WebApi.FailedToResolveActionDownloadInfoException)
+                    {
+                        throw new WebApi.FailedToResolveActionDownloadInfoException("Failed to resolve action download info", ex);
+                    }
+                }
```
jww3, line 105:
```suggestion
                catch (WebApi.FailedToResolveActionDownloadInfoException ex)
                {
                    throw new WebApi.FailedToResolveActionDownloadInfoException("Failed to resolve action download info", ex);
                }
```

Otherwise, we're silently swallowing all other types of exception.

thboop: +1.

I think we also want to include in this block the areas that log this as an infrastructure error from [here](https://github.com/actions/runner/blob/9958fc0374820841ec0c131fa1259a442593fb1b/src/Runner.Worker/JobExtension.cs#LL436-L440C27)

src/Runner.Worker/ActionManager.cs
```diff
@@ -683,7 +715,8 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
                         // * Policy validation failed
                         if (ex is WebApi.UnresolvableActionDownloadInfoException)
                         {
-                            throw;
+                            throw new WebApi.UnresolvableActionDownloadInfoException("Unable to resolve action download info", ex);
```
jww3, line 718: What's the benefit of wrapping the exception with different messaging?  Does it make the problem easier to spot in telemetry?  or logs?

src/Runner.Worker/ActionManager.cs
```diff
@@ -174,7 +187,26 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
             if (repositoryActions.Count > 0)
             {
                 // Get the download info
-                var downloadInfos = await GetDownloadInfoAsync(executionContext, repositoryActions);
+                var downloadInfos = new Dictionary<string, WebApi.ActionDownloadInfo>
+                {
+                    ["patterns"] = new WebApi.ActionDownloadInfo()
```
jww3, line 192: Why inject this seed data only to clear it on line 197?

src/Runner.Worker/ActionManager.cs
```diff
@@ -174,7 +187,26 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
...
+                };
+
+                try
+                {
+                    downloadInfos.Clear();
+                    if (downloadInfos.Count == 0)
```
jww3, line 198: This condition will always be true after a call to `Clear()`

src/Runner.Worker/ActionManager.cs
```diff
@@ -137,6 +152,8 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
 #endif
             }
             return new PrepareResult(containerSetupSteps, result.PreStepTracker);
+
+
```
TingluoHuang, line 156:
```suggestion
```

src/Runner.Worker/ActionRunner.cs
```diff
@@ -89,7 +90,16 @@ public async Task RunAsync()
...
-                var prepareResult = await actionManager.PrepareActionsAsync(ExecutionContext, compositeHandlerData.Steps, ExecutionContext.Id);
+                PrepareResult prepareResult = default(PrepareResult);
+                try
+                {
+                    prepareResult = await actionManager.PrepareActionsAsync(ExecutionContext, compositeHandlerData.Steps, ExecutionContext.Id);
+                }
+                catch (WebApi.FailedToResolveActionDownloadInfoException ex)
+                {
+                    throw new WebApi.FailedToResolveActionDownloadInfoException("Failed to resolve action download info", ex);
+
```
TingluoHuang, line 101:
```suggestion
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -100,7 +101,21 @@ public async Task<PrepareResult> PrepareActionsAsync(IExecutionContext execution
             }
             IEnumerable<Pipelines.ActionStep> actions = steps.OfType<Pipelines.ActionStep>();
             executionContext.Output("Prepare all required actions");
-            var result = await PrepareActionsRecursiveAsync(executionContext, state, actions, depth, rootStepId);
+            PrepareActionsState result = new PrepareActionsState();
+            try
+            {
+                result = await PrepareActionsRecursiveAsync(executionContext, state, actions, depth, rootStepId);
+
```
TingluoHuang, line 108:
```suggestion
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -173,8 +190,8 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
 
             if (repositoryActions.Count > 0)
             {
-                // Get the download info
-                var downloadInfos = await GetDownloadInfoAsync(executionContext, repositoryActions);
+
+
```
TingluoHuang, line 194:
```suggestion
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -184,6 +201,8 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
                     {
                         continue;
                     }
+                    // Get the download info
+                    var downloadInfos = await GetDownloadInfoAsync(executionContext, repositoryActions);
```
TingluoHuang, line 205: why did we move this down into the `foreach` loop? this would mean we will call the actions service multiple times to get the same response.

src/Runner.Worker/ActionManager.cs
TingluoHuang: +1, why do we need to wrap with the same exception type?

src/Runner.Worker/ActionManager.cs
```diff
@@ -683,7 +702,8 @@ private async Task BuildActionContainerAsync(IExecutionContext executionContext,
                         // * Policy validation failed
                         if (ex is WebApi.UnresolvableActionDownloadInfoException)
                         {
-                            throw;
+                            throw new WebApi.UnresolvableActionDownloadInfoException("Unable to resolve action download info", ex);
+                            // throw;
```
TingluoHuang, line 706:
```suggestion
```

src/Runner.Worker/ActionRunner.cs
TingluoHuang: Why do we need to catch and rethrow `FailedToResolveActionDownloadInfoException` with a different error message?

src/Runner.Worker/ActionManager.cs
vmjoseph: That makes sense. I'd originally moved this when I was considering a different approach to how we handled the actions service call. Reverted!

src/Runner.Worker/ActionManager.cs
vmjoseph: My original intent was to make logs easier to read, but after reading up more on how c# handles exceptions, it looks like it's a waste of resources to wrap the exception. Removed!

src/Runner.Worker/ActionRunner.cs
```diff
@@ -89,7 +89,15 @@ public async Task RunAsync()
...
                 var actionManager = HostContext.GetService<IActionManager>();
-                var prepareResult = await actionManager.PrepareActionsAsync(ExecutionContext, compositeHandlerData.Steps, ExecutionContext.Id);
+                PrepareResult prepareResult = default(PrepareResult);
+                try
+                {
+                    prepareResult = await actionManager.PrepareActionsAsync(ExecutionContext, compositeHandlerData.Steps, ExecutionContext.Id);
+                }
+                catch (WebApi.FailedToResolveActionDownloadInfoException ex)
+                {
+                    throw new WebApi.FailedToResolveActionDownloadInfoException("Failed to resolve action download info", ex);
```
TingluoHuang, line 99: why do we need to catch the exception and rethrow the same exception with a different message?

src/Runner.Worker/ActionManager.cs
```diff
@@ -173,6 +187,7 @@ private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionC
 
             if (repositoryActions.Count > 0)
             {
+
```
TingluoHuang, line 190:
```suggestion
```

src/Runner.Worker/ActionRunner.cs
```diff
@@ -10,8 +10,8 @@
 using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
 using GitHub.Runner.Worker.Handlers;
+using WebApi = GitHub.DistributedTask.WebApi;
 using Pipelines = GitHub.DistributedTask.Pipelines;
-
```
TingluoHuang, line 14: add this line back.

TingluoHuang (approved):

zezo2019 (approved):

PR #2489, opened 2023-03-16
Audit `curl` command line switches

jww3:


src/Misc/externals.sh
```diff
@@ -60,6 +60,7 @@ function acquireExternalTool() {
             #      -S Show error. With -s, make curl show errors when they occur
             #      -L Follow redirects (H)
             #      -o FILE    Write to FILE instead of stdout
+            # REVIEW
             curl -fkSL -o "$partial_target" "$download_source" 2>"${download_target}_download.log" || checkRC 'curl'
```
jww3, line 64: I'm not convinced that these are the right options to pass to `curl`

Specifically, I think `-f` overrides `-S`  

For example, I saw a [CI failure today](https://github.com/actions/runner/actions/runs/4435962942/jobs/7783789866#step:3:37) related to this `curl` call and I wasn't able to get any real insight from the logs.

We might also want to consider leveraging curl's `--retry` option(s) here.

jww3: Retry has been added, but there's still the question about `-f` and -S`

PR #2494, opened 2023-03-17
ADR: Runner Image Tags

Link-:
### Context

[Rendered](https://github.com/actions/runner/blob/535d9da5149db056771a49e0734adf4e4bbd301e/docs/adrs/2494-runner-image-tags.md)


docs/adrs/2494-runner-image-tags.md
```diff
@@ -0,0 +1,62 @@
...
+    containers:
+    - name: runner
+      image: "ghcr.io/actions/actions-runner"
+      tag: "v2.300.0"
+      command: ["/home/runner/run.sh"]
+```
+
+### Option 2: Keep the `latest` tag
+
+Keeping the `latest` tag is also a reasonable option especially if we don't expect to make any breaking changes to the runner image. We could enhance this by adding a [NOTES.txt](https://helm.sh/docs/chart_template_guide/notes_files/) to the helm chart which will be displayed to the user after a successful helm install/upgrade. This will help users understand the implications of using the `latest` tag and how to pin a specific version of the runner image.
```
TingluoHuang, line 49: if we keep the `latest` tag, we should also change the release workflow to delay when we tag the image as `latest` from release creation time to finish rollout to service.

Link-: Great point @TingluoHuang - I'll add it

nikola-jokic (approved):

rentziass (approved):

PR #2497, opened 2023-03-20
Fix JIT configurations on Windows

MatisseHack:
JIT configurations don't work on Windows because all configuration files get written as plain text. On Windows, the runner then tries to decrypt the `.credentials_rsaparams` file using `ProtectedData.Unprotect`, which fails if the file contents were never encrypted.

I've tested this patch on Windows and Linux and everything seems to work as expected.

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -220,7 +220,7 @@ namespace GitHub.Runner.Listener
 #endif
                             File.WriteAllBytes(configFile, configContent);
                             File.SetAttributes(configFile, File.GetAttributes(configFile) | FileAttributes.Hidden);
-                            Trace.Info($"Save {configContent.Length} chars to '{configFile}'.");
+                            Trace.Info($"Saved {configContent.Length} bytes to '{configFile}'.");
                         }
                     }
                     catch (Exception ex)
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -210,8 +211,14 @@ public async Task<int> ExecuteCommand(CommandSettings command)
...
+                            var configContent = Convert.FromBase64String(config.Value);
+#if OS_WINDOWS
+                            if (configFile == HostContext.GetConfigFile(WellKnownConfigFile.RSACredentials))
+                            {
+                                configContent = ProtectedData.Protect(configContent, null, DataProtectionScope.LocalMachine);
+                            }
+#endif
+                            File.WriteAllBytes(configFile, configContent);
                             File.SetAttributes(configFile, File.GetAttributes(configFile) | FileAttributes.Hidden);
                             Trace.Info($"Save {configContent.Length} chars to '{configFile}'.");
```
TingluoHuang, line 223: `configContent.Length` might be off since configContent become bytes now.

MatisseHack: Good point! Is it enough to update the log message?

```suggestion
                            Trace.Info($"Saved {configContent.Length} bytes to '{configFile}'.");
```

TingluoHuang: looks good. 👍

TingluoHuang (approved):

PR #2500, opened 2023-03-22
Listen directly to Broker for messages

luketomlinson:
This PR adds 3 things:

1) BrokerHttpClient (just the client to call the `/message` endpoint)
2) BrokerServer (following the *Server pattern. This maintains a connection with the server and uses the http client to make the requests)
3) BrokerMessageListener (wraps the BrokerServer and continuously polls for messages)

We will set the `ServerUrlV2` in a followup PR.

### DIFF
--- a/src/Runner.Listener/BrokerMessageListener.cs
+++ b/src/Runner.Listener/BrokerMessageListener.cs
@@ -1,0 +1,217 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Common;
+using GitHub.Runner.Listener.Configuration;
+using GitHub.Runner.Sdk;
+using GitHub.Services.Common;
+using GitHub.Services.OAuth;
+
+namespace GitHub.Runner.Listener
+{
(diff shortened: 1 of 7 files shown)


src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -0,0 +1,199 @@
...
+                        //retry after a random backoff to avoid service throttling
+                        //in case of there is a service error happened and all agents get kicked off of the long poll and all agent try to reconnect back at the same time.
+                        if (continuousError <= 5)
+                        {
+                            // random backoff [15, 30]
+                            _getNextMessageRetryInterval = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(30), _getNextMessageRetryInterval);
+                        }
+                        else
+                        {
+                            // more aggressive backoff [30, 60]
```
lokesh755, line 124: Basically we'll retry forever in case of retriable exceptions? 
Nit: Regarding comment, It's less aggressive right? since we're increasing the retry intervals.

luketomlinson: Yep I think so. This is copied behavior from the existing Message listener

luketomlinson: I guess it could be read as "more aggressively backing off by increasing the time" 😄

luketomlinson: I could have some maximum of `continuousError` where I abort if it reaches that

lokesh755 (approved):

PR #2505, opened 2023-03-24
Register Runners against V2 servers

luketomlinson:
This PR registers runners against the new endpoint. The biggest changes are:

- Sending the runner's public key, encoded as an XML string up to the registration call

- Parsing the authorization object from the runner registration response and saves it for refreshing tokens

- Saves the `ServerUrlV2` in the RunnerSettings to connect to the new server for messages

### DIFF
--- a/src/Sdk/DTWebApi/WebApi/Runner.cs
+++ b/src/Sdk/DTWebApi/WebApi/Runner.cs
@@ -8,6 +8,9 @@ namespace GitHub.DistributedTask.WebApi
 
         public class Authorization
         {
+            /// <summary>
+            /// The url to refresh tokens
+            /// </summary> 
             [JsonProperty("authorization_url")]
             public Uri AuthorizationUrl
             {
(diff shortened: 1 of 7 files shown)


src/Sdk/DTWebApi/WebApi/Runner.cs
```diff
@@ -1,18 +1,73 @@
...
+            get;
+            internal set;
+        }
+
+        [JsonProperty("authorization")]
+        public Authorization RunnerAuthorization
+        {
+            get;
+            internal set;
+        }
```
jww3, line 61: Do we need to mark these new fields as optional in order to support all rollback and rollforward scenarios?

src/Sdk/DTWebApi/WebApi/Runner.cs
```diff
@@ -1,18 +1,73 @@
...
+
+        public void ApplyToTaskAgent(TaskAgent agent)
```
jww3, line 63: This feels like it might be blurring encapsulation lines.  Maybe it's better as a static method in a static class?  Or maybe it's better as  a method on `TaskAgent`?

src/Sdk/DTWebApi/WebApi/Runner.cs
luketomlinson: They should be optional by default: Screenshot 2023-03-27 at 11 22 49 AM

src/Sdk/DTWebApi/WebApi/Runner.cs
luketomlinson: great point. Updated.

luketomlinson: I opted to just pull this out of the method and make it explicit at the call site rather than burying it

jww3 (approved):

PR #2510, opened 2023-03-28
Update Runner to send step updates to Results

yacaovsnc:
* Refactored `jobServer` so `Results` code are in `ResultsServer` now.
* Updated `ResultsServer` to send steps update to Results
* Results upload is best effort and any exception should skip subsequent uploads
* Add authenticated proxy support
* Buffer request for `WinHttpHandler`

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -931,8 +931,6 @@ namespace GitHub.Runner.Common
         public long TotalLines { get; set; }
     }
 
-
-
     internal class ConsoleLineInfo
     {
         public ConsoleLineInfo(Guid recordId, string line, long? lineNumber)
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -514,22 +514,14 @@ namespace GitHub.Runner.Common
                         }
                         catch (Exception ex)
                         {
-                            var issue = new Issue() { Type = IssueType.Warning, Message = $"Caught exception during file upload to results. {ex.Message}" };
-                            issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.ResultsUploadFailure;
-
-                            var telemetryRecord = new TimelineRecord()
-                            {
-                                Id = Constants.Runner.TelemetryRecordId,
-                            };
-                            telemetryRecord.Issues.Add(issue);
-                            QueueTimelineRecordUpdate(_jobTimelineId, telemetryRecord);
-
                             Trace.Info("Catch exception during file upload to results, keep going since the process is best effort.");
                             Trace.Error(ex);
                             errorCount++;
 
                             // If we hit any exceptions uploading to Results, let's skip any additional uploads to Results
                             _resultsClientInitiated = false;
+
+                            SendResultsTelemetry(ex);
                         }
                     }
 
@@ -545,6 +537,19 @@ namespace GitHub.Runner.Common
                     await Task.Delay(_delayForResultsUploadDequeue);
                 }
             }
+        }
+
+        private void SendResultsTelemetry(Exception ex)
+        {
+            var issue = new Issue() { Type = IssueType.Warning, Message = $"Caught exception with results. {ex.Message}" };
+            issue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.ResultsUploadFailure;
+
+            var telemetryRecord = new TimelineRecord()
+            {
+                Id = Constants.Runner.TelemetryRecordId,
+            };
+            telemetryRecord.Issues.Add(issue);
+            QueueTimelineRecordUpdate(_jobTimelineId, telemetryRecord);
         }
 
         private async Task ProcessTimelinesUpdateQueueAsync(bool runOnce = false)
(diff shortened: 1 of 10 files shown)


src/Runner.Worker/JobRunner.cs
```diff
@@ -49,6 +49,8 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
                 await runServer.ConnectAsync(systemConnection.Url, jobServerCredential);
                 server = runServer;
+
+                _jobServerQueue.Start(message);
```
brittanyellich, line 53: Just verifying that this change should be here as it doesn't look related, but maybe it is?

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -126,6 +127,46 @@ public class CreateMetadataResponse
...
+        public string Name;
+        [DataMember]
+        public Status Status;
+        [DataMember]
+        public string StartedAt;
+        [DataMember]
+        public string CompletedAt;
+    }
+
+    public enum Status
```
brittanyellich, line 162: It may be helpful to link to somewhere to indicate why were are choosing these particular status numbers? Otherwise the context on the strange ordering of the numbers may get lost. Unless there is already an explanation in results.

brittanyellich (approved):

src/Runner.Worker/JobRunner.cs
yacaovsnc: Ah, good eye!  I forgot to remove it.  This will be in next PR where we may skip uploading with `JobServer` but need to start `ResultsServer`.

src/Sdk/WebApi/WebApi/Contracts.cs
yacaovsnc: We have a list of `Status` and the numbers are consecutive, but only those are relevant to steps.  🤷🏼

src/Sdk/Common/Common/VssHttpMessageHandler.cs
TingluoHuang, line 1: undo the change if nothing changed. 😄

src/Sdk/Common/Common/VssHttpRequestSettings.cs
TingluoHuang, line 1: same, undo if not needed.

src/Sdk/Common/Common/VssHttpMessageHandler.cs
yacaovsnc: Looks like a linter change. I will undo.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -111,8 +113,11 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest)
                 !string.IsNullOrEmpty(resultsReceiverEndpoint))
             {
                 Trace.Info("Initializing results client");
-                _jobServer.InitializeResultsClient(new Uri(resultsReceiverEndpoint), accessToken);
-                _resultsClientInitiated = true;
+                if (_resultsServer != null)
```
TingluoHuang, line 116: `_resultsServer ` should never be NULL since it's created by `public override void Initialize(IHostContext hostContext)`

src/Runner.Common/JobServerQueue.cs
```diff
@@ -612,6 +620,21 @@ private async Task ProcessTimelinesUpdateQueueAsync(bool runOnce = false)
...
                             await _jobServer.UpdateTimelineRecordsAsync(_scopeIdentifier, _hubName, _planId, update.TimelineId, update.PendingRecords, default(CancellationToken));
+                            try
+                            {
+                                if (_resultsClientInitiated)
+                                {
+                                    await _resultsServer.UpdateResultsWorkflowStepsAsync(_scopeIdentifier, _hubName, _planId, update.TimelineId, update.PendingRecords, default(CancellationToken));
+                                }
+                            }
+                            catch (Exception e)
+                            {
```
TingluoHuang, line 631: do we want to report the error to the actions service as telemetry? like how we did for log upload?

src/Runner.Common/JobServerQueue.cs
yacaovsnc: Removed null guard.

src/Runner.Common/ResultsServer.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Services.Results.Client;
+
+namespace GitHub.Runner.Common
+{
+    [ServiceLocator(Default = typeof(ResultServer))]
+    public interface IResultsServer : IRunnerService, IAsyncDisposable
```
TingluoHuang, line 12: do you need `IAsyncDisposable`?

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -126,6 +127,46 @@ public class CreateMetadataResponse
...
+    {
+        StatusUnknown = 0,
+        StatusInProgress = 3,
+        StatusPending = 5,
+        StatusCompleted = 6
```
TingluoHuang, line 167: why not 0, 1, 2, 3?

yacaovsnc: Those numbers are defined elsewhere.  I removed the ones that are not applicable to runner.

src/Runner.Common/ResultsServer.cs
yacaovsnc: I was going to move the web socket client to here in the future so I left this.  I will remove this for now since it's not needed.

TingluoHuang (approved):

PR #2516, opened 2023-04-03
Bypass job server when run service is enabled

yacaovsnc:
* Add websocket client in `ResultsServer`
* Adding boolean to enable upload to Actions service in `JobServer`, this is only used for Run svc messages.


src/Runner.Worker/JobRunner.cs
```diff
@@ -49,6 +49,8 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
                 await runServer.ConnectAsync(systemConnection.Url, jobServerCredential);
                 server = runServer;
+
+                _jobServerQueue.Start(message, enableActionsUpload: false);
```
jherns, line 53: `_jobServerQueue`  isn't initialized here because it is only initialized in the else block

src/Runner.Common/WebsocketFeedServer.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Threading.Tasks;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Sdk;
+using GitHub.Services.Common;
+using GitHub.Services.WebApi.Utilities.Internal;
+
+namespace GitHub.Runner.Common
+{
+    [ServiceLocator(Default = typeof(WebsocketFeedServer))]
+    public interface IWebsocketFeedServer : IRunnerService, IAsyncDisposable
```
TingluoHuang, line 17: is this just copy from what's deleted from `IJobServer.cs`?

src/Runner.Worker/JobRunner.cs
```diff
@@ -49,6 +49,9 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 VssCredentials jobServerCredential = VssUtil.GetVssCredential(systemConnection);
                 await runServer.ConnectAsync(systemConnection.Url, jobServerCredential);
                 server = runServer;
+
+                _jobServerQueue = HostContext.GetService<IJobServerQueue>();
+                _jobServerQueue.Start(message, enableActionsUpload: false);
```
TingluoHuang, line 54: the name `enableActionsUpload` is kind of confusing. :D 
my first read is something about uploading an Action as an artifact or something.

src/Runner.Common/WebsocketFeedServer.cs
yacaovsnc: Mostly yes with a slight twist - the fallback with http long poll only works with Actions Service, so I had to refactor that portion a little bit.

src/Runner.Worker/JobRunner.cs
yacaovsnc: 😄  Do you have a suggestion?  The intention is to skip any http interactions with Actions service with this bool flag.

jherns: Does `enableTimelineRecordsUpload` make sense?

yacaovsnc: It's not only timeline records upload though, we also create attachments and upload log content.  

Maybe `enableInteractionsWithActions`? 🤷🏼

src/Runner.Common/WebsocketFeedServer.cs
TingluoHuang: do we know whether we still need the fallback for the console log?

src/Runner.Worker/JobRunner.cs
TingluoHuang: `resultServiceOnly = true`? and have it default to `false` so you don't have to change the other caller. 😄

src/Runner.Common/WebsocketFeedServer.cs
TingluoHuang: it's a little bit weird that we have `IWebsocketFeedServer`, `IJobServer`, and `IResultServer`, but `IWebsocketFeedServer` somehow depends on `IJobServer`. 😕

TingluoHuang: is it possible to leave the current code in the `IJobServer`, and create a copy of it without the fallback logic control which one to initialize/use inside the `JobServerQueue` base on the new parameter you added to `JobServerQueue.Start()`?
This would be safer for deployment.

TingluoHuang: It should not be that bad, in jobServerQueue, it's basically if/else between JobServer vs. ResultServer.

src/Runner.Worker/JobRunner.cs
yacaovsnc: Renamed the variable.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -599,7 +619,7 @@ private async Task ProcessTimelinesUpdateQueueAsync(bool runOnce = false)
 
                         foreach (var detailTimeline in update.PendingRecords.Where(r => r.Details != null))
                         {
-                            if (!_allTimelines.Contains(detailTimeline.Details.Id))
+                            if (!_allTimelines.Contains(detailTimeline.Details.Id) && !_resultsServiceOnly)
```
TingluoHuang, line 622:
```suggestion
                            if (!_resultsServiceOnly && !_allTimelines.Contains(detailTimeline.Details.Id))
```

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +118,134 @@ public void InitializeResultsClient(Uri uri, string token)
...
+            }
+        }
+
+        private async Task ConnectWebSocketClient(string feedStreamUrl, TimeSpan delay)
+        {
+            try
+            {
+                Trace.Info($"Attempting to start websocket client with delay {delay}.");
+                await Task.Delay(delay);
+                await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), default(CancellationToken));
```
TingluoHuang, line 174: check the other code in job server, I recently made a change to give the connectasync a 10 or 30 seconds timeout.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -194,6 +204,9 @@ public async Task ShutdownAsync()
             Trace.Info($"Disposing job server ...");
             await _jobServer.DisposeAsync();
 
+            Trace.Info($"Disposing Results server ...");
```
TingluoHuang, line 207:
```suggestion
            Trace.Info($"Disposing results server ...");
```

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +118,134 @@ public void InitializeResultsClient(Uri uri, string token)
 
             throw new InvalidOperationException("Results client is not initialized.");
         }
+
+        public void InitializeWebsocketClient(ServiceEndpoint serviceEndpoint)
```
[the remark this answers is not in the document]
yacaovsnc: We need to, since Results is already in production and we may switch to use Results for live console logs even with the legacy `ServiceEndpoint`.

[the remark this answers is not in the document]
TingluoHuang: what do you need for `ServiceEndpoint`? I assume you just need a URL and a token, `ServiceEndpoint` is an over kill for those 2 things. 😄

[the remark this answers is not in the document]
yacaovsnc: Alright I moved the parsing out from `ResultsServer` to `JobServerQueue` to consolidate the client initializations.  I hope in the future we can define a new contract between the server and runner, and then all the parsing and initialization could happen in `ResultsServer` so we don't pollute `JobServerQueue`.

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +122,128 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                        var lastChunk = i + (1 * 1024) >= jsonDataBytes.Length;
+                        var chunk = new ArraySegment<byte>(jsonDataBytes, i, Math.Min(1 * 1024, jsonDataBytes.Length - i));
+                        await _websocketClient.SendAsync(chunk, WebSocketMessageType.Text, endOfMessage: lastChunk, cancellationToken);
+                    }
+
+                }
+                catch (Exception ex)
+                {
+                    _failedAttemptsToPostBatchedLinesByWebsocket++;
+                    Trace.Info($"Caught exception during append web console line to websocket, let's fallback to sending via non-websocket call (total calls: {_totalBatchedLinesAttemptedByWebsocket}, failed calls: {_failedAttemptsToPostBatchedLinesByWebsocket}, websocket state: {this._websocketClient?.State}).");
```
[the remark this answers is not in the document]
yacaovsnc: That's a really good question...  hmmm,  I feel we could retry connection for couple more times, but if this still doesn't work, there won't be any fallback.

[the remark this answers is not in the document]
TingluoHuang: can we check telemetry and see if the current failback to REST for the jobserver ever brings any value? ex: how many jobs actually need to fallback to REST API, maybe the answer is we don't need any fallback if the websocket is stable.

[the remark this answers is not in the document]
yacaovsnc: We aren't planning for any fallback in Results at the moment.  I do see quite a few calls to this REST API today, but I also suspect that's because we don't retry deliver console log via websocket connection, and fallback with the first error.  I am going to add retries here first.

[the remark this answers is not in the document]
yacaovsnc: > What should be the correct behavior for websocket exceptions when using the result service?

Also this codepath is exercised only with the new server logic.  We would still use `jobServer` with current server logic, even if live log through Results (the `feedStreamUrl` points to new results).

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                    userAgentValues.AddRange(UserAgentUtility.GetDefaultRestUserAgent());
+                    userAgentValues.AddRange(HostContext.UserAgents);
+                    this._websocketClient.Options.SetRequestHeader("User-Agent", string.Join(" ", userAgentValues.Select(x => x.ToString())));
+
+                    // during initialization, retry upto 3 times to setup connection
+                    this._websocketConnectTask = ConnectWebSocketClient(liveConsoleFeedUrl, delay, retryConnection: true);
+                }
+                else
+                {
+                    Trace.Info($"No FeedStreamUrl found, so we will use Rest API calls for sending feed data");
```
TingluoHuang, line 151: feels like these traces are out of date. 😄

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+            GC.SuppressFinalize(this);
+
+            return ValueTask.CompletedTask;
+        }
+
+        private void InitializeWebsocketClient(string liveConsoleFeedUrl, string accessToken, TimeSpan delay)
+        {
+            if (!string.IsNullOrEmpty(accessToken))
+            {
+                if (!string.IsNullOrEmpty(liveConsoleFeedUrl))
```
TingluoHuang, line 134: maybe just check both `if (!string.IsNullOrEmpty(accessToken) && !string.IsNullOrEmpty(liveConsoleFeedUrl))` since we should never land in the method if we don't have that info?

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                {
+                    // let's ensure we use the right scheme
+                    liveConsoleFeedUrl = liveConsoleFeedUrl.Replace("https://", "wss://").Replace("http://", "ws://");
```
TingluoHuang, line 137: do we still need to do this URL magic for URL from result service?
we were doing it due to the limitation of legacy LocationService.

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+            int retries = 0;
+
+            do
+            {
+                try
+                {
+                    Trace.Info($"Attempting to start websocket client with delay {delay}.");
+                    await Task.Delay(delay);
+                    using var connectTimeoutTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(30));
+                    await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), connectTimeoutTokenSource.Token);
```
TingluoHuang, line 172:
```suggestion
                    await this._websocketClient.ConnectAsync(new Uri(feedStreamUrl), connectTimeoutTokenSource.Token);
                    connected = true;
```

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                    while (!delivered && retries < 3)
+                    {
+                        try
+                        {
+                            await _websocketClient.SendAsync(chunk, WebSocketMessageType.Text, endOfMessage: lastChunk, cancellationToken);
+                            delivered = true;
+                        }
+                        catch (Exception ex)
+                        {
+                            if (_websocketClient != null)
```
TingluoHuang, line 219: i think we can remove the check for `if (_websocketClient != null)` since we already checked it in line 197

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +117,138 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (!delivered)
+            {
+                // Giving up for good if we still can't deliver a message after 3 tries
+                _websocketClient = null;
```
TingluoHuang, line 235: should we give up or keep retrying?
assume there is a network blip at the beginning of the job, then you won't get any live console log throughout the entire job even after the network blip recovery.

src/Runner.Common/ResultsServer.cs
yacaovsnc: This is required.  If the case of failure we will reinitialize the webclient, so it's possible this is still null from line 179.

yacaovsnc: Arrrr, you are right, I had this check in the wrong place.  🤦🏼‍♂️

src/Runner.Common/ResultsServer.cs
yacaovsnc: I added a timer to track the last failure - so if 10 minutes passed, we will try again.

src/Runner.Common/ResultsServer.cs
yacaovsnc: I will make server changes.

src/Runner.Common/ResultsServer.cs
yacaovsnc: I still want to trace a different error message.  Refactored this check.

yacaovsnc:
Need to queue the CI build again.

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +119,143 @@ public void InitializeResultsClient(Uri uri, string token)
...
+            var userAgentValues = new List<ProductInfoHeaderValue>();
+            userAgentValues.AddRange(UserAgentUtility.GetDefaultRestUserAgent());
+            userAgentValues.AddRange(HostContext.UserAgents);
+            this._websocketClient.Options.SetRequestHeader("User-Agent", string.Join(" ", userAgentValues.Select(x => x.ToString())));
+
+            // during initialization, retry upto 3 times to setup connection
+            this._websocketConnectTask = ConnectWebSocketClient(liveConsoleFeedUrl, delay, retryConnection: true);
+        }
+
+        private async Task ConnectWebSocketClient(string feedStreamUrl, TimeSpan delay, bool retryConnection = false)
```
TingluoHuang, line 158: I didn't find anyone pass `retryConnection: false`? do we still need the parameter?

yacaovsnc: Line 155 we pass in `true`; `false` is default and we use it in line 226.

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +119,143 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                                delivered = true;
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            var delay = BackoffTimerHelper.GetRandomBackoff(MinDelayForWebsocketReconnect, MaxDelayForWebsocketReconnect);
+                            Trace.Info($"Websocket is not open, let's attempt to connect back again with random backoff {delay} ms.");
+                            Trace.Error(ex);
+                            retries++;
+                            InitializeWebsocketClient(_liveConsoleFeedUrl, _token, delay);
```
yacaovsnc, line 226: We don't want to retry connection setup because we are already inside a retry loop.  We need to retry sending the chunk.

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +119,143 @@ public void InitializeResultsClient(Uri uri, string token)
...
+
+        public async Task<bool> AppendLiveConsoleFeedAsync(Guid scopeIdentifier, string hubName, Guid planId, Guid timelineId, Guid timelineRecordId, Guid stepId, IList<string> lines, long? startLine, CancellationToken cancellationToken)
+        {
+            if (_websocketConnectTask != null)
+            {
+                await _websocketConnectTask;
+            }
+
+            bool delivered = false;
+            int retries = 0;
```
TingluoHuang, line 193: should we reset these before `while (!delivered && retries < 3)`?
What's our retry strategy? 3 retries per chunk or 3 retries per line

src/Runner.Common/ResultsServer.cs
```diff
@@ -94,5 +119,143 @@ public void InitializeResultsClient(Uri uri, string token)
...
+                    : new TimelineRecordFeedLinesWrapper(stepId, lines);
+                var jsonData = StringUtil.ConvertToJson(linesWrapper);
+                var jsonDataBytes = Encoding.UTF8.GetBytes(jsonData);
+                // break the message into chunks of 1024 bytes
+                for (var i = 0; i < jsonDataBytes.Length; i += 1 * 1024)
+                {
+                    var lastChunk = i + (1 * 1024) >= jsonDataBytes.Length;
+                    var chunk = new ArraySegment<byte>(jsonDataBytes, i, Math.Min(1 * 1024, jsonDataBytes.Length - i));
+
+                    while (!delivered && retries < 3)
```
TingluoHuang, line 210: it seems like we sill stop sending any chunks if the first chunk success.

src/Runner.Common/ResultsServer.cs
yacaovsnc: 3 retries per method invocation.  If there is a connection issue, I think we should fallback a little bit aggressively.  If we retry 3 times per chunk, we may try too many times from this call as we could have many chunks, it's unknown to us.

yacaovsnc: 3 retries per method invocation.  If there is a connection issue, I think we should fallback a little bit aggressively.  If we retry 3 times per chunk, we may try too many times from this call as we could have many chunks, it's unknown to us.

src/Runner.Common/ResultsServer.cs
yacaovsnc: Ah refactor fail - moved the condition around too many times.

src/Runner.Common/ResultsServer.cs
TingluoHuang: `ConnectWebSocketClient` or `InitializeWebsocketClient` then?

yacaovsnc: I want to reinitialize the websocketclient as that's cleanest.  We are doing the [same](https://github.com/actions/runner/blob/main/src/Runner.Common/JobServer.cs#L275) in `JobServer` as well so.

yacaovsnc: Ah, what I meant was I don't want to enter another retry loop in the `InitializeWebsocketClient` class so using the `retryConnection = false` default value here.

src/Runner.Common/ResultsServer.cs
TingluoHuang: does the method `ConnectWebSocketClient` get called in anywhere in this file other than lin 155?

yacaovsnc: 🤦🏼‍♂️  Updated.

Issue #716, opened 2020-09-22
${{ github.action_path }} makes no sense when action is run inside self-hosted custom container.


sebastian-freitag-pylot:
**Describe the bug**
${{ github.action_path }} points to a location in the host filesystem of the github runner host. Not to /__w/_actions/...

**To Reproduce**
1. Create a custom container.
2. Create a workflow that uses the custom docker container on a self-hosted runner.
3. Try to follow the guide for the new composite actions that requires to use ${{ github.action_path }}, for example to locate a shell script of a composite action.

**Expected behavior**
The shell script is found and executed, just like when doing this with a normal ubuntu-latest runner.

## Runner Version and Platform
v2.273.4 on Linux.

OS of the machine running the runner? OSX/Windows/Linux/...
Linux.

## What's not working?
``` 
(...)
##[error]Process completed with exit code 127.
Run pylot-tech/actions_checkout@v1
/__w/_temp/615e5631-1155-478e-93e8-8e5cc51fb88e.sh: line 1: /home/github-runner/actions-runner-new/_work/_actions/pylot-tech/actions_checkout/v1/clone.sh: No such file or directory
```

AltiusRupert:
Hi Sebastian,,
SImilar issue here.
Did you figure out a way of running your script ?

AltiusRupert:
See also #649

sebastian-freitag-pylot:
> Hi Sebastian,,
> SImilar issue here.
> Did you figure out a way of running your script ?

By **not** using ${{ github.action_path }} and instead hard-coding the location:
```
runs:
  using: "composite"
  steps:
    - run: /__w/_actions/pylot-tech/actions_checkout/v1/clone.sh
      shell: bash
```

Ugly, quick solution.

AltiusRupert:
Hi Sebastian,
Thanks for your answer.

I too found this same workaround in the meantime.

zotrix:
seems like env variable GITHUB_ACTION_PATH show to correct place.
so step could be called via GITHUB_ACTION_PATH instead of ${{ github.action_path }}
```
  - run: ${GITHUB_ACTION_PATH}/setup.sh
    shell: sh
```

travislikestocode:
Here's a workaround that will allow you to use existing composite actions without modifying them. This should work on self-hosted runners using a BASH shell. Note that this includes a workaround for #691 as well with the step ID `get-user`. The one for this bug is the `workdir` step. I'm including it in case someone finds this bug first.

```
[...]

PR #2517, opened 2023-04-04
Extend github context with host-workspace

nikola-jokic:
We encountered an issue where step host was not used when evaluating user inputs.
The PR #1762 fixed this problem, but it introduced another problem for hash files.

hashFiles operates on the host, regardless of whether you are running inside the container or not.

Since inputs are translated based on the step host, we lose the ability to know how to translate back from the container path to the host path.

This change aims to add another context variable to the runner context named `host-workspace` that will be set to the runner host workspace regardless of whether the action is executed inside the container or not.
Example of the composite action:
```yaml
name: "test"
runs:
  using: "composite"
  steps:
  - run: |
      echo "${{github.action_path}}"
      echo "${{github.workspace}}" 
      echo "${{github.event_path}}"
    shell: bash
    name: "Just echo"
  - run: echo ${{ hashFiles('**/*.lock')}}
    shell: bash
    name: "test"
```

Repro of this issue with StepHost evaluation and hash files:
image

After applying this change:
image

Related to #716

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -1355,7 +1355,7 @@ namespace GitHub.Runner.Worker
         {
             foreach (var key in dict.Keys.ToList())
             {
-                if (key == PipelineTemplateConstants.HostWorkDirectory)
+                if (key == PipelineTemplateConstants.HostWorkspace)
                 {
                     // The HostWorkspace context var is excluded so that there is a var that always points to the host path. 
                     // This var can be used to translate back from container paths, e.g. in HashFilesFunction, which always runs on the host machine
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -1355,7 +1355,7 @@ namespace GitHub.Runner.Worker
         {
             foreach (var key in dict.Keys.ToList())
             {
-                if (key == PipelineTemplateConstants.HostWorkspace)
+                if (key == PipelineTemplateConstants.HostWorkDirectory)
                 {
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


abhijeetdhoke:
@nikola-jokic  could you please help us to fix this issue we are stuck from last week on same stage, I dont have resolution for this.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1355,6 +1355,10 @@ private static void ResolvePathsInExpressionValuesDictionary(this IExecutionCont
         {
             foreach (var key in dict.Keys.ToList())
             {
+                if (key == PipelineTemplateConstants.HostWorkspace)
+                {
+                    continue;
```
fhammerl, line 1360:
```suggestion
                if (key == PipelineTemplateConstants.HostWorkspace)
                {
                    // The HostWorkspace context var is excluded so that there is a var that always points to the host path. 
                    // This var can be used to translate back from container paths, e.g. in HashFilesFunction, which always runs on the host machine
                    continue;
```

src/Runner.Worker/Expressions/HashFilesFunction.cs
```diff
@@ -31,6 +33,23 @@ public sealed class HashFilesFunction : Function
...
+
+            templateContext.ExpressionValues.TryGetValue(PipelineTemplateConstants.Runner, out var runnerContextData);
+            ArgUtil.NotNull(runnerContextData, nameof(runnerContextData));
+            var runnerContext = runnerContextData as DictionaryContextData;
+            ArgUtil.NotNull(runnerContext, nameof(runnerContext));
+            runnerContext.TryGetValue(PipelineTemplateConstants.HostWorkspace, out var hostWorkspace);
+
+            if (hostWorkspace != null)
+            {
+                var hostWorkspaceData = hostWorkspace as StringContextData;
```
fhammerl, line 45:
```suggestion
            // Depends on the 'DistributedTask.UseContainerPathForTemplate' feature flag
            if (hostWorkspace != null)
            {
                var hostWorkspaceData = hostWorkspace as StringContextData;
```

fhammerl (approved):
LGTM

PR #2523, opened 2023-04-06
Feat: Enable Amazon Linux (Fedora)

AllanOricil:
@nikola-jokic 

While configuring a self-hosted runner using an Amazon Linux image, I had problems installing Dotnet Core 6.0 dependencies. The`./bin/installdependencies.sh` does not work https://github.com/amazonlinux/amazon-linux-2023 because there is no `/etc/redhat-release` file. 

image

To fix this issue, I added a new branch in this script that run when `/etc/amazon-linux-release` is present in the distro. 

This is the content of `/etc/` in Amazon's Fedora distro.
image

These are the contents of `/etc/amazon-linux-release` and `/etc/amazon-linux/release-cpe`
image

This is the content of `/etc/os-release`
image

`dnf` is available in this distro
image

In this distro, there are no `/etc/fedora-release` and `/etc/redhat-release` files
image

OPTIONAL CHANGE IN THE DOCS:
You would also need to update this step in the doc, if people decide to use this amazon image.
image

Where it says `shasum -a 256 -C` has to be `sha256sum -C` because this image does not have `shasum` binary installed.


AllanOricil:
well. Maybe this could be useful one day so I will leave this open

DARKo6o6o6:
Fix it you have my permission

iambibhas:
Could use this. Trying to run the runner on Amazon linux and it fails. Have to install everything explicitly.

AllanOricil:
Could someone review this little change, please?

PR #2529, opened 2023-04-12
Resolve Actions Directly From Launch for Run Service Jobs

jherns:
- Create `LaunchHttpClient`
- Create `LaunchServer`
- Initialize `LaunchServer` within `JobServerQueue`
- Initialize `JobServerQueue` when running Run Service Jobs
- Store the job request type in the Global Context so that it can be used in `ActionManager`
- When running a Run Service Job, use the Launch Client to resolve actions instead of Action Service


src/Sdk/Sdk.csproj
```diff
@@ -8,7 +8,7 @@
         <NoWarn>NU1701;NU1603</NoWarn>
         <Version>$(Version)</Version>
         <DefineConstants>TRACE</DefineConstants>
-        <LangVersion>7.3</LangVersion>
+        <LangVersion>8.0</LangVersion>
```
TingluoHuang, line 11: do you have to bump this?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace GitHub.Services.Launch.Contracts
+{
+    [DataContract]
+    public class ActionReferenceRequest
+    {
+        [DataMember(EmitDefaultValue = false, Name = "action")]
+        public string Action { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "version")]
+        public string Version { get; set; }
```
TingluoHuang, line 14: this is currently called `Ref` instead of `Version`, what's the reason for the rename?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [DataMember(EmitDefaultValue = false, Name = "path")]
+        public string Path { get; set; }
```
TingluoHuang, line 17: Does the server care about the `Path` of the action?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+        public string ResolvedName { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "resolved_sha")]
+        public string ResolvedSha { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "tar_url")]
+        public string TarUrl { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "version")]
+        public string Version { get; set; }
```
TingluoHuang, line 46: same question about `version`  vs `ref`

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [DataMember(EmitDefaultValue = false, Name = "zip_url")]
+        public string ZipUrl { get; set; }
+    }
+
+    [DataContract]
+    public class ActionDownloadAuthenticationResponse
+    {
+        [DataMember(EmitDefaultValue = false, Name = "expires_at")]
+        public DateTime ExpiresAt { get; set; }
```
TingluoHuang, line 56: do we really care about the token's expiresAt?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [DataMember(EmitDefaultValue = false, Name = "version")]
+        public string Version { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "path")]
+        public string Path { get; set; }
+    }
+
+    [DataContract]
+    public class ActionReferenceRequestList
```
TingluoHuang, line 21: why do we need an object to wrap around a list?
Can we just send the List as json to service?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        [DataMember(EmitDefaultValue = false, Name = "expires_at")]
+        public DateTime ExpiresAt { get; set; }
+
+        [DataMember(EmitDefaultValue = false, Name = "token")]
+        public string Token { get; set; }
+    }
+
+    [DataContract]
+    public class ActionDownloadInfoResponseCollection
```
TingluoHuang, line 63: same question, why do we need this wrapper around a dictionary?

src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -0,0 +1,118 @@
...
+using GitHub.DistributedTask.WebApi;
+using GitHub.Services.Launch.Contracts;
+
+using Sdk.WebApi.WebApi;
+
+namespace GitHub.Services.Launch.Client
+{
+    public class LaunchHttpClient : RawHttpClientBase
+    {
+
```
TingluoHuang, line 20:
```suggestion
```

src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -0,0 +1,118 @@
...
+            bool disposeHandler)
+            : base(baseUrl, pipeline, disposeHandler)
+        {
+            m_token = token;
+            m_launchServiceUrl = baseUrl;
+            m_formatter = new JsonMediaTypeFormatter();
+        }
+
+        // Resolve Actions
+        private async Task<T> GetLaunchSignedURLResponse<R, T>(Uri uri, R request, CancellationToken cancellationToken)
```
TingluoHuang, line 34: move the private method down to the file.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -741,6 +741,9 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
             // File table
             Global.FileTable = new List<String>(message.FileTable ?? new string[0]);
 
+            // What type of job request is running (i.e. Run Service vs. pipelines)
+            Global.Variables.Set("job_request_type", message.MessageType);
```
TingluoHuang, line 745: put `job_request_type` into the constants file.
also, name it `system.jobRequestType`.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -66,6 +66,7 @@ public sealed class JobServerQueue : RunnerService, IJobServerQueue
         // common
         private IJobServer _jobServer;
         private IResultsServer _resultsServer;
+        private ILaunchServer _launchServer;
```
TingluoHuang, line 69: `_launchServer` should not be initialized as part of `jobserverqueue`, I would put it in `jobrunner.cs` and similar to where we initialize the `JobServer`

src/Runner.Common/JobServerQueue.cs
```diff
@@ -109,25 +111,34 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest, bool resultServic
...
                 }
 
-                _resultsServer.InitializeResultsClient(new Uri(resultsReceiverEndpoint), liveConsoleFeedUrl, accessToken);
-                _resultsClientInitiated = true;
+                jobRequest.Variables.TryGetValue("system.github.launch_endpoint", out VariableValue launchEndpointVariable);
+                var launchReceiverEndpoint = launchEndpointVariable?.Value;
+                if (!string.IsNullOrEmpty(launchReceiverEndpoint))
+                {
+                    Trace.Info("Initializing launch client");
+                    _launchServer.InitializeLaunchClient(new Uri(launchReceiverEndpoint), accessToken);
```
TingluoHuang, line 140: we are using a JWT token from Token service to auth with Laucn?

src/Sdk/WebApi/WebApi/LaunchContracts.cs
jherns: This is copied pretty much directly from https://github.com/github/actions-dotnet/blob/179ed33568dd1ad80c457fb1c328c964e372accf/Launch/Sdk/Server/Models/GitHub.cs#L90-L152. Since the runner is now the one receiving the payload from launch, it needs to parse the JSON payload the same way. I don't believe this is a rename

src/Runner.Common/JobServerQueue.cs
jherns: Yes, that's correct

src/Sdk/Sdk.csproj
jherns: Without bumping, we can't use Nullable Reference Types, which would mean that we can't just copy the deserialization for the actions resolution response from actions-dotnet

src/Runner.Common/JobServerQueue.cs
jherns: Isn't this following the exact same pattern as both the `JobServer` and the `ResultsServer`?

TingluoHuang: not really, I think.
`await runServer.ConnectAsync(systemConnection.Url, jobServerCredential);` in `JobRunner.cs`

JobServerQueue.cs is for stuff that is supposed to run background for reporting status/log/etc, any IXXXServer that is used by the job should be initially in the JobRunner.cs

src/Runner.Worker/JobRunner.cs
```diff
@@ -58,6 +58,21 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                    systemConnection.Authorization.Parameters.TryGetValue("AccessToken", out var accessToken) &&
+                    !string.IsNullOrEmpty(accessToken) &&
+                    !string.IsNullOrEmpty(launchReceiverEndpoint))
+                {
+                    var launchServer = HostContext.GetService<ILaunchServer>();
+                    if (!string.IsNullOrEmpty(launchReceiverEndpoint))
+                    {
+                        Trace.Info("Initializing launch client");
+                        launchServer.InitializeLaunchClient(new Uri(launchReceiverEndpoint), accessToken);
+                    }
```
TingluoHuang, line 74:
```suggestion
                    Trace.Info("Initializing launch client");
                    var launchServer = HostContext.GetService<ILaunchServer>();
                    launchServer.InitializeLaunchClient(new Uri(launchReceiverEndpoint), accessToken);
```

src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -0,0 +1,117 @@
...
+            : base(baseUrl, pipeline, disposeHandler)
+        {
+            m_token = token;
+            m_launchServiceUrl = baseUrl;
+            m_formatter = new JsonMediaTypeFormatter();
+        }
+
+        public async Task<ActionDownloadInfoCollection> GetResolveActionsDownloadInfoAsync(Guid planId, Guid jobId, ActionReferenceList actionReferenceList, CancellationToken cancellationToken)
+        {
+
```
TingluoHuang, line 34:
```suggestion
```

src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -0,0 +1,117 @@
...
+            var GetResolveActionsDownloadInfoURLEndpoint = new Uri(m_launchServiceUrl, $"/actions/build/{planId.ToString()}/jobs/{jobId.ToString()}/runnerresolve/actions");
+
```
TingluoHuang, line 36:
```suggestion
```

TingluoHuang (approved):

PR #2531, opened 2023-04-13
Also send in conclusion for steps

yacaovsnc:
We also need to send the step result back to Results server.


lokesh755 (approved):

f56id:
No hack no scam oor spam

PR #2547, opened 2023-04-17
Rename AcquireJobRequest::StreamID to AcquireJobRequest::JobMessageID

jww3:
Rename AcquireJobRequest::StreamID to AcquireJobRequest::JobMessageID to match corresponding server-side update.

### DIFF
--- a/src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
+++ b/src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
@@ -7,7 +7,29 @@ namespace GitHub.Actions.RunService.WebApi
     [DataContract]
     public class AcquireJobRequest
     {
+        [DataMember(Name = "jobMessageId", EmitDefaultValue = false)]
+        public string JobMessageID { get; set; }
+
+        // The earliest versions of this type serialized the JobMessageID under a field named "streamId"
+        // This private property is expected to be a short-lived shim to aid in transitioning over to the new schema.
         [DataMember(Name = "streamId", EmitDefaultValue = false)]
-        public string StreamID { get; set; }
+        private string DeprecatedJobMessageIdAlias
+        {
+            get
+            {
+                return this.JobMessageID;
+            }
+            set
+            {
+                // Note that this is a private property, so the only way to reach this setter is via deserialization.
+                // Deserializtion is NOT an expected use case on the Runner side as of early 2023.
+                // In the unlikely event that deserializtion does become a valid use case, we include this guard
+                // to ensure that a valid JobMessageID can't be overwritten by this otherwise unreachable code path.
+                if (string.IsNullOrEmpty(this.JobMessageID))
+                {
+                    this.JobMessageID = value;
+                }
+            }
+        }
     }
 }
--- a/src/Sdk/RSWebApi/RunServiceHttpClient.cs
+++ b/src/Sdk/RSWebApi/RunServiceHttpClient.cs
@@ -64,7 +64,7 @@ namespace GitHub.Actions.RunService.WebApi
             HttpMethod httpMethod = new HttpMethod("POST");
             var payload = new AcquireJobRequest
             {
-                StreamID = messageId
+                JobMessageID = messageId
             };
 
             requestUri = new Uri(requestUri, "acquirejob");


src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
```diff
@@ -7,7 +7,29 @@ namespace GitHub.Actions.RunService.WebApi
     [DataContract]
     public class AcquireJobRequest
```
yaananth, line 8: This might be cleaner? 

```suggestion
    using System;
using System.Runtime.Serialization;

[DataContract]
public class AcquireJobRequest
{
    [DataMember(Name = "jobMessageId", EmitDefaultValue = false)]
    public string JobMessageID { get; set; }

    [DataMember(Name = "streamId", EmitDefaultValue = false)]
    private string DeprecatedJobMessageIdAlias { get; set; }

    [OnSerializing]
    private void OnSerializing(StreamingContext context)
    {
        DeprecatedJobMessageIdAlias = JobMessageID;
    }


    [OnDeserializing]
    private void OnDeserializing(StreamingContext context)
    {
        if (string.IsNullOrEmpty(JobMessageID) && !string.IsNullOrEmpty(DeprecatedJobMessageIdAlias))
        {
            JobMessageID = DeprecatedJobMessageIdAlias;
        }
    }
}

```

yaananth (commented):
runner folks owns the code, but left a suggestion

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
```diff
@@ -7,7 +7,29 @@ namespace GitHub.Actions.RunService.WebApi
...
+
+        // The earliest versions of this type serialized the JobMessageID under a field named "streamId"
+        // This private property is expected to be a short-lived shim to aid in transitioning over to the new schema.
         [DataMember(Name = "streamId", EmitDefaultValue = false)]
-        public string StreamID { get; set; }
+        private string DeprecatedJobMessageIdAlias
+        {
+            get
+            {
+                return this.JobMessageID;
```
jww3, line 20: "streamId" gets serialized out with whatever value is stored in this.JobMessageID 

(Note that it doesn't have its own backing field.  It piggy-backs on JobMessageID.)

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
```diff
@@ -7,7 +7,29 @@ namespace GitHub.Actions.RunService.WebApi
...
+            set
+            {
+                // Note that this is a private property, so the only way to reach this setter is via deserialization.
+                // Deserializtion is NOT an expected use case on the Runner side as of early 2023.
+                // In the unlikely event that deserializtion does become a valid use case, we include this guard
+                // to ensure that a valid JobMessageID can't be overwritten by this otherwise unreachable code path.
+                if (string.IsNullOrEmpty(this.JobMessageID))
+                {
+                    this.JobMessageID = value;
+                }
```
jww3, line 31: It's worth pointing out here that the main factor at play in this unlikely scenario is the order in which the fields are deserialized.  Experimentation reveals it depends more on the raw JSON order than anything else.  Setting the `Order` property in the `DataMember` attribute did not produce a reliable guarantee that one field would always be deserialized before another.

jww3: I also considered just making this setter a `throw` since we don't anticipate this code path to ever be exercised.

jww3:
follow-up PR is here:  https://github.com/actions/runner/pull/2549

src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -64,7 +64,7 @@ public class RunServiceHttpClient : RawHttpClientBase
             HttpMethod httpMethod = new HttpMethod("POST");
             var payload = new AcquireJobRequest
             {
-                StreamID = messageId
+                JobMessageID = messageId
```
TingluoHuang, line 67: will this cause any compat break on the server side, feels like either we accept the break or send both fields in a short period of time so we can release the runner ahead of the server.

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
jww3: I added unit tests to prove `streamId` is serialized in every scenario where `JobMessageID` has a non-empty value.

src/Sdk/RSWebApi/RunServiceHttpClient.cs
jww3: Yeah, the easiest way is to release the runner ahead of the server and send both fields for a short period of time.

The corresponding server change is here:  https://github.com/github/actions-run-service/pull/525/files#diff-9a30818e05fad1a13d0fcde9639c0ac3ac29a7902e95eeba19659808b2192b45

src/Test/L0/Sdk/RSWebApi/AcquireJobRequestL0.cs
```diff
@@ -0,0 +1,62 @@
...
+    public void VerifyDeserialization()
+    {
+        var serializer = new DataContractJsonSerializer(typeof(AcquireJobRequest));
+        var variations = new Dictionary<string, string>()
+        {
+            ["{'streamId': 'legacy', 'jobMessageId': 'new'}"] = "new",
+            ["{'jobMessageId': 'new', 'streamId': 'legacy'}"] = "new",
+            ["{'jobMessageId': 'new'}"] = "new",
+            ["{'streamId': 'legacy'}"] = "legacy"
+        };
```
jww3, line 45: I should have probably used a list of Tuple's for predictable ordering.

src/Sdk/RSWebApi/RunServiceHttpClient.cs
TingluoHuang: Do you want to include that in the PR? sending both `JobMessageId` and `StreamId`

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
```diff
@@ -7,7 +7,29 @@ namespace GitHub.Actions.RunService.WebApi
     [DataContract]
     public class AcquireJobRequest
     {
+        [DataMember(Name = "jobMessageId", EmitDefaultValue = false)]
+        public string JobMessageID { get; set; }
```
TingluoHuang, line 11: `JobMessageID`->`JobMessageId`

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
TingluoHuang:
```
[DataContract]
public class AcquireJobRequest
{
    [DataMember(Name = "jobMessageId", EmitDefaultValue = false)]
    public string JobMessageId { get; set; }

    [DataMember(Name = "streamId", EmitDefaultValue = false)]
    public string StreamId { get; set; }
}
```

Can we simply do this ☝️  and send both to service in this PR, dropping the one we don't want in the next runner release?

yaananth: +1, this is probably the right way: https://github.com/github/actions-run-service/pull/520#discussion_r1171204673

We can't depend on deployment order, for things like these, we need to provide full compat and then clean up next release

jww3: I talked with Ting.  I'm still confident my proposed solution will work just as well, but am more interested in moving forward here, so if Ting's suggestion is the consensus best approach, so be it.

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
jww3: fixed

src/Sdk/RSWebApi/RunServiceHttpClient.cs
jww3: Both fields are sent.

TingluoHuang (approved):

src/Sdk/RSWebApi/Contracts/AcquireJobRequest.cs
yaananth: In my mind the best path is: (1) deploy run service to accept both (2) runner just does the new way (3) after runner is release run service will only accept new things

Could you explain what you are thinking off from your side for posterity?

PR #2568, opened 2023-04-26
Add orchestrationId to useragent for better correlation.

TingluoHuang:
With `OrchestrationId` in the user-agent, we can get a better correlation in our telemetry to figure out the source of each HTTP request (runner+job).

### DIFF
--- a/src/Runner.Worker/JobRunner.cs
+++ b/src/Runner.Worker/JobRunner.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/JobRunner.cs
```diff
@@ -42,6 +43,13 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
             DateTime jobStartTimeUtc = DateTime.UtcNow;
             IRunnerService server = null;
 
+            // add orchestration id to useragent for better correlation.
+            if (message.Variables.TryGetValue(Constants.Variables.System.OrchestrationId, out VariableValue orchestrationId) &&
```
TingluoHuang, line 47: I still need to do a server PR to add the variable to the job message.

cory-miller (dismissed):

cory-miller (approved):

PR #2578, opened 2023-05-02
When inferring additional secrets from multi-line content, ignore very short lines.

jww3:


rentziass (approved):
This looks good to me!

src/Sdk/DTLogging/Logging/SecretMasker.cs
```diff
@@ -40,6 +40,19 @@ private SecretMasker(SecretMasker copy)
             }
         }
 
+        /// <summary>
+        /// Provide callers with a recommendation on what to consider a secret.
+        /// This is helpful in cases where JSON (for example) is broken into multiple lines
+        /// and we don't want to start treating standalone JSON control characters as secrets.
+        /// </summary>
+        public int DerivedSecretRecommendedMinimumLength
```
TingluoHuang, line 48: i would just define this on some file not under `src/SDK` since those better deployed with the server code first.

src/Runner.Worker/Worker.cs
```diff
@@ -156,8 +156,9 @@ private void InitializeSecretMasker(Pipelines.AgentJobRequestMessage message)
                     HostContext.SecretMasker.AddValue(value);
 
                     // Also add each individual line. Typically individual lines are processed from STDOUT of child processes.
-                    var split = value.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
-                    foreach (var item in split)
+                    var auxiliarySecrets = value.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
```
TingluoHuang, line 159: you can also get the `DerivedSecretRecommendedMinimumLength` from something inside the message, so you can configure the value from the service via feature flag/registry setting instead of a hardcode `3`

jmarshall:
We ran into this today, with the scrubber redacting `{` and `}` in JSON output, and the resulting `Skip output 'foo' since it may contain secret` preventing the workflow from running as expected.

Is this fix or an equivalent likely to be merged?

PR #2591, opened 2023-05-08
Trace WebSocket exception into verbose level to reduce noise in diag log.

TingluoHuang:
Live console output are best effort.
We use websocket to send out those console output, the websocket will be closed by the service once every 30 minutes and cause some huge non-actionable exception message get traced in the diag log.

I am moving the exception trace into `Verbose` level, so it' only shows up in when tracelevel set to `debug` (Not the default).

https://github.com/github/c2c-actions-runtime/issues/2439


takost (approved):
LGTM!

PR #2601, opened 2023-05-14
Enable linux/arm64 docker build and publish

dee-kryvenko:
Hey all - I am not sure if it was intentionally not enabled, but on the off chance it wasn't...


deviantintegral:
I'm prototyping https://github.com/actions/actions-runner-controller by pointing it to my local colima instance on my Mac. I was surprised to see everything running through emulation and not native binaries. It looks like the underlying issue is that the runner is only amd64, so all binaries inside of it fall back to that too.

ajschmidt8:
I hope this gets merged soon. My team is currently building our own `arm64` binaries for our use case. It would be nice if these were published from the source repository.

@nikola-jokic, @TingluoHuang, any thoughts on when/if this PR can be merged?

nikola-jokic:
Hey everyone,

We are planning to address this issue, hopefully before the next release :relaxed: 

However, we will create a separate PR to address this issue since we also need to publish images with correct labels and have it be part of the release process

nikola-jokic (approved):
LGTM! @TingluoHuang I tested it on my local repo and just moved these changes there. Everything works :smile: Do you mind also taking a look just in case. I'm confident this is all we need

images/Dockerfile
```diff
@@ -1,25 +1,30 @@
 # Source: https://github.com/dotnet/dotnet-docker
 FROM mcr.microsoft.com/dotnet/runtime-deps:6.0-jammy as build
 
+ARG TARGETOS
+ARG TARGETARCH
 ARG RUNNER_VERSION
-ARG RUNNER_ARCH="x64"
 ARG RUNNER_CONTAINER_HOOKS_VERSION=0.3.2
+ARG RUNNER_ARCH="x64"
```
TingluoHuang, line 8: do we still this as an ARG?

nikola-jokic: I think it should be safe to remove it
With this change, it is calculated always, and I tested it on my fork just in case

TingluoHuang (approved):

wyatterp365:
<details><summary>Details</summary>
<p>
 

- [ ] **@`g😆`**



</p>
</details>

PR #2605, opened 2023-05-16
Dont log template errors

JoannaaKL:
### Context:
When evaluating composite actions template issues are logged twice (they're added to _timelineUpdateQueue and to _webConsoleLineQueue). This pr removes the issue duplicates when merging timelines.
Closes https://github.com/github/c2c-actions-runtime/issues/2381 

### Before the fix, we can see duplicated annotations
Screenshot 2023-05-17 at 10 45 48
### After the fix
Screenshot 2023-05-17 at 10 41 44

### Todo:
- [ ] Add more tests for composite template with full template evaluation
- [ ] Better document handling of composite actions - when the template has errors it fails on template load. The embedded context is not even created and CompositeActionsHandler is not used

### DIFF
--- a/src/Runner.Common/JobServerQueue.cs
+++ b/src/Runner.Common/JobServerQueue.cs
@@ -799,6 +799,7 @@ namespace GitHub.Runner.Common
                 Trace.Verbose($"    Record: t={record.RecordType}, n={record.Name}, s={record.State}, st={record.StartTime}, {record.PercentComplete}%, ft={record.FinishTime}, r={record.Result}: {record.CurrentOperation}");
                 if (record.Issues != null)
                 {
+                    removeDuplicatedIssues(record);
                     foreach (var issue in record.Issues)
                     {
                         String source;
@@ -817,6 +818,18 @@ namespace GitHub.Runner.Common
             }
 
             return mergedRecords;
+        }
+
+        private void removeDuplicatedIssues(TimelineRecord timelineRecord)
+        {
+            Dictionary<string, Issue> issuesMap = new Dictionary<string, Issue>();
+            foreach (var i in timelineRecord.Issues)
+            {
+                issuesMap.TryAdd(i.Message, i);
+            }
+            timelineRecord.Issues.Clear();
+
+            timelineRecord.Issues.AddRange(issuesMap.Select(entry => entry.Value).ToList());
         }
 
         private async Task UploadFile(UploadFileInfo file)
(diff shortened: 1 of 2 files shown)


src/Runner.Common/JobServerQueue.cs
```diff
@@ -819,6 +820,18 @@ private List<TimelineRecord> MergeTimelineRecords(List<TimelineRecord> timelineR
             return mergedRecords;
         }
 
+        private void removeDuplicatedIssues(TimelineRecord timelineRecord)
```
jww3, line 823: nit:  C# naming standards are PascalCase, not camelCase.

JoannaaKL: Ah right, golang went too deep :D

src/Runner.Common/JobServerQueue.cs
```diff
@@ -819,6 +820,18 @@ private List<TimelineRecord> MergeTimelineRecords(List<TimelineRecord> timelineR
...
+        {
+            Dictionary<string, Issue> issuesMap = new Dictionary<string, Issue>();
+            foreach (var i in timelineRecord.Issues)
+            {
```
jww3, line 827: The message text is probably not the right key, here.  What about legitimate duplicates  (i.e. the same error happens several times)?

It'd be nice if `Issue` included a high-precision timestamp or something so that we could distinguish between those.

JoannaaKL: That's a broader question of how to we define the identity of an `Issue`. Message is the most obvious one because that's what an user effectively see, I can add `Type`, `Category` and `IsInfrastructureIssue` to the comparison to include all fields (or just object's HashCode). Timestamp is not ideal because the main problem is that the same Issue is reported in different places. So we still will have duplicated messages, but with distinct timestamps. :D

jww3: One potentially simple fix might be to change `TimelineRecord::Issues` to be backed by an order-preserving `Set` instead of a `List`.  That would automatically cull duplicates if the same issue reference was added twice.

JoannaaKL: I thought about it but ideally we should fix the real issue, which is in this case: why are duplicates even added? Errors are being logged to the context all over the place and that's the result.
This pr is just a duck tape to fix this particular bug tbh. :D

fhammerl: The PR is purely for templating issues though, right? Templating issues have unique messages because of the `/foo/path (Line: 8, Col: 7)` format. I can't think of a legitimate duplicate templating issue that would help the action author here.

JoannaaKL: Regarding removing 'legit' duplicates: issue message contains the line and column numbers, so even if two exceptions are the same they won't be deduplicated because effectively the messages will be different.

PR #2627, opened 2023-05-26
Fixed a bug where a misplaced `=` character could bypass heredoc-style processing.

jww3:
Fixes https://github.com/github/c2c-actions/issues/6910

GitHub Docs for context:  https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -322,21 +322,9 @@ namespace GitHub.Runner.Worker
                     var equalsIndex = line.IndexOf("=", StringComparison.Ordinal);
                     var heredocIndex = line.IndexOf("<<", StringComparison.Ordinal);
 
-                    // Normal style NAME=VALUE
-                    if (equalsIndex >= 0 && (heredocIndex < 0 || equalsIndex < heredocIndex))
-                    {
-                        var split = line.Split(new[] { '=' }, 2, StringSplitOptions.None);
-                        if (string.IsNullOrEmpty(line))
-                        {
-                            throw new Exception($"Invalid format '{line}'. Name must not be empty");
-                        }
-
-                        key = split[0];
-                        output = split[1];
-                    }
-
-                    // Heredoc style NAME<<EOF
-                    else if (heredocIndex >= 0 && (equalsIndex < 0 || heredocIndex < equalsIndex))
+                    // Heredoc style NAME<<EOF (where EOF is typically randomly-generated Base64 and may include an '=' character)
+                    // see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
+                    if (heredocIndex >= 0 && (equalsIndex < 0 || heredocIndex < equalsIndex))
                     {
                         var split = line.Split(new[] { "<<" }, 2, StringSplitOptions.None);
                         if (string.IsNullOrEmpty(split[0]) || string.IsNullOrEmpty(split[1]))
@@ -363,6 +351,18 @@ namespace GitHub.Runner.Worker
                         }
 
                         output = endIndex > startIndex ? text.Substring(startIndex, endIndex - startIndex) : string.Empty;
+                    }
+                    // Normal style NAME=VALUE
+                    else if (equalsIndex >= 0 && heredocIndex < 0)
+                    {
+                        var split = line.Split(new[] { '=' }, 2, StringSplitOptions.None);
+                        if (string.IsNullOrEmpty(line))
+                        {
+                            throw new Exception($"Invalid format '{line}'. Name must not be empty");
+                        }
+
+                        key = split[0];
+                        output = split[1];
                     }
                     else
                     {
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -281,7 +281,7 @@ namespace GitHub.Runner.Worker
         }
     }
 
-    public sealed class EnvFileKeyValuePairs: IEnumerable<KeyValuePair<string, string>>
+    public sealed class EnvFileKeyValuePairs : IEnumerable<KeyValuePair<string, string>>
     {
         private IExecutionContext _context;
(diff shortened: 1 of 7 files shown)


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -322,21 +322,9 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
                     var equalsIndex = line.IndexOf("=", StringComparison.Ordinal);
                     var heredocIndex = line.IndexOf("<<", StringComparison.Ordinal);
 
-                    // Normal style NAME=VALUE
-                    if (equalsIndex >= 0 && (heredocIndex < 0 || equalsIndex < heredocIndex))
```
jww3, line 326: I basically just swapped the order of this conditional and the `else if` that follows it.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -364,6 +352,18 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
 
                         output = endIndex > startIndex ? text.Substring(startIndex, endIndex - startIndex) : string.Empty;
                     }
+                    // Normal style NAME=VALUE
+                    else if (equalsIndex >= 0 && heredocIndex < 0)
```
jww3, line 356: Now we can be stricter here:

Note that we're checking

```
if (equalsIndex >= 0 && heredocIndex < 0)
```

whereas before it was:
```
if (equalsIndex >= 0 && (heredocIndex < 0 || equalsIndex < heredocIndex))
```

src/Test/L0/Worker/CreateStepSummaryCommandL0.cs
```diff
@@ -167,21 +167,6 @@ public void CreateStepSummaryCommand_ScrubSecrets()
...
-        {
-            if (string.IsNullOrEmpty(newline))
-            {
-                newline = Environment.NewLine;
-            }
-
-            var encoding = new UTF8Encoding(true); // Emit BOM
-            var contentStr = string.Join(newline, content);
-            File.WriteAllText(path, contentStr, encoding);
-        }
```
jww3, line 183: There were four verbatim copies of this method floating around, so I defined it once-and-only-once in TestUtil.cs.

I also dropped the ability to pass an arbitrary string in as a newline.  (Several tests were passing a single SPACE which was such a subtle difference that it made it really hard to decipher why two seemingly-identical tests would behave differently.)

The new version in `TestUtil.cs` accepts an enum for specifying the LineEnding style.


```
public enum LineEndingType
{
    Native,
    Linux   = 0x__0A,
    Windows = 0x0D0A
}

public static void WriteContent(
    string path, 
    IEnumerable<string> content, 
    LineEndingType lineEnding = LineEndingType.Native)
```

src/Test/L0/Worker/SaveStateFileCommandL0.cs
```diff
@@ -311,15 +311,8 @@ public void SaveStateFileCommand_Heredoc_MissingNewLine()
...
-                    "line one",
-                    "line two",
-                    "line three",
-                    "EOF",
-                };
-                WriteContent(stateFile, content, " ");
+                string content = "MY_OUTPUT<<EOF line one line two line three EOF";
+                TestUtil.WriteContent(stateFile, content);
                 var ex = Assert.Throws<Exception>(() => _saveStateFileCommand.ProcessCommand(_executionContext.Object, stateFile, null));
                 Assert.Contains("Matching delimiter not found", ex.Message);
```
jww3, line 317: As I mentioned earlier, it wasn't easy to spot the difference between this test case and the one that begins on line 190 and I was puzzling over why that one is expected to succeed and this one is expected to fail.

The difference is just the third arg (`" "`) passed to `WriteContent` which "tricks" `WriteContent` into mashing all the content together on a single line.

I think it's clearer to just be explicit about that:

```
string content = "MY_OUTPUT<<EOF line one line two line three EOF";
```

src/Test/L0/Worker/SetEnvFileCommandL0.cs
```diff
@@ -250,75 +247,144 @@ public void SetEnvFileCommand_Heredoc_SkipEmptyLines()
...
             using (var hostContext = Setup())
             {
                 var envFile = Path.Combine(_rootDirectory, "heredoc");
                 var content = new List<string>
                 {
-                    "MY_ENV<<=EOF",
-                    "hello",
-                    "one",
-                    "=EOF",
-                    "MY_ENV_2<<<EOF",
```
jww3, line 276: It wasn't clear to me that the `=` on line 272 and the final `<` on line 276 are actually considered part of the "end marker", so I split these two cases out into a separate unit test, `SetEnvFileCommand_Heredoc_EndMarkerVariations` that leverages `[InlineData("=EOF")][InlineData("<EOF")]` to make things clearer.

src/Test/L0/Worker/SetEnvFileCommandL0.cs
jww3, line 1: TODO:  Copy the new unit tests here (or find a way to declare them all in one place).

src/Test/L0/Worker/SetOutputFileCommandL0.cs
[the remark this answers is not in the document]
jww3: done

src/Test/L0/Worker/SaveStateFileCommandL0.cs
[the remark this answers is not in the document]
jww3: done

src/Test/L0/Worker/FileCommandTestBase.cs
jww3, line 1: The next three files in this PR were nearly identical.  I took all the duplicated code and moved it into this file.  That allowed me to pare down the three files that follow.

rentziass (approved):
LGTM, thank you for looking into it!

Link- (approved):
I don't see any glaring issues, let's move this forward!

PR #2635, opened 2023-05-31
Move "Using V2 Flow" log to Trace instead of WriteLine

luketomlinson:
There is no need to pollute the output of Runner Registration with this tracing. Move it to the logs.


JoannaaKL (approved):
lgtm

PR #2637, opened 2023-06-01
Log template errors as debug message

JoannaaKL:
**Background**
When parsing a workflow template there are two objects at play: `ExecutionContext` and `TemplateContext`. `ExecutionContext` stores all errors (named issues) that are later sent as job results. `TemplateContext` on the other side is an object used, as the name suggests, during the template parsing. It also stores template errors together with field ids and names, to provide a meaningful message, which workflow line was invalid. 

The duplication comes from the fact that during the template reading:
1. first we’re loading the template using [TemplateReader.Read](https://github.com/actions/runner/blob/efffbaeabc6d53c4c1ec05b11cea58331ff38e3c/src/Runner.Worker/ActionManifestManager.cs#L87) and here template errors are being written to the `TemplateContext` for the first time
1. after we’re done with loading all tokens we go through all TemplateContext errors and add them to the ExecutionContext errors list [here](https://github.com/actions/runner/blob/efffbaeabc6d53c4c1ec05b11cea58331ff38e3c/src/Runner.Worker/ActionManifestManager.cs#L144) (screenshot with more logs is at the bottom in expandable section) 
	
All that would be fine if not the small fact, that in step **1** `TemplateContext` already added errors to the ExecutionContext: internally it is using `TraceWriter.Error` method that is…writing errors to the `ExecutionContext`:
```
internal void Error(TemplateValidationError error)
{
            Errors.Add(error); <——— add error to Template errors (we iterate over them in step 2) 
            TraceWriter.Error(error.Message); <—— TraceWriter writes errors to ExecutionContext directly causing duplication 
}
```
[link](https://github.com/actions/runner/blob/efffbaeabc6d53c4c1ec05b11cea58331ff38e3c/src/Runner.Worker/ExecutionContext.cs#LL1404C13-L1404C30)

We could deduplicate issues in ExecutionContext based on error position and message, but that would not fix the underlying issue of logging two times the same error or fix the real issue, that is writing the same errors in multiple places.
Since the workflow parser code is used in other classes and there's a risk of regression, I have added a feature flag `DistributedTask.LogTemplateErrorsToTraceWriter` that will be passed from actions service. This way only one method in `TemplateReader` will stop duplicating errors. With time we can expand this logic and disable TraceWriter error logging in other places.
Long term solution would be code a refactoring, since `TemplateContext` has access to the `TraceWriter`, it can manipulate the state of `ExecutionContext`. These two should either be separated or merged into one object.

Closes https://github.com/github/c2c-actions-runtime/issues/2381
Follow-up: https://github.com/github/actions-dotnet/pull/15290

**Before**
Screenshot 2023-05-31 at 16 01 33
**After**
Screenshot 2023-05-31 at 16 01 05

<details>
<summary>Workflow run with more debug logs</summary>

[Branch](https://github.com/actions/runner/tree/DuplicatedErrorWithLogging)
  
Screenshot 2023-06-13 at 09 03 12

</details>

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -1402,14 +1402,13 @@ namespace GitHub.Runner.Worker
 
         public void Error(string format, params Object[] args)
         {
-            _executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Runner.Features.LogTemplateErrorsAsDebugMessage, out var logErrorsAsDebug);
-            if (StringUtil.ConvertToBoolean(logErrorsAsDebug, defaultValue: false))
+
+            if (logTemplateErrorsAsDebugMessage())
             {
                 _executionContext.Debug(string.Format(CultureInfo.CurrentCulture, format, args));
             }
             else _executionContext.Error(string.Format(CultureInfo.CurrentCulture, format, args));
         }
-
 
         public void Info(string format, params Object[] args)
         {
@@ -1421,6 +1420,13 @@ namespace GitHub.Runner.Worker
             // todo: switch to verbose?
             _executionContext.Debug(string.Format(CultureInfo.CurrentCulture, $"{format}", args));
         }
+
+        private bool logTemplateErrorsAsDebugMessage()
+        {
+            _executionContext.Global.EnvironmentVariables.TryGetValue(Constants.Runner.Features.LogTemplateErrorsAsDebugMessage, out var logErrorsAsDebug);
+            return StringUtil.ConvertToBoolean(logErrorsAsDebug, defaultValue: false);
+        }
+
     }
 
     public static class WellKnownTags
@@ -1432,4 +1438,5 @@ namespace GitHub.Runner.Worker
         public static readonly string Notice = "##[notice]";
         public static readonly string Debug = "##[debug]";
     }
+
 }
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -1423,7 +1423,7 @@ namespace GitHub.Runner.Worker
 
         private bool logTemplateErrorsAsDebugMessage()
         {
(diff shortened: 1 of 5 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1401,7 +1402,12 @@ internal TemplateTraceWriter(IExecutionContext executionContext)
 
         public void Error(string format, params Object[] args)
         {
-            _executionContext.Error(string.Format(CultureInfo.CurrentCulture, format, args));
+
+            if (logTemplateErrorsAsDebugMessages())
```
TingluoHuang, line 1406: i think the proper fix would be change `ActionsManifestManager` to properly handle exception `TemplateValidationException`

TingluoHuang: https://github.com/actions/runner/commit/690d4f769b706f8c68632a3b3945e296dbe86a56

JoannaaKL: When I'm reading [these lines](https://github.com/actions/runner/blob/690d4f769b706f8c68632a3b3945e296dbe86a56/src/Runner.Worker/ActionManifestManager.cs#LL133C6-L133C6):
```
catch (Exception ex) when (!(ex is TemplateValidationException))
{
                Trace.Error(ex);
                templateContext.Errors.Add(ex);
}
```
If an error is not a template validation exception add it to template context errors. Looking at just human-level language understanding I'd expect the contrary to happen, meaning add an error if the template failed validation.

src/Sdk/DTObjectTemplating/ObjectTemplating/TemplateContext.cs
TingluoHuang, line 1: Try to not change anything under `src/Sdk` within the runner, these files are shared with the service and they should be keep in sync. If you really plan to change this file, you will need to change it on the service first and rollout out to production, then copy the changed version to the runner.

src/Runner.Worker/ExecutionContext.cs
TingluoHuang, line 1: Undo change in this file?

PR #2640, opened 2023-06-02
Goodbye BOM - Convert (C# source files) from UTF-8 with BOM to standard UTF-8

fhammerl:
What separates UTF-8 from UTF-8 BOM is the 3 byte order markers at the beginning of the file (aka first line).
I ran the bash script below to convert from UTF-8 with BOM to standard UTF-8.

```bash
# Find all files ending in .cs in project
find . -name "*.cs" | while read file; do
    # Remove BOM markers if exist
    sed -i '1s/^\xEF\xBB\xBF//' $file
done
```

## Merge conflicts with existing PRs
None*:
1. Files that the user has converted to NOBOM according to the linter have no conflicts
2. Files that the user has not converted will have no conflict

*would have conflict with files converted to UTF-16 or UTF-32 due to their boms, but we should have none of these


JoannaaKL (approved):
lgtm and thanks for looking into this!

PR #2643, opened 2023-06-03
Unlock Node.js 18 & 20

smorimoto:
In this PR, we will unlock Node.js 18 and 20, but there are two key points here.
The first is that, until now, we've been using a special build from the Node.js team for ARM Windows, but starting from 20, builds for ARM exist by default. So if you don't bump the 16 version, we have to make a special arm build just for 18, which is already out of support. I've asked the Node.js team to help us by providing special builds.
Second, the build for Alpine used to provide a special build from the actions team via blob storage, but that can be resolved using the musl build from the Node.js team.
You're starting to see a point to discuss here, as well. That is, how do you go about Node.js 18? I don't see any good reason to include it. The build for ARM Windows is missing, and 16 and 20 are enough. Doing so can reduce both the scripts and the work in the unofficial build repository. This makes a lot of sense.

### DIFF
--- a/src/Misc/externals.sh
+++ b/src/Misc/externals.sh
@@ -4,8 +4,9 @@ PRECACHE=$2
 
 NODE_URL=https://nodejs.org/dist
 UNOFFICIAL_NODE_URL=https://unofficial-builds.nodejs.org/download/release
-NODE12_VERSION="12.22.7"
-NODE16_VERSION="16.16.0"
+NODE16_VERSION="16.20.0"
+NODE18_VERSION="18.16.0"
+NODE20_VERSION="20.2.0"
 
 get_abs_path() {
   # exploits the fact that pwd will print abs path when no args
@@ -137,10 +138,12 @@ function acquireExternalTool() {
 
 # Download the external tools only for Windows.
 if [[ "$PACKAGERUNTIME" == "win-x64" || "$PACKAGERUNTIME" == "win-x86" ]]; then
-    acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/$PACKAGERUNTIME/node.exe" node12/bin
-    acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/$PACKAGERUNTIME/node.lib" node12/bin
     acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.exe" node16/bin
     acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.lib" node16/bin
+    acquireExternalTool "$NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.exe" node18/bin
+    acquireExternalTool "$NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.lib" node18/bin
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/$PACKAGERUNTIME/node.exe" node20/bin
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/$PACKAGERUNTIME/node.lib" node20/bin
     if [[ "$PRECACHE" != "" ]]; then
         acquireExternalTool "https://github.com/microsoft/vswhere/releases/download/2.6.7/vswhere.exe" vswhere
     fi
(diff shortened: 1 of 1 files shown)


src/Misc/externals.sh
```diff
@@ -151,36 +154,46 @@ if [[ "$PACKAGERUNTIME" == "win-arm64" ]]; then
     # todo: replace these with official release when available
     acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.exe" node16/bin
     acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.lib" node16/bin
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.exe" node18/bin
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.lib" node18/bin
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/$PACKAGERUNTIME/node.exe" node20/bin
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/$PACKAGERUNTIME/node.lib" node20/bin
```
smorimoto, line 160: 🎉 https://nodejs.org/dist/v20.2.0/win-arm64/

src/Misc/externals.sh
```diff
@@ -151,36 +154,46 @@ if [[ "$PACKAGERUNTIME" == "win-arm64" ]]; then
...
-    acquireExternalTool "$NODE_URL/v${NODE12_VERSION}/node-v${NODE12_VERSION}-linux-x64.tar.gz" node12 fix_nested_dir
-    acquireExternalTool "https://vstsagenttools.blob.core.windows.net/tools/nodejs/${NODE12_VERSION}/alpine/x64/node-v${NODE12_VERSION}-alpine-x64.tar.gz" node12_alpine
     acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/node-v${NODE16_VERSION}-linux-x64.tar.gz" node16 fix_nested_dir
-    acquireExternalTool "https://vstsagenttools.blob.core.windows.net/tools/nodejs/${NODE16_VERSION}/alpine/x64/node-v${NODE16_VERSION}-alpine-x64.tar.gz" node16_alpine
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE16_VERSION}/node-v${NODE16_VERSION}-linux-x64-musl.tar.gz" node16_alpine
+    acquireExternalTool "$NODE_URL/v${NODE18_VERSION}/node-v${NODE18_VERSION}-linux-x64.tar.gz" node18 fix_nested_dir
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE18_VERSION}/node-v${NODE18_VERSION}-linux-x64-musl.tar.gz" node18_alpine
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/node-v${NODE20_VERSION}-linux-x64.tar.gz" node20 fix_nested_dir
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE20_VERSION}/node-v${NODE20_VERSION}-linux-x64-musl.tar.gz" node20_alpine
```
smorimoto, line 186: The official Node.js Alpine image uses the unofficial musl build. This is much more reliable than the in-house build.

src/Misc/externals.sh
```diff
@@ -151,36 +154,46 @@ if [[ "$PACKAGERUNTIME" == "win-arm64" ]]; then
...
     acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/node-v${NODE16_VERSION}-darwin-x64.tar.gz" node16 fix_nested_dir
+    acquireExternalTool "$NODE_URL/v${NODE18_VERSION}/node-v${NODE18_VERSION}-darwin-x64.tar.gz" node18 fix_nested_dir
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/node-v${NODE20_VERSION}-darwin-x64.tar.gz" node20 fix_nested_dir
 fi
 
 if [[ "$PACKAGERUNTIME" == "osx-arm64" ]]; then
-    # node.js v12 doesn't support macOS on arm64.
     acquireExternalTool "$NODE_URL/v${NODE16_VERSION}/node-v${NODE16_VERSION}-darwin-arm64.tar.gz" node16 fix_nested_dir
+    acquireExternalTool "$NODE_URL/v${NODE18_VERSION}/node-v${NODE18_VERSION}-darwin-arm64.tar.gz" node18 fix_nested_dir
+    acquireExternalTool "$NODE_URL/v${NODE20_VERSION}/node-v${NODE20_VERSION}-darwin-arm64.tar.gz" node20 fix_nested_dir
```
smorimoto, line 176: 🎉 https://nodejs.org/dist/v18.16.0/

src/Misc/externals.sh
```diff
@@ -151,36 +154,46 @@ if [[ "$PACKAGERUNTIME" == "win-arm64" ]]; then
     # todo: replace these with official release when available
     acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.exe" node16/bin
     acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE16_VERSION}/$PACKAGERUNTIME/node.lib" node16/bin
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.exe" node18/bin
+    acquireExternalTool "$UNOFFICIAL_NODE_URL/v${NODE18_VERSION}/$PACKAGERUNTIME/node.lib" node18/bin
```
smorimoto, line 158: NOTE: They do NOT yet exist. For 16, we can avoid the problem by keeping the version.
Consult here: https://github.com/nodejs/unofficial-builds/issues/71

smorimoto: Published!

smorimoto:
We are now fully ready to go ahead!

domenkozar:
This needs to happen since nodes 16 is insecure.

smorimoto:
@domenkozar I have to agree with that, but the Actions team seems less willing to accept such a change 😮‍💨

RaitoBezarius:
Thank you for your changes @smorimoto -- it's awesome! We will probably take this patch home in nixpkgs for our packages.

takost:
JFYI, @smorimoto  latest runner release ([2.308.0](https://github.com/actions/runner/releases/tag/v2.308.0)) contains latest node16 version and node20 version

Obagoal1 (approved):
Next play online sportybet aviator next

PR #2650, opened 2023-06-07
Send environment url to Run Service

heavymachinery:
Related issue: https://github.com/github/actions-service-team/issues/3857

Since the environment url is evaluated on the runner, we need to send the final evaluated url to Run Service once the job is complete.


ericsciple (dismissed):

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -1177,7 +1177,7 @@ private async Task ForceFailJob(IRunnerService server, Pipelines.AgentJobRequest
                         jobAnnotations.Add(unhandledAnnotation.Value);
                     }
 
-                    await runServer.CompleteJobAsync(message.Plan.PlanId, message.JobId, TaskResult.Failed, outputs: null, stepResults: null, jobAnnotations: jobAnnotations, CancellationToken.None);
+                    await runServer.CompleteJobAsync(message.Plan.PlanId, message.JobId, TaskResult.Failed, outputs: null, stepResults: null, jobAnnotations: jobAnnotations, environmentUrl: "", CancellationToken.None);
```
TingluoHuang, line 1180:
```suggestion
                    await runServer.CompleteJobAsync(message.Plan.PlanId, message.JobId, TaskResult.Failed, outputs: null, stepResults: null, jobAnnotations: jobAnnotations, environmentUrl: null, CancellationToken.None);
```

src/Runner.Worker/JobRunner.cs
```diff
@@ -283,14 +284,21 @@ private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecution
             // Make sure we don't submit secrets as telemetry
             MaskTelemetrySecrets(jobContext.Global.JobTelemetry);
 
+            // Get environment url
+            var environmentUrl = "";
+            if (jobContext?.ActionsEnvironment?.Url is StringToken urlStringToken)
```
TingluoHuang, line 289:
```suggestion
            if (jobContext.ActionsEnvironment?.Url is StringToken urlStringToken)
```

`jobContext` should never be null

src/Runner.Worker/JobRunner.cs
```diff
@@ -283,14 +284,21 @@ private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecution
             // Make sure we don't submit secrets as telemetry
             MaskTelemetrySecrets(jobContext.Global.JobTelemetry);
 
+            // Get environment url
+            var environmentUrl = "";
```
TingluoHuang, line 288:
```suggestion
            string environmentUrl = null;
```

TingluoHuang (approved):

PR #2656, opened 2023-06-13
Switch TemplateTraceWriter on ExecutionContext to not create error annotations on TraceError

JoannaaKL:
More details here
https://github.com/actions/runner/pull/2637

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -1401,7 +1401,15 @@ namespace GitHub.Runner.Worker
 
         public void Error(string format, params Object[] args)
         {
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1414,8 +1422,16 @@ public void Verbose(string format, params Object[] args)
             // todo: switch to verbose?
             _executionContext.Debug(string.Format(CultureInfo.CurrentCulture, $"{format}", args));
         }
+
+        private bool logTemplateErrorsAsDebugMessages()
+        {
+            if (_executionContext.Global == null || _executionContext.Global.EnvironmentVariables == null) return false;
```
TingluoHuang, line 1428: will these ever be NULL?

TingluoHuang (approved):

PR #2660, opened 2023-06-15
Reduce token service and unnecessary calls - send token to redirects

yaananth:
## Before

...
- Call to pipelines messages with a valid token
- 302 to broker
- Call to broker with OUT any token (default `AllowAutoRedirect` behavior)
- Call to token service to get a token
- Call to broker with a valid token
- after 30 seconds, long poll terminates
- 🔁
...

## Now

...
- Call to pipelines messages with a valid token
- 302 to broker
- Call to broker with a valid token
- after 30 seconds, long poll terminates
- 🔁
...

### DIFF
--- a/src/Runner.Sdk/Util/VssUtil.cs
+++ b/src/Runner.Sdk/Util/VssUtil.cs
@@ -61,7 +61,7 @@ namespace GitHub.Runner.Sdk
                 // prefer environment variable
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
-            
+
             if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("USE_BROKER_FLOW")))
             {
                 settings.AllowAutoRedirectForBroker = true;
--- a/src/Runner.Sdk/Util/VssUtil.cs
+++ b/src/Runner.Sdk/Util/VssUtil.cs
@@ -62,11 +62,6 @@ namespace GitHub.Runner.Sdk
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
 
-            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("USE_BROKER_FLOW")))
-            {
-                settings.AllowAutoRedirect = true;
-            }
-
             // Remove Invariant from the list of accepted languages.
             //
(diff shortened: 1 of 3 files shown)


src/Runner.Sdk/Util/VssUtil.cs
```diff
@@ -62,11 +62,6 @@ public static void InitializeVssClientSettings(List<ProductInfoHeaderValue> addi
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
 
-            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("USE_BROKER_FLOW")))
```
yaananth, line 65: we are taking over here, so disabling default behavior

src/Sdk/Common/Common/VssHttpMessageHandler.cs
```diff
@@ -211,14 +211,34 @@ protected override void Dispose(Boolean disposing)
 
                     traceInfo?.TraceBufferedRequestTime();
 
-                    // ConfigureAwait(false) enables the continuation to be run outside any captured 
+                    // ConfigureAwait(false) enables the continuation to be run outside any captured
                     // SyncronizationContext (such as ASP.NET's) which keeps things from deadlocking...
-                    response = await m_messageInvoker.SendAsync(request, tokenSource.Token).ConfigureAwait(false);
+
+                    var tmpResponse = await m_messageInvoker.SendAsync(request, tokenSource.Token).ConfigureAwait(false);
```
TingluoHuang, line 217: can we run this only under 4-nine?

src/Sdk/Common/Common/VssHttpMessageHandler.cs
```diff
@@ -211,14 +211,34 @@ protected override void Dispose(Boolean disposing)
...
+
+                    if (tmpResponse.StatusCode == HttpStatusCode.Redirect)
```
TingluoHuang, line 219: i would suggest we don't change the default redirect behavior on the `VssHttpMessageHandler` since the handler is used for all outbound Http request from runner, not only to broker.

Consider uncontrolled network environment like GHES, I would keep the redirect behavior as close as to the C# default which is clear out auth header and require re-auth.

src/Sdk/Common/Common/VssHttpMessageHandler.cs
yaananth: done

TingluoHuang (approved):

PR #2668, opened 2023-06-22
Fix error message reported on non-local action setup

### LINKED ISSUE #2513: Message is wrong for non local actions: `Did you forget to run actions/checkout before running your local action?"`
https://github.com/actions/runner/blob/1ceb1a67f270537fe80da23445fa78bfb97503ea/src/Runner.Worker/ActionManager.cs#L510

I'm _pretty_ sure that the runner knows the difference between a local action and a non-local action.

Steps:
1. Create a https://github.com/check-spelling/redpanda/actions/runs/4576350730/workflow that points to `actions/cache/restore@v2` instead of `actions/cache/restore@v3`
2. Trigger a run: https://github.com/check-spelling/redpanda/actions/runs/4576350730/jobs/8080370474

For non-local actions, something like:
```c#
 throw new NotSupportedException($"Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '{fullPath}'. Did you specify the wrong repository/branch/tag/path?"); 
```

Seems like a better message.

nikola-jokic:
Fixes #2513

The PrepareRepositoryActionAsync stops if the action is local (self alias).

So the error message was misleading that the checkout is needed even though the action is not related to this repository.

The result of this change would look something like this:
image

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -1136,7 +1136,7 @@ namespace GitHub.Runner.Worker
                 {
                     reference = $"{reference}@{repositoryReference.Ref}";
                 }
-                throw new InvalidOperationException($"Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action '{reference.ToString()}'.");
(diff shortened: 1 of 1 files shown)


fhammerl (approved):
LGTM, will merge for next release version

PR #2678, opened 2023-07-04
Add warning to notify about forcing node16 

takost:
Add warning message to notify that Action that uses node12 was forced to run on node16.

Screen Shot 2023-07-06 at 17 11 39 PM

### DIFF
--- a/src/Runner.Worker/Handlers/HandlerFactory.cs
+++ b/src/Runner.Worker/Handlers/HandlerFactory.cs
@@ -72,7 +72,7 @@ namespace GitHub.Runner.Worker.Handlers
                         if (repoAction != null)
                         {
                             var warningActions = new HashSet<string>();
-                            if (executionContext.Global.Variables.TryGetValue("Node16ForceActionsWarnings", out var node16ForceWarnings))
+                            if (executionContext.Global.Variables.TryGetValue(Constants.Runner.EnforcedNode12DetectedAfterEndOfLifeEnvVariable, out var node16ForceWarnings))
                             {
                                 warningActions = StringUtil.ConvertFromJson<HashSet<string>>(node16ForceWarnings);
                             }
--- a/src/Runner.Worker/Handlers/HandlerFactory.cs
+++ b/src/Runner.Worker/Handlers/HandlerFactory.cs
@@ -67,11 +67,17 @@ namespace GitHub.Runner.Worker.Handlers
                     var isLocalOptOut = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.AllowActionsUseUnsecureNodeVersion));
                     bool isOptOut = isWorkflowOptOutSet ? StringUtil.ConvertToBoolean(workflowOptOut) : isLocalOptOut;
                     if (!isOptOut)
-                    {   
+                    {
                         var repoAction = action as Pipelines.RepositoryPathReference;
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -67,7 +67,21 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                            if (string.IsNullOrEmpty(repoAction.Name))
+                            {
+                                repoActionFullName = repoAction.Path; // local actions don't have a 'Name'
+                            }
+                            else
+                            {
+                                repoActionFullName = $"{repoAction.Name}/{repoAction.Path ?? string.Empty}".TrimEnd('/') + $"@{repoAction.Ref}";
+                            }
+                        }
+                        executionContext.Warning(string.Format(Constants.Runner.EnforcedNode12DetectedAfterEndOfLife, repoActionFullName));
```
jww3, line 84: `repoActionFullName` may still be empty here.

The output will appear as:

```
The action  uses node12 which is deprecated and will be forced to run on node16. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/"
```

Is that OK?

fhammerl: Is that the case of running a local action like this?
```yaml
#...someyaml
uses: ./
```

takost: I decided to aggregate warnings from all steps to one warning. I attached screenshot with new warning and this example includes local action

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -68,6 +68,28 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
                     bool isOptOut = isWorkflowOptOutSet ? StringUtil.ConvertToBoolean(workflowOptOut) : isLocalOptOut;
                     if (!isOptOut)
                     {
+                        var repoAction = action as Pipelines.RepositoryPathReference;
+                        if (repoAction != null)
+                        {
+                            var warningActions = new HashSet<string>();
+                            if (executionContext.Global.Variables.TryGetValue("Node16ForceActionsWarnings", out var node16ForceWarnings))
```
JoannaaKL, line 75: Could we extract string `Node16ForceActionsWarnings` to constants? :) https://github.com/actions/runner/blob/d4bbbb841987d6c2d4830b05e124735ac1f6d6be/src/Runner.Common/Constants.cs#L157

src/Runner.Worker/JobRunner.cs
```diff
@@ -374,6 +374,12 @@ private async Task<TaskResult> CompleteJobAsync(IJobServer jobServer, IExecution
                 jobContext.Warning(string.Format(Constants.Runner.Node12DetectedAfterEndOfLife, actions));
             }
 
+            if (jobContext.Global.Variables.TryGetValue("Node16ForceActionsWarnings", out var node16ForceWarnings))
```
JoannaaKL, line 377: Same here, let's use constants :)

fhammerl (approved):
LGTM

domenkozar:
Note that node16 has almost reached [EOL](https://nodejs.org/en/blog/announcements/nodejs16-eol).

PR #2683, opened 2023-07-07
Calculate docker instance label based on the hash of the config

### LINKED ISSUE #2647: network label same for each workflows
**Describe the bug**
Hi
When I launch 2 workflows at the same time that use service container (to run postgres, redis an rabbitmq for example), they have the same label. 
As you can see, the 1st workflow creates a network with this label : 
```
Create local container network
  /usr/bin/docker network create --label c3f261 github_network_02a5eecd05de4c4f
```
But the second one is going to destroy the first : 
```
Clean up resources from previous jobs
  /usr/bin/docker ps --all --quiet --no-trunc --filter "label=c3f261"
  /usr/bin/docker network prune --force --filter "label=c3f261"
  Deleted Networks:
  github_network_02a5eecd05de4c4f
```
**To Reproduce**
1. In my case, when I launch two workflows at the same time, the second one always destroy the first 

**Expected behavior**
How to modify this label so the workflows can be no dependant ? 

My runners are currently hosted on an ECS cluster 

## What's not working?
The first job : 
image

The second job that destroys the first one : 
image

nikola-jokic:
Fixes #2647

The issue can occur when multiple runners are sharing the same docker socket, and they are executing within a container.
The well known directory is going to be the same for both runners, assuming they are executing the same image.

So when docker instance is calculated, the paths within the container are going to be exactly the same, one runner can remove resources owned by the other runner.

The fix uses the `.runner` file, which is unique per runner to calculate instance label, which should eliminate collision issues.


fhammerl (approved):
LGTM

This will make the docker instance label unique even if the runners sharing the same docker socket through docker-from-docker are under the same path in their respective containers.

Hashing the entire .runner file means that changes to this file (e.g. runnergroupname) will be reflected as a change in docker instance label as well - we found this acceptable as this change would happen outside of a running job, which means the docker network created based on this label in a previous job has been disposed.

PR #2694, opened 2023-07-13
Update Docker Version in Images

ajschmidt8:
The Docker version that's currently being used by the images is quite a bit out of date.

This PR updates the Docker version to the latest version.

Release notes are linked below for reference:

- https://docs.docker.com/engine/release-notes/24.0/
- https://docs.docker.com/engine/release-notes/23.0/
- https://docs.docker.com/engine/release-notes/20.10/


ajschmidt8:
@TingluoHuang, @fhammerl, @takost, any thoughts on this change? Can this be merged?

images/Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG RUNNER_VERSION
 ARG RUNNER_CONTAINER_HOOKS_VERSION=0.3.2
-ARG DOCKER_VERSION=20.10.23
+ARG DOCKER_VERSION=24.0.5
```
TingluoHuang, line 8: i would suggest we go to `23.0.6` which matches the hosted runner image https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#tools

ajschmidt8: @TingluoHuang, updated!

TingluoHuang (approved):

PR #2705, opened 2023-07-21
Draft: Create host directories when mounting container volumes

olemorud:
Relying on Docker to create `hostpath` when using the flag `-v hostpath:containerpath` ~~is undocumented behavior and~~ is not compatible with other containerization tools that people might want to use (even if they aren't explicitly supported).

### DIFF
--- a/src/Runner.Worker/ContainerOperationProvider.cs
+++ b/src/Runner.Worker/ContainerOperationProvider.cs
@@ -297,15 +297,28 @@ namespace GitHub.Runner.Worker
             ArgUtil.NotNull(githubContext, nameof(githubContext));
             var workingDirectory = githubContext["workspace"] as StringContextData;
             ArgUtil.NotNullOrEmpty(workingDirectory, nameof(workingDirectory));
-            container.MountVolumes.Add(new MountVolume(HostContext.GetDirectory(WellKnownDirectory.Work), container.TranslateToContainerPath(HostContext.GetDirectory(WellKnownDirectory.Work))));
+
+            var directories = new List<String> {
+                WellKnownDirectory.Work,
+                WellKnownDirectory.Temp,
+                WellKnownDirectory.Actions,
+                WellKnownDirectory.Tools,
+            };
+
+            foreach (var d in directories)
+            {
+                var dir = HostContext.GetDirectory(d);
+                // Not creating the host path for a volume is an undocumented bug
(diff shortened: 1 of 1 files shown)


olemorud:
@ChristopherHX 
> But it is documented here by docker: https://docs.docker.com/engine/reference/commandline/run/#volume
> 
> > When the host directory of a bind-mounted volume doesn’t exist, Docker automatically creates this directory on the host for you. In the example above, Docker creates the /doesnt/exist folder before starting your container.

You are right 😅 Thanks for pointing this out.

Creating directories is not supported by Podman however (by design), making the setup of self-hosted runners on any RHEL-derivative more painful. Podman is also rootless by default and is a much safer option for most self-hosted runners. This PR fixes the only annoying compatibility issue without introducing any breaking changes for Docker users (directories will still be created as usual)

> You can use https://github.com/actions/runner-container-hooks to customize how containers are managed by the actions/runner while keeping actions/runner unchanged.
> 
> For example you have to already provide a symlink/wrapper or it wouldn't even try to use your unspecified container runtime.

You can just symlink `/usr/bin/docker` to whatever container runtime you want as long as the interface is compatible and not touch hooks at all. In this case you need to symlink it to a wrapper script that creates the missing directories before passing the arguments onto Podman, which is a trivial but annoying workaround (around 15 lines of python).

I am not sure how this would be easier to implement in a `run_container_step` hook, where in the examples it seems that you have to write and transpile several hundred lines of typescript to modify the container behavior.

nathanweeks:
I've temporarily coded around this in the runner by overwriting the /usr/bin/docker script (installed by podman-docker, which internally calls podman) with a version that adds the relevant mkdir commands, though ideally each development team that uses rootless podman runners for enhanced security wouldn't have to independently discover and code around this issue. Creating the non-existent host directories beforehand as proposed by this PR seems like a reasonable solution that shouldn't cause issues with runners that use real Docker. 

The docker `-v` semantics of creating any non-existent host directories were recognized as a misfeature and briefly deprecated, but subsequently un-deprecated to preserve backwards compatibility, though it was at least not carried forward to the  newer`--mount type=bind` syntax (https://github.com/moby/moby/issues/13121).

The podman project maintainers have declared their intent not to emulate `docker -v` just because GitHub Actions expects it (https://github.com/containers/podman/issues/6234#issuecomment-1644804689).

PR #2707, opened 2023-07-21
Update heredoc detection to add additional edge cases

benwells:
The following examples should all detected as heredoc

1. "=" at the end of the key

```
test=<<EOF
hello world
EOF
```
2. any number of spaces between the equals-containing key and <<

```
test=      <<EOF
hello world
EOF
```

3. case where there the delimeter itself has an equals (think base64 encoded random string)

```
MY_KEY<<sdfsdf=
hellow world
sdfsdf=
```

We also want to correctly detect "normal" single line key/value style where "<<" is in the value. 

`stdout=hi there << hello world`

### DIFF
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -396,7 +396,7 @@ namespace GitHub.Runner.Worker
         }
 
         // accepts a string, two indexes, and returns true if only whitespace chars exist between those two indexes (non-inclusive)
-        private static bool CheckIfOnlyWhitespaceBetweenPositions(string str, int pos1, int pos2)
+        private static bool OnlyContainsWhiteSpaceBetweenPositions(string str, int pos1, int pos2)
         {
             // Check if the provided positions are valid
             if (pos1 < 0 || pos2 < 0 || pos1 >= str.Length || pos2 >= str.Length)
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -329,11 +329,11 @@ namespace GitHub.Runner.Worker
                     var isHeredoc = heredocIndex >= 0 &&
                     (
                         equalsIndex < 0 ||
-                        
+                        heredocIndex < equalsIndex ||
                         (
                             heredocIndex > equalsIndex &&
                             OnlyContainsWhiteSpaceBetweenPositions(line, equalsIndex, heredocIndex)
-                        ) || heredocIndex < equalsIndex
+                        ) 
                     );
                     
                     if (isHeredoc)
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -396,7 +396,7 @@ namespace GitHub.Runner.Worker
         }
 
         // accepts a string, two indexes, and returns true if only whitespace chars exist between those two indexes (non-inclusive)
-        private static bool CheckIfOnlyWhitespaceBetweenSymbols(string str, int pos1, int pos2)
+        private static bool CheckIfOnlyWhitespaceBetweenPositions(string str, int pos1, int pos2)
         {
             // Check if the provided positions are valid
             if (pos1 < 0 || pos2 < 0 || pos1 >= str.Length || pos2 >= str.Length)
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -383,6 +395,34 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
...
+                throw new ArgumentOutOfRangeException("Whitespace check: Positions should be within the string length.");
+            }
+
+            // Ensure pos1 is always the smaller position
+            if (pos2 < pos1)
+            {
+                int temp = pos2;
+                pos2 = pos1;
+                pos1 = temp;
+            }
```
cory-miller, line 413: I am not sure we should flip these. If equalsIndex is after heredocIndex, then we have a string like:

```
FOO << =EOF
abc
=EOF
```

I think we can just check `if pos2 < pos1` and throw an argument exception.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -322,9 +322,21 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
...
+                    // "key=<<EOF is also considered heredoc where the key contains an =.
+                    // Also we detect heredoc if whitespace chars exist between = and <<. (ex: "key      <<EOF")
+                    var isHeredoc = heredocIndex >= 0 &&
+                    (
+                        equalsIndex < 0 ||
+                        (
+                            heredocIndex > equalsIndex &&
+                            OnlyContainsWhiteSpaceBetweenPositions(line, equalsIndex, heredocIndex)
+                        ) ||
+                        heredocIndex < equalsIndex
```
cory-miller, line 336: To go along with my other comment about flipping indices, I think we can flip these to short circuit the case where heredocIndex is before equalsIndex (meaning equals was possibly part of the delimiter).
```suggestion
                        heredocIndex < equalsIndex ||
                        (
                            heredocIndex > equalsIndex &&
                            OnlyContainsWhiteSpaceBetweenPositions(line, equalsIndex, heredocIndex)
                        )
```
```suggestion
                        (
                            heredocIndex > equalsIndex &&
                            OnlyContainsWhiteSpaceBetweenPositions(line, equalsIndex, heredocIndex)
                        ) ||
                        heredocIndex < equalsIndex
```

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -383,6 +395,34 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
             return GetEnumerator();
         }
 
+        // accepts a string, two indexes, and returns true if only whitespace chars exist between those two indexes (non-inclusive)
+        private static bool CheckIfOnlyWhitespaceBetweenPositions(string str, int pos1, int pos2)
```
cory-miller, line 399:
```suggestion
        private static bool OnlyContainsWhiteSpaceBetweenPositions(string str, int pos1, int pos2)
```

src/Test/L0/Worker/FileCommandTestBase.cs
```diff
@@ -238,6 +238,10 @@ protected void TestHeredoc_EdgeCases()
                     "MY_KEY_4<<EOF",
                     "EOF EOF",
                     "EOF",
+                    "MY_KEY_5=abc << def",
+                    "MY_KEY_6=    <<EOF",
+                    "white space test",
+                    "EOF"
```
cory-miller, line 244:
```suggestion
                    "EOF"
                    "MY_KEY_7 << =EOF="
                    "abc"
                    "=EOF="
```

src/Test/L0/Worker/FileCommandTestBase.cs
```diff
@@ -247,6 +251,8 @@ protected void TestHeredoc_EdgeCases()
                 Assert.Equal($"hello=two", _store["MY_KEY_2"]);
                 Assert.Equal($" EOF", _store["MY_KEY_3"]);
                 Assert.Equal($"EOF EOF", _store["MY_KEY_4"]);
+                Assert.Equal($"abc << def", _store["MY_KEY_5"]);
+                Assert.Equal($"white space test", _store["MY_KEY_6="]);
```
cory-miller, line 255:
```suggestion
                Assert.Equal($"white space test", _store["MY_KEY_6="]);
                Assert.Equal($"abc", _store["MY_KEY_7"]);
```

PR #2713, opened 2023-07-24
Allow retry if Runner sees 404 when downloading actions

yacaovsnc:
Sometimes we get 404s when the token's permission is not yet replicated fully.  The 404 maybe temporary and we should retry in those cases.

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -848,8 +848,15 @@ namespace GitHub.Runner.Worker
                                     }
                                     else if (response.StatusCode == HttpStatusCode.NotFound)
                                     {
-                                        // It doesn't make sense to retry in this case, so just stop
-                                        throw new ActionNotFoundException(new Uri(link));
+                                        if (retryCount == 2)
+                                        {
+                                            throw new ActionNotFoundException(new Uri(link));
+                                        }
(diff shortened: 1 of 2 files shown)


TingluoHuang:
Should we fix the Token? 😄 4XX means client side error, why retry on those?

src/Runner.Worker/ActionManager.cs
```diff
@@ -848,8 +848,15 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
                                     }
                                     else if (response.StatusCode == HttpStatusCode.NotFound)
                                     {
-                                        // It doesn't make sense to retry in this case, so just stop
-                                        throw new ActionNotFoundException(new Uri(link));
+                                        if (retryCount == 2)
```
TingluoHuang, line 851:
```
else if (response.StatusCode == HttpStatusCode.NotFound && retryCount >=2)
{
    // It doesn't make sense to retry in this case, so just stop
    throw new ActionNotFoundException(new Uri(link));
}
```

bryanmacfarlane:
yeah, this seems wrong to retry

bryanmacfarlane:
if the service is unreliable then lets fix that.  404 means it was conclusively not found.  it's not an error.

PR #2714, opened 2023-07-25
Pass timeout in executioncontext instead of stepsrunner

fhammerl:


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -421,8 +421,6 @@ private async Task RunStepAsync(IStep step)
         {
             Trace.Info($"Starting: {step.DisplayName}");
             step.ExecutionContext.Debug($"Starting: {step.DisplayName}");
-            // composite steps inherit the timeout from the parent, set by https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes
-            step.ExecutionContext.SetTimeout(step.ExecutionContext.Parent.GetRemainingTimeout());
```
fhammerl, line 425: In previous fix I treated `step.ExecutionContext.Parent.ExecutionContext` as the EC of workflow or composite step creating `step`, but it is actually the EC of the job because comp steps are created by `Root`

nikola-jokic (approved):

JoannaaKL (dismissed):

nikola-jokic (approved):

PR #2717, opened 2023-07-26
Remove node12

takost:
This PR removes node12 from runners. 
We already forced all actions to be run on node16, but we had logic that allowed to opt-out this enforcement and continue execution on node12. This change removes node12 and the opt-out logic, so all actions will be executed on node16


TingluoHuang (approved):

src/Runner.Common/Constants.cs
```diff
@@ -171,7 +171,6 @@ public static class Features
...
             public static readonly string UnsupportedSummarySize = "$GITHUB_STEP_SUMMARY upload aborted, supports content up to a size of {0}k, got {1}k. For more information see: https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions#adding-a-markdown-summary";
             public static readonly string SummaryUploadError = "$GITHUB_STEP_SUMMARY upload aborted, an error occurred when uploading the summary. For more information see: https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions#adding-a-markdown-summary";
-            public static readonly string Node12DetectedAfterEndOfLife = "Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: {0}. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.";
```
TingluoHuang, line 174: do we want to replace all Node12xxxx with Node16?

takost: Yes, but as we will work on node20 next, I decided that it will be easier to clean up code now and add the warning regarding node16 later.

JoannaaKL (approved):

AvaStancu (approved):

PR #2726, opened 2023-07-31
Expose results service endpoint as environment variable

robherley:
We'll need to expose the results service public endpoint for upcoming work with toolkit and eventually the upload/download artifact actions.


robherley:
🤔 Linter is complaining about file encoding

image

Wonder how that happened lol

TingluoHuang (approved):

PR #2727, opened 2023-08-01
Update HTTPEventSourceListener to trace the right events.

TingluoHuang:
Dotnet core changed their network tracing event source name as part of dotnet core 5.0

https://github.com/dotnet/runtime/issues/37428

We need to react these changes otherwise we are not going to get any trace from the network layer from dotnet core.

TLDR:

`Microsoft-System-Net-Http` -> `System.Net.Http` and `Private.InternalDiagnostics.System.Net.Http`
`Microsoft-System-Net-Security` -> `System.Net.Security` and `Private.InternalDiagnostics.System.Net.Security`

I noticed this when check support issue https://github.com/github/c2c-actions-support/issues/2691

### DIFF
--- a/src/Runner.Listener/Checks/CheckUtil.cs
+++ b/src/Runner.Listener/Checks/CheckUtil.cs
@@ -351,21 +351,39 @@ namespace GitHub.Runner.Listener.Check
         private readonly Dictionary<string, HashSet<string>> _ignoredEvent = new()
         {
             {
-                "Microsoft-System-Net-Http",
+                "System.Net.Http",
                 new HashSet<string>
                 {
                     "Info",
                     "Associate",
-                    "Enter",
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Checks/CheckUtil.cs
```diff
@@ -351,21 +351,39 @@ public sealed class HttpEventSourceListener : EventListener
...
         {
             {
-                "Microsoft-System-Net-Http",
+                "System.Net.Http",
                 new HashSet<string>
                 {
                     "Info",
                     "Associate",
-                    "Enter",
-                    "Exit"
```
TingluoHuang, line 360: these 2 events no longer exists in the newer version dotnet.

PR #2771, opened 2023-08-16
Query runner by name on server side.

TingluoHuang:
Once the runner-admin service support query runner by name, we can move the query to use server side filter instead of fetching all the runners from the server and do the filter on the client side.


vanZeben (approved):
LGTM

PR #2775, opened 2023-08-16
Filter NODE_OPTIONS from env for file output

cory-miller:
This matches the deprecated command equivalent as seen [here](https://github.com/actions/runner/blob/1183100ab8a6af99165a383ec79ff6c5fc261bb2/src/Runner.Worker/ActionCommandManager.cs#L268-L283).


src/Runner.Worker/FileCommandManager.cs
```diff
@@ -322,9 +349,21 @@ public EnvFileKeyValuePairs(IExecutionContext context, string filePath)
                     var equalsIndex = line.IndexOf("=", StringComparison.Ordinal);
                     var heredocIndex = line.IndexOf("<<", StringComparison.Ordinal);
 
-                    // Heredoc style NAME<<EOF (where EOF is typically randomly-generated Base64 and may include an '=' character)
-                    // see https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings
-                    if (heredocIndex >= 0 && (equalsIndex < 0 || heredocIndex < equalsIndex))
+                    // Normal style NAME=VALUE
```
TingluoHuang, line 352: what are these changes?

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -275,10 +275,37 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                var isBlocked = false;
+                foreach (var blocked in _setEnvBlockList)
+                {
+                    if (string.Equals(blocked, pair.Key, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Log Telemetry and let user know they shouldn't do this
+                        var issue = new Issue()
+                        {
+                            Type = IssueType.Error,
+                            Message = $"Can't update {blocked} environment variable using ::set-env:: command."
```
TingluoHuang, line 287: might want to update the error message since we are not doing `::set-env::`

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -275,10 +275,37 @@ public void ProcessCommand(IExecutionContext context, string filePath, Container
...
+                var isBlocked = false;
+                foreach (var blocked in _setEnvBlockList)
+                {
+                    if (string.Equals(blocked, pair.Key, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Log Telemetry and let user know they shouldn't do this
+                        var issue = new Issue()
+                        {
+                            Type = IssueType.Error,
+                            Message = $"Can't store {blocked} output parameter using GITHUB_OUTPUT command."
```
TingluoHuang, line 287:
```suggestion
                            Message = $"Can't store {blocked} output parameter using $GITHUB_ENV file."
```

src/Test/L0/Worker/SetEnvFileCommandL0.cs
```diff
@@ -183,6 +183,47 @@ public void SetEnvFileCommand_Simple_SpecialCharacters()
             }
         }
 
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void SetOutputFileCommand_BlockListItemsFiltered()
```
TingluoHuang, line 190: do you need to update the test to verify the ENV instead of output?

cory-miller: Fixed the test name. The test body used `_setEnvFileCommand.ProcessCommand`

TingluoHuang (approved):

rymndhng:
@cory-miller @TingluoHuang I started seeing these errors on our worker (we set NODE_OPTIONS by writing to $GITHUB_ENV). 

Can you clarify why setting this envvar is not allowed?

synaptiko:
@cory-miller Our whole CI pipeline is failing since today morning due to this. We need to increase Node.js's memory and without it our commands our failing on OOMs.

Up until this point setting `NODE_OPTIONS` worked fine. We use this:
```
echo "NODE_OPTIONS=--max-old-space-size=${MAX_OLDSPACE_SIZE_MB}" >> $GITHUB_ENV
```

I checked both PRs related to blocking `NODE_OPTIONS` but there is no explanation why it's blocked and what is an alternative way to set it. Can you please shed some light on this topic? I tried to google it as well but no luck.

The problem is that our deployments are blocked by this quite hidden change. And it took us a while to find this PR as well.

There is no note about `NODE_OPTIONS` being special in the official docs either: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

cc @pleszkowicz @icaliskanoglu

cory-miller:
For the deprecated `set-env` command we received bug bounty reports showing exploits of arbitrary code execution using `NODE_OPTIONS`. This led to the initial filtering for `set-env`. When introducing `GITHUB_ENV` for some reason this filter was added despite also being vulnerable to the initial report. This led to subsequent reports related to arbitrary code execution hence the need to add back the filtering.

I will look at the docs and update them.

vlkl-sap:
The workaround works. What is the threat model here?

eluchsinger:
This workaround is not working for me anymore... :/

tr4l:
> are omissions from this blacklist now also considered security vulns by GitHub

No, you can send report with other env variable leading to code execution, and they will not be considered as vulns.
Which I can understand for stuff like `LD_PRELOAD` (where you also need a file).

However, `BASH_ENV` don't need anything, and will work as long as you have a "run" (on linux, with default bash shell)

The special issue with `NODE_OPTIONS`, is that almost all action are written in node. So if you set this env var, you have code execution in almost all step that are an action. My last random statistic show me that they are more "run" than "use action" on workflow. And "if" they are `GITHUB_ENV` injection, pretty good chance that one of the next step will be a "run"...

EvanCarroll:
I doubt that's the case. Why not just run the processes specified in the action without the variable? Why ambient `run` steps? And if that is the reason, it's just a weird user experience. You can't reason about the the actual security practice it's just because "almost all" code in actions/middleware is in node so this variable gets unique treatment.

tr4l:
It's weird, it's not really efficient. But this is the reason.

It's also weird that "action/checkout" did not work when that env variable is badly set (especially that people may not be aware that checkout is done with node).

I agree that (most?) env var should just not passed to action, they are "input" if we want to pass thing.

PR #2791, opened 2023-08-24
Support replacing runners in v2 flow

luketomlinson:
We previously implemented adding a runner with the Runner Admin flow, but we didn't support replace. This PR adds that ability


src/Runner.Common/RunnerDotcomServer.cs
```diff
@@ -15,9 +15,10 @@ namespace GitHub.Runner.Common
     [ServiceLocator(Default = typeof(RunnerDotcomServer))]
     public interface IRunnerDotcomServer : IRunnerService
     {
-        Task<List<TaskAgent>> GetRunnersAsync(int runnerGroupId, string githubUrl, string githubToken, string agentName);
+        Task<List<TaskAgent>> GetRunnerByNameAsync(string githubUrl, string githubToken, string agentName);
```
TingluoHuang, line 18: i merged https://github.com/actions/runner/pull/2771 and you can rebase on it.

src/Runner.Common/RunnerDotcomServer.cs
```diff
@@ -157,9 +168,15 @@ public async Task<DistributedTask.WebApi.Runner> AddRunnerAsync(int runnerGroupI
...
 
+            if (replace)
+            {
+                bodyObject.Add("runner_id", agent.Id);
+                bodyObject.Add("replace", replace);
+            }
+
             var body = new StringContent(StringUtil.ConvertToJson(bodyObject), null, "application/json");
 
             return await RetryRequest<DistributedTask.WebApi.Runner>(githubApiUrl, githubToken, RequestType.Post, 3, "Failed to add agent", body);
```
TingluoHuang, line 182: should the create use `POST`, update use `PATCH`, and replace use `PUT`?

TingluoHuang (approved):

PR #2802, opened 2023-08-25
Automatically update hashes if new hash is needed

vanZeben:
Presently this workflow is really tedious. When a pull request is made that requires a dotnet-runtime hash or an external hash update, you have to look at the `build.yml` workflow, and wait for each of the (presently) 7 platforms to run, one of them will fail with a "you need to manually update the hash in this file to X", which causes all the others to be skipped. You then have to copy that hash, and paste it in the referenced platform file, and push your changes, and then repeat for the rest of the 7 platforms. 

This workflow change, makes is so that on pull requests only (will not influence main branch changes), if a hash update is required, the workflow will still fail out, but it will just automatically push the file updates to the correct hashes for you in a single commit. This automated push will also re-trigger the same job to run allowing for the L0's to properly be run and tested after the computed hashes are updated.

An example PR where this has happened can be found here: https://github.com/bbq-beets/vanzeben-sandbox/pull/41

And the associated run that produced the has update commit can be found here: https://github.com/bbq-beets/vanzeben-sandbox/actions/runs/5980042098?pr=41

--- 
For a Forked PR, this functionality is unavailable due to [this comment](https://github.com/actions/runner/pull/2802#discussion_r1306058903). So in that case we error the workflow out saying that hashes need to be updated

An example PR: https://github.com/bbq-beets/vanzeben-sandbox/pull/37
An example run for this: https://github.com/bbq-beets/vanzeben-sandbox/actions/runs/5979995564?pr=37


.github/workflows/build.yml
```diff
@@ -68,25 +71,57 @@ jobs:
...
+          ./src/Misc/contentHash/externals/${{ matrix.runtime }}
+          ./src/Misc/contentHash/dotnetRuntime/${{ matrix.runtime }}
+        key: compute-hashes-${{ matrix.runtime }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
+    - name: Create an warning annotation if computed hashes need to be updated        
+      if: github.event_name == 'pull_request' && steps.compute-hash.outcome != 'success' 
+      shell: bash
+      run: |
+        echo "::warning ::Computed hashes do not match, we will automatically update these for you, you can safely ignore the errors on this job"
+        # We need to exit with a failure code so the commit can be process through
+        exit 1
```
vanZeben, line 118: This is actually required, and _will_ unfortunately cause two independent errors in the run. We need to know that the job has failed, and the failure in `compute-hash` has `continue-on-error` set, which means that if there isn't a second `exit 1` this job will be marked as successful. 
Screenshot 2023-08-25 at 2 47 14 PM

The way around this, would be to instead of `exit $NEEDS_UPDATE` in `compute-hash` to just output that variable for other steps to pickup, @TingluoHuang what would be your preference on this?

.github/workflows/build.yml
```diff
@@ -106,3 +141,83 @@ jobs:
...
+    - name: Fetch cached computed hashes
+      if: steps.cache-restore-linux-x64.outputs.cache-hit == 'true' ||
+          steps.cache-restore-linux-arm64.outputs.cache-hit == 'true' ||
+          steps.cache-restore-linux-arm.outputs.cache-hit == 'true' ||
+          steps.cache-restore-win-x64.outputs.cache-hit == 'true' ||
+          steps.cache-restore-win-arm64.outputs.cache-hit == 'true' ||
+          steps.cache-restore-osx-x64.outputs.cache-hit == 'true' ||
+          steps.cache-restore-osx-arm64.outputs.cache-hit == 'true'
+      shell: bash
+      run: |
```
TingluoHuang, line 219: this might not working if the PR is from a fork.

vanZeben: Talked offline. Some context here about forks that I didn't know.

When you fork a repo and present a change back to the origin repo, that change isn't copied into a branch sourced on the origin repo. That branch still exists in your fork and only your fork. Meaning that the GITHUB_TOKEN that is on origin, doesn't have permission to push changes to your branch in your forked repo. 

For now, Ill have this run either by workflow_dispatch, or automatically on non-forked branches.

vanZeben: The Fork behaviour has changed, an example PR for a fork can be seen here: https://github.com/bbq-beets/vanzeben-sandbox/pull/37

TLDR: we error the workflow out, and add a comment to the PR about the error

.github/workflows/build.yml
vanZeben: I changed this to just use an output variable to reduce the error's for readability

.github/workflows/build.yml
```diff
@@ -106,3 +158,92 @@ jobs:
...
+        Environments=( "linux-x64" "linux-arm64" "linux-arm" "win-x64" "win-arm64" "osx-x64" "osx-arm64" )
```
TingluoHuang, line 244: what is this `Environments` for?

vanZeben: Ahh good catch nothing at this point. Originally I was going to try and pull all the caches in a single step but that wasn't working. I'll remove this

PR #2814, opened 2023-08-31
Delegating handler for Http redirects

paveliak:
This PR adds a delegating handler that performs redirection of Http requests if Actions service responds with one of 30x codes.

There is a safety switch disabling redirection handlers (controlled by the `GITHUB_ACTIONS_RUNNER_NO_HTTP_REDIRECTS` environment variable). It does not however control server behavior and if this environment is set and server responds with a redirect then job most likely will fail.

### DIFF
--- a/src/Runner.Worker/JobRunner.cs
+++ b/src/Runner.Worker/JobRunner.cs
@@ -84,7 +84,14 @@ namespace GitHub.Runner.Worker
                 Trace.Info($"Creating job server with URL: {jobServerUrl}");
                 // jobServerQueue is the throttling reporter.
                 _jobServerQueue = HostContext.GetService<IJobServerQueue>();
-                VssConnection jobConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, new DelegatingHandler[] { new ThrottlingReportHandler(_jobServerQueue) });
+                var delegatingHandlers = new List<DelegatingHandler>() { new ThrottlingReportHandler(_jobServerQueue) };
+                message.Variables.TryGetValue("system.github.enable_http_redirects", out VariableValue enableHttpRedirects);
+                if (StringUtil.ConvertToBoolean(enableHttpRedirects?.Value) &&
+                    !StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_NO_HTTP_REDIRECTS")))
+                {
+                    delegatingHandlers.Add(new RedirectMessageHandler(Trace));
+                }
+                VssConnection jobConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, delegatingHandlers);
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/JobRunner.cs
```diff
@@ -84,7 +84,14 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 Trace.Info($"Creating job server with URL: {jobServerUrl}");
                 // jobServerQueue is the throttling reporter.
                 _jobServerQueue = HostContext.GetService<IJobServerQueue>();
-                VssConnection jobConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, new DelegatingHandler[] { new ThrottlingReportHandler(_jobServerQueue) });
+                var delegatingHandlers = new List<DelegatingHandler>() { new ThrottlingReportHandler(_jobServerQueue) };
+                message.Variables.TryGetValue("system.github.enable_http_redirects", out VariableValue enableHttpRedirects);
+                if (StringUtil.ConvertToBoolean(enableHttpRedirects?.Value) &&
```
TingluoHuang, line 89: is it possible to only enable this for the actions resolve endpoint? or we plan to have the redirect for all requests to Actions service at some point?

src/Runner.Worker/JobRunner.cs
```diff
@@ -84,7 +84,14 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 Trace.Info($"Creating job server with URL: {jobServerUrl}");
                 // jobServerQueue is the throttling reporter.
                 _jobServerQueue = HostContext.GetService<IJobServerQueue>();
-                VssConnection jobConnection = VssUtil.CreateConnection(jobServerUrl, jobServerCredential, new DelegatingHandler[] { new ThrottlingReportHandler(_jobServerQueue) });
+                var delegatingHandlers = new List<DelegatingHandler>() { new ThrottlingReportHandler(_jobServerQueue) };
+                message.Variables.TryGetValue("system.github.enable_http_redirects", out VariableValue enableHttpRedirects);
```
TingluoHuang, line 88: i think we normally just use the feature flag name as the variable in other case.
ex: `DistributedTask.EnhancedAnnotations` both on the server and the runner, so it's easy to search and cleanup later.

src/Runner.Worker/JobRunner.cs
paveliak: It would look "hacky" if we try to scope it to a single endpoint. Besides, Actions service controls whether it redirects request or not so what does it buy for us if we add additional constraints on the client side? `VssConnection` is used for accessing Actions service only so it's not like we install system-wide redirection handler that would also engage when talking to other services.

TingluoHuang (approved):

PR #2828, opened 2023-09-05
Throw NonRetryableException on GetNextMessage from broker as needed.

TingluoHuang:
The runner exit itself as expected instead of a infinite restart loop.
```
@TingluoHuang ➜ /workspaces/actions/runner/_layout $ ./run.sh 
Current runner version: '2.308.0'
2023-09-05 18:10:10Z: Listening for Jobs
2023-09-05 18:10:56Z: Runner connect error: Failed to get job message: Error: Not Found. Retrying until reconnected.
An error occurred: Get next message failed with non-retryable error.
Runner listener exit with terminated error, stop the service, no retry needed.
Exiting runner...
```

### DIFF
--- a/src/Runner.Listener/BrokerMessageListener.cs
+++ b/src/Runner.Listener/BrokerMessageListener.cs
@@ -108,7 +108,7 @@ namespace GitHub.Runner.Listener
 
                     if (!IsGetNextMessageExceptionRetriable(ex))
                     {
-                        throw;
+                        throw new NonRetryableException("Get next message failed with non-retryable error.", ex);
                     }
                     else
                     {


src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -108,7 +108,7 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
 
                     if (!IsGetNextMessageExceptionRetriable(ex))
                     {
-                        throw;
+                        throw new NonRetryableException("Get next message failed with non-retryable error.", ex);
```
TingluoHuang, line 111: We didn't do this with the other Listener since we are preventing the loop during CreateSeesion.

Since the broker listener doesn't really have a concept for create session, i need to break out the loop at here for now.

PR #2850, opened 2023-09-15
handle secret ending with &+

lokesh755:
## Context

This pull request addresses an edge case in the `PowerShellPostAmpersandEscape` method within the `ValueEncoders` class. Previously, when a secret string ended with `&+`, an `ArgumentOutOfRangeException` was thrown. This PR ensures that the method handles this scenario gracefully.

Related to https://github.com/github/c2c-actions-support/issues/2857


Link-:
Fixed in https://github.com/actions/runner/pull/2875

PR #2857, opened 2023-09-18
Allow use action archive cache to speed up workflow jobs.

TingluoHuang:
To avoid network roundtrip for downloading each action's archive and speed up workflow jobs, we are introduce a new ENV to the runner, and the runner will try to find the action tarball/zipball of the resolved version in the action archive cache folder and avoid downloading it from the service.

The cache folder will be point by `ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE` env and have the following structure:

For Linux/macOS:
```
./actions_checkout/
  ./SHA1.tar.gz
  ./SHA2.tar.gz
  ./SHA3.tar.gz
  ./SHA4.tar.gz
  ....
./actions_cache/
  ./SHA1.tar.gz
  ./SHA2.tar.gz
  ./SHA3.tar.gz
  ./SHA4.tar.gz
  ....
```

For Windows:
```
./actions_checkout/
  ./SHA1.zip
  ./SHA2.zip
  ./SHA3.zip
  ./SHA4.zip
  ....
./actions_cache/
  ./SHA1.zip
  ./SHA2.zip
  ./SHA3.zip
  ./SHA4.zip
  ....
```

The tarball or zipball will be the content from https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-tar

https://github.com/github/actions-assignment/issues/17

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -792,8 +792,6 @@ namespace GitHub.Runner.Worker
             string link = downloadInfo?.TarballUrl;
 #endif
 
-            // i will remove the trace.info later, it here for better diff.
-            Trace.Info($"Save archive '{link}' into {archiveFile}.");
             try
             {
                 var useActionArchiveCache = false;
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -796,7 +796,48 @@ namespace GitHub.Runner.Worker
             Trace.Info($"Save archive '{link}' into {archiveFile}.");
             try
             {
-                await DownloadRepositoryArchive(executionContext, link, downloadInfo.Authentication?.Token, archiveFile);
+                var useActionArchiveCache = false;
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.UseActionArchiveCache") == true)
+                {
+                    var hasActionArchiveCache = false;
+                    var actionArchiveCacheDir = Environment.GetEnvironmentVariable(Constants.Variables.Agent.ActionArchiveCacheDirectory);
+                    if (!string.IsNullOrEmpty(actionArchiveCacheDir) &&
+                        Directory.Exists(actionArchiveCacheDir))
+                    {
+                        hasActionArchiveCache = true;
+                        Trace.Info($"Check if action archive '{downloadInfo.ResolvedNameWithOwner}@{downloadInfo.ResolvedSha}' already exists in cache directory '{actionArchiveCacheDir}'");
+#if OS_WINDOWS
+                        var cacheArchiveFile = Path.Combine(actionArchiveCacheDir, downloadInfo.ResolvedNameWithOwner.Replace(Path.DirectorySeparatorChar, '_').Replace(Path.AltDirectorySeparatorChar, '_'), $"{downloadInfo.ResolvedSha}.zip");
+#else
+                        var cacheArchiveFile = Path.Combine(actionArchiveCacheDir, downloadInfo.ResolvedNameWithOwner.Replace(Path.DirectorySeparatorChar, '_').Replace(Path.AltDirectorySeparatorChar, '_'), $"{downloadInfo.ResolvedSha}.tar.gz");
+#endif
+                        if (File.Exists(cacheArchiveFile))
+                        {
+                            try
+                            {
+                                Trace.Info($"Found action archive '{cacheArchiveFile}' in cache directory '{actionArchiveCacheDir}'");
+                                File.Copy(cacheArchiveFile, archiveFile);
+                                useActionArchiveCache = true;
+                                executionContext.Debug($"Copied action archive '{cacheArchiveFile}' to '{archiveFile}'");
+                            }
+                            catch (Exception ex)
+                            {
+                                Trace.Error($"Failed to copy action archive '{cacheArchiveFile}' to '{archiveFile}'. Error: {ex}");
+                            }
+                        }
+                    }
+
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -795,97 +792,11 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             string link = downloadInfo?.TarballUrl;
 #endif
 
+            // i will remove the trace.info later, it here for better diff.
             Trace.Info($"Save archive '{link}' into {archiveFile}.");
             try
             {
-                int retryCount = 0;
```
TingluoHuang, line 801: i move the entire download logic behind the new `DownloadRepositoryArchive()`
Should not have impact on functionality.

src/Runner.Worker/ActionManager.cs
```diff
@@ -796,7 +796,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             Trace.Info($"Save archive '{link}' into {archiveFile}.");
             try
             {
-                await DownloadRepositoryArchive(executionContext, link, downloadInfo.Authentication?.Token, archiveFile);
+                var useActionArchiveCache = false;
+                if (executionContext.Global.Variables.GetBoolean("DistributedTask.UseActionArchiveCache") == true)
```
TingluoHuang, line 798: FF controlled by the service.

src/Runner.Worker/ActionManager.cs
```diff
@@ -795,97 +792,50 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
             string link = downloadInfo?.TarballUrl;
 #endif
 
-            Trace.Info($"Save archive '{link}' into {archiveFile}.");
```
TingluoHuang, line 798: the diff is not readable.
I break the change into 3 commits, which might explain the diff better.
- https://github.com/actions/runner/pull/2857/commits/41a2d7cb15c3bffb83866cddf2039b04005ac103  move code around and should not change any behavior.
- https://github.com/actions/runner/pull/2857/commits/f687e155e5e1ec0ab0ddd7c66a9e52bdd99b145f change behind FF that use local actions cache instead of downloading from the service.
- https://github.com/actions/runner/pull/2857/commits/95ee6ccda1a08d43d79603f721ec2443d978c724 cleanup.

johnsudol (approved):

alexklibisz:
This is exactly what we've been looking for to mitigate some API rate limiting issues. Can anyone provide an estimate for when this would go out in a release?

TingluoHuang:
@alexklibisz do you have a log shows me where you are getting rate limiting? i want to double check to make sure this feature actually can fix your problem.
If you are not feel comfortable of posting log here, you can email it to me with `tingluohuang` @ `github.com`

alexklibisz:
I don't have the exact log on hand, but it's something like `"Github API rate limit exceeded while cloning action "actions/checkout@v3"`. This is on a GHE deployment. I haven't seen this happen on public Github. But it seems like we could avoid a lot of redundant requests by just caching the downloaded actions, as in this PR.

retzero:
@TingluoHuang 
Thank you for this great feature.
How can I enable this variable `UseActionArchiveCache` to `true`? I'm running GHES ans self-hosted runners.

```
                if (executionContext.Global.Variables.GetBoolean("DistributedTask.UseActionArchiveCache") == true)
```

TingluoHuang:
This feature will not help the Rate limit problem happen during `Resolving` Action (translate the v3 tag to a SHA for a given action), but it will help avoid the download of the action if the local cache exists.
@retzero The feature flag (the `DistributedTask.UseActionArchiveCache` variable) will be enabled in the next GHES release once we verify everything is working as expected with github.com

alexklibisz:
> This feature will not help the Rate limit problem happen during Resolving Action (translate the v3 tag to a SHA for a given action)

That's too bad. What if we explicitly use `@SHA` instead of `@v3`? Would that eliminate an API call?

TingluoHuang:
@alexklibisz that's not a bad idea, we can make change to let the runner skip the resolving API call if you are referencing actions via `@SHA` and the local cache has it already.
Let me talk with my team to see what's their thought on this. 🙇

alexklibisz:
> @alexklibisz that's not a bad idea, we can make change to let the runner skip the resolving API call if you are referencing actions via `@SHA` and the local cache has it already. Let me talk with my team to see what's their thought on this. 🙇

Thanks! These rate limiting issues have been pretty disruptive, so anything that can reduce API calls when the jobs are starting is helpful. 

For context, we are using Actions Runner Controller with Horizontal Runner Autoscalers and Runner Deployments on top of AWS spot instances. We really like the scalability and the ephemeral nature of the runners. But as we've onboarded more projects, we're getting disrupted by these API rate limits basically daily. The API rate limits seem reasonable. It seems like there is some room for optimization in the runner initialization. I think these optimizations will be important as Github has officially adopted/recommended Actions Runner Controller and teams are starting to adopt it.

alexklibisz:
@TingluoHuang Can you help me understand one thing more precisely:

> but it will help avoid the download of the action if the local cache exists.

Does downloading the action count against the API rate limit? In other words, will the change in this PR help at all with rate limit issues?

TingluoHuang:
@alexklibisz bad news is my change won't help rate limiting at all, there are some action policy we need to check when the runner try to use certain actions, so we have to call the resolving actions endpoint. 😞 

If you are hitting rate limiting on your GHES, you can reach out to GitHub support, and they should be able to help you bump the rate limit setting on your GHES instance.

PR #2871, opened 2023-09-20
Change alpine from vst blobs to OSS gha alpine build

vanZeben:
Remove dependency on vstagenttools blob storage for alpine-nodejs and use our own republished alpine-nodejs binary from https://github.com/actions/alpine_nodejs

Issue: https://github.com/github/c2c-actions-runtime/issues/2076


Link- (approved):
🥳

PR #2878, opened 2023-09-24
Set `ImageOS` environment variable in runner images

int128:
`ImageOS` environment variable is provided in GitHub-hosted runners. It would be nice if it is provided in the self-hosted runner images as well, because some actions such as ruby/setup-ruby depend on it.

See also the following issues:

- https://github.com/actions/runner-images/issues/345
- https://github.com/actions/actions-runner-controller/issues/715


int128:
@Link- @nikola-jokic @TingluoHuang 
I'd like to ask the maintainers to review this change. If an issue is needed, I will write it.

nikola-jokic (approved):
LGTM! Sorry for the delay and thank you for the PR!

PR #2887, opened 2023-09-26
Add warning about node16 deprecation

takost:
Add warning about node16 deprecation. This warning will be enabled on October 23rd, according to https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
Right now it will be under FF `DistributedTask.AddWarningToNode16Action`


src/Runner.Worker/Handlers/NodeScriptActionHandler.cs
```diff
@@ -134,6 +134,29 @@ public async Task RunAsync(ActionRunStage stage)
...
+            if(Data.NodeVersion == Constants.Runner.DeprecatedNodeVersion && (ExecutionContext.Global.Variables.GetBoolean(Constants.Runner.Features.Node16Warning) ?? false))
+            {
+                var repoAction = Action as RepositoryPathReference;
+                var warningActions = new HashSet<string>();
+                if (ExecutionContext.Global.Variables.TryGetValue(Constants.Runner.DeprecatedNodeDetectedAfterEndOfLifeActions, out var deprecatedNodeWarnings))
+                {
+                    warningActions = StringUtil.ConvertFromJson<HashSet<string>>(deprecatedNodeWarnings);
+                }
+
+                var repoActionFullName = "";
```
JoannaaKL, line 146: I'd propose below code, it's a bit simpler and reduces amount of mutable vars. Plus repoActionFullName is used only in this if-else block anyway.
```suggestion
                 if (string.IsNullOrEmpty(repoAction.Name))
                {
                  // local actions don't have a 'Name'
                   warningActions.Add(repoAction.Path);
                }
                else
                {
                     warningActions.Add($"{repoAction.Name}/{repoAction.Path ?? string.Empty}".TrimEnd('/') + $"@{repoAction.Ref}");
                }
```

takost: Thanks for you comment. Applied 👍

JoannaaKL (approved):

ajschmidt8:
cross-posting the issue I created below for visibility:

- https://github.com/actions/runner/issues/2906

The transition from Node 16 to Node 20 conflicts with the `actions/runner` claim of support for CentOS 7.

PR #2893, opened 2023-09-27
Do not give up if Results is powering logs

yacaovsnc:
Do not give up if Results is powering logs.  Continue to upload the next file in that case.


DARKo6o6o6 (approved):

DARKo6o6o6:
You have My permission Sir

pje:
The logical change makes sense to me!

One request: if we're renaming variables, can we do it everywhere all at once? `s/resultService/resultsService/g`

👉 https://github.com/search?q=repo%3Ayacaovsnc%2Frunner%20resultServiceOnly&type=code

yacaovsnc:
Thanks @pje!  I think I got all the name variables.  Grep doesn't show any more instance of the old `ResultServiceOnly` name anymore.  🙇🏼‍♂️

pje:
@yacaovsnc Yep, you're right! I messed up the new search syntax and was looking at the main branch. 🤦 

Looks good! 👍

pje (approved):

PR #2896, opened 2023-09-28
Update message when runners are deleted

thboop:
This pr elaborates on why a runner registration may have been deleted, so that users aren't surprised when they see this has happened.


sethrylan (approved):
LGTM

PR #2898, opened 2023-09-29
Incorrect returncode by run.sh

shahidhs-ibm:
The `config.sh` script return error code 131 when .NET is not found or not properly set through environment variables. However same is not true for `run.sh` script. It proceeds to execute and returns 0 in the end even when .NET is missing. 

FYI, we are trying on non-Intel architecture where download of .NET runtime fails to download it from download link.


shahidhs-ibm:
Any update??

PR #2908, opened 2023-10-04
Use RawHttpMessageHandler and VssHttpRetryMessageHandler in ResultsHttpClient

yacaovsnc:
Consolidate http client implementation in `ResultsHttpClient`.  Use the standard message handlers in Results as well.

### DIFF
--- a/src/Runner.Common/ResultsServer.cs
+++ b/src/Runner.Common/ResultsServer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Net.WebSockets;
 using System.Security;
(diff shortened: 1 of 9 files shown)


src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
```diff
@@ -68,6 +68,20 @@ public class RawHttpClientBase : IDisposable
             m_formatter = new VssJsonMediaTypeFormatter();
         }
 
+        protected RawHttpClientBase(
+            Uri baseUrl,
+            HttpMessageHandler pipeline,
+            int timeoutSeconds,
+            bool disposeHandler)
+        {
+            m_client = new HttpClient(pipeline, disposeHandler);
```
TingluoHuang, line 77: i would suggest we use the same message handler, and not the raw httpclient

src/Sdk/Common/Common/Authentication/NoOpCredentials.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace GitHub.Services.Common
+{
+    // Set of classes used to bypass token operations
+    // Results Service and External services follow a different auth model but
+    // we are required to pass in a credentials object to create a RawHttpMessageHandler
+    public class NoOpCredentials : FederatedCredential
```
TingluoHuang, line 10: do you need to have `FedCred` or just use `IssuedTokenCredential`?

TingluoHuang: then you can always return false for `IsAuthenticationChallenge()`

src/Runner.Sdk/Util/VssUtil.cs
```diff
@@ -85,6 +85,35 @@ public static void InitializeVssClientSettings(List<ProductInfoHeaderValue> addi
             VssCredentials credentials,
             IEnumerable<DelegatingHandler> additionalDelegatingHandler = null,
             TimeSpan? timeout = null)
+        {
```
TingluoHuang, line 88: is these change just moving content around?  create the new `GetHttpRequestSettings` for re-use

src/Sdk/Common/Common/Authentication/NoOpCredentials.cs
yacaovsnc: FedCred, it's on the base constructor.  

```
        public RawHttpMessageHandler(
            FederatedCredential credentials,
            RawClientHttpRequestSettings settings,
            HttpMessageHandler innerHandler)
```

src/Runner.Common/ResultsServer.cs
```diff
@@ -52,8 +53,8 @@ public sealed class ResultServer : RunnerService, IResultsServer
 
         public void InitializeResultsClient(Uri uri, string liveConsoleFeedUrl, string token)
         {
-            var httpMessageHandler = HostContext.CreateHttpClientHandler();
-            this._resultsClient = new ResultsHttpClient(uri, httpMessageHandler, token, disposeHandler: true);
+            this._resultsClient = CreateHttpClient(uri, token);
```
TingluoHuang, line 56: should we have 2 clients for result server, one with Token, one with the NoOpsCredential?

src/Runner.Sdk/Util/VssUtil.cs
yacaovsnc: Yup, line 116 is the new method and it's reused in the `ResultsServer`.

PR #2913, opened 2023-10-05
Send os and arch during long poll

luketomlinson:
We want the runner to self report it's os/architecture. This allows us to make certain decisions about upgrades, etc.

### DIFF
--- a/src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
+++ b/src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
@@ -461,6 +461,8 @@ namespace GitHub.DistributedTask.WebApi
             long? lastMessageId = null,
             TaskAgentStatus? status = null,
             string runnerVersion = null,
+            string os = null,
+            string architecture = null,
             object userState = null,
             CancellationToken cancellationToken = default)
         {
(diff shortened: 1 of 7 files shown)


src/Runner.Common/RunnerDotcomServer.cs
```diff
@@ -163,6 +164,8 @@ private async Task<DistributedTask.WebApi.Runner> AddOrReplaceRunner(int runnerG
                         {"ephemeral", agent.Ephemeral},
                         {"labels", agent.Labels},
                         {"public_key", publicKey},
+                        {"os", VarUtil.OS},
```
TingluoHuang, line 167: This or the combined package name, ex Linux-x64

TingluoHuang (approved):

src/Runner.Common/RunnerDotcomServer.cs
luketomlinson: We send back just the `os` in the REST API, so it's easy enough to concatenate them if we need to. https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28

TingluoHuang (dismissed):

src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
```diff
@@ -483,6 +485,16 @@ public TaskAgentHttpClientBase(Uri baseUrl, HttpMessageHandler pipeline, bool di
                 queryParams.Add("runnerVersion", runnerVersion);
             }
 
+            if (os != null)
+            {
+                queryParams.Add("os", os);
```
TingluoHuang, line 490: make sure you double check configure the runner with your change with github.com
I am 99.9% sure the extra query param will just no-opt on the service side.

TingluoHuang (approved):

src/Sdk/DTGenerated/Generated/TaskAgentHttpClientBase.cs
luketomlinson: done

PR #2915, opened 2023-10-09
Trim whitespace in `./Misc/contentHash/dotnetRuntime/*`

TingluoHuang:


src/Test/L0/PackagesTrimL0.cs
```diff
@@ -273,5 +273,29 @@ public async Task RunnerLayoutParts_CheckExternalsHash()
                 Assert.True(string.Equals(hashResult, File.ReadAllText(externalsHashFile).Trim()), $"Hash mismatch for externals. You might need to update `Misc/contentHash/externals/{BuildConstants.RunnerPackage.PackageName}` or check if `hashFiles.ts` ever changed recently.");
             }
         }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Common")]
+        public Task RunnerLayoutParts_ContentHashFilesNoNewline()
```
TingluoHuang, line 280: add a test to make sure we don't add any whitespace in the future...

luketomlinson (approved):

PR #2844, opened 2023-09-11
Update default version to node20

takost:
Make node20 default node version for runner internal operations. This can be opted-out by starting Runner with `ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION=node16` or if `DistributedTask.ForceInternalNodeVersionOnRunnerTo16` FF set to `true` (PR: https://github.com/github/actions-dotnet/pull/16155).

PR #2918, opened 2023-10-10
Revert "Update default version to node20 (#2844)"

takost:
This reverts commit acdc6edf7ce3964c759db8217771aa9cd8e16b51.
Before merging update of internal operations, we should make sure that customers are aware about this change and are able to work with it


dmtzs:
Just a question, this means that when using for example actions/checkout@v4 we will have again the issue of the GLIB versions not installed as required by node version 20? @takost 
Thank you in advance

PR #2923, opened 2023-10-12
Write archive hash/size to log after extraction error

larsxschneider:
I would like to understand the corrupt archives better. Let's log their size. Maybe this gives us a new clue in what way these archives are corrupted.

⚠️ Untested! How do I test this properly? ⚠️


carlosmn:
What would help with figuring out if something's gone really wrong on the codeload side would be to store the error message from tar somewhere.

AvaStancu:
@larsxschneider Regarding testing, would extracting the new logic into a separate function that for ex. receives the `archiveFile` and returns the `sha256hash` / `fi.Length`, so you can unit test it, help? what else would you need to add a test?

TingluoHuang:
@larsxschneider I will take over the PR with more changes (ex: send the output from `tar` on failures) and get it tested and merged.

TingluoHuang:
Taking over with https://github.com/actions/runner/pull/2939

PR #2940, opened 2023-10-19
Bump node 20 to v20.8.1

Julien-Marcou:
Bumping node 20, as there was 2 security releases since the v20.5.0:

- https://nodejs.org/en/blog/release/v20.5.1
- https://nodejs.org/en/blog/release/v20.8.1


TingluoHuang:
Take over with https://github.com/actions/runner/pull/2945

PR #2942, opened 2023-10-21
Use the correct unicode checkmark symbol in success messages

Infinoid:
✓ is U+2713, check mark.
√ is U+221A, square root (radical symbol).


Infinoid:
Note that if you don't like the look of a normal check mark, there are some other options available as well.  [Wikipedia has a list.](https://en.wikipedia.org/wiki/Check_mark#Unicode)  And there's also the thumbs up symbol.

I just don't like the square root symbol, my eyes always try to read it as a math equation.

PR #2950, opened 2023-10-25
Fix docker buildx

Wielewout:
In https://github.com/actions/runner/pull/2901 `buildx` was again added. However it was not included in the final layer, hence it is currently still missing unfortunately.


Wielewout:
Fixed in https://github.com/actions/runner/pull/2952

PR #2952, opened 2023-10-25
Fix `buildx` installation

ajschmidt8:
This PR is a continuation of #2901.

I hadn't tested that change locally and didn't realize the Dockerfile did a multi-stage build.

This PR fixes the `buildx` installation to ensure that the binary is copied to the final stage.

I tested these changes locally to confirm that it works as expected.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -54,6 +54,7 @@ RUN adduser --disabled-password --gecos "" --uid 1001 runner \
 WORKDIR /home/runner
 
 COPY --chown=runner:docker --from=build /actions-runner .
+COPY --from=build /usr/local/lib/docker/cli-plugins/docker-buildx /usr/local/lib/docker/cli-plugins/docker-buildx
 
 RUN install -o root -g root -m 755 docker/* /usr/bin/ && rm -rf docker
 


ajschmidt8:
cc: @TingluoHuang.

If we could get a patch release for `2.311.0` with this change that would be awesome.

images/Dockerfile
```diff
@@ -54,6 +54,7 @@ RUN adduser --disabled-password --gecos "" --uid 1001 runner \
 WORKDIR /home/runner
 
 COPY --chown=runner:docker --from=build /actions-runner .
+COPY --from=build /usr/local/lib/docker/cli-plugins/docker-buildx /usr/local/lib/docker/cli-plugins/docker-buildx
```
TingluoHuang, line 57: do you also need `--chown=runner:docker`?

ajschmidt8: I don't think that's necessary. It worked fine in my local testing.

Also, in our custom images, we've been using a `root:root` owned `docker-buildx` file without issues.

TingluoHuang (approved):

chrisdleech-pol:
This seems to be causing an issue for some container actions. Example when using hadolint-action in a workflow.

``` 
Build container for action use: '/home/runner/_work/_actions/hadolint/hadolint-action/v3.1.0/Dockerfile'.
  ERROR: mkdir /home/runner/.docker/buildx: read-only file system
  ```

ajschmidt8:
> This seems to be causing an issue for some container actions. Example when using hadolint-action in a workflow.
> 
> ```
> Build container for action use: '/home/runner/_work/_actions/hadolint/hadolint-action/v3.1.0/Dockerfile'.
>   ERROR: mkdir /home/runner/.docker/buildx: read-only file system
> ```

As far I as know, this hasn't been released yet. So I'm not sure if this is what's causing your problem.

chrisdleech-pol:
Fair enough. The latest image for 2.311.0 was updated 1 day ago, and I assumed this part of that update

jlehnGIT:
@ajschmidt8 was curious if this was related...We tried to run runner version 3.310.2 from current 3.209.0 and something has changed to where what was working fine now is failing and seems to be related possibly to Docker. Was wondering if this error indicates related to this issue or not please?

image

ajschmidt8:
This PR hasn't been included in any runner release yet. I would recommend reviewing the release changelogs to identify any changes that may have affected your workflows.

Disclaimer, I don't work for GitHub and therefore don't provide any support for these changes.

jlehnGIT:
Thank you and yes understand. Was seeing if you may have an idea of what may be happening. Thank you.

omer2500:
> > This seems to be causing an issue for some container actions. Example when using hadolint-action in a workflow.
> > ```
> > Build container for action use: '/home/runner/_work/_actions/hadolint/hadolint-action/v3.1.0/Dockerfile'.
> >   ERROR: mkdir /home/runner/.docker/buildx: read-only file system
> > ```
> 
> As far I as know, this hasn't been released yet. So I'm not sure if this is what's causing your problem.

For anyone facing this issue
add this to your runner env:

  - name: BUILDX_CONFIG
    value: /tmp/buildx
This will tell him to go to a different folder
works on 2.313
should work on earlier versions with the build

@ajschmidt8 maybe we should set it in a folder that we know is not read-only by default?

PR #2970, opened 2023-11-02
Send disableUpdate as query parameter 

luketomlinson:
Similar to https://github.com/actions/runner/pull/2913, we now send disableUpdate as a query parameter

### DIFF
--- a/src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
+++ b/src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
@@ -61,6 +61,7 @@ namespace GitHub.Actions.RunService.WebApi
             TaskAgentStatus? status,
             string os = null,
             string architecture = null,
+            bool? disableUpdate = null,
             CancellationToken cancellationToken = default
         )
         {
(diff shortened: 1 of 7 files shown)


src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -87,6 +88,11 @@ public class BrokerHttpClient : RawHttpClientBase
                 queryParams.Add("architecture", architecture);
             }
 
+            if (disableUpdate != null)
+            {
+                queryParams.Add("disableUpdate", disableUpdate.Value.ToString().ToLower());
```
TingluoHuang, line 93: do we need to always provide `disableUpdate=false` as default, or the service will handle the default case when the request doesn't have the query parameter?

TingluoHuang (approved):

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
luketomlinson: yeah we assume false when it's not there

PR #2978, opened 2023-11-09
Issue 2892: Add `RUNNER_TOOL_CACHE` to image to support runner tools caching

### LINKED ISSUE #2892: Action Runner image doesn't support setup-ruby
**Describe the bug**
When using the action runner image, the [setup-ruby](https://github.com/ruby/setup-ruby) action is not supported out of the box.

**To Reproduce**
1. Use the latest version of the image (in our case, installed as part of ARC using the [getting started guide](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller))
2. Add the setup-ruby action to a workflow running on the self hosted runner:
  ```
  - name: Setup Ruby and install gems
    uses: ruby/setup-ruby@v1
    with:
      bundler-cache: true
  ```
3. See the error
  ```
  Run ruby/setup-ruby@v1
  with:
    bundler-cache: true
    ruby-version: default
  
Using 3.1.3 as input from file .tool-versions
Modifying PATH
  Entries added to PATH to use selected Ruby:
    /opt/hostedtoolcache/Ruby/3.1.3/x64/bin
Error: Error: EACCES: permission denied, mkdir '/opt/hostedtoolcache'
```
**Expected behavior**
As recommended in [the guides](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller#about-the-runner-container-image:~:text=This%20image%20contains,node%20for%20Node.), I expect the setup action for ruby to be able to complete correctly. It looks like however that the expected global `RUNNER_TOOL_CACHE` path is not available on the container, but this path is required because prebuild rubies are generated in this path. See this section for more info: https://github.com/ruby/setup-ruby#using-self-hosted-runners

239.0

Linux.

## What's not working?
The `Error: Error: EACCES: permission denied, mkdir '/opt/hostedtoolcache'` doesn't exist, and the runner cannot write to it.
image

AmitBenAmi:
# Description

The `RUNNER_TOOL_CACHE` (`/opt/hostedtoolcache`) is a mandatory and required path to exist in the Github runners for specific set-up actions (for example, [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby).

It is also written in [their documentation for self-hosted runners](https://github.com/ruby/setup-ruby#using-self-hosted-runners) that this path is required.

I used the same mechanism for creating this path as in the `actions-runner-controller` project [runners' Dockerfile](https://github.com/actions/actions-runner-controller/blob/862bc1a9dd8ab1077fbd797557c15e266d27afdb/runner/actions-runner-dind.ubuntu-20.04.dockerfile#L90).

fixes #2892


AmitBenAmi:
@nikola-jokic I wanted to ask for a maintainer review on this, please?
Thank you!

diclophis:
@TingluoHuang Is this something that could be prioritized and reviewed... it is a major blocker to utilizing `setup/ruby` actions in self-hosted runners ... it really seems like it should work on the surface, but it does not because the vanilla `ghcr.io/actions/actions-runner:latest` image is broken.

diclophis:
@luketomlinson @rentziass @AllanGuigou I see you all have recently approved some PRs.

It would be greatly appreciated if we could get a review on this (either reject or approve) so that we can address any remaining concerns and hopefully work towards a fix to support `setup-ruby` gha actions

see https://github.com/actions/runner/issues/2892 for the main problem

GN93:
We're struggling with the same problem.. Is there any chance for a fix?

reitermarkus:
How is this still not merged after over two years?

For now I came up with with the following workaround using the `gha-runner-scale-set-experimental` Helm chart and the following values:

```yaml

runner:
  env:
    # FIXME: https://github.com/actions/runner/pull/2978
    - name: RUNNER_TOOL_CACHE
      value: /opt/hostedtoolcache
    # FIXME: Ruby default encoding is ASCII if this is not set.
    - name: LANG
      value: en_US.UTF-8
  container:
    command:
      - bash
      - -c
      - |
        sudo mkdir -p "${RUNNER_TOOL_CACHE}"
        sudo chgrp runner "${RUNNER_TOOL_CACHE}"
        sudo chmod g+rwx "${RUNNER_TOOL_CACHE}"

        /home/runner/run.sh
 ```

diclophis:
@salmanmkc / @ericsciple / @TingluoHuang any luck on this PR?

PR #2989, opened 2023-11-15
Use ENV to set LogLevel / TraceLevel in Listener and Worker log files and stdout when allowed.

fhammerl:
Legacy GITHUB_ACTIONS_RUNNER_TRACE still takes precedence for back compat.


booleanbetrayal:
I just stumbled upon this PR and it looks like the original `DefaultTraceLevel` assignment is just wrong and will always assign verbose if provided. Maybe this PR should replace that assignment with this fix?

fhammerl:
Closing PR as there aren't plans to introduce this configuration option to the runner at this time.

booleanbetrayal:
I went ahead and opened a new issue here - https://github.com/actions/runner/issues/3045

PR #3016, opened 2023-11-29
Correctly  detect arm64

arianvp:
Without this it misdetects arm64 hosts as x86


arianvp:
This is still an issue. Can you please reopen and merge?

PR #3017, opened 2023-12-01
Add flag to prevent references to 'self' local actions

gateixeira:
This pull request implements a customer request to prevent local actions from being executed.

The main use case is to have a more clear boundary of what can be executed by a given user.
As GitHub does not allow to limit that with a policy at the application level, we decided to extend the runner code.

Customer's main goal is to prevent execution of local actions as well as pre-establish an allowlist of synced actions that can be executed. **This pull request targets only the former.** The latter is achieved via a `ACTIONS_RUNNER_HOOK_JOB_STARTED` script that checks for the downloaded actions against such allow list.

---

### Copilot-generated description:

This pull request primarily introduces a new feature that prevents local actions from being executed in certain contexts. A new constant `ActionsPreventLocalAction` is added in the `Constants.cs` file. The `ActionManager.cs` file is updated to check this constant and throw an exception if local actions are attempted. Two new tests are also added in `ActionManagerL0.cs` to verify the correct behavior when the `ActionsPreventLocalAction` is set to true and false.

Prevention of local actions:

* <a href="diffhunk://#diff-a52d1672f99c1570465e6dea6567eec5c1d10028cdf0c94c8dd4f5d708e5b4abR250">`src/Runner.Common/Constants.cs`</a>: Added a new constant `ActionsPreventLocalAction` to the `Actions` class.
* <a href="diffhunk://#diff-5b282ef924f268139b4015eb491825d3e35bfa6cbe46962e3bef3bd8fa05dfccL4-L9">`src/Runner.Worker/ActionManager.cs`</a>: Removed unused imports.
* <a href="diffhunk://#diff-5b282ef924f268139b4015eb491825d3e35bfa6cbe46962e3bef3bd8fa05dfccR72-R85">`src/Runner.Worker/ActionManager.cs`</a>: Updated `PrepareActionsAsync` method to check the `ActionsPreventLocalAction` constant and throw an exception if local actions are attempted in a context where they are not allowed.

Testing:

* <a href="diffhunk://#diff-f7999ffa872c1d85e7368f4c351978e3d2a4ad3e3b9a27547234ede8e26bdaf5R1150-R1224">`src/Test/L0/Worker/ActionManagerL0.cs`</a>: Added two new tests `PrepareActions_PreventLocalActions_False` and `PrepareActions_PreventLocalActions_True` to verify the correct behavior when the `ActionsPreventLocalAction` is set to true and false.


EricDales:
Hello, 

any news on the review of this PR ?

PR #3033, opened 2023-12-11
Use Azure SDK to upload files to Azure Blob

yacaovsnc:
This pull request changes `ResultsHttpClient.cs` to use [Azure Storage Blobs SDK](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage/Azure.Storage.Blobs) to upload logs and summaries. 

This allows us to control `timeout` and `retries` independently since we were using the same http layer for both communications to results service and Azure blob storage.  Now we can rely on the SDK to manage details such as headers and query params.

Using Azure specific SDK means we need to add additional logic for other blob providers down the road.  This is an acceptable tradeoff.

Here are the most important changes:

Integration of Azure Storage Blobs SDK:

* <a href="diffhunk://#diff-64a5051fbf21116833fbb1e196c6cb27686093c28f681fdf560bde506f73666dR16">`src/Sdk/Sdk.csproj`</a>: Added a package reference to the Azure Storage Blobs SDK.
* <a href="diffhunk://#diff-733082d8c98a759445365fda97e77de1455e1d048ded4fa1bb4642a4c9a68bddL3-L12">`src/Sdk/WebApi/WebApi/ResultsHttpClient.cs`</a>: Imported the Azure Storage Blobs SDK namespaces.

Refactoring of methods for uploading files to Azure Blob Storage:

* <a href="diffhunk://#diff-733082d8c98a759445365fda97e77de1455e1d048ded4fa1bb4642a4c9a68bddL164-L230">`src/Sdk/WebApi/WebApi/ResultsHttpClient.cs`</a>: Replaced the HTTP request-based file upload methods with methods that use the Azure Storage Blobs SDK. This includes changes to `UploadBlockFileAsync`, `CreateAppendFileAsync`, and `UploadAppendFileAsync`. 
* <a href="diffhunk://#diff-733082d8c98a759445365fda97e77de1455e1d048ded4fa1bb4642a4c9a68bddL424-L428">`src/Sdk/WebApi/WebApi/ResultsHttpClient.cs`</a>: Removed constants related to Azure Blob Storage headers.


src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
TingluoHuang, line 1: do you want to FF this change?

TingluoHuang: ex: ResultsHttpClient.cs (existing) and ResultsAzureHttpClient.cs (new)

yacaovsnc: I can try to add a feature flag here, it would be passed in - the flag must be in the `ResultsHttpClient` though as we need to first make a call to Results to know the blob provider is Azure, and both http calls live here.  I considered extract the Azure specific logic to another class, but since this is the only provider we support for now, I was thinking we could do the refactoring later on.

src/Runner.Common/JobServerQueue.cs
```diff
@@ -134,8 +134,8 @@ public void Start(Pipelines.AgentJobRequestMessage jobRequest, bool resultsServi
                 {
                     liveConsoleFeedUrl = feedStreamUrl;
                 }
-
-                _resultsServer.InitializeResultsClient(new Uri(resultsReceiverEndpoint), liveConsoleFeedUrl, accessToken);
+                jobRequest.Variables.TryGetValue("system.github.results_upload_with_sdk", out VariableValue resultsUseSdkVariable);
+                _resultsServer.InitializeResultsClient(new Uri(resultsReceiverEndpoint), liveConsoleFeedUrl, accessToken, Convert.ToBoolean(resultsUseSdkVariable?.Value));
```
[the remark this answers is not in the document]
yacaovsnc: Is this a Runner helper method?  Does it handle null values?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -161,73 +164,164 @@ private async Task JobLogUploadCompleteAsync(string planId, string jobId, long l
...
-            };
-
-            if (blobStorageType == BlobStorageTypes.AzureBlobStorage)
-            {
-                request.Content.Headers.Add(Constants.AzureBlobTypeHeader, Constants.AzureBlockBlob);
+                throw new Exception($"SAS url is empty");
             }
 
-            using (var response = await SendAsync(request, HttpCompletionOption.ResponseHeadersRead, userState: null, cancellationToken))
+            var blobUri = new Uri(url);
```
TingluoHuang, line 174: maybe use `URLbuilder` to help you play with URI?
what are we trying to do here?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -251,23 +345,22 @@ public async Task UploadStepSummaryAsync(string planId, string jobId, Guid stepI
...
             {
-                var response = await UploadBlockFileAsync(uploadUrlResponse.SummaryUrl, uploadUrlResponse.BlobStorageType, fileStream, cancellationToken);
+                await UploadBlockFileAsync(uploadUrlResponse.SummaryUrl, uploadUrlResponse.BlobStorageType, fileStream, cancellationToken);
             }
 
             // Send step summary upload complete message
             await StepSummaryUploadCompleteAsync(planId, jobId, stepId, fileSize, cancellationToken);
         }
 
-        private async Task<HttpResponseMessage> UploadLogFile(string file, bool finalize, bool firstBlock, string sasUrl, string blobStorageType,
```
TingluoHuang, line 261: we we no longer need to return ResponseMessage?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -422,6 +514,9 @@ public static class Constants
         public static readonly string ResultsProtoApiV1Endpoint = "twirp/github.actions.results.api.v1.WorkflowStepUpdateService/";
         public static readonly string WorkflowStepsUpdate = ResultsProtoApiV1Endpoint + "WorkflowStepsUpdate";
 
+        public static readonly int DefaultNetworkTimeoutInSeconds = 30;
+        public static readonly int DefaultBlobUploadRetries = 3;
```
TingluoHuang, line 518: do you want to hard code these in the runner or pass it down from the service as variable?

TingluoHuang: it depends on whether you plan to tweak these in the future.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
yacaovsnc: Correct, we were never reading response anyway.  We always throw exceptions if something's wrong.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
yacaovsnc: Hard code those seems okay.  We have 100s timeout hardcoded before this as well.  

Maybe we could override those with env variables, passing down from server maybe a bit overkill?

TingluoHuang: i am fine with either way, it's up to you. 😄

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
yacaovsnc: The SDK requires the blob uri and the credentials as constructor argument to create the blob client, the service returns a single SAS url from the server.  So we must parse the response from Results service and separate the blob path and the SAS params.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
yacaovsnc: I am going to leave this at 30s for now to collect more data.  We can always revert back to the current behavior with the feature flag.  Since this is only used to upload logs and summaries, I think 30s should be plenty time.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
TingluoHuang: it's up to you, but with UriBuilder, you can just set `Query=empty` and get the Uri object without manually construct `var blobPath = $"{blobUri.Scheme}://{blobUri.Authority}/{blobUri.AbsolutePath}";`

TingluoHuang (dismissed):

TingluoHuang (approved):

Issue #3075, opened 2024-01-02
--check should test live logs


congyiwu:
**Describe the bug**
I was looking at a support ticket where the customer is trying to set up DNS-based firewall rules and can't get live logs working b/c `results-receiver.actions.githubusercontent.com` is blocked (they manually tested w/ `curl`).

`--check` did not detect this b/c that domain isn't tested in https://github.com/actions/runner/blob/main/src/Runner.Listener/Checks/ActionsCheck.cs

**Expected behavior**
`--check` should test the domain for the live logs API too

Version of your runner?
2.311.0 on Ubuntu 22.04. Note that I'm actually running this on a GitHub-Hosted Runner, to see if it will work for troubleshooting firewall rules w/ [private networking](https://docs.github.com/en/actions/using-github-hosted-runners/connecting-to-a-private-network/about-private-networking-with-github-hosted-runners)

OS of the machine running the runner? Linux

## What's not working?
Please include error messages and screenshots.

## Job Log Output
If applicable, include the relevant part of the job / step log output here.  All sensitive information should already be masked out, but please double-check before pasting here.

## Runner and Worker's Diagnostic Logs
If applicable, add relevant diagnostic log information.  Logs are located in the runner's `_diag` folder. The runner logs are prefixed with `Runner_` and the worker logs are prefixed with `Worker_`. Each job run correlates to a worker log.  All sensitive information should already be masked out, but please double-check before pasting here.

congyiwu:
Closing as fixed by #3075

PR #3078, opened 2024-01-03
Extend `--check` to check Results-Receiver service.

TingluoHuang:
Add new check follow by https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github

Fixing: https://github.com/actions/runner/issues/3075


thboop (approved):
lgtm

PR #3083, opened 2024-01-08
Remove USE_BROKER_FLOW usage

luketomlinson:
This `USE_BROKER_FLOW ` flag allowed client-side redirects. It should not be needed anymore so removing for now.

### DIFF
--- a/src/Runner.Sdk/Util/VssUtil.cs
+++ b/src/Runner.Sdk/Util/VssUtil.cs
@@ -62,11 +62,6 @@ namespace GitHub.Runner.Sdk
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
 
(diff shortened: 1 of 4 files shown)


JoannaaKL (approved):

src/Runner.Sdk/Util/VssUtil.cs
```diff
@@ -62,11 +62,6 @@ public static void InitializeVssClientSettings(List<ProductInfoHeaderValue> addi
                 settings.SendTimeout = TimeSpan.FromSeconds(Math.Min(Math.Max(httpRequestTimeoutSeconds, 100), 1200));
             }
 
-            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("USE_BROKER_FLOW")))
```
sethrylan, line 65: Should we also remove in the Runner.Listener?

https://github.com/actions/runner/blob/3f3d9b0d99755c1d329c01328f9eab26a955be65/src/Runner.Listener/MessageListener.cs#L186

luketomlinson: yep. missed that

PR #3102, opened 2024-01-18
Fix release workflow.

TingluoHuang:
Fix regression caused by https://github.com/actions/runner/pull/3074


winlover32 (approved):

diogotorres97:
@TingluoHuang @thboop @winlover32 can we have a release (patch) with the revert container hook and latest changes?

thboop (approved):
lgtm

PR #3103, opened 2024-01-18
Implement Broker Redirects for Session and Messages

luketomlinson:


src/Runner.Listener/MessageListener.cs
```diff
@@ -98,6 +102,14 @@ public async Task<Boolean> CreateSessionAsync(CancellationToken token)
                                                         taskAgentSession,
                                                         token);
 
+                    // if (_session.SessionMigrationURI != null)
```
luketomlinson, line 105: sessions not implemented yet on broker

src/Runner.Listener/MessageListener.cs
```diff
@@ -228,6 +240,24 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
+                        await brokerServer.ConnectAsync(migrationMessage.BrokerBaseUrl, _creds);
+                        message = await brokerServer.GetRunnerMessageAsync(token, 
+                                                                        _session.SessionId, 
+                                                                        runnerStatus, 
+                                                                        BuildConstants.RunnerPackage.Version,
+                                                                        VarUtil.OS, 
+                                                                        VarUtil.OSArchitecture, 
+                                                                        _settings.DisableUpdate);
+                    }
+
```
pje, line 260: I'm admittedly new to the Runner code, but don't we need to save some state so that _all future `get /message`_ requests _also_ get routed to the broker?

luketomlinson: That's a great question. I actually am intentionally _NOT_ doing that. This is so that we can roll the feature flag forward and back as necessary and the changes get reflected ASAP.

src/Runner.Listener/MessageListener.cs
```diff
@@ -228,6 +240,24 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
+                    
+                    if (message != null && message.MessageType == BrokerMigrationMessage.MessageType)
+                    {
+                        Trace.Info("BrokerMigration message received. Polling Broker for messages...");
+
+                        var migrationMessage = JsonUtility.FromString<BrokerMigrationMessage>(message.Body);                        
+                        var brokerServer = HostContext.GetService<IBrokerServer>();
+
+                        await brokerServer.ConnectAsync(migrationMessage.BrokerBaseUrl, _creds);
+                        message = await brokerServer.GetRunnerMessageAsync(token,
```
pje, line 252: If the runner ever polls with the wrong sessionID for its runnerID, we'd planned for Broker to respond [403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). Am I right that that 403 would be caught [here in BrokerHttpClient](https://github.com/actions/runner/blob/40f813e0fb2a3b500858c907723f3efabc9d979b/src/Sdk/WebApi/WebApi/BrokerHttpClient.cs#L123-L126)?

pje: If so, Runner should try to recreate the session after a 403. (Maybe that already happens somewhere in Runner's code, like I said I'm new to the runner.)

src/Runner.Listener/MessageListener.cs
pje: Ah, right, so _every_ time the runner polls it'll be redirected to Broker like this.

src/Runner.Listener/MessageListener.cs
luketomlinson: If we get a 403, the runner will exit and restart and start the whole `MessageListener` flow over again, thus attempting to create a new sessions

pje: @luketomlinson Just a heads up that we actually have the broker responding `409 Conflict` in this case: https://github.com/github/actions-broker-listener/pull/1031/files#diff-37c615ad33e1d15f935093bd9fc81ccecce06ed47240d27cacce9ced12d79ee3R303

src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -74,6 +74,7 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                 try
                 {
                     message = await _brokerServer.GetRunnerMessageAsync(_getMessagesTokenSource.Token,
+                                                                        null,
```
pje, line 77: We should add the sessionID here, right?

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -111,5 +117,33 @@ public class BrokerHttpClient : RawHttpClientBase
...
+
+            if (result.IsSuccess)
+            {
+                return result.Value;
+            }
+
+            if (result.StatusCode == HttpStatusCode.Forbidden)
+            {
+                throw new AccessDeniedException(result.Error);
+            }
```
pje, line 144: Broker responds with 401s or 400s right now, no 403s: https://github.com/github/actions-broker-listener/blob/eb08f036c89ce6e385e62ddbeae5f84074d7d94c/internal/middleware/jwt_validation_middleware.go#L47-L65

pje: Also! Broker responds `409 Conflict` when there's already an active session. https://github.com/github/actions-broker-listener/pull/1031/files#diff-37c615ad33e1d15f935093bd9fc81ccecce06ed47240d27cacce9ced12d79ee3R303

src/Runner.Common/BrokerServer.cs
```diff
@@ -44,11 +45,20 @@ private void CheckConnection()
...
         }
 
-        public Task<TaskAgentMessage> GetRunnerMessageAsync(CancellationToken cancellationToken, TaskAgentStatus status, string version, string os, string architecture, bool disableUpdate)
+        public Task<TaskAgentSession> CreateSessionAsync(CancellationToken cancellationToken, TaskAgentSession session)
+        {
+            CheckConnection();
+            var jobMessage = RetryRequest<TaskAgentSession>(
+                async () => await _brokerHttpClient.CreateSessionAsync(session, cancellationToken), cancellationToken);
+
+            return jobMessage;
```
pje, line 54:
```suggestion
            var session = RetryRequest<TaskAgentSession>(
                async () => await _brokerHttpClient.CreateSessionAsync(session, cancellationToken), cancellationToken);

            return session;
```

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -111,5 +117,33 @@ public class BrokerHttpClient : RawHttpClientBase
...
+            var result = await SendAsync<TaskAgentSession>(
+                new HttpMethod("POST"),
+                requestUri: requestUri,
+                content: requestContent,
+                cancellationToken: cancellationToken);
+
+            if (result.IsSuccess)
+            {
+                return result.Value;
+            }
```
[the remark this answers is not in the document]
luketomlinson: Yeah deep in the chain it uses:
```
//
        // Summary:
        //     Gets a value that indicates if the HTTP response was successful.
        //
        // Returns:
        //     true if System.Net.Http.HttpResponseMessage.StatusCode was in the range 200-299;
        //     otherwise, false.
        public bool IsSuccessStatusCode { get; }
 ```

So 201 should be fine

File: src/Runner.Listener/BrokerMessageListener.cs (hunk 4 above)
Comment (luketomlinson) [reply src/Runner.Listener/BrokerMessageListener.cs:77]: yes. Though I'm kind of thinking BrokerMessageListener might eventually be deprecated.

File: src/Runner.Listener/BrokerMessageListener.cs [hunk 8]
```diff
@@ -36,13 +44,129 @@ public override void Initialize(IHostContext hostContext)
 
         public async Task<Boolean> CreateSessionAsync(CancellationToken token)
         {
-            await RefreshBrokerConnection();
-            return await Task.FromResult(true);
+            Trace.Entering();
```
luketomlinson, line 47: this is all copied from MessageListener. Eventually, one of these will be deprecated, but keeping them in sync for now

src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -196,6 +321,72 @@ private bool IsGetNextMessageExceptionRetriable(Exception ex)
             }
         }
 
+        private bool IsSessionCreationExceptionRetriable(Exception ex)
```
luketomlinson, line 324: also copied

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
luketomlinson: 👍 Runner already handles the 409 conflict exception

lokesh755 (approved):

PR #3113, opened 2024-01-26
Make embedded timeline record has same order as its parent record.

TingluoHuang:
https://github.com/github/actions-results-team/issues/1871


thboop (approved):
lgtm nice catch

PR #3114, opened 2024-01-29
Upload the diagnostic logs to the Results Service

jtamsut:
## Background 

This change uploads diagnostic logs to the Results Service. This is done in two steps:

1. First a signed URL (SAS URL) must be obtained from results 
2. Using the scoped URL the runner will upload the diagnostic log to blob storage

This work follows similar patterns to those uploading completed job logs.

## Related

Closes https://github.com/github/actions-results-team/issues/2185


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -968,6 +968,8 @@ public void QueueAttachFile(string type, string name, string filePath)
             }
 
             _jobServerQueue.QueueFileUpload(_mainTimelineId, _record.Id, type, name, filePath, deleteSource: false);
+
+            _jobServerQueue.QueueResultsUpload(_record.Id, name, filePath, type, deleteSource: false, finalize: true, firstBlock: true, totalLines: 0);
```
TingluoHuang, line 972: i would suggest don't change this since you are mixing Results related method into a Actions service method.

Here is what i would suggest:
- In `ExecutionContext`, add `void QueueDiagnosticLogFile(string name, string filePath);` to upload diag log only to Results, similar to how the existing `void QueueSummaryFile(string name, string filePath, Guid stepRecordId);` set up.
- In `DiagnosticLogManager`, call the `ExecutionContext.QueueDiagnosticLogFile()`

This way you get clear isolation for Actions service vs. Results, and easy for us to delete legacy code later.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -81,6 +81,32 @@ private async Task<GetSignedStepLogsURLResponse> GetStepLogUploadUrlAsync(string
             return await GetResultsSignedURLResponse<GetSignedStepLogsURLRequest, GetSignedStepLogsURLResponse>(getStepLogsSignedBlobURLEndpoint, cancellationToken, request);
         }
 
+        private async Task<GetSignedDiagnosticLogsURLResponse> GetSignedDiagnosticLogsURL(string planId, string jobId, CancellationToken cancellationToken)
```
robherley, line 84: I think this and the method below it are doing the same thing. This one can be deleted

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -982,6 +983,19 @@ public void QueueSummaryFile(string name, string filePath, Guid stepRecordId)
             _jobServerQueue.QueueResultsUpload(stepRecordId, name, filePath, ChecksAttachmentType.StepSummary, deleteSource: false, finalize: true, firstBlock: true, totalLines: 0);
         }
 
+        public void QueueDiagnosticLogFile(string name, string filePath)
+        {
+            ArgUtil.NotNullOrEmpty(name, nameof(name));
+            ArgUtil.NotNullOrEmpty(filePath, nameof(filePath));
+
```
TingluoHuang, line 990:
```suggestion
```

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -982,6 +983,19 @@ public void QueueSummaryFile(string name, string filePath, Guid stepRecordId)
...
+
+            if (!File.Exists(filePath))
+            {
+                throw new FileNotFoundException($"Can't upload diagnostic log file: {filePath}. File does not exist.");
+            }
+            _jobServerQueue.QueueResultsUpload(_record.Id, name, filePath, CoreAttachmentType.DiagnosticLog, deleteSource: false, finalize: true, firstBlock: true, totalLines: 0);
```
TingluoHuang, line 996: can you create a new type? 🙇 
ex: `CoreAttachmentType.ResultDiagnosticLog = "Results.Core.DiagnosticLog"`

src/Runner.Common/JobServerQueue.cs
```diff
@@ -551,6 +551,10 @@ private async Task ProcessResultsUploadQueueAsync(bool runOnce = false)
                             {
                                 await UploadSummaryFile(file);
                             }
+                            if (string.Equals(file.Type, CoreAttachmentType.DiagnosticLog, StringComparison.OrdinalIgnoreCase))
```
TingluoHuang, line 554: change this once you create the `CoreAttachmentType.ResultsDiagnosticLog`

TingluoHuang (approved):

PR #3119, opened 2024-01-30
Add `Content-Type` to BlockBlob upload

bethanyj28:
The default content type for BlockBlob uploads is `application/octet-stream`, which signals to many browsers that they should download the content. This adds the content type of `text/plain` to hint that the browser should open the content in a new tab.

Closes https://github.com/github/actions-results-team/issues/2137


src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -234,6 +234,7 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
                 if (blobStorageType == BlobStorageTypes.AzureBlobStorage)
                 {
                     request.Content.Headers.Add(Constants.AzureBlobTypeHeader, Constants.AzureBlockBlob);
+                    request.Content.Headers.Add("Content-Type", "text/plain");
```
yacaovsnc, line 237: Do we need to update the SDK path?  Or that's handled automatically for us?  In that case, once we rollout the SDK change, does this become a none issue?

bethanyj28: Added to the sdk case, thank you!

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -214,9 +214,16 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
             if (m_useSdk && blobStorageType == BlobStorageTypes.AzureBlobStorage)
             {
                 var blobClient = GetBlobClient(url);
+								var uploadOptions = new BlobUploadOptions
```
bethanyj28, line 217: Weird spacing - will fix

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -214,9 +214,16 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
...
+                {
+                    HttpHeaders = new BlobHttpHeaders
+                    {
+                        ContentType = "text/plain"
+                    }
+                };
                 try
                 {
-                    await blobClient.UploadAsync(file, cancellationToken);
+                    await blobClient.UploadAsync(file, uploadOptions, cancellationToken);
```
TingluoHuang, line 226: do you need to check what is the default option is and merge with it instead of just overwrite it?

TingluoHuang: not quite sure i follow your new change.

my initial comment is basically try to make sure the SDK don't have any special initialized `BlobUploadOptions`.

Based on the source code, maybe this is what you want?

Not sure... we can chat more...
```
await blobClient.UploadAsync(file, new BlobUploadOptions()
                    {
                        HttpHeaders = new BlobHttpHeaders
                        {
                            ContentType = "text/plain"
                        },
                        Conditions = new BlobRequestConditions
                        {
                            IfNoneMatch = new ETag("*")
                        }
                    }, cancellationToken);
```

bethanyj28: Gotcha! That makes sense to me. I was thinking specifically around the HTTP headers but I do recall seeing that in the source code of the SDK when I looked.

jtamsut:
@bethanyj28 How are we handling blob downloads that actually should have a content type of `application/octet-stream`? For example, we want this for downloading raw logs.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -214,9 +214,15 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
             if (m_useSdk && blobStorageType == BlobStorageTypes.AzureBlobStorage)
             {
                 var blobClient = GetBlobClient(url);
+                var uploadOptions = new BlobUploadOptions();
+                string extension = System.IO.Path.GetExtension(file.Name);
+                if (extension == ".txt")
+                {
+                    uploadOptions.HttpHeaders.ContentType = "text/plain";
+                }
```
bethanyj28, line 222: > @bethanyj28 How are we handling blob downloads that actually should have a content type of `application/octet-stream`? For example, we want this for downloading raw logs.

@jtamsut this is the way we chatted about earlier, let me know if you have something else in mind.

jtamsut: Ah I missed this. Okay this looks good to me!

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
bethanyj28: Found this issue which was helpful for wrapping my head around the options: https://github.com/Azure/azure-sdk-for-net/issues/9823. Thanks for calling it out!

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -214,9 +214,23 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
             if (m_useSdk && blobStorageType == BlobStorageTypes.AzureBlobStorage)
             {
                 var blobClient = GetBlobClient(url);
+                var httpHeaders = new BlobHttpHeaders();
+                var conditions = new BlobRequestConditions
+                {
+                    IfMatch = new ETag("*")
+                };
+                string extension = System.IO.Path.GetExtension(file.Name);
+                if (extension == ".txt")
```
TingluoHuang, line 223: is this check required, since we don't have this in the non-SDK version?

bethanyj28: This is needed to support https://github.com/actions/runner/pull/3114 since those should be of type `application/octet-stream`. I didn't add it to the non-sdk version since that should be rolled out by the time the diagnostic change is out, but happy to add to both if I misunderstood! CC @yacaovsnc

TingluoHuang: I am not quite sure whether we use `.txt` in the runner, ex: are those log file ends up with `.txt`?

bethanyj28: Yep! Our log files our `.txt` (that's the `job-logs.txt`): https://github-grid.enterprise.slack.com/files/U02V5TYNARX/F06GB94RRK4/-.txt

Also have validated the change that it does set the content-type correctly.

The diagnostic logs are `.zip`

bethanyj28: Chatted offline and updated to not use the runner file extension, but the blob url instead!

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -227,9 +227,22 @@ private async Task UploadBlockFileAsync(string url, string blobStorageType, File
             if (m_useSdk && blobStorageType == BlobStorageTypes.AzureBlobStorage)
             {
                 var blobClient = GetBlobClient(url);
+                var httpHeaders = new BlobHttpHeaders();
+                var conditions = new BlobRequestConditions
+                {
+                    IfNoneMatch = new ETag("*")
+                };
+                if (url.Contains(".txt")) // check if blob path is .txt file
```
TingluoHuang, line 235: is it possible to let the caller to decide what content type to use?
Ex: 
```
private async Task UploadBlockFileAsync(string url, string blobStorageType, FileStream file, bool plainTextContent, CancellationToken cancellationToken)
```

TingluoHuang (approved):

TingluoHuang (approved):

Issue #2906, opened 2023-10-04
Composite Actions using Node 20 don't work with CentOS 7


ajschmidt8:
## Problem

As mentioned in the GitHub blog post below, Actions is currently transitioning from Node 16 to Node 20.

- https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

However, Node 20 seems to be incompatible with CentOS 7, which is listed as a compatible OS for self-hosted runners below:

- https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#linux

This incompatibility seems to stem from the fact that NodeJS version 20 supports `glibc >=2.28` (as seen in the link below), but CentOS 7 only has `glibc==2.17`.

- https://github.com/nodejs/node/blob/v20.0.0/BUILDING.md#platform-list

The consequence of this is that composite actions that have been updated to use Node 20 no longer work with CentOS 7 host machines or containers.

Here is an example of the error message that results from trying to run a composite action using Node 20 on a CentOS 7 container:

- https://github.com/rapidsai/dask-cuda/actions/runs/6402970861/job/17380789311?pr=1255#step:19:8

As a result of this incompatibility, we've had to intentionally downgrade our composite actions to a version that uses Node 16 so that our workflows continue to work.

## To Reproduce
Run the following workflow to exhibit the error message

```yaml
name: centos7

on:
  push:

jobs:
  run:
    runs-on: ubuntu-latest
    container: centos:7
    steps:
      - uses: actions/checkout@v4
      - run: echo "hello, world"
```
![image](https://github.com/actions/runner/assets/7400326/fb0ca403-beae-4944-9a0e-636c1da3581b)

## Expected behavior
Since the documentation below states that self-hosted runners support CentOS 7, I would expect the above workflow to run to completion without any errors

- https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#linux

## Runner Version and Platform
Ubuntu 22.04, runner version `2.309.0`

firewave:
This also affects `ubuntu:16:04` (https://github.com/danmar/cppcheck/actions/runs/7677304162/job/20926809147?pr=5917) and `ubuntu:18.04` (https://github.com/danmar/cppcheck/actions/runs/7677304162/job/20926809224?pr=5917).

TomasNovo:
hey @ajschmidt8 ,
[...]

PR #3128, opened 2024-02-05
use a glibc 2.17 compatible node20 build on x86_64 (better idea inside!)

reaperhulk:
GitHub uses official nodejs.org builds in most cases, but also uses the "unofficial node" builds (which are also supplied by node.js) for win-arm64. This PR uses the unofficial node.js build for 20.x that links against glibc 2.17. This restores compatibility with older Linux distributions (CentOS7) that are still commonly in use.

While this "fixes" the problem for x86_64, the issue also exists for AArch64, and the correct solution is actually for GH to produce its own node builds using the `--fully-static` flag. GitHub could produce `--fully-static` builds for Linux x86_64 and AArch64 and then they would no longer need to worry about this problem (both old distributions and alpine).

Considering you're already building musl node builds via https://github.com/actions/alpine_nodejs please consider just altering this to use `--fully-static`, adding AArch64, and then using those binaries in all cases. To demonstrate this I've forked that repo and built a [fully static node.js](https://github.com/reaperhulk/static_nodejs/releases/tag/v20.8.1) which runs on all glibc distributions as well.

refs: https://github.com/actions/runner/issues/2906


esabol:
This is the ideal solution, and it's been sitting here since February. Does anyone from GitHub actually care about CI?

PR #3135, opened 2024-02-07
Process `snapshot` tokens

davidomid:
Closes: https://github.com/github/hosted-runners/issues/526 

Summary of changes: 

- Added deserialisation of a new `TemplateToken` called `snapshot` to `AgentJobRequestMessage`. 
- Added evaluation of `snapshot` from `TemplateToken` to a new `Snapshot` type. 
- Added a new `SnapshotOperationProvider` service, responsible for processing `Snapshot` objects. 
- Updated `JobExtension` to evaluate a `Snapshot` if one is present in the job template, then register a new post-job step for invoking `SnapshotOperationProvider` with it. 
- Added new L0 tests for `JobExtension` and `SnapshotOperationProvider`. 
- Updated existing tests for regression.


src/Runner.Worker/JobExtension.cs
```diff
@@ -404,6 +405,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                                                                           data: (object)jobHookData));
                     }
 
+                    // Register custom image creation post-job step if the "snapshot" token is present in the message.
```
ericsciple, line 408: Can you confirm the order? I would imagine the post steps run in the following order:
- Snapshot
- Completed hook
- Job cleanup

src/Runner.Worker/JobExtension.cs
```diff
@@ -17,6 +17,7 @@
 using GitHub.Runner.Common.Util;
 using GitHub.Runner.Sdk;
 using GitHub.Services.Common;
+using Newtonsoft.Json;
```
ericsciple, line 20: curious is this used?

ericsciple (dismissed):
Thanks! This makes sense to me. Although I would like to get a second review from someone who works in the runner more recently.

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -346,6 +346,39 @@ internal static ContainerRegistryCredentials ConvertToContainerCredentials(Templ
             return result;
         }
 
+        internal static Snapshot ConvertToJobSnapshotRequest(TemplateContext context, TemplateToken token)
```
TingluoHuang, line 349: are these changes copied from the service?
we we expect the golang parser handles this?

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -71,7 +71,8 @@
           "env": "job-env",
           "outputs": "job-outputs",
           "defaults": "job-defaults",
-          "steps": "steps"
+          "steps": "steps",
+          "snapshot": "snapshot"
```
TingluoHuang, line 75: is this schema change coming from the actions-service or the parser?
i am trying to find the source and make sure they match each other.

src/Runner.Worker/JobExtension.cs
```diff
@@ -404,6 +405,18 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                                                                           data: (object)jobHookData));
                     }
 
+                    // Register custom image creation post-job step if the "snapshot" token is present in the message.
+                    var snapshotRequest = templateEvaluator.EvaluateJobSnapshotRequest(message.Snapshot, jobContext.ExpressionValues, jobContext.ExpressionFunctions);
```
TingluoHuang, line 409: i think this needs to move up before we calling `// Register Job Completed hook if the variable is set`

basically, snapshot first, then `job_complete_hook`

src/Runner.Worker/SnapshotOperationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    public async Task CreateSnapshotRequestAsync(IExecutionContext executionContext, Snapshot snapshotRequest)
+    {
+        var snapshotRequestFilePath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Root), ".snapshot", "request.json");
+        var snapshotRequestDirectoryPath = Path.GetDirectoryName(snapshotRequestFilePath);
+        if (snapshotRequestDirectoryPath != null)
+        {
+            Directory.CreateDirectory(snapshotRequestDirectoryPath);
+        }
+
+        var snapshotRequestJson = JsonConvert.SerializeObject(snapshotRequest);
```
TingluoHuang, line 27: you can use `StringUtil.ConvertToJson()` instead of `JsonConvert.SerializeObject`

src/Runner.Worker/SnapshotOperationProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+public interface ISnapshotOperationProvider : IRunnerService
+{
+    Task CreateSnapshotRequestAsync(IExecutionContext executionContext, Snapshot snapshotRequest);
+}
+
+public class SnapshotOperationProvider : RunnerService, ISnapshotOperationProvider
+{
+    public async Task CreateSnapshotRequestAsync(IExecutionContext executionContext, Snapshot snapshotRequest)
+    {
+        var snapshotRequestFilePath = Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Root), ".snapshot", "request.json");
```
TingluoHuang, line 20: is the runner root directory the right place for putting this file?

or the WellKnownDirectory.Temp might be better since it get cleanup before each job.

src/Runner.Worker/SnapshotOperationProvider.cs
TingluoHuang: you can also just use `IOUtil.SaveObject()` to do the josn serialization and write to disk in one line.

src/Runner.Worker/JobExtension.cs
davidomid: Apologies, it was used originally before I migrated the logic to `SnapshotOperationProvider` and I forgot to remove the reference.

src/Sdk/DTPipelines/workflow-v1.0.json
davidomid: It was added to the C# parser in this PR: https://github.com/github/actions-workflow-parser/pull/406

The Go parser hasn't been updated yet, that's happening in a separate batch: https://github.com/github/hosted-runners/issues/528

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
davidomid: I **_think_** the answer to these questions is similar to this comment: https://github.com/actions/runner/pull/3135/files#r1489319642

But please let me know if you need additional information. 

For context, we're unsure at this stage if we'll need expression support at some point and @ericsciple advised that it made sense to send the entire `TemplateToken` down to the runner and process it there instead of sending an evaluated `Snapshot` object. This was recommended as we would have the ability to check for expression tokens in the future. I'm happy to make changes if this wasn't the optimal approach though.

src/Runner.Worker/SnapshotOperationProvider.cs
davidomid: Thanks for the advice! 

I opted for just using `IOUtil.SaveObject`. Originally I thought it made sense to have the operations separated as I was logging the snapshot object JSON afterwards but I ended up removing that as it seemed redundant. The behaviour should be adequately tested by the L0 tests.

src/Runner.Worker/JobExtension.cs
davidomid: Thanks, I think you're right. I'll go ahead and move that up.

src/Runner.Worker/JobExtension.cs
davidomid: @TingluoHuang pointed out that the order doesn't seem correct, so I've addressed that concern here: https://github.com/actions/runner/pull/3135/files#r1486509000

ericsciple (approved):
Thanks! This makes sense to me. Will defer to Ting whether he has any remaining concerns.

PR #3139, opened 2024-02-08
Remove dockerd but leave docker cli in the image. Bump docker version

nikola-jokic:
To avoid alerts raised by scans, this change removes all other utilities except for docker cli. Other utilities are not used, since docker socket is mounted to the image to be executed in `dind` mode.

Also, docker version is bumped.


images/Dockerfile
```diff
@@ -5,7 +5,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG RUNNER_VERSION
 ARG RUNNER_CONTAINER_HOOKS_VERSION=0.5.1
-ARG DOCKER_VERSION=25.0.2
+ARG DOCKER_VERSION=25.0.3
```
nikola-jokic, line 8: This is just a bump, and can be excluded. Without this bump the vulnerability is not reported, but I thought why not use the latest patch version when we are here :relaxed:

Issue #2380, opened 2023-01-18
safe_sleep.sh inadvertently top 100% CPU 


leleobhz:
**Describe the bug**

safe_sleep.sh is built as while loop aiming to be [safer than count on sleep or ping binaries](https://github.com/actions/runner/issues/1673). While loop the way function was implemented leads to 100% CPU usage without try another alternatives (Eg. Testing if sleep is available).

**To Reproduce**
1. Start runner
2. Wait

**Expected behavior**

A lesser CPU usage from runner while idle.

**Actual behavior**

![image](https://user-images.githubusercontent.com/201189/213169313-9b8c1c85-d436-4d8d-af3d-3b1b150d8ae1.png)

## Runner Version and Platform
Linux 2.300.2 running on Docker.

fhammerl:
Thanks @leleobhz, labeled the issue.

`good-first-issue` `keep`

This barebones implementation for `safe_sleep.sh` was chosen for its broad availability on most platforms.

A better 'safe sleep' must
- NOT use 100% CPU
- Be compatible with all platforms the runner supports (some installations might not have `sleep` or `ping`)

A good start would be to check for the existence of such utils before resorting to the above CPU intensive counting

LiranBarton:
Hey @fhammerl, was this bug fixed? 

We have runners in ASG that scales to MAX because of this script.
Any suggestions ? (did not fully understand why we need this script - can we remove it on boot?)

leleobhz:
> We have runners in ASG that scales to MAX because of this script. Any suggestions ? (did not fully understand why we need this script - can we remove it on boot?)

@LiranBarton hello!

This script as far I understood the reasons aims to have a minimal `sleep()` implementation for environments that does'nt have one using pure shell. 

Main questions here is this implementation lead's to CPU ocupation and may be uses only if sleep are not present on system, as last fallback. 

I suggest this way to handle sleep() issues if it's really needed (But personally I can't see why sleep does need a pure-shell implementation here since we have both bash or busybox that someway implement this function. Also, maybe embed busybox a better solution instead custom implementation).

tlhakhan:
I also noticed high CPU utilization on my nodes as well, and it ended up being the `safe_sleep.sh` script.  See below of a snippet of my `top` output.

```
top - 00:34:09 up 17 days, 16:44,  2 users,  load average: 4.13, 4.24, 4.16
Tasks: 380 total,   5 running, 375 sleeping,   0 stopped,   0 zombie
[...]

PR #3143, opened 2024-02-10
Reduce CPU usage of safe_sleep.sh

fahhem:
Addresses https://github.com/actions/runner/issues/2380

### DIFF
--- a/src/Misc/layoutroot/safe_sleep.sh
+++ b/src/Misc/layoutroot/safe_sleep.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+if [ -x "$(command -v sleep)" ]; then
+  sleep $1
+  exit 0
+fi
+if [ -x "$(command -v ping)" ]; then
+  ping -c $1 127.0.0.1 > /dev/null
+  exit 0
+fi
+
 SECONDS=0
 while [[ $SECONDS != $1 ]]; do
     :


Dr-Emann:
@TjlHope had a nice fallback that only relies on bash, so could replace the rest of the script if `sleep` or `ping` isn't available: https://github.com/actions/runner/issues/3792#issuecomment-3289007717

fahhem:
Nobody from the project has reviewed this, so I'm not going to change things until I get a sense of what they expect here. If they prefer the bash-only option from #3792, then I'm happy to switch to that

ephraimduncan (approved):

5HT:
So it is still not reduced :-)

src/Misc/layoutroot/safe_sleep.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+if [ -x "$(command -v sleep)" ]; then
+  sleep $1
+  exit 0
+fi
+if [ -x "$(command -v ping)" ]; then
+  ping -c $1 127.0.0.1 > /dev/null
```
z1bbo: should this be `ping -c $(( $1 + 1 )) 127.0.0.1 > /dev/null`?
you need to add one because the first ping is executed immediately

fahhem: You're right, here's the patch so I can merge it from this comment

```suggestion
  # Add 1 since the first ping is executed immediately.
  ping -c $(( $1 + 1 )) 127.0.0.1 > /dev/null
```

AbdelrahmanHafez:
CI is waiting since 2024 for `SECONDS` to become `5000`, now we're at 57,865,643, but we'll get there, eventually.

Screenshot 2025-12-10 at 7 28 37 PM

src/Misc/layoutroot/safe_sleep.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 
+if [ -x "$(command -v sleep)" ]; then
+  sleep $1
```
Rudxain, line 4: [ShellCheck complains about this](https://www.shellcheck.net/wiki/SC2086)

src/Misc/layoutroot/safe_sleep.sh
```diff
@@ -1,6 +1,16 @@
...
+fi
+if [ -x "$(command -v ping)" ]; then
+  # Add 1 since the first ping is executed immediately.
+  ping -c $(( $1 + 1 )) 127.0.0.1 > /dev/null
+  exit 0
+fi
+
 SECONDS=0
-while [[ $SECONDS != $1 ]]; do
+while [[ $SECONDS -lt $1 ]]; do
```
Rudxain, line 14: [`+` and `-lt` are vulnerable to code-injection](https://mywiki.wooledge.org/BashProgramming/05#Arithmetic_Expansion) ([SC](https://github.com/koalaman/shellcheck/issues/3067))

Previously [mentioned here](https://github.com/actions/runner/pull/3157#issuecomment-3638892152)

src/Misc/layoutroot/safe_sleep.sh
Rudxain: BTW, it's also recommended to add `--` between the cmd and the arg. But since [the POSIX version doesn't accept any options](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sleep.html#tag_20_111_04), the `--` could make it crash on some systems

cubernetes:
What speaks against `read -t5`? Or some variation thereof, like `read -t5</dev/zero` or `read -t5<"$character_special_device_that_blocks_infinitely"`.

If something like that worked, you wouldn't need to rely on sleep at all, and it's been in bash since version 2.04-devel or sth (i.e., forever).

Rudxain:
https://github.com/actions/runner/pull/4146#issuecomment-3643997864

PR #3144, opened 2024-02-10
Run actions runner on FreeBSD Linux compatibility layer

ChanTsune:
I tried to run a self-hosted runner on the FreeBSD Linux compatibility layer. When I attempted to run it, I encountered an error stating that the 'tar' command does not exist.
After investigating the cause, I discovered that Directory.GetFiles was not functioning as expected in WhichUtil.Which. On FreeBSD's Linux compatibility layer, Directory.GetFiles does not work properly and consistently returns an empty array. Therefore, I implemented the equivalent functionality using File.Exists.

After overcoming this issue, the runner functions more or less as expected on FreeBSD Linux compatibility layer.

### DIFF
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -36,8 +36,8 @@ namespace GitHub.Runner.Sdk
             {
                 if (!string.IsNullOrEmpty(pathSegment) && Directory.Exists(pathSegment))
                 {
-                    string[] matches = null;
 #if OS_WINDOWS
+                    string[] matches = null;
                     string pathExt = Environment.GetEnvironmentVariable("PATHEXT");
                     if (string.IsNullOrEmpty(pathExt))
                     {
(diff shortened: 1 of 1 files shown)


thresheek:
Hi @ChanTsune !

While I'm not an actions developer, I am also interested in this merged for similar reasons.  The problem is that it no longer applies since https://github.com/actions/runner/commit/5107c5efb2b122903d95f59d492dfc9438419230.  I've adapted the patch and now my freebsd-based worker seems to be ok:

```
diff --git a/src/Runner.Sdk/Util/WhichUtil.cs b/src/Runner.Sdk/Util/WhichUtil.cs
index ef7683a..653d00b 100644
--- a/src/Runner.Sdk/Util/WhichUtil.cs
+++ b/src/Runner.Sdk/Util/WhichUtil.cs
@@ -95,13 +95,11 @@ namespace GitHub.Runner.Sdk
 #else
                     try
                     {
-                        foreach (var file in Directory.EnumerateFiles(pathSegment, command))
+                        string commandPath = Path.Join(pathSegment, command);
+                        if (File.Exists(commandPath) && IsPathValid(commandPath, trace))
                         {
-                            if (IsPathValid(file, trace))
-                            {
-                                trace?.Info($"Location: '{file}'");
-                                return file;
-                            }
+                            trace?.Info($"Location: '{commandPath}'");
+                            return commandPath;
                         }
                     }
                     catch (UnauthorizedAccessException ex)
```

Hope this helps

thresheek:
FYI v2.321.0 with that patch applied now needs linux_base-rl9 and segfaults on a start.

I think I'm giving up with this approach and will try to migrate our freebsd-based runners to github-act-runner instead.

Issue #2511, opened 2023-03-29
Runner (v2.303) not working on AmazonLinux2023 due to missing  ICU package


jalbstmeijer:
**Describe the bug**
The runner cannot be started due to a missing requirement.

Couldn't find a valid ICU package installed on the system

**To Reproduce**
1. start a AmazonLinux2023 AMI, for instance ami-00169914e6299b8e0 in eu-west-1
2. download and unpack actions-runner-linux-x64-2.303.0.tar.gz
3. run after config.sh, run.sh and see error

**Expected behavior**
Running runner

## Runner Version and Platform
runner-v2.303
AmazonLinux2023

## What's not working?

```
./run.sh

Libicu's dependencies is missing for Dotnet Core 6.0
Execute sudo ./bin/installdependencies.sh to install any missing Dotnet Core 6.0 dependencies.
Process terminated. Couldn't find a valid ICU package installed on the system. Please install libicu using your package manager and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.
   at System.Environment.FailFast(System.String)
   at System.Globalization.GlobalizationMode+Settings..cctor()
   at System.Globalization.CultureData.CreateCultureWithInvariantData()
   at System.Globalization.CultureData.get_Invariant()
   at System.Globalization.CultureInfo..cctor()
   at System.Globalization.CultureInfo.get_CurrentCulture()
   at System.Text.RegularExpressions.Regex.Init(System.String, System.Text.RegularExpressions.RegexOptions, System.TimeSpan, System.Globalization.CultureInfo)
   at System.Text.RegularExpressions.Regex..ctor(System.String, System.Text.RegularExpressions.RegexOptions, System.TimeSpan, System.Globalization.CultureInfo)
   at System.Text.RegularExpressions.Regex..ctor(System.String, System.Text.RegularExpressions.RegexOptions)
   at GitHub.Runner.Sdk.RunnerWebProxy..ctor()
   at GitHub.Runner.Common.HostContext..ctor(System.String, System.String)
   at GitHub.Runner.Listener.Program.Main(System.String[])
/tmp/actions-runner/run-helper.sh: line 36:  2220 Aborted                 (core dumped) "$DIR"/bin/Runner.Listener run $*
Exiting with unknown error code: 134
Exiting runner...

```

softlberton:
Hi @jalbstmeijer,

[...]

PR #3155, opened 2024-02-16
Parse os-release to identify OS by ID_LIKE field (i.e. make `install-dependencies.sh` work on Amazon Linux 2023)

stewartsmith:
The /etc/os-release file has evolved over the years, with current documentation being at
https://www.freedesktop.org/software/systemd/man/latest/os-release.html

Part of this spec is to also look at /usr/lib/os-release as image based Linux distributions that can exist without anything in /etc may put the os-release file there.

In the docs, Example 3 shows how to parse this file with shell. We can use that example and look at ID_LIKE to better determine if the OS is a Debian like system, or a Fedora like system, and differentiate between how RHEL / CentOS and friends are like Fedora and how non-RHEL, non-CentOS is also like Fedora.

The practical end-outcome of this is that it will now correctly detect Amazon Linux 2023, pick 'dnf' as the package manager, and install the dependencies.

Fixes: https://github.com/actions/runner/issues/2511

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -36,13 +36,19 @@ function print_rhel6errormessage()
     echo "https://github.com/dotnet/core/blob/master/Documentation/build-and-install-rhel6-prerequisites.md"
 }
 
-if [ -e /etc/os-release ]
+# As per https://www.freedesktop.org/software/systemd/man/latest/os-release.html
+# Snippet licensed under:
+# # SPDX-License-Identifier: MIT-0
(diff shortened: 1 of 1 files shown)


stewartsmith:
After reading the comments in https://github.com/actions/runner/issues/2511 more closely, this Pull Request probably also fixes the Amazon Linux 2 issue mentioned there, as it has `ID_LIKE="centos rhel fedora"` which would get matched.

softlberton:
LFG !!!

sjethvani (approved):

stewartsmith:
Any chance of this being merged?

stewartsmith:
> Any chance of this being merged?

I guess not :(

PR #3157, opened 2024-02-17
Update safe_sleep.sh for bug when scheduler is paused for more than 1 second

horner:
fixed a rare condition where the comparison may not be made within the second and then this will loop forever (or at least until a rollover of SECONDS)  ;-)

sleep was replaced in https://github.com/actions/runner/pull/1707

But if safe_sleep.sh is running when a vm is suspended it will not do the comparison within the second and therefore loops forever when the vm is resumed.

### DIFF
--- a/src/Misc/layoutroot/safe_sleep.sh
+++ b/src/Misc/layoutroot/safe_sleep.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 SECONDS=0
-while [[ $SECONDS != $1 ]]; do
+while [[ $SECONDS -lt $1 ]]; do
     :
 done


horner:
Screenshot 2024-02-17 at 11 19 18 AM

andrewrk:
> fixed a rare condition

I assume you mean "rare" as in, "not finished cooking" because it happens all the damn time

paralin:
Is looping infinitely really an effective sleep?

AXDOOMER:
Why are they not using the `sleep` command?

NicolasDorier:
I would suggest to just use `sleep` if available instead of doing something as extremely inefficient as that just so the 0.0001% of systems who doesn't have `sleep` can work.

Or better, just drop a `sleep` script in the `PATH` of the systems not supporting sleep by default... This might literally saves us from climate change.

bybatkhuu:
Came from "primeagen" content, holys**t it's real???
WOW!

symmetricalbyte:
_**“A remarkable demonstration of how far one can stretch the boundaries of conventional engineering wisdom.”**_

jcastrence:
Ironically named "safe_sleep.sh" 😭

Technically, I guess it is safer to simply _not run your code at all_.

Rudxain:
Sadly, this is now [vulnerable to code-injection](https://mywiki.wooledge.org/BashProgramming/05#Arithmetic_Expansion) ([see SC issue](https://github.com/koalaman/shellcheck/issues/3067)), because `!=` is a string comparison and `-lt` is a numeric comparison

> Bash is taking inspiration from PHP's insanity

latekvo:
@Rudxain nice catch, I tested it on my end, and it actually allows for RCE 

But as far as I see, `safe_sleep.sh` is never used with user input, so I don't think it's that serious. Still should be fixed in case it's exposed to the user in the future though.

ngxson:
```sh
read -t $1
```

No external program needed

msschambach:
> Came from "primeagen" content, holys**t it's real??? WOW!

Seems like a response to the vid 😄 👉🏾  https://github.com/actions/runner/pull/4146

andrzejsokolowski:
Doesn't this still take up an entire core of a CPU? Is this malicious code designed to waste as much processor compute time? Isn't this running this comparison billions of times a second? As other people pointed out in this thread, why not use sleep or other alternatives?

pasabanov:
> Isn't this running this comparison billions of times a second?

Not billions - bash is not that fast, but probably millions.
This still takes up an entire thread of the CPU.

> why not use sleep or other alternatives?

This is a reasonable question. They previously had code that checked for programs like `sleep` or `ping` to use them if available, but they replaced it with this abomination.

whoreson:
put me in the screenshot

lachesis:
Under what circumstance is `sleep` actually not available?

PR #3163, opened 2024-02-21
handle broker run service exception handling

yaananth:
Closes https://github.com/github/actions-runtime/issues/4348

There is no reason to log `Job server does not support handling unhandled exception yet`, specially if it's out of space, it's not going to help.

So made changes to skip it/remove it, we don't plan to do that for run-service today, we will circle back if we need to call `forcefailjob` everytime, if so we will just call it.


TingluoHuang (approved):

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -546,13 +546,27 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
-                                        await ForceFailJob(jobServer, message, detailInfo);
+                                        case IJobServer js:
+                                        {
+                                            await LogWorkerProcessUnhandledException(js, message, unhandledExceptionIssue);
+                                            // Go ahead to finish the job with result 'Failed' if the STDERR from worker is System.IO.IOException, since it typically means we are running out of disk space.
+                                            if (detailInfo.Contains(typeof(System.IO.IOException).ToString(), StringComparison.OrdinalIgnoreCase))
+                                            {
+                                                Trace.Info($"Finish job with result 'Failed' due to IOException.");
+                                                await ForceFailJob(js, message);
+                                            }
```
yaananth, line 561: same as before , no change here

src/Runner.Listener/JobDispatcher.cs
```diff
@@ -546,13 +546,27 @@ private async Task RunAsync(Pipelines.AgentJobRequestMessage message, string orc
...
+
+                                            break;
+                                        }
+                                        case IRunServer rs:
+                                            await ForceFailJob(rs, message, unhandledExceptionIssue);
```
TingluoHuang, line 566: only do this for `if (detailInfo.Contains(typeof(System.IO.IOException).ToString(), StringComparison.OrdinalIgnoreCase))` to match existing behavior?

yaananth: existing one already does failing job, inside `LogWorkerProcessUnhandledException`:

```
if (message.Variables.TryGetValue("DistributedTask.MarkJobAsFailedOnWorkerCrash", out var markJobAsFailedOnWorkerCrash) &&
                    StringUtil.ConvertToBoolean(markJobAsFailedOnWorkerCrash?.Value))
                {
                    Trace.Info("Mark the job as failed since the worker crashed");
                    jobRecord.Result = TaskResult.Failed;
                    // mark the job as completed so service will pickup the result
                    jobRecord.State = TimelineRecordState.Completed;
                }
```

yaananth: that ff is ON, so I think we still match with existing behavior overall

TingluoHuang (dismissed):

TingluoHuang (approved):

PR #3173, opened 2024-02-27
Add ARM64/AARCH64 Support to create-latest-svc.sh Script

roznai:
This pull request aims to enhance the `create-latest-svc.sh` script by adding support for ARM64 architecture.

The current script does not fully support setting up GitHub Actions runners on ARM64-based systems. This PR introduces changes that allow the script to correctly identify and work with ARM64 architecture, thereby enabling users to set up self-hosted runners on ARM64 machines.

Changes include:

- Added a condition to check if the system architecture is either `arm64` or `aarch64`.
- Based on the architecture, the appropriate runner package is downloaded and installed.

These changes will extend the usability of GitHub Actions runners to a wider range of systems and provide better support for ARM64 machines.


zikphil:
Any update? Seems pretty straight forward

Issue #3171, opened 2024-02-26
Incorrect order of steps and incomplete summary


AGelzer:
## Describe the bug
Summary has incorrect order of steps and is not complete for composite action (since February 22, 2024)

## To Reproduce
Run action with multiple steps and composite action.

[Workflow](https://github.com/AGelzer/actions-summary/blob/3d9db3b87809942c8f0bccbc5fc77b0fe39a707e/.github/workflows/summary-composite.yml)

[Composite Action (AGelzer/actions-summary-composite)](https://github.com/AGelzer/actions-summary-composite/blob/0b6b6e6c929c33718aa69ea1805df0d78f0085d3/action.yml)

See Job Run:
https://github.com/AGelzer/actions-summary/actions/runs/8046453616/job/21973698353

## Expected behavior
Summary:

```
### Step: 1. Before

### Composite Step: 1.

### Composite Step: 2.

### Composite Step: 3.

### Step: 1. After

### Step: 2. After

### Step: 3. After
```

## Runner Version and Platform
Current runner version: '2.313.0'
Ubuntu 22.04.4 LTS

## What's not working?
Summary has incorrect order of steps and is not complete for composite action

Summary:
```
### Composite Step: 1.

### Step: 1. Before
```

See Job Run:
https://github.com/AGelzer/actions-summary/actions/runs/8046453616/job/21973698353

## Job Log Output
<details>
<summary>Run Log</summary>
```
2024-02-26T09:19:49.6386442Z Requested labels: ubuntu-22.04
2024-02-26T09:19:49.6386786Z Job defined at: AGelzer/actions-summary/.github/workflows/summary-composite.yml@refs/heads/main
2024-02-26T09:19:49.6386918Z Waiting for a runner to pick up this job...
2024-02-26T09:19:50.3226455Z Job is waiting for a hosted runner to come online.
2024-02-26T09:19:54.3004791Z Job is about to start running on the hosted runner: GitHub Actions 19 (hosted)
2024-02-26T09:19:55.9601200Z Current runner version: '2.313.0'
2024-02-26T09:19:55.9624342Z ##[group]Operating System
2024-02-26T09:19:55.9624921Z Ubuntu
2024-02-26T09:19:55.9625393Z 22.04.4
2024-02-26T09:19:55.9625716Z LTS
2024-02-26T09:19:55.9626006Z ##[endgroup]
2024-02-26T09:19:55.9626503Z ##[group]Runner Image
[...]

PR #3174, opened 2024-02-27
fix summaries for actions results

SrRyan:
Addresses a bug that was reported around summaries where we don't respect the embedded Id on uploads. This can cause conflicts during the upload due to how the IDs are utilized by GitHub Actions.


yacaovsnc (approved):

TingluoHuang (approved):

TingluoHuang:
https://github.com/actions/runner/issues/3171

PR #3191, opened 2024-03-07
Handle new non-retryable exception type

thyeggman:
Add new code to handle additional non-retryable exceptions from Actions


sethrylan (approved):
LGTM

PR #3192, opened 2024-03-07
Add ability to enforce actions to run on node20

takost:
Add two ways to enforce actions execute on `node20`:
1. Add FF `DistributedTask.ForceGithubJavascriptActionsToNode20` for Actions Service side. When this FF will be turned on it will make all javascript actions run on `node20`. In this case we will show warning, that actions was enforced: 

image 
This enforcement can be opted-out by setting `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` env var to `true`

2. Add env variable `FORCE_JAVASCRIPT_ACTIONS_TO_NODE20` so customers can try to enforce actions execution on `node20`. They can try to set this var on workflow level or as local environment variable.

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -813,12 +813,6 @@ namespace GitHub.Runner.Worker
                 Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion, "node16");
             }
 
-            var forceActionsToNode20 = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.ManualForceActionsToNode20));
-            if ((Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode20") ?? false) || forceActionsToNode20)
-            {
-                Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedActionsNodeVersion, "node20");
-            }
-
             // Environment variables shared across all actions
             Global.EnvironmentVariables = new Dictionary<string, string>(VarUtil.EnvironmentVariableKeyComparer);
 
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -813,6 +813,12 @@ namespace GitHub.Runner.Worker
                 Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion, "node16");
             }
 
+            var forceActionsToNode20 = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.ManualForceActionsToNode20));
+            if ((Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode20") ?? false) || forceActionsToNode20)
+            {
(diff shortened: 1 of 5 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -813,6 +813,12 @@ public void InitializeJob(Pipelines.AgentJobRequestMessage message, Cancellation
                 Environment.SetEnvironmentVariable(Constants.Variables.Agent.ForcedInternalNodeVersion, "node16");
             }
 
+            var forceActionsToNode20 = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.ManualForceActionsToNode20));
```
ericsciple, line 816: Curious why we need an env var to control this? Is this for customers to have a way to test their changes?

ericsciple: Curious because this is checking machine process variables here, not `vars` or `env` from the workflow context

Wondering whether we expect customers to set an environment variable on their machines? Customers who use hosted runners will not have a way to do this.

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -84,6 +84,45 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
                     }
                     nodeData.NodeVersion = "node16";
                 }
+
+                var localForceActionsToNode20 = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.ManualForceActionsToNode20));
```
ericsciple, line 88: Wondering why we do the following two checks during `ExecutionContext.InitializeJob` and also here:
- `Environment.GetEnvironmentVariable(Constants.Variables.Agent.ManualForceActionsToNode20)`
- `Global.Variables.GetBoolean("DistributedTask.ForceGithubJavascriptActionsToNode20")`

Seems like having the logic either in InitializeJob or here would be sufficient. Do we need both?

Also, here we do an additionally check and prefer the workflow-level `env`, but we don't check that during `InitializeJob`. Should we add that to `InitializeJob`?

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -84,6 +84,45 @@ public sealed class HandlerFactory : RunnerService, IHandlerFactory
...
+                            {
+                                repoActionFullName = repoAction.Path; // local actions don't have a 'Name'
+                            }
+                            else
+                            {
+                                repoActionFullName = $"{repoAction.Name}/{repoAction.Path ?? string.Empty}".TrimEnd('/') + $"@{repoAction.Ref}";
+                            }
+
+                            warningActions.Add(repoActionFullName);
+                            executionContext.Global.Variables.Set("Node20ForceActionsWarnings", StringUtil.ConvertToJson(warningActions));
```
ericsciple, line 121: Should this use the const `EnforcedNode16DetectedAfterEndOfLifeEnvVariable`?

src/Runner.Worker/ExecutionContext.cs
takost: Ops, I missed checking workflow variable here. Fixed

src/Runner.Worker/Handlers/HandlerFactory.cs
takost: I reverted changes in `ExecutionContext`. We don't need them. So we need this logic only here for customer's actions

ericsciple (approved):

PR #3209, opened 2024-03-15
Pin System.Private.Uri in SDK to fix CVEs reported

nikola-jokic:


drew-viles:
It's a tad concerning that this has been sat here since March and there have been releases since then 😬 
It does address a HIGH CVE CVE-2019-0980/CVE-2019-0981. 

Is this going to be accepted merged in soon or does more need to be done?

PR #3215, opened 2024-03-21
Add new SessionConflict return code

eeSquared:
Bubbles up a new return code `SessionConflict` to allow hosted compute to handle this case differently than other failures.

### DIFF
--- a/src/Misc/layoutroot/run-helper.cmd.template
+++ b/src/Misc/layoutroot/run-helper.cmd.template
@@ -49,5 +49,10 @@ if %ERRORLEVEL% EQU 4 (
   exit /b 1
 )
 
+if %ERRORLEVEL% EQU 5 (
+  echo "Runner listener exit with Session Conflict error, stop the service, no retry needed."
+  exit /b 0
+)
+
 echo "Exiting after unknown error code: %ERRORLEVEL%"
 exit /b 0
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -146,6 +146,13 @@ namespace GitHub.Runner.Common
                 }
             }
 
+            public enum CreateSessionResult
+            {
+                Success,
+                Failure,
+                SessionConflict
+            }
+
             public static class ReturnCode
             {
                 public const int Success = 0;
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -153,6 +153,7 @@ namespace GitHub.Runner.Common
                 public const int RetryableError = 2;
                 public const int RunnerUpdating = 3;
                 public const int RunOnceRunnerUpdating = 4;
+                public const int SessionConflict = 5;
             }
 
             public static class Features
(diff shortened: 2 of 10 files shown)


src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -42,7 +42,7 @@ public override void Initialize(IHostContext hostContext)
             _brokerServer = HostContext.GetService<IBrokerServer>();
         }
 
-        public async Task<Boolean> CreateSessionAsync(CancellationToken token)
+        public async Task<int> CreateSessionAsync(CancellationToken token)
```
ericsciple, line 45: At this layer, I wonder whether an enum would be more appropriate? I'm not sure it's appropriate for this layer to understand application return codes.

src/Runner.Common/Constants.cs
```diff
@@ -153,6 +153,7 @@ public static class ReturnCode
                 public const int RetryableError = 2;
                 public const int RunnerUpdating = 3;
                 public const int RunOnceRunnerUpdating = 4;
+                public const int SessionConflict = 5;
```
TingluoHuang, line 156: make sure you double check all the entry script we have to make sure it handles the new exit code as expected.
Most customer use the `./run.sh` or the service entrypoint `.js`

ericsciple: Does `run.sh` get updated?

TingluoHuang: there is a `run-helper.sh` which contains the core logic today and it should get updated on runner restart.

src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -42,7 +42,7 @@ public override void Initialize(IHostContext hostContext)
             _brokerServer = HostContext.GetService<IBrokerServer>();
         }
 
-        public async Task<Boolean> CreateSessionAsync(CancellationToken token)
+        public async Task<Constants.Runner.CreateSessionResult> CreateSessionAsync(CancellationToken token)
```
ericsciple, line 45: Do you mind moving the enum to a different place? This method return signature looks funny because the result value by definition is not a constant (i.e. it can be one of multiple different values).

Maybe the Runner.Listener namespace would be a good place to define the enum since usage is localized to that namespace.

src/Misc/layoutroot/run-helper.cmd.template
```diff
@@ -49,5 +49,10 @@ if %ERRORLEVEL% EQU 4 (
   exit /b 1
 )
 
+if %ERRORLEVEL% EQU 5 (
+  echo "Runner listener exit with Session Conflict error, stop the service, no retry needed."
+  exit /b 0
```
ericsciple, line 54: It would be good to get a review from Ting regarding these scripts. I looked over the changes and makes sense to me.

Odd that we return exit code `0` here but that looks to be consistent with other non-retriable errors.

TingluoHuang (approved):

PR #3226, opened 2024-04-02
Update supported distros to match new docs

BagToad:
Update supported distros to match [the docs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners)

Previously: https://github.com/actions/runner/pull/3040


pje (approved):
Thanks for the update, @BagToad!

PR #3232, opened 2024-04-05
backoff if we retried polling for more than 50 times in less than 30m…

aiqiaoy:
…inutes

We should not really be making more than 40 long polls in 30mins so backoff if we tried more than 50 times. This is a temporary fix until we figure out why message queue cancellation doesn't always work.

### DIFF
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -361,15 +361,6 @@ namespace GitHub.Runner.Listener
                 if (message == null)
                 {
                     continuousEmptyMessage++;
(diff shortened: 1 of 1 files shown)


rentziass (approved):
LGTM!

src/Runner.Listener/MessageListener.cs
```diff
@@ -359,15 +360,26 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
...
                         Trace.Info($"No message retrieved from session '{_session.SessionId}' within last 30 minutes.");
                         heartbeat.Restart();
+                        continuousEmptyMessage = 0;
                     }
                     else
                     {
                         Trace.Verbose($"No message retrieved from session '{_session.SessionId}'.");
                     }
+                    
+                    if (continuousEmptyMessage > 50)
```
aiqiaoy, line 375: Moved this check to after the "No message retrieved" trace on 372.

thboop (dismissed):
lgtm

thboop (approved):
lgtm

brewagebear:
@aiqiaoy  
> "This is a temporary fix until we figure out why message queue cancellation needsn't always work." 

May I know the related issues?

PR #3248, opened 2024-04-19
Cleanup enabled feature flags.

TingluoHuang:


src/Runner.Listener/JobDispatcher.cs
TingluoHuang, line 1: hide whitespace.

src/Runner.Worker/ActionManager.cs
```diff
@@ -483,10 +483,6 @@ public Definition LoadAction(IExecutionContext executionContext, Pipelines.Actio
                             {
                                 // Load stored Ids for later load actions
                                 compositeAction.Steps[i].Id = _cachedEmbeddedStepIds[action.Id][i];
-                                if (string.IsNullOrEmpty(executionContext.Global.Variables.Get("DistributedTask.EnableCompositeActions")) && compositeAction.Steps[i].Reference.Type != Pipelines.ActionSourceType.Script)
-                                {
-                                    throw new Exception("`uses:` keyword is not currently supported.");
```
TingluoHuang, line 488: we use this to throw when using the runner with older GHES.
The oldest supported GHES is current 3.9 and it supports composite action.

ericsciple (approved):

PR #3269, opened 2024-05-01
Preserve dates when deserializing job message from Run Service

ericsciple:


src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -9,13 +9,23 @@
 using GitHub.Services.Common;
 using GitHub.Services.OAuth;
 using GitHub.Services.WebApi;
+using Newtonsoft.Json;
```
luketomlinson, line 12: I think we need to do this in `TaskAgentHttpClient` too, to fix the same issue from Actions Service

src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -174,5 +184,11 @@ public class RunServiceHttpClient : RawHttpClientBase
                     throw new Exception($"Failed to renew job: {result.Error}");
             }
         }
+
+        protected override async Task<T> ReadJsonContentAsync<T>(HttpResponseMessage response, CancellationToken cancellationToken = default(CancellationToken))
```
ericsciple, line 188: This is a virtual method on `RawHttpClientBase`

https://github.com/actions/runner/blob/04b07b6675c56da40e532ab73eeb4287bf223e34/src/Sdk/WebApi/WebApi/RawHttpClientBase.cs#L235

src/Sdk/DTWebApi/WebApi/ActionsRunServerHttpClient.cs
```diff
@@ -0,0 +1,95 @@
...
+            return SendAsync<Pipelines.AgentJobRequestMessage>(
+                httpMethod,
+                locationId,
+                routeValues: routeValues,
+                version: new ApiResourceVersion(6.0, 1),
+                userState: userState,
+                cancellationToken: cancellationToken);
+        }
+
+        protected override async Task<T> ReadJsonContentAsync<T>(HttpResponseMessage response, CancellationToken cancellationToken = default(CancellationToken))
```
ericsciple, line 89: Trying to isolate the change to this one call since this flow is controlled by a feature flag

luketomlinson (approved):

crohr:
@ericsciple could this explain issues with jobs not being picked up / picked up late? https://github.com/runs-on/runs-on/discussions/75

PR #3273, opened 2024-05-03
Make it easy to install `git` on an Action Runner Image

jww3:
The version of `git` (`v2.34.1`) associated with `ubuntu:jammy` is quite old at this point.  Calling `apt update` out-of-the-box won't cause any newer version to be discovered.

This change allows users of the actions/runner container image to easily install more modern versions of git (e.g. `v2.43.0`) should they desire.

Note that we don't actually install `git`.  We simply get the prerequisites out of the way.

## Related Issues
- https://github.com/actions/checkout/issues/1708

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -39,11 +39,13 @@ ENV RUNNER_MANUALLY_TRAP_SIG=1
 ENV ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1
 ENV ImageOS=ubuntu22
 
-RUN apt-get update -y \
-    && apt-get install -y --no-install-recommends \
-    sudo \
-    lsb-release \
+RUN apt update -y \
+    && apt install -y --no-install-recommends sudo lsb-release software-properties-common \
     && rm -rf /var/lib/apt/lists/*
+
+# Configure git-core/ppa based on guidance here:  https://git-scm.com/download/linux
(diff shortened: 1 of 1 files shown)


jww3:
While we're at it, should we also perform [these prerequisites](https://github.com/cli/cli/blob/trunk/docs/install_linux.md#debian-ubuntu-linux-raspberry-pi-os-apt) to facilitate installation of the `gh cli`?

JAORMX (approved):

images/Dockerfile
```diff
@@ -39,12 +39,14 @@ ENV RUNNER_MANUALLY_TRAP_SIG=1
 ENV ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1
 ENV ImageOS=ubuntu22
 
-RUN apt-get update -y \
-    && apt-get install -y --no-install-recommends \
-    sudo \
-    lsb-release \
+RUN apt update -y \
+    && apt install -y --no-install-recommends sudo lsb-release software-properties-common \
     && rm -rf /var/lib/apt/lists/*
```
jww3, line 44: I'm combining these into a single line to aid readability.  (One command per line).

The net change here is to install `software-properties-common`.  (We still install `sudo` and `lsb-release`.)

`software-properties-common` is required for the new instruction on line 47, `add-apt-repository`

TingluoHuang (approved):

PR #3303, opened 2024-05-23
Add new libicu versions for Ubuntu 24.04 (74) and debian experimental (73,75)

EraYaN:
libicu74 is in fact supported for 24.04: https://github.com/dotnet/core/blob/main/release-notes/6.0/linux-packages.md#ubuntu-2404-noble

Also update some of the documentation to reflect these changes and the changes to the .NET name.

Arguably this should just use the official .deb packages from Microsoft for the SDK on any modern Ubuntu version and all others that they support native packing for.


AndreasBergmeier6176:
Can we get that reviewed?

PR #3317, opened 2024-06-06
fix: actions cache does not work

### LINKED ISSUE #3316: Actions cache does not actually work
**Describe the bug**

https://github.com/actions/runner/blob/edfdbb966118990c9898c6723121f5610188c155/src/Runner.Common/HostContext.cs#L280-L283

Specifies that the actions cache is a subfolder of the work folder, but the work folder is always totally reset.

**To Reproduce**
1. Run a workflow that uses any action -> observe in the "Set up Job" part of the run that it downloads the action(s)
(ie, the text from https://github.com/actions/runner/blob/edfdbb966118990c9898c6723121f5610188c155/src/Runner.Worker/ActionManager.cs#L778)
2. Run same work -> observe in the "Set up Job" part of the run that it downloads the action(s)

**Expected behavior**
1. same
2. Action is already downloaded -> ie, the text from https://github.com/actions/runner/blob/edfdbb966118990c9898c6723121f5610188c155/src/Runner.Worker/ActionManager.cs#L770

latest, any os

## Job Log Output
```
Download action repository 'actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808' (SHA:65462800fd760344b1a7b4382951275a0abb4808)
```

## Suggested fix:

https://github.com/actions/runner/blob/edfdbb966118990c9898c6723121f5610188c155/src/Runner.Common/HostContext.cs#L282

Change to .Tools instead of .Work

davhdavh:
fixes #3316


davhdavh:
Am I missing anything for a review?

davhdavh:
review?

davhdavh:
now possible with https://github.com/marketplace/actions/save-actions-cache

PR #3319, opened 2024-06-06
Image size reduction

ccosby:
* Revert BUILDX_VERSION to 0.13.1 (0.13.2 does not exist upstream)
* Extract Docker in build stage, copy to main stage (keeps /home/runner/docker out of main image)
* Install ppa:git-core/ppa in build stage, copy apt lists and gpg keys to main stage (keeps very many packages out of main image)


ccosby:
This reduces the overall on-disk image size by ~301MB.
Sizes for x86_64:
  v2.317.0 = 1.24GB
  this = 939MB

ccosby:
@nikola-jokic You seem to be the one committing to the Docker image most often, so I ask you this. Who is the reviewer for Docker scoped pull requests? Almost 2 months without anybody looking at or commenting on a PR is a pretty long time.

images/Dockerfile
```diff
@@ -1,36 +1,34 @@
...
-    && if [ "$RUNNER_ARCH" = "amd64" ]; then export DOCKER_ARCH=x86_64 ; fi \
-    && if [ "$RUNNER_ARCH" = "arm64" ]; then export DOCKER_ARCH=aarch64 ; fi \
-    && curl -fLo docker.tgz https://download.docker.com/${TARGETOS}/static/stable/${DOCKER_ARCH}/docker-${DOCKER_VERSION}.tgz \
-    && tar zxvf docker.tgz \
-    && rm -rf docker.tgz \
-    && mkdir -p /usr/local/lib/docker/cli-plugins \
-    && curl -fLo /usr/local/lib/docker/cli-plugins/docker-buildx \
-        "https://github.com/docker/buildx/releases/download/v${BUILDX_VERSION}/buildx-v${BUILDX_VERSION}.linux-${TARGETARCH}" \
-    && chmod +x /usr/local/lib/docker/cli-plugins/docker-buildx
+ && if [ "$RUNNER_ARCH" = "amd64" ]; then export RUNNER_ARCH=x64 ; fi \
```
joh-klein, line 26: why did you remove the arm64 condition?

images/Dockerfile
```diff
@@ -39,28 +37,35 @@ ENV RUNNER_MANUALLY_TRAP_SIG=1
...
+# Install docker and cli-plugins as a copy from docker:cli image
+COPY --from=docker-cli --chown=root:root --chmod=0555 /usr/local/bin/docker /usr/bin/docker
+COPY --from=docker-cli --chown=root:root --chmod=0555 /usr/local/libexec/docker/cli-plugins/docker-buildx /usr/local/libexec/docker/cli-plugins/docker-buildx
 
-WORKDIR /home/runner
+# Install a sane set of base utilities
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends sudo lsb-release gpg-agent software-properties-common curl git jq unzip \
+ && apt-get clean all \
+ && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
```
joh-klein, line 52: why remove the lists here? wouldn't that mean that the next `apt-get update` has to download all of them again?

images/Dockerfile
ccosby: Didn't intend to. This is why we code review! Added it back. Will push an update.

ccosby: RUNNER_ARCH for arm64 wasn't in the upstream for actions-runner-$TARGETOS-$RUNNER_ARCH-$RUNNER_VERSION.tar.gz. I didn't remove it. The upstream link for that uses x64 and arm64. The former is set by the conditional, the latter is the same as TARGETARCH.

images/Dockerfile
ccosby: I was optimizing for size. Yes, it has to download again, but the cache hit rate on lists is very low anyway. I'll rework this section.

ccosby: Upstream image (mcr.microsoft.com/dotnet/runtime-deps:6.0-jammy) also ships with an empty /var/lib/apt/lists directory. I like it this way. Willing to discuss though.

images/Dockerfile
ccosby: FWIW...the diff tool isn't doing us any favors. Look at the whole file in context and you will see what I'm talking about. It's diffing the docker cli install and actions-runner install here, so it looks like I remove it.

images/Dockerfile
joh-klein: Maybe there is a misunderstanding. I'm not against `apt-get clean all && rm -rf /var/lib/apt/lists/* /var/cache/apt/*` – I was just wondering, why you did it multiple times in a row and not just once at the end.

images/Dockerfile
joh-klein: I see it now – thanks for taking the time to explain it! That is a strange coincidence.

images/Dockerfile
ccosby: No reason. That's an easy fix-up. It only has an effect at build time. I'll move the upgrade into the top block and save a layer at the same time. But first, lunch.

ccosby: All done. Have a look.

images/Dockerfile
```diff
@@ -1,36 +1,34 @@
...
-    && rm runner.tar.gz
 
-RUN curl -f -L -o runner-container-hooks.zip https://github.com/actions/runner-container-hooks/releases/download/v${RUNNER_CONTAINER_HOOKS_VERSION}/actions-runner-hooks-k8s-${RUNNER_CONTAINER_HOOKS_VERSION}.zip \
-    && unzip ./runner-container-hooks.zip -d ./k8s \
-    && rm runner-container-hooks.zip
+# Configure git-core/ppa based on guidance here: https://git-scm.com/download/linux
+# The second `apt-get update` isn't strictly necessary, but let's leave it here to force the
+# image build to break if the ppa is suddenly unavailable.
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends curl unzip sudo lsb-release gpg-agent software-properties-common \
```
joh-klein, line 21: Could you maybe add the comment

`# 'gpg-agent' and 'software-properties-common' are needed for the 'add-apt-repository' command that follows`

here so it is clear, why you need this?

But why the others? Is `sudo` and `lsb-release` already needed at this stage?

images/Dockerfile
```diff
@@ -39,28 +37,30 @@ ENV RUNNER_MANUALLY_TRAP_SIG=1
...
-RUN adduser --disabled-password --gecos "" --uid 1001 runner \
-    && groupadd docker --gid 123 \
-    && usermod -aG sudo runner \
-    && usermod -aG docker runner \
-    && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers \
-    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers
+# Install a sane set of base utilities and upgrade all system packages
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends sudo lsb-release gpg-agent software-properties-common curl git jq unzip \
+ && apt-get upgrade -y \
```
joh-klein, line 51: This is probably a matter of style – but as far as I can see (and what I read from them), microsoft keeps the base image very up to date – so, maybe this is not really needed?

ccosby:
@joh-klein Since the whole Dockerfile has been updated, I went all-in. I took into account your suggestions. Have a look.

Issue #1199, opened 2021-07-12
Can't Login ghcr.io for docker registry with Github Enterprise server


pyama86:
**Describe the bug**
H!
I'm using Github Enterprise Server and GitHub Actions.
I realized message `Docker login for 'ghcr.io' failed with exit code 1` and I did a little research on this.
I think you not considered pta token for docker registry token.
https://github.com/actions/runner/blob/main/src/Runner.Worker/ContainerOperationProvider.cs#L506
Seemingly  login with Github Enterprise Token but actions should use github.com token.
Can I do any workarounds?

Using version.
GHES: 3.1
runner: v2.278.0

**To Reproduce**
2. Run '....'
3. See error

**Expected behavior**

## Runner Version and Platform
Version of your runner?

OS of the machine running the runner? OSX/Windows/Linux/...

## What's not working?
Please include error messages and screenshots.

## Job Log Output
If applicable, include the relevant part of the job / step log output here.  All sensitive information should already be masked out, but please double-check before pasting here.

## Runner and Worker's Diagnostic Logs
If applicable, add relevant diagnostic log information.  Logs are located in the runner's `_diag` folder. The runner logs are prefixed with `Runner_` and the worker logs are prefixed with `Worker_`. Each job run correlates to a worker log.  All sensitive information should already be masked out, but please double-check before pasting here.

TingluoHuang:
@pyama86 do you mind sharing a sample workflow YAML file and workflow log?

pyama86:
@TingluoHuang 
I could send log to you, but Isn't it self-evident that GHES users can't have a github.com token?

hsbt:
Hi, I'm colleague with @pyama86 . I did escalate this issue to GHES support team. 

@TingluoHuang Can you contact them? Thanks.

thboop:
This is a bug, we should not use a GHAE/GHES github token to login to ghcr.io. This makes it fail for public containers because auth fails even though these containers are public.

PR #3326, opened 2024-06-10
Make Settings.IsHostedServer value consistent with UrlUtil.IsHostedServer

alixinne:
When running GitHub Enterprise Server with a pool of JIT-configured runners (such as what [philips-labs/terraform-aws-github-runner](https://github.com/philips-labs/terraform-aws-github-runner/) provides), #1199 still occurs. The fix in #1291 uses the "IsHostedServer" property from the settings which is **not** set by the GHES endpoint for JIT configuration.

This change computes the default value of IsHostedServer from the GitHub server URL, as is already done in [lots of other parts of the code](https://github.com/search?q=repo%3Aactions%2Frunner%20IsHostedServer&type=code) to make it consistent.

### DIFF
--- a/src/Runner.Common/ConfigurationStore.cs
+++ b/src/Runner.Common/ConfigurationStore.cs
@@ -62,7 +62,16 @@ namespace GitHub.Runner.Common
             get
             {
                 // Old runners do not have this property. Hosted runners likely don't have this property either.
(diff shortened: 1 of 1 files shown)


alixinne:
This is actually not stale.

Issue #3308, opened 2024-05-29
SIGTERM kills the job instead of waiting for job completion


jaswanthikolla:
May be this can be a feature request but I don't see any discussions ( of any kind)  at [here](https://github.com/orgs/community/discussions/categories/actions-and-packages?discussions_q=category%3Aactions-and-packages+sort%3Adate_created),  so not sure where to log it.  And also, It's existing functionality too.

**Describe the bug**
In _summerwind_ images, If we receive a SIGTERM the runners waits [default 15 mins](https://github.com/actions/actions-runner-controller/blob/9afd93065fa8b1f87296f0dcdf0c2753a0548cb7/runner/graceful-stop.sh#L5) before killing a running job.  With Scaled Sets ( ARSS),  helmcharts use `ghcr.io/actions/actions-runner` images with `run.sh` entrypoint which lacks the graceful termination functionality.  This functionality is expected as consistent between ARC and scaled sets. 

**To Reproduce**
1. Launch the Runner with a label.
2. Run a workflow with the assigned label. Make sure to have a `step` of bash `sleep` statement to simulate long running job.
3. Send SIGTERM signal. On kubernetes send a pod termination signal.
4. Check if the workflow succeeded or not.

**Expected behavior**
These  2 different behaviors are expected based on running is running the job or not.
-  Run the runner without a job assigned, SIGTERM should be propagated and runner should exit.
-  Run the runner with a job assigned, SIGTERM should not be propagated and runner shouldn't  exit unless job is finished with a reasonable grace period.

## Runner Version and Platform
2.316.1

OS of the machine running the runner? Linux - Ubuntu 22

## What's not working?
SIGTERM handling

jaswanthikolla:
I am ready to raise a PR for this if https://github.com/actions/runner/issues/3330 can be solved.

PR #3340, opened 2024-06-14
[Issue:3308] SIGTERM Graceful shutdown functionality

jaswanthikolla:
This fixes the https://github.com/actions/runner/issues/3308  by handling graceful shutdown of the runner. It does following. 
1. If the runner is just listening for jobs and Idle, It will just exit.
2. If the runner is running a job, It will wait `RUNNER_GRACEFUL_STOP_TIMEOUT` seconds before terminating or job completion  whichever happens first.

### DIFF
--- a/src/Misc/layoutroot/run.sh
+++ b/src/Misc/layoutroot/run.sh
@@ -26,11 +26,47 @@ run() {
     done
 }
 
+handle_sigterm() {
+    # Default graceful stop timeout is 3 seconds
+    RUNNER_GRACEFUL_STOP_TIMEOUT=${RUNNER_GRACEFUL_STOP_TIMEOUT:-3}
+    echo "Received SIGTERM, " \
+        "Graceful shutdown in $RUNNER_GRACEFUL_STOP_TIMEOUT Secs ..."
+
+    if [ -n "$RUNNER_TOKEN" ]; then
+        idle_runner="/runner/config.sh remove --token $RUNNER_TOKEN"
+    else
+        # workaround for Issue#3330
+        # For the case JITCONFIG is used instead of reg token.
+        # Fallback to check if worker is running,race condition prone.
+        worker_process_id=$(pgrep Runner.Worker)
+        idle_runner="test -z \"$worker_process_id\""
+    fi
+
+    if ! eval $idle_runner; then
+        echo "Running a job, waiting for $RUNNER_GRACEFUL_STOP_TIMEOUT s to finish.."
+        i=0
+        while [[ $i -lt $RUNNER_GRACEFUL_STOP_TIMEOUT ]]; do
+            # While waiting, if runner is stops.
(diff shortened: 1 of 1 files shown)


nicolaswebb (approved):

jaswanthikolla:
Any ETA on when can we expect a review on this PR?

ccincotti3:
This would be really great to get in assuming it works, we're also experiencing this.

moosh3:
Would love to see this merged

joosangkim:
This PR is an essential bug fix for using github runner with Karpenter.

jaswanthikolla:
Karpenter support is essential to save significant amount of across all companies.  We save easily 200k+ per year, Scaling that across 1000's of tech companies, Karpenter support can easily save 50 million+ and CO2 Emissions. 

Can we be more responsible users, and prioritize reviewing and merging this PR.

jaswanthikolla:
Karpenter support is essential to save significant cost savings across all companies.  We save easily $200k+ per year, Scaling that across 1000's of tech companies, Karpenter support can easily save $50 million+ and CO2 Emissions. 

Can we be more responsible users, and prioritize reviewing and merging this PR.

velkovb:
Upvote for the PR. We ended up implementing a custom image and baking in the script. However, we noticed that it is not behaving properly in `dind` runners as the signal is only captured on the `runner` container and the docker socket dies. Moving `dind` to a sidecar container has solved it for us - https://github.com/actions/actions-runner-controller/pull/3842

alec-drw:
@velkovb could I inquire as to the errors you saw when the runner did not capture the signal correctly? I have observed behavior in with ephemeral pvc's get stuck in the `Released` state after docker fails to cleanly shutdown, leading to an eventual break in the storage provisioner. 

Have been leaning towards using the Kubernetes buildkit driver as the solution, but a side car would certainly be easier

velkovb:
> @velkovb could I inquire as to the errors you saw when the runner did not capture the signal correctly? I have observed behavior in with ephemeral pvc's get stuck in the `Released` state after docker fails to cleanly shutdown, leading to an eventual break in the storage provisioner.
> 
> Have been leaning towards using the Kubernetes buildkit driver as the solution, but a side car would certainly be easier

We were seeing errors that connection to the docker socket was lost during an image build. We get a SIGTERM signal and the runner container handles it properly but the dind one doesn't and terminates so docker host disappears and build breaks.

jaswanthikolla:
I did look into sidecar while working on this, But they need 1.29 which is still not a production as earlier versions are not forced by EOL upgrade.

jaswanthikolla:
> However, we noticed that it is not behaving properly in dind runners as the signal is only captured on the runner container and the docker socket dies

that's a different issue, and fixed in PR https://github.com/actions/actions-runner-controller/pull/3601

PR #3342, opened 2024-06-17
Receive error body from Run Service

ericsciple:
Receive error body from Run Service so we can distinguish responses from proxies/etc

On an E2E Codespace, I tested an error response from each affected REST API (acquire job, complete job, renew job)

### DIFF
--- a/src/Sdk/RSWebApi/RunServiceHttpClient.cs
+++ b/src/Sdk/RSWebApi/RunServiceHttpClient.cs
@@ -86,6 +86,7 @@ namespace GitHub.Actions.RunService.WebApi
                 httpMethod,
                 requestUri: requestUri,
                 content: requestContent,
+                readErrorBody: true,
                 cancellationToken: cancellationToken);
 
             if (result.IsSuccess)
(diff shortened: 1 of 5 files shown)


src/Runner.Common/RunServer.cs
```diff
@@ -62,7 +62,9 @@ public Task<AgentJobRequestMessage> GetJobMessageAsync(string id, CancellationTo
             CheckConnection();
             return RetryRequest<AgentJobRequestMessage>(
                 async () => await _runServiceHttpClient.GetJobMessageAsync(requestUri, id, VarUtil.OS, cancellationToken), cancellationToken,
-                shouldRetry: ex => ex is not TaskOrchestrationJobAlreadyAcquiredException);
+                shouldRetry: ex => ex is not TaskOrchestrationJobNotFoundException &&
```
ericsciple, line 65: revert for now

src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -86,21 +86,43 @@ static RunServiceHttpClient()
...
                 content: requestContent,
+                readErrorContent: true,
                 cancellationToken: cancellationToken);
 
             if (result.IsSuccess)
             {
                 return result.Value;
             }
 
+            if (TryParseErrorContent(result.ErrorContent, out RunServiceError error))
```
ericsciple, line 97: apply to the rest of run service

src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -169,19 +208,37 @@ static RunServiceHttpClient()
...
             {
                 case HttpStatusCode.NotFound:
                     throw new TaskOrchestrationJobNotFoundException($"Job not found: {jobId}");
-                default:
-                    throw new Exception($"Failed to renew job: {result.Error}");
+            }
+
+            if (!string.IsNullOrEmpty(result.ErrorBody))
+            {
+                throw new Exception($"Failed to renew job: {result.Error}. {result.ErrorBody}");
```
ericsciple, line 237: todo: truncate ErrorBody if greater than 1000 chars

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
```diff
@@ -101,15 +101,55 @@ public void Dispose()
             }
         }
 
+        protected async Task<RawHttpClientResult> Send2Async(
```
yaananth, line 104: Should we better name `Send2Async`? This is RawHttp, so not every call needs to have json response like run-service does, in that sense, we might want to keep `SendAsync` and this new one as well?

yaananth (approved):

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
yaananth: or since we have `try*` anyway, why do we need `send2`? we plan to delete `send` one later? 🤔

TingluoHuang (approved):

src/Sdk/WebApi/WebApi/RawHttpClientBase.cs
ericsciple: Open to changing this. I played around with a few different names but everything was horrible.

E.g. "SendAndReceiveWrapperResult". Such a long, aesthetically displeasing name.

PR #3354, opened 2024-06-21
Backoff to avoid excessive retries to Run Service in a duration

ericsciple:
Helps avoid excessive calls to Run Service when encountering non-retriable errors from /acquirejob. Normally we rely on the HTTP clients to back off between retry attempts. However, acquiring a job involves calls to both Run Serivce and Broker. And Run Service and Broker communicate with each other in an async fashion.

When Run Service encounters a non-retriable error, it sends an async message to Broker. The runner will, however, immediately call Broker to get the next message. If the async event from Run Service to Broker has not yet been processed, the next message from Broker may be the same job message.

The error throttler helps us back off when encountering successive, non-retriable errors from /acquirejob.

### DIFF
--- a/src/Runner.Common/RunServer.cs
+++ b/src/Runner.Common/RunServer.cs
@@ -62,7 +62,10 @@ namespace GitHub.Runner.Common
             CheckConnection();
             return RetryRequest<AgentJobRequestMessage>(
                 async () => await _runServiceHttpClient.GetJobMessageAsync(requestUri, id, VarUtil.OS, cancellationToken), cancellationToken,
-                shouldRetry: ex => ex is not TaskOrchestrationJobAlreadyAcquiredException);
+                shouldRetry: ex =>
+                    ex is not TaskOrchestrationJobNotFoundException &&          // HTTP status 404
+                    ex is not TaskOrchestrationJobAlreadyAcquiredException &&   // HTTP status 409
+                    ex is not TaskOrchestrationJobUnprocessableException);      // HTTP status 422
         }
 
         public Task CompleteJobAsync(
--- a/src/Runner.Listener/ErrorThrottler.cs
+++ b/src/Runner.Listener/ErrorThrottler.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Runner.Common;
+using GitHub.Services.Common;
+
+namespace GitHub.Runner.Listener
+{
+    [ServiceLocator(Default = typeof(ErrorThrottler))]
+    public interface IErrorThrottler : IRunnerService
+    {
+        void Reset();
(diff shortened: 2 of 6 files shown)


src/Runner.Common/RunServer.cs
```diff
@@ -62,7 +62,10 @@ public Task<AgentJobRequestMessage> GetJobMessageAsync(string id, CancellationTo
             CheckConnection();
             return RetryRequest<AgentJobRequestMessage>(
                 async () => await _runServiceHttpClient.GetJobMessageAsync(requestUri, id, VarUtil.OS, cancellationToken), cancellationToken,
-                shouldRetry: ex => ex is not TaskOrchestrationJobAlreadyAcquiredException);
+                shouldRetry: ex =>
```
ericsciple, line 65: The problem with not retrying these errors, means we immediately call to Broker to get the next message, which might be the same job message.

When one of these errors is encountered from Run Service, an async message is sent to Broker. We don't want to keep hammering the service in a tight loop if there are delays processing the message queue.

src/Runner.Listener/ErrorThrottler.cs
```diff
@@ -0,0 +1,43 @@
...
+                return;
+            }
+
+            TimeSpan backoff = BackoffTimerHelper.GetExponentialBackoff(
+                attempt: _count - 1,
+                minBackoff: s_minBackoff,
+                maxBackoff: s_maxBackoff,
+                deltaBackoff: s_backoffCoefficient);
+            Trace.Warning($"Back off {backoff.TotalSeconds} seconds before next attempt.");
+            await Task.Delay(backoff, token);
```
TingluoHuang, line 40: FYI, `Task.Delay` will throw on cancellation token fire.

src/Runner.Listener/ErrorThrottler.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                return;
+            }
+
+            TimeSpan backoff = BackoffTimerHelper.GetExponentialBackoff(
+                attempt: _count - 1,
+                minBackoff: s_minBackoff,
+                maxBackoff: s_maxBackoff,
+                deltaBackoff: s_backoffCoefficient);
+            Trace.Warning($"Back off {backoff.TotalSeconds} seconds before next attempt.");
```
TingluoHuang, line 39: also trace the `_count` might be useful as well.

src/Runner.Listener/ErrorThrottler.cs
ericsciple: I think that's OK, but I switched to Host.Delay so I could unit test this

src/Runner.Listener/ErrorThrottler.cs
ericsciple: good idea, fixed

TingluoHuang (approved):

PR #3364, opened 2024-06-27
Apply error throttler for unexpected exceptions as well

ericsciple:


src/Runner.Listener/Runner.cs
```diff
@@ -595,6 +595,7 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                                         catch (Exception ex)
                                         {
                                             Trace.Error($"Caught exception from acquiring job message: {ex}");
+                                            await _acquireJobThrottler.IncrementAndWaitAsync(messageQueueLoopTokenSource.Token);
```
ericsciple, line 598: Also incremented on line 592; rest on line 584

Introduces an exponential backoff algorithm for successive errors, up to 1m max delay. Will alleviate a little bit more load on our service during sustained errors.

PR #3413, opened 2024-08-02
Warn for soon to be deprecated OS versions

ericsciple:

PR #3422, opened 2024-08-06
.NET 8 OS compatibility test

ericsciple:
Based on feature flag state, performs a .NET 8 compatibility test during job-setup and optionally adds a warning if the test fails.

Also, reverted previous implementation:
- https://github.com/actions/runner/pull/3413


src/Runner.Worker/OSWarningChecker.cs
```diff
@@ -80,8 +80,70 @@ public async Task CheckOSAsync(IExecutionContext context, IList<OSWarning> osWar
...
+                        arguments: string.Empty,
+                        environment: null,
+                        cancellationToken: context.CancellationToken);
+
+                    var outputStr = string.Join("\n", output).Trim();
+                    if (exitCode != 0 || !string.Equals(outputStr, "Hello from .NET 8!", StringComparison.Ordinal))
+                    {
+                        if (!omitAnnotation)
+                        {
+                            context.Warning("The runner is not compatible with .NET 8.");
```
ericsciple, line 132: Have the server send this message, which will solve the following problems:
- Act as a feature flag
- Can update from server to include a link to the changelog, etc.

src/Runner.Worker/OSWarningChecker.cs
```diff
@@ -13,75 +13,77 @@ namespace GitHub.Runner.Worker
...
                             }
                         }
+                    };
+
+                    int exitCode = await process.ExecuteAsync(
+                        workingDirectory: HostContext.GetDirectory(WellKnownDirectory.Root),
+                        fileName: Path.Combine(HostContext.GetDirectory(WellKnownDirectory.Bin), "testDotNet8Compatibility", $"TestDotNet8Compatibility{IOUtil.ExeExtension}"),
+                        arguments: string.Empty,
+                        environment: null,
+                        cancellationToken: context.CancellationToken);
```
TingluoHuang, line 65: i would consider set a static timeout about 10s, so the checking won't block the job forever when something went wrong...

src/Runner.Worker/OSWarningChecker.cs
```diff
@@ -13,75 +13,77 @@ namespace GitHub.Runner.Worker
...
-                    Trace.Error("An error occurred while checking OS warnings for file '{0}' and regex '{1}'.", osWarning.FilePath, osWarning.RegularExpression);
-                    Trace.Error(ex);
-                    context.Global.JobTelemetry.Add(new JobTelemetry() { Type = JobTelemetryType.General, Message = $"An error occurred while checking OS warnings for file '{osWarning.FilePath}' and regex '{osWarning.RegularExpression}': {ex.Message}" });
-                }
+            }
+            catch (Exception ex)
+            {
+                Trace.Error("An error occurred while testing .NET 8 compatibility'");
+                Trace.Error(ex);
+                context.Global.JobTelemetry.Add(new JobTelemetry() { Type = JobTelemetryType.General, Message = $"An error occurred while testing .NET 8 compatibility; exception type '{ex.GetType().FullName}'; message: {ex.Message}" });
```
TingluoHuang, line 85: do we want to log the `output` into the telemetry, so we know what's failing?

TingluoHuang: maybe we got some output from the process, but the `process.ExecuteAsync` somehow throw later.

TingluoHuang (approved):

PR #3427, opened 2024-08-12
.NET 8 compat test adjustments: 1) do not trim SDK, 2) support pattern to match output, 3) modify output truncation length

ericsciple:


src/Runner.Worker/OSWarningChecker.cs
```diff
@@ -83,14 +96,15 @@ public async Task CheckOSAsync(IExecutionContext context)
...
                 Trace.Error(ex);
-                context.Global.JobTelemetry.Add(new JobTelemetry() { Type = JobTelemetryType.General, Message = $".NET 8 OS compatibility test encountered exception type '{ex.GetType().FullName}', message: '{ex.Message}', process output: '{GetShortOutput(output)}'" });
+                context.Global.JobTelemetry.Add(new JobTelemetry() { Type = JobTelemetryType.General, Message = $".NET 8 OS compatibility test encountered exception type '{ex.GetType().FullName}', message: '{ex.Message}', process output: '{GetShortOutput(context, output)}'" });
             }
         }
 
-        private static string GetShortOutput(List<string> output)
+        private static string GetShortOutput(IExecutionContext context, List<string> output)
         {
+            var length = context.Global.Variables.System_DotNet8CompatibilityOutputLength ?? 200;
```
TingluoHuang, line 105: given we knows that 200 is not enough, should we bump it higher by default? :D

src/TestDotNet8Compatibility/TestDotNet8Compatibility.csproj
```diff
@@ -6,7 +6,6 @@
     <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64;osx-arm64;win-arm64</RuntimeIdentifiers>
     <SelfContained>true</SelfContained>
     <PublishTrimmed>true</PublishTrimmed>
```
TingluoHuang, line 8: removed the wrong one?

ericsciple: oh snap, good catch

src/Runner.Worker/OSWarningChecker.cs
ericsciple: might be enough after publishing non-trimmed

TingluoHuang (approved):

SidhantaKanha (commented):
#

PR #3454, opened 2024-09-05
Handle Error Body in Responses from Broker

luketomlinson:
To distinguish between errors from proxies/in-between services, we need to parse error bodies from Broker to be sure the error is the one from broker.

### DIFF
--- a/src/Runner.Common/BrokerServer.cs
+++ b/src/Runner.Common/BrokerServer.cs
@@ -92,7 +92,7 @@ namespace GitHub.Runner.Common
 
         public bool ShouldRetryException(Exception ex)
         {
-            if (ex is AccessDeniedException ade && ade.ErrorCode == 1)
+            if (ex is AccessDeniedException ade)
             {
                 return false;
             }
--- a/src/Sdk/RSWebApi/Contracts/BrokerError.cs
+++ b/src/Sdk/RSWebApi/Contracts/BrokerError.cs
@@ -8,8 +8,8 @@ namespace GitHub.Actions.RunService.WebApi
         [DataMember(Name = "source", EmitDefaultValue = false)]
         public string Source { get; set; }
 
-        [DataMember(Name = "code", EmitDefaultValue = false)]
-        public int Code { get; set; }
+        [DataMember(Name = "errorType", EmitDefaultValue = false)]
+        public string Type { get; set; }
 
         [DataMember(Name = "errorMessage", EmitDefaultValue = false)]
         public string Message { get; set; }
--- a/src/Sdk/RSWebApi/Contracts/BrokerError.cs
+++ b/src/Sdk/RSWebApi/Contracts/BrokerError.cs
@@ -1,0 +1,17 @@
+﻿using System.Runtime.Serialization;
+
+namespace GitHub.Actions.RunService.WebApi
+{
+    [DataContract]
+    public class BrokerError
+    {
+        [DataMember(Name = "source", EmitDefaultValue = false)]
+        public string Source { get; set; }
+
+        [DataMember(Name = "code", EmitDefaultValue = false)]
+        public int Code { get; set; }
+
+        [DataMember(Name = "errorMessage", EmitDefaultValue = false)]
+        public string Message { get; set; }
+    }
+}
--- a/src/Sdk/RSWebApi/Contracts/BrokerError.cs
+++ b/src/Sdk/RSWebApi/Contracts/BrokerError.cs
@@ -8,8 +8,11 @@ namespace GitHub.Actions.RunService.WebApi
         [DataMember(Name = "source", EmitDefaultValue = false)]
         public string Source { get; set; }
 
-        [DataMember(Name = "errorType", EmitDefaultValue = false)]
-        public string Type { get; set; }
+        [DataMember(Name = "errorKind", EmitDefaultValue = false)]
+        public string ErrorKind { get; set; }
+
+        [DataMember(Name = "statusCode", EmitDefaultValue = false)]
+        public int StatusCode { get; set; }
 
         [DataMember(Name = "errorMessage", EmitDefaultValue = false)]
         public string Message { get; set; }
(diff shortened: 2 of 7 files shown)


src/Runner.Common/BrokerServer.cs
```diff
@@ -92,7 +92,7 @@ public Task ForceRefreshConnection(VssCredentials credentials)
 
         public bool ShouldRetryException(Exception ex)
         {
-            if (ex is AccessDeniedException ade && ade.ErrorCode == 1)
+            if (ex is AccessDeniedException ade)
```
luketomlinson, line 95: I don't think we'd want to retry regular 403s?

lokesh755 (approved):

src/Runner.Common/BrokerServer.cs
ericsciple: +1 403 is conclusive

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -103,24 +103,31 @@ public async Task<TaskAgentMessage> GetRunnerMessageAsync(
...
                 return result.Value;
             }
 
-            // the only time we throw a `Forbidden` exception from Listener /messages is when the runner is
-            // disable_update and is too old to poll
-            if (result.StatusCode == HttpStatusCode.Forbidden)
+            if (TryParseErrorBody(result.ErrorBody, out BrokerError brokerError))
             {
-                throw new AccessDeniedException($"{result.Error} Runner version v{runnerVersion} is deprecated and cannot receive messages.")
+                if (result.StatusCode == HttpStatusCode.Forbidden && brokerError.Code == 1)
```
ericsciple, line 116: In RunServiceError, the `Code` is the HTTP status code. It helps us distinguish the HTTP status code returned by Run Service any intermediate layer.

Here is looks like `Code` means something else. Any reason not to follow the same pattern across our all of services?

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -103,24 +103,31 @@ public async Task<TaskAgentMessage> GetRunnerMessageAsync(
...
-                    ErrorCode = 1
-                };
+                    throw new AccessDeniedException($"{result.Error} Runner version v{runnerVersion} is deprecated and cannot receive messages.")
+                    {
+                        ErrorCode = 1
+                    };
+                }
+            }
+
+            if (result.StatusCode == HttpStatusCode.Forbidden)
```
ericsciple, line 125: Is this temporary back-compat logic? If so, can we comment for readers, e.g. [like this](https://github.com/actions/runner/blob/36c66c808363839ea7069bef87fd4572c8bb6c54/src/Sdk/RSWebApi/RunServiceHttpClient.cs#L110)

ericsciple: Guessing in the future we rely on the message body to create a strongly typed exception (`AccessDeniedException` in this case). Otherwise we consider it to be a generic exception.

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -172,5 +179,11 @@ public async Task DeleteSessionAsync(
 
             throw new Exception($"Failed to delete broker session: {result.Error}");
         }
+
+        protected static bool TryParseErrorBody(string errorBody, out BrokerError error)
+        {
+            var validator = new Func<BrokerError, bool>(e => e.Source != null && e.Source == "actions-broker-listener");
+            return RawHttpClientBase.TryParseErrorBody(errorBody, validator, out error);
```
ericsciple, line 186: Consider copy/pasting [this impl](https://github.com/actions/runner/blob/36c66c808363839ea7069bef87fd4572c8bb6c54/src/Sdk/RSWebApi/RunServiceHttpClient.cs#L251) and avoiding callback indirection (arguably decreases readability) 🤷

ericsciple: (the http client classes have a lot of layers already)

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
ericsciple: Along that same line, wondering what `1` indicates. If we need another error code in addition to the HTTP status code, wondering whether a friendly code (i.e. string) would be better.

luketomlinson: Yeah that's a fair question. The `AccessDeniedException` has a Code=1, which I was going to match. But I guess we're not tied to that.

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
luketomlinson: updated

src/Runner.Common/BrokerServer.cs
ericsciple: Unclear who sets `ErrorCode = 1` today. Do we need to worry about back-compat or is this change OK to make?

ericsciple, line 95: Should we catch a specific exception type `RunnerVersionTooOldException`, since AccessDenied is ambiguous

src/Sdk/RSWebApi/Contracts/BrokerError.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Runtime.Serialization;
+
+namespace GitHub.Actions.RunService.WebApi
+{
+    [DataContract]
+    public class BrokerError
+    {
+        [DataMember(Name = "source", EmitDefaultValue = false)]
+        public string Source { get; set; }
+
```
ericsciple, line 10: Should we add statusCode to match RunServiceError

src/Sdk/RSWebApi/Contracts/BrokerError.cs
```diff
@@ -0,0 +1,17 @@
...
+        [DataMember(Name = "errorType", EmitDefaultValue = false)]
+        public string Type { get; set; }
```
ericsciple, line 12: Have you considered ErrorKind? to avoid system.type confusion

ericsciple (approved):

PR #3455, opened 2024-09-05
Capture actions_type after resolving alpine container.

TingluoHuang:
Instead of capturing what's in the `action.yml` for `actions_type`, we will use the final resolved result, ex: after force node version and check with alpine compatibility.

We want to see what `node` version the step really used.


pje (approved):
Makes sense! :shipit:

aiqiaoy (approved):

PR #3456, opened 2024-09-06
Cleanup back-compat code for interpreting Run Service status codes

ericsciple:


TingluoHuang (approved):

src/Sdk/RSWebApi/RunServiceHttpClient.cs
```diff
@@ -107,15 +107,6 @@ public async Task<AgentJobRequestMessage> GetJobMessageAsync(
                 }
             }
 
-            // Temporary back compat
-            switch (result.StatusCode)
```
ericsciple, line 111: Tested E2E to make sure everything is still working as expected. Specifically I tested the 422 UnprocessableEntity scenario above. Easy to repro that scenario via reusable workflow with a required secret, with the caller providing an expression for the secret value that maps to empty/nil.

ericsciple:
```yaml
name: CI
on:
  workflow_dispatch:
jobs:
  build:
    uses: ./.github/workflows/reuse.yml
    secrets: 
      foo: ${{ github.nosuch }}
```

```yaml
on:
  workflow_call:
    secrets:
      foo:
        description: asdf
        required: true
jobs:
  build:
    runs-on: self-hosted
    steps:
      - run: echo hi
        env:
          FOO_SECRET: ${{ secrets.foo }}
```

PR #3468, opened 2024-09-19
Adding Snapshot additional mapping tokens

rajrku:
Closes: https://github.com/github/actions-larger-runners/issues/3170

Adding new mapping tokens to snapshot yaml keyword
Added tests to validate the mapping token changes to snapshot keyword

Format:
```
- snapshot:
   - image-name: blah-blah
   - version: <major-version>.*
   - if: <conditions-similar-to-job-if>
  ```

### DIFF
--- a/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
+++ b/src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
@@ -399,10 +399,7 @@ namespace GitHub.DistributedTask.Pipelines.ObjectTemplating
         {
             var versionSegments = versionString.Split(".");
 
-            if (versionSegments.Length != 2 || 
-                !versionSegments[1].Equals("*") ||
-                !Int32.TryParse(versionSegments[0], result: out int parsedMajor) ||
-                parsedMajor <= 0)
+            if (versionSegments.Length != 2 || !versionSegments[1].Equals("*") || !Int32.TryParse(versionSegments[0], result: out int parsedMajor) || parsedMajor <= 0)
             {
                 return false;
             }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/Sdk/DTPipelines/workflow-v1.0.json [hunk 1]
```diff
@@ -162,18 +162,35 @@
         "snapshot-mapping"
       ]
     },
-
+
```
TingluoHuang, line 165: remove the whitespace

src/Sdk/DTPipelines/workflow-v1.0.json
```diff
@@ -162,18 +162,35 @@
...
+          "if": "snapshot-if",
+          "version": {
+            "type": "non-empty-string",
+            "required": false
           }
         }
       }
     },
 
+    "snapshot-if": {
```
TingluoHuang, line 182: do you need to have 
```
 "always(0,0)",
        "failure(0,0)",
        "cancelled(0,0)",
        "success(0,0)",
```
like other `if` expression?

src/Sdk/DTPipelines/Pipelines/ObjectTemplating/PipelineTemplateConverter.cs
```diff
@@ -376,7 +389,26 @@ internal static Snapshot ConvertToJobSnapshotRequest(TemplateContext context, Te
...
+                Condition = condition,
+                Version = version
+            };
+        }
+
+        private static bool IsSnapshotImageVersionValid(string versionString)
+        {
+            var versionSegments = versionString.Split(".");
+
+            if (versionSegments.Length != 2 ||
```
TingluoHuang, line 403: so we only support A.B and not A.B.C for version?
This is different compare to `https://github.com/github/actions-larger-runners/issues/3170`

rajrku: Not yet, we want to start with <major>.* first and get feedback from customers. And then decide if we want to go beyond.

src/Sdk/DTPipelines/workflow-v1.0.json
rajrku: To start with no. After discussing this with @ericsciple, we want to start small and extend it later. So we chose only to include what we think is sufficient.

TingluoHuang (approved):

Issue #385, opened 2020-03-24
FreeBSD support


asomers:
**Describe the enhancement**
Support building the runner on FreeBSD

**Additional information**
I think FreeBSD has all the libraries that the runner needs.  And while the dotnet-sdk isn't availble from Microsoft, there is already a package for it: lang/linux-dotnet-sdk .  So it should be possible to build the runner on FreeBSD.

TingluoHuang:
We can't really do that for mainly 2 reasons. 
- Supportability, without office support from dotnet core team, debugging any native dependency problem will be a nightmare.
- Legal, we (GitHub) might run into legal problem when we consume and redistribute 3rd party open source project.

But, we are totally fine with forking this repository and try yourselves, we are generally doing do on holding server/client compat.

Hope this can help. 😄

Ornias1993:
Addition:
lang/linux-dotnet-sdk is based on a very old version. Considering the DotNet Core team is working on version 5 by now, I doubt anyone is willing to spend time debuging it, as @TingluoHuang explained.

TingluoHuang:
We can add support if we do https://github.com/actions/runner/issues/243 or Dotnet 5 supports FreeBSD

Neilpang:
Sorry to bother you.  
but I just created an action to make it possible to use FreeBSD in your workflow.

https://github.com/vmactions/freebsd-vm

davidchisnall:
I'd love to see FreeBSD support, but there's a long tail of other platforms.  For [Microsoft/snmalloc](https://github.com/Microsoft/snmalloc), we support Windows, macOS, Linux, OpenEnclave, and FreeBSD and have had external contributions for NetBSD, Haiku, OpenBSD, DragonflyBSD and Solaris.  We support x86, ARM, MIPS, and RISC-V, (most in both 32- and 64-bit variants) and have CHERI versions in development for ARM, MIPS and RISC-V.

We would love to spin up CI for more of these, and can run a lot of them in Azure to provide self-hosted runners (at least for a weekly test run, if not on every PR), but porting .NET to all of them and having supported releases is a phenomenal amount of work, particularly over the next couple of years as we grow the amount of CI that we want to run on experimental hardware.

[...]

PR #3512, opened 2024-10-19
Initial support building the runner on FreeBSD

ChanTsune:
I tried to address #385.

The dotnet binary for FreeBSD is not available on https://dotnetcli.azureedge.net, so I am using the system's dotnet.

I have adjusted the build process to support FreeBSD, following the examples of other operating systems.
The following steps have been tested and confirmed to pass successfully.

```
./dev.sh layout Release freebsd-x64
./dev.sh package Release
```

I have confirmed that `./dev.sh test` passes, except for tests that require pre-built binaries such as node.
However, the operation of the runner on FreeBSD is not fully guaranteed, so please note that this is a PR that enables building for FreeBSD.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -6,6 +6,10 @@
     <BUILD_OS Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">Linux</BUILD_OS>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'false'">
+    <BUILD_OS Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::FreeBSD)))' == 'true'">FreeBSD</BUILD_OS>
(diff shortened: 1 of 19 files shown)


er2off (commented):
Generally LGTM but I made some changes

- In dev.sh shebang should be changed to `#!/usr/bin/env bash` for better cross-platform support (or use `bash ./dev.sh`).
  (This change also could be made for all other sh/bash run scripts)
  Also script fails to run after "Skip installing dotnet, use system dotnet"
  so you need to add `mkdir -p "${DOTNETSDK_INSTALLDIR}"` after (also someone from Microsoft, please add dotnet to Azure).

- Dotnet version provided in global.json doesn't exist in FreeBSD packages, there are only (at this moment)
  9.0.105 (package lang/dotnet)
  and 8.0.106 (package lang/dotnet8)

  Of course dotnet 9 fails to restore packages as target is net8.0, dotnet8 package seems OK to build everything.

- Tests depends on currently non-existed packages like `_layout/externals/node20/bin/node` and latest release of runner (for self updater test), this was already said.
  But I noticed that `src/Runner.Listener/Configuration/ServiceControlManager.cs` wasn't fully ported.
  Test said `Missing ServiceLocatorAttribute for interface 'GitHub.Runner.Listener.Configuration.ILinuxServiceControlManager'. Add the attribute to the interface or whitelist the interface in the test.`
  and that's true, service manager for FreeBSD init wasn't written in this PR.

- Just a note: self updater requires bash which isn't part of FreeBSD base system. As long as run scripts depends on bash (so you need to install it to make it work), this isn't important.

PR #3536, opened 2024-11-03
Handle runner not found

ericsciple:


pje:
So is the intention: "404s should never be retryable from the runner's perspective"?

src/Sdk/WebApi/WebApi/BrokerHttpClient.cs
```diff
@@ -115,6 +115,7 @@ public async Task<TaskAgentMessage> GetRunnerMessageAsync(
             {
                 switch (brokerError.ErrorKind)
                 {
+                    case BrokerErrorKind.RunnerNotFound:
```
luketomlinson, line 118: I wouldn't re-use the same AccessDeniedException with ErrorCode=1. That means the runner deprecation error in a few other places. Can we throw another error here instead? I don't really care what, but probably should be different than the Runner version being too old?

ericsciple: Oh interesting. I thought this was a generic mechanism to bubble out. Will look again.

ericsciple:
> So is the intention: "404s should never be retryable from the runner's perspective"?

Yes if the runner doesn't exist anymore, then shut down

luketomlinson (approved):

PR #3540, opened 2024-11-05
Decrease docker image size by 200MB

duboisf:
Copying and then deleting the docker directory in 2 steps (therefore, layers) means the docker directory is still present in the previous layer.

We uncompress the docker.tgz in the /tmp directory to be able to copy the docker binaries directly to their target destination.

```
$ docker images | grep actions | grep 2.320        
ghcr.io/duboisf/actions-runner         2.320.1-beta    18404fab5680   17 minutes ago   1.2GB
ghcr.io/actions/actions-runner         2.320.0         1ad90d9a13d2   4 weeks ago      1.4GB
```

[dive](https://github.com/wagoodman/dive) differences:

image


duboisf:
Tested with our self-hosted runners, so far so good.

duboisf:
@TingluoHuang what is the process for getting this reviewed?

PR #3545, opened 2024-11-07
Publish job telemetry to run-service.

TingluoHuang:
https://github.com/github/actions-runtime/issues/4771

### DIFF
--- a/src/Runner.Worker/JobRunner.cs
+++ b/src/Runner.Worker/JobRunner.cs
@@ -282,7 +282,7 @@ namespace GitHub.Runner.Worker
 
             var jobQueueTelemetry = await ShutdownQueue(throwOnFailure: false);
             // include any job telemetry from the background upload process.
-            if (jobQueueTelemetry.Count > 0)
+            if (jobQueueTelemetry?.Count > 0)
             {
                 jobContext.Global.JobTelemetry.AddRange(jobQueueTelemetry);
             }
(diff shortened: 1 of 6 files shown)


src/Runner.Worker/JobRunner.cs
```diff
@@ -279,7 +280,12 @@ private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecution
             jobContext.Debug($"Finishing: {message.JobDisplayName}");
             TaskResult result = jobContext.Complete(taskResult);
 
-            await ShutdownQueue(throwOnFailure: false);
+            var jobQueueTelemetry = await ShutdownQueue(throwOnFailure: false);
```
TingluoHuang, line 283: match the behavior for sending job complete to actions service.

src/Runner.Worker/JobRunner.cs
```diff
@@ -329,49 +342,7 @@ private async Task<TaskResult> CompleteJobAsync(IJobServer jobServer, IExecution
 
             if (_runnerSettings.DisableUpdate == true)
             {
-                try
-                {
```
TingluoHuang, line 333: i moved this to its own method in the bottom of the file, so i can easily diff the different between send job completion to run-service vs. actions-service.

ericsciple (approved):

PR #3546, opened 2024-11-08
Fetch repo-level runner groups from API in v2 flow

lucavallin:
Fetch repo-level runner groups from API in v2 flow.

### DIFF
--- a/src/Runner.Common/RunnerDotcomServer.cs
+++ b/src/Runner.Common/RunnerDotcomServer.cs
@@ -46,7 +46,11 @@ namespace GitHub.Runner.Common
             var githubApiUrl = "";
             var gitHubUrlBuilder = new UriBuilder(githubUrl);
             var path = gitHubUrlBuilder.Path.Split('/', '\\', StringSplitOptions.RemoveEmptyEntries);
-            if (path.Length == 1)
+            var isOrgRunner = path.Length == 1;
+            var isRepoOrEnterpriseRunner = path.Length == 2;
+            var isRepoRunner = !string.Equals(path[0], "enterprises", StringComparison.OrdinalIgnoreCase);
+
+            if (isOrgRunner)
             {
                 // org runner
                 if (UrlUtil.IsHostedServer(gitHubUrlBuilder))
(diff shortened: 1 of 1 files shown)


src/Runner.Common/RunnerDotcomServer.cs
```diff
@@ -46,7 +46,11 @@ public async Task<List<TaskAgent>> GetRunnerByNameAsync(string githubUrl, string
             var githubApiUrl = "";
             var gitHubUrlBuilder = new UriBuilder(githubUrl);
             var path = gitHubUrlBuilder.Path.Split('/', '\\', StringSplitOptions.RemoveEmptyEntries);
-            if (path.Length == 1)
+            var isOrgRunner = path.Length == 1;
+            var isRepoOrEnterpriseRunner = path.Length == 2;
+            var isRepoRunner = !string.Equals(path[0], "enterprises", StringComparison.OrdinalIgnoreCase);
```
luketomlinson, line 51: Would it be better to say 
```csharp
var isRepoRunner = isRepoOrEnterpriseRunner && !string.Equals(path[0], "enterprises", StringComparison.OrdinalIgnoreCase);
```
Otherwise, `isRepoRunner` would be true even in the org case (may not matter how you're using it, but it'd be good to not have a wrong value if we use it in the future)

src/Runner.Common/RunnerDotcomServer.cs
```diff
@@ -90,7 +102,11 @@ public async Task<List<TaskAgentPool>> GetRunnerGroupsAsync(string githubUrl, st
...
```
luketomlinson, line 107: same logic question here

src/Runner.Common/RunnerDotcomServer.cs
lucavallin: Updated!

src/Runner.Common/RunnerDotcomServer.cs
lucavallin: Updated!

luketomlinson (approved):

Issue #3505, opened 2024-10-11
Runner fails to pass check on arm (armv7l/armhf)


satmandu:
<!--
👋 You're opening a bug report against the GitHub Actions **runner application**.

🛑 Please stop if you're not certain that the bug you're seeing is in the runner application - if you have general problems with actions, workflows, or runners, please see the [GitHub Community Support Forum](https://github.community/c/code-to-cloud/52) which is actively monitored.  Using the forum ensures that we route your problem to the correct team.  😃
-->

**Describe the bug**

Having built an ubuntu 24.04/24.10 armhf based docker container and installed the runner, I get an error trying to start the container. Things work just fine with an arm64 container.

This is what I see:
```
docker@e80c31f556be:/$ /start.sh

Host:                                                                                                                                         
  Version:      8.0.10                                                                                                                        
  Architecture: arm                                                                                                                           
  Commit:       81cabf2857                                                                                                                    
  RID:          linux-arm                                                                                                                     
                                                                                                                                              
.NET SDKs installed:                                                                                                                          
  No SDKs were found.                                                                                                                         
                                                                                                                                              
.NET runtimes installed:                                                                                                                      
  Microsoft.NETCore.App 8.0.10 [/home/docker/.dotnet/shared/Microsoft.NETCore.App]            
                                                                                                                                              
Other architectures found:                                                                                                                    
  None                                                                                                                                        
                                                                                                                                              
Environment variables:                                                                                                                        
  DOTNET_ROOT       [/home/docker/.dotnet]                                                                                                    

global.json file:
  Not found

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
  https://aka.ms/dotnet/download
                                                                                                              
                                                                                                                                              
--------------------------------------------------------------------------------                                                              
|        ____ _ _   _   _       _          _        _   _                      |                                                              
|       / ___(_) |_| | | |_   _| |__      / \   ___| |_(_) ___  _ __  ___      |                                                              
|      | |  _| | __| |_| | | | | '_ \    / _ \ / __| __| |/ _ \| '_ \/ __|     |                                                              
|      | |_| | | |_|  _  | |_| | |_) |  / ___ \ (__| |_| | (_) | | | \__ \     |                                                              
|       \____|_|\__|_| |_|\__,_|_.__/  /_/   \_\___|\__|_|\___/|_| |_|___/     |                                                              
|                                                                              |                                                              
|                       Self-hosted runner registration                        |                                                              
|                                                                              |                                                              
[...]

PR #3575, opened 2024-11-15
Bump runner to dotnet 9.

TingluoHuang:
Fixes https://github.com/actions/runner/issues/3505

https://github.com/dotnet/core/discussions/9285

Make sure the runner is still working on ARM32 with Ubuntu24.04


rekha-prakash-maersk:
Hi @TingluoHuang , Thanks for the PR, looking for this feature as GE already rolled out 24.04 and I am using this OS runner image but need the 24.04. any idea when this will be available ?

TingluoHuang:
Sorry for the delay, we need to run extra tests to make sure dotnet9 won't cause any new problem with customers. 😢

PR #3588, opened 2024-11-21
Ignore error when fail to report worker crash.

TingluoHuang:
This fix will make sure we stop renew lock after the worker crash for whatever reason.

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -545,28 +545,36 @@ namespace GitHub.Runner.Listener
                                     detailInfo = string.Join(Environment.NewLine, workerOutput);
                                     Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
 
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -545,28 +545,36 @@ await processChannel.SendAsync(
                                     detailInfo = string.Join(Environment.NewLine, workerOutput);
                                     Trace.Info($"Return code {returnCode} indicate worker encounter an unhandled exception or app crash, attach worker stdout/stderr to JobRequest result.");
 
-                                    var jobServer = await InitializeJobServerAsync(systemConnection);
-                                    var unhandledExceptionIssue = new Issue() { Type = IssueType.Error, Message = detailInfo };
-                                    unhandledExceptionIssue.Data[Constants.Runner.InternalTelemetryIssueDataKey] = Constants.Runner.WorkerCrash;
-                                    switch (jobServer)
+                                    try
```
TingluoHuang, line 548: hide whitespace for better diff view.

yaananth (approved):

Issue #3523, opened 2024-10-23
GitHub actions package has vulnerabilities


kantkrishan:
**Describe the bug**
’m trying to build a Docker image using the GitHub Actions runner as the base image. After building it, I scan the image with Twistlock, which identifies several vulnerabilities listed in the “What's not working” section.

The Dockerfile includes the following commands:
```
FROM ghcr.io/actions/actions-runner:latest

USER runner
```

**To Reproduce**

	1.	Use the provided Dockerfile to build the image.
	2.	Scan the image using Twistlock.
	3.	Observe the reported vulnerabilities.

**Expected behavior**
The image should not contain critical or high-severity CVEs.

## Runner Version and Platform
2.320.0 

OS of the machine running the runner? OSX/Windows/Linux/...
Linux

## What's not working?
Several high-severity vulnerabilities are detected. We would like to avoid the presence of critical and high-severity CVEs. Below are the details of these vulnerabilities—please assist in resolving them in a future version:
CVE-2023-42282 - Critical
CVE-2023-50782 - high
CVE-2023-49083 - high
CVE-2022-29217 - high
CVE-2022-25883 - high
CVE-2019-0981   - high
CVE-2019-0980  - high
CVE-2023-0286 - high

## Job Log Output
Not necessary.

## Runner and Worker's Diagnostic Logs
Not required.

marcellodesales:
Those are spreading to the latest base images... Even https://github.com/actions/runner/releases/tag/v2.325.0 is affected by it...

PR #3610, opened 2024-12-05
Update libraries to resolve known vulnerabilities

joey-squid:
This updates System.Text.Encodings.Web, System.Private.Uri, System.Formats.Asn1 and System.IO.Compression.ZipFile, all of which have known vulnerabilities, some dating back a few years. See also issue #3523 which has some overlap with these changes.


TingluoHuang:
i think we have bumped our dependency to latest version as part of the work to consume dotnet 8.

PR #3622, opened 2024-12-12
Allow the use of '--network' container option

marko-k0:
# Problem

User is not able to use `host` network type if for whatever reason that is what is wanted (e.g. IPv6).

# Solution

Do not create a network if `--network` (e.g. `--network host`) is used under container options.


zoltan:
any update on this from the github team?

zoltan:
is this solved? else why close it?

PR #3645, opened 2025-01-09
Print immutable action package details in set up job logs

heavymachinery:
Soon the server will begin to respond with additional details about the package when an immutable action is downloaded. If the package details are available, we want to print them in the `Set up job` logs so it is clear when a workflow run used an immutable action.

[Related internal tracking issue](https://github.com/github/package-registry-team/issues/8276)


src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -108,6 +109,21 @@ private static ActionDownloadInfo ToServerData(ActionDownloadInfoResponse action
...
+            {
+                return null;
+            }
+
+            return new ActionDownloadPackageDetails
+            {
+                Version = actionDownloadPackageDetails.Version,
+                ManifestDigest = actionDownloadPackageDetails.ManifestDigest
+            };
+        }
```
thyeggman, line 125: Here's a more concise way to write this method:

```suggestion
        private static ActionDownloadPackageDetails? ToServerData(ActionDownloadPackageDetailsResponse? actionDownloadPackageDetails) =>
            actionDownloadPackageDetails == null ? null : new ActionDownloadPackageDetails
            {
                Version = actionDownloadPackageDetails.Version,
                ManifestDigest = actionDownloadPackageDetails.ManifestDigest
            };
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -775,7 +775,19 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
                 IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
                 Directory.CreateDirectory(destDirectory);
-                executionContext.Output($"Download action repository '{downloadInfo.NameWithOwner}@{downloadInfo.Ref}' (SHA:{downloadInfo.ResolvedSha})");
+
+                if (downloadInfo.PackageDetails != null) 
+                {
+                    executionContext.Output($"##[group]Download immutable action package '{downloadInfo.NameWithOwner}@{downloadInfo.Ref}'");
+                    executionContext.Output($"Version: {downloadInfo.PackageDetails.Version}");
+                    executionContext.Output($"Digest: {downloadInfo.PackageDetails.ManifestDigest}");
+                    executionContext.Output($"Commit SHA: {downloadInfo.ResolvedSha}");
```
conorsloan, line 784: Nit: Should we call this "source commit SHA"? It might make it slightly clearer that it's not the actual commit contents

conorsloan: Also worth noting: we could consider not including this since it's basically what the package manifest _says_ its source commit is - we can't verify that. However because it's used for the whole action caching thing currently, it likely makes sense to just keep it in there.

conorsloan (approved):
One comment but feel free to ignore! We might want product input on the specifics of the output, maybe @tinaheidinger in the absence of @Steve-Glass , but he'll be back next week anyway.

src/Runner.Worker/ActionManager.cs
heavymachinery: Good call, changed it to "source commit SHA" 👍 IMO I would still like to include it just for reference, and because folks will probably expect it to be there since we include it in the existing logs

src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
heavymachinery: Fair enough 😄 I will probably leave it as is, personally I find that reads a little less clearly, and it is less consistent with existing code in that file: https://github.com/actions/runner/blob/ea2794d5d103e0ab3e8387e4ad1fd5613fd64086/src/Sdk/WebApi/WebApi/LaunchHttpClient.cs#L98-L110

ericsciple (approved):

src/Runner.Worker/ActionManager.cs
```diff
@@ -775,7 +775,19 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
                 // make sure we get a clean folder ready to use.
                 IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
                 Directory.CreateDirectory(destDirectory);
-                executionContext.Output($"Download action repository '{downloadInfo.NameWithOwner}@{downloadInfo.Ref}' (SHA:{downloadInfo.ResolvedSha})");
+
+                if (downloadInfo.PackageDetails != null)
```
ericsciple, line 779: I'm guessing we have a feature flag on the server controlling whether the new PackageDetails property is added.

ericsciple: Also want to make sure the payload from the server is backwards compatible with older runners.

heavymachinery: Yep it's all feature flagged & backwards compatible 👍 Thanks for the 👀 on this!

PR #3696, opened 2025-02-03
Do not retry CompleteJobAsync for known non-retryable errors

ericsciple:


src/Runner.Worker/JobRunner.cs
```diff
@@ -318,9 +318,30 @@ private async Task<TaskResult> CompleteJobAsync(IRunServer runServer, IExecution
...
                     return result;
                 }
+                catch (VssUnauthorizedException ex) when (jobContext.Global.Variables.GetBoolean(Constants.Runner.Features.SkipRetryCompleteJobUponKnownErrors) ?? false)
+                {
+                    Trace.Error($"Catch exception while attempting to complete job {message.JobId}, job request {message.RequestId}.");
+                    Trace.Error(ex);
+                    exceptions.Add(ex);
+                    break;
+                }
+                catch (TaskOrchestrationJobNotFoundException ex) when (jobContext.Global.Variables.GetBoolean(Constants.Runner.Features.SkipRetryCompleteJobUponKnownErrors) ?? false)
```
yaananth, line 338: do we need this `jobContext.Global.Variables.GetBoolean(Constants.Runner.Features.SkipRetryCompleteJobUponKnownErrors)`? Only be thrown by `CompleteJob2Async` which is already guraded anyway right?

yaananth (approved):

src/Runner.Worker/JobRunner.cs
ericsciple: Yes because there are retries at different layers for job-completed. In the case of a conclusive response from Run Service, we need to stop retries at both this layer and at the HTTP client layer.

PR #3721, opened 2025-02-21
Update Dockerfile: set default locale to en-us.UTF-8

aibaars:
This should avoid problems with non-ascii characters in file names, as reported by https://github.com/github/codeql-action/issues/2777


jglick:
`C.UTF-8` would also be appropriate for a build tool image.

PR #3726, opened 2025-02-26
Install Docker Compose plugin into Actions Runner image

villapx-path:
We (Path Robotic) have started to use ARC in a k8s cluster for our GitHub Actions workflows, and in attempting porting some of our workflows, I noticed that the Docker Compose plugin is not installed in the ARC runner image. This PR installs the Docker Compose plugin, using the same method for how it's installed in the [Actions Runner Controller image](https://github.com/actions/actions-runner-controller/blob/master/runner/actions-runner.ubuntu-22.04.dockerfile#L83-L91).

I know that the alternative solution here is to create our own runner image based `FROM ghcr.io/actions/runner`, but I am hoping to avoid the extra maintenance on our side of keeping our derived image up to date.

The addition of Docker Compose increases the on-disk/uncompressed image size by 13 MiB, from 1.55 GiB up to 1.68 GiB.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -6,6 +6,7 @@ ARG TARGETARCH
 ARG RUNNER_VERSION
 ARG RUNNER_CONTAINER_HOOKS_VERSION=0.6.1
 ARG DOCKER_VERSION=27.5.1
+ARG DOCKER_COMPOSE_VERSION=v2.33.1
 ARG BUILDX_VERSION=0.20.1
 
 RUN apt update -y && apt install curl unzip -y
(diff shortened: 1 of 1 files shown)


villapx-path:
@TingluoHuang Is this something you all would be interested in?

smaug87 (approved):

smaug87 (approved):

smaug87 (commented):
> [`_beauty_`]

villapx-path:
@TingluoHuang Hi there, just hoping to bump this in case you didn't have time to take a look before.

Cheers!

PR #3773, opened 2025-03-28
Small runner code cleanup.

TingluoHuang:
In prepare for my up coming change, I am making small changes to reduce future PR's diff.

### DIFF
--- a/src/Runner.Listener/BrokerMessageListener.cs
+++ b/src/Runner.Listener/BrokerMessageListener.cs
@@ -29,6 +29,7 @@ namespace GitHub.Runner.Listener
         private TaskAgentSession _session;
         private IRunnerServer _runnerServer;
         private IBrokerServer _brokerServer;
+        private ICredentialManager _credMgr;
         private readonly Dictionary<string, int> _sessionCreationExceptionTracker = new();
         private bool _accessTokenRevoked = false;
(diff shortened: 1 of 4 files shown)


src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -329,7 +330,7 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
             }
         }
 
-        public async Task RefreshListenerTokenAsync(CancellationToken cancellationToken)
```
TingluoHuang, line 332: removed the `CancellationToken cancellationToken` from the interface method since no one is using it.

src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -432,17 +433,10 @@ ex is AccessDeniedException ||
 
         private async Task RefreshBrokerConnectionAsync()
         {
-            var configManager = HostContext.GetService<IConfigurationManager>();
-            _settings = configManager.LoadSettings();
-
-            if (string.IsNullOrEmpty(_settings.ServerUrlV2))
```
TingluoHuang, line 438: we don't need to check this again, we are in BrokerMessageListener only if we have the ServerUrlV2 in the first place.

ericsciple (approved):

PR #3777, opened 2025-03-31
Add option in OAuthCred to load authUrlV2.

TingluoHuang:
Here is the normally flow we auth with service today:
- Load `.credentials` file to create a `VssCredential` object
- Use the created `VssCredential` to `ConnectAsync` for different server, ex: RunnerServer, BrokerServer, RunServer, ActionsRunServer, etc.

The service will send down a newer version of the `.credentials` config that looks like the following:
```
{
  "scheme": "OAuth",
  "data": {
    "clientId": "GUID",
    "authorizationUrl": "https://tokenghub.actions.githubusercontent.com/_apis/oauth2/token/GUID",
    "authorizationUrlV2": "https://runner-auth.actions.githubusercontent.com/",
    "requireFipsCryptography": "True"
  }
}
```

I am changing the runner.listener to realize there might be 2 types of `VssCredential`, one created with `authorizationUrl`, another created with `authorizationUrlV2`.

Different `VssCredential` will be used for `ConnectAsync` for different server.
- RunnerServer: `authorizationUrl`
- ActionsRUnServer: `authorizationUrl`
- BorkerServer: `authorizationUrlV2`
- RunServer: `authorizationUrlV2`

This PR is not making real change of using `authorizationUrlV2`, since we pass `allowAuthUrlV2: false` in all the places we create `VssCredential` via `VssCredentials GetVssCredentials(IHostContext context, bool allowAuthUrlV2);`

https://github.com/github/actions-fusion/issues/2105

### DIFF
--- a/src/Runner.Listener/Configuration/CredentialManager.cs
+++ b/src/Runner.Listener/Configuration/CredentialManager.cs
@@ -13,7 +13,7 @@ namespace GitHub.Runner.Listener.Configuration
     public interface ICredentialManager : IRunnerService
     {
         ICredentialProvider GetCredentialProvider(string credType);
-        VssCredentials LoadCredentials();
+        VssCredentials LoadCredentials(bool allowAuthUrlV2);
     }
 
     public class CredentialManager : RunnerService, ICredentialManager
@@ -40,7 +40,7 @@ namespace GitHub.Runner.Listener.Configuration
             return creds;
         }
 
-        public VssCredentials LoadCredentials()
+        public VssCredentials LoadCredentials(bool allowAuthUrlV2)
         {
             IConfigurationStore store = HostContext.GetService<IConfigurationStore>();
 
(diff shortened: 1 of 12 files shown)


src/Runner.Listener/Configuration/CredentialManager.cs
```diff
@@ -51,21 +51,16 @@ public VssCredentials LoadCredentials()
 
             CredentialData credData = store.GetCredentials();
             var migratedCred = store.GetMigratedCredentials();
-            if (migratedCred != null)
+            if (migratedCred != null &&
+                migratedCred.Scheme == Constants.Configuration.OAuth)
```
TingluoHuang, line 55: Hosted runner will have different auth schema.

src/Runner.Listener/RunnerConfigUpdater.cs
```diff
@@ -197,6 +197,16 @@ private async Task UpdateRunnerCredentialsAsync(string serviceType, string confi
                     await ReportTelemetryAsync($"Credential clientId in refreshed config '{refreshedClientId ?? "Empty"}' does not match the current credential clientId '{clientId}'.");
                     return;
                 }
+
+                //  make sure the credential authorizationUrl in the refreshed config match the current credential authorizationUrl for OAuth auth scheme
+                var authorizationUrl = _credData.Data.GetValueOrDefault("authorizationUrl", null);
```
TingluoHuang, line 202: this will make sure the `.credentials_migrated` is identical with `.credentials` for us to safely fallback.

ericsciple (approved):

PR #3786, opened 2025-04-03
Enable auth migration based on config refresh.

TingluoHuang:
After refresh `.credentials` with the service, we can enable or disable auth migration based on whether `authorizationurlv2` exists or not.

### DIFF
--- a/src/Runner.Common/BrokerServer.cs
+++ b/src/Runner.Common/BrokerServer.cs
@@ -37,6 +37,7 @@ namespace GitHub.Runner.Common
 
         public async Task ConnectAsync(Uri serverUri, VssCredentials credentials)
         {
+            Trace.Entering();
             _brokerUri = serverUri;
 
             _connection = VssUtil.CreateRawConnection(serverUri, credentials);
--- a/src/Runner.Listener/MessageListener.cs
+++ b/src/Runner.Listener/MessageListener.cs
@@ -245,7 +245,6 @@ namespace GitHub.Runner.Listener
                     // Decrypt the message body if the session is using encryption
                     message = DecryptMessage(message);
 
-
                     if (message != null && message.MessageType == BrokerMigrationMessage.MessageType)
                     {
                         var migrationMessage = JsonUtility.FromString<BrokerMigrationMessage>(message.Body);
@@ -305,6 +304,10 @@ namespace GitHub.Runner.Listener
                 {
                     Trace.Error("Catch exception during get next message.");
                     Trace.Error(ex);
+
+                    // clear out potential message for broker migration,
+                    // in case the exception is thrown from get message from broker-listener.
+                    message = null;
 
                     // don't retry if SkipSessionRecover = true, DT service will delete agent session to stop agent from taking more jobs.
                     if (ex is TaskAgentSessionExpiredException && !_settings.SkipSessionRecover && (await CreateSessionAsync(token) == CreateSessionResult.Success))
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
-using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -31,6 +30,8 @@ namespace GitHub.Runner.Listener
         private ITerminal _term;
         private bool _inConfigStage;
         private ManualResetEvent _completedCommand = new(false);
+        private IRunnerServer _runnerServer;
+        private RunnerSettings _runnerSettings;
 
         // <summary>
         // Helps avoid excessive calls to Run Service when encountering non-retriable errors from /acquirejob.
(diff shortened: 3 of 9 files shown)


src/Runner.Common/BrokerServer.cs
```diff
@@ -37,6 +37,7 @@ public sealed class BrokerServer : RunnerService, IBrokerServer
 
         public async Task ConnectAsync(Uri serverUri, VssCredentials credentials)
         {
+            Trace.Entering();
```
TingluoHuang, line 40: add verbose trace.

src/Runner.Listener/MessageListener.cs
```diff
@@ -306,6 +305,10 @@ public async Task<TaskAgentMessage> GetNextMessageAsync(CancellationToken token)
                     Trace.Error("Catch exception during get next message.");
                     Trace.Error(ex);
 
+                    // clear out potential message for broker migration,
+                    // in case the exception is thrown from get message from broker-listener.
+                    message = null;
```
TingluoHuang, line 310: in case we got migration message from pipelines, but fail to get message broker-listener.
we currently return the migration message to the caller which the caller don't know how to handle the migration message.

src/Runner.Listener/Runner.cs
```diff
@@ -50,7 +51,8 @@ public override void Initialize(IHostContext hostContext)
         {
             base.Initialize(hostContext);
             _term = HostContext.GetService<ITerminal>();
-            _acquireJobThrottler = HostContext.CreateService<IErrorThrottler>();
```
TingluoHuang, line 53: we should always use `GetService` instead of `CreateService` for singleton service.

src/Runner.Listener/Runner.cs
```diff
@@ -252,7 +254,7 @@ public async Task<int> ExecuteCommand(CommandSettings command)
                     }
                 }
 
-                RunnerSettings settings = configManager.LoadSettings();
+                _runnerSettings = configManager.LoadSettings();
```
TingluoHuang, line 257: make this a global variable, so we can reuse it later (next PR) without reloading it from disk.

src/Runner.Listener/RunnerConfigUpdater.cs
```diff
@@ -211,7 +211,17 @@ private async Task UpdateRunnerCredentialsAsync(string serviceType, string confi
...
-            await ReportTelemetryAsync("Runner credentials updated successfully.");
+
+            if (refreshedCredConfig.Data.ContainsKey("authorizationUrlV2"))
+            {
+                HostContext.EnableAuthMigration("Credential file updated");
+                await ReportTelemetryAsync("Runner credentials updated successfully. Auth migration is enabled.");
+            }
+            else
+            {
+                HostContext.DeferAuthMigration(TimeSpan.FromDays(365), "Credential file does not contain authorizationUrlV2");
```
TingluoHuang, line 222: this is acting as disable the migration via FF from the service.

src/Test/L0/TestHostContext.cs
```diff
@@ -103,8 +103,8 @@ public async Task Delay(TimeSpan delay, CancellationToken token)
                 handler(this, new DelayEventArgs(delay, token));
             }
 
-            // Delay zero
-            await Task.Delay(TimeSpan.Zero);
```
TingluoHuang, line 107: `zero`  will make the test case stuck in case we use the HostContext.Delay in some background task.

src/Runner.Listener/MessageListener.cs
ericsciple: Good find

src/Runner.Listener/Runner.cs
ericsciple: This was originally intended not to be a singleton. Why switch to singleton?

ericsciple: > The error throttler helps us back off when encountering successive, non-retriable errors from /acquirejob.

src/Runner.Listener/Runner.cs
ericsciple: nit: _instance_ not _global_

ericsciple: Many methods throughout this class receive `RunnerSettings` as a parameter. Will those all switch to use the instance variable instead?

src/Runner.Listener/RunnerConfigUpdater.cs
ericsciple: I'm fuzzy on this detail. This PR seems fine for now. I will catch up on this detail on the next PR.

ericsciple (approved):

src/Runner.Listener/Runner.cs
TingluoHuang: i didn't find any place we use `IErrorThrottler` outside of the Runner, so i thought we only need one.
I guess, if the original idea is to create `IErrorThrottler` as needed, then `CreateService` make more sense.

src/Runner.Listener/Runner.cs
TingluoHuang: nvm, i am going to switch back, the instance var might cause more changes...

src/Runner.Listener/Runner.cs
TingluoHuang: i changed it back for now.

ericsciple (approved):

PR #3787, opened 2025-04-03
Support auth migration using authUrlV2 in Runner/MessageListener.

TingluoHuang:
We use `var cred=LoadCredentials(false)` for any connection to `RunnerServer` and `ActionsRunServer`.
We use `var cred=LoadCredentials(true)` for any connection to `BrokerServer` and `RunServer`.

In case any request to `BrokerServer` and `RunServer` hit exception, invoke `HostContext.DeferAuthMigration()` as needed.

https://github.com/github/actions-fusion/issues/2105

### DIFF
--- a/src/Runner.Listener/BrokerMessageListener.cs
+++ b/src/Runner.Listener/BrokerMessageListener.cs
@@ -26,6 +26,7 @@ namespace GitHub.Runner.Listener
         private TaskAgentStatus runnerStatus = TaskAgentStatus.Online;
         private CancellationTokenSource _getMessagesTokenSource;
         private VssCredentials _creds;
+        private VssCredentials _credsV2;
         private TaskAgentSession _session;
         private IRunnerServer _runnerServer;
         private IBrokerServer _brokerServer;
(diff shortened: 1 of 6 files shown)


src/Runner.Listener/BrokerMessageListener.cs
```diff
@@ -88,7 +90,8 @@ public async Task<CreateSessionResult> CreateSessionAsync(CancellationToken toke
                 try
                 {
                     Trace.Info("Connecting to the Broker Server...");
-                    await _brokerServer.ConnectAsync(new Uri(serverUrlV2), _creds);
+                    _credsV2 = _credMgr.LoadCredentials(allowAuthUrlV2: true);
```
ericsciple, line 93: Is this bool used?

It looks like it's passed to `GetVssCredentials` but unused within that method
- https://github.com/actions/runner/blob/c1095ae2d100886a7865f466357f09ef30dec1b9/src/Runner.Listener/Configuration/CredentialProvider.cs#L36C85-L36C99

TingluoHuang: it's only used in `OAuthCredential` 
https://github.com/actions/runner/blob/c1095ae2d100886a7865f466357f09ef30dec1b9/src/Runner.Listener/Configuration/OAuthCredential.cs

PR #3804, opened 2025-04-10
Allow enable auth migration by default.

TingluoHuang:
This pull request introduces changes to enable authentication migration by default when certain conditions are met. The most important changes include modifications to the `EnableAuthMigration` method, updates to the `ExecuteCommand` method, and the addition of a new unit test to verify the new functionality.

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -268,7 +268,7 @@ namespace GitHub.Runner.Common
                 }
             }
 
-            _trace.Info($"Enable auth migration at {_deferredAuthMigrationTime.ToString("O")}.");
+            _trace.Info($"Enable auth migration at {DateTime.UtcNow.ToString("O")}.");
             AuthMigrationChanged?.Invoke(this, new AuthMigrationEventArgs(trace));
         }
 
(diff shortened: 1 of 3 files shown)


src/Runner.Common/HostContext.cs
```diff
@@ -268,7 +268,7 @@ public void EnableAuthMigration(string trace)
                 }
             }
 
-            _trace.Info($"Enable auth migration at {_deferredAuthMigrationTime.ToString("O")}.");
```
TingluoHuang, line 271: copilot used the wrong time before... `_deferredAuthMigrationTime` is a time in the future, i want to log the time at this point.

src/Runner.Listener/Runner.cs
```diff
@@ -309,6 +309,15 @@ public async Task<int> ExecuteCommand(CommandSettings command)
                         _term.WriteLine("https://docs.github.com/en/actions/hosting-your-own-runners/autoscaling-with-self-hosted-runners#using-ephemeral-runners-for-autoscaling", ConsoleColor.Yellow);
                     }
 
+                    var cred = store.GetCredentials();
+                    if (cred != null &&
+                        cred.Scheme == Constants.Configuration.OAuth &&
+                        cred.Data.ContainsKey("EnableAuthMigrationByDefault"))
```
TingluoHuang, line 315: `EnableAuthMigrationByDefault` in the `.credentials` will make the runner to try migration without waiting for service.

pje (approved):

src/Runner.Listener/Runner.cs
TingluoHuang: it should never be null, however, i added check since this a new code path, and i can't patch the runner after it releases.

PR #3811, opened 2025-04-16
Feature-flagged support for `JobContext.CheckRunID`

pje:
If the feature flag is enabled upstream, the job context will now contain a value for `check_run_id`, which workflow authors can use in job contexts like `${{ job.check_run_id }}`.


src/Runner.Worker/JobContext.cs
```diff
@@ -56,5 +56,31 @@ public DictionaryContextData Container
                 }
             }
         }
+
+        public double? CheckRunID
```
ericsciple, line 60: nit: `CheckRunId` per csharp naming conventions

ericsciple: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalizing-compound-words-and-common-terms

ericsciple (dismissed):

ericsciple (approved):

Felixoid:
How can this kind of information be used?

It's already been 3 months since the merge, and `job` context still does not have `check_run_id` https://github.com/Felixoid/actions-experiments/actions/runs/16805751735/job/47597603856#step:5:329

PR #3812, opened 2025-04-16
Prevent overwriting ACTIONS_RESULTS_URL if already set

sunziping2016:


sunziping2016:
Sorry. This PR is not intended.

PR #3830, opened 2025-05-02
Implement termination grace period support for the runner

TingluoHuang:
`export ACTIONS_RUNNER_TERMINATION_GRACE_PERIOD_SECONDS=100`

The runner will not cancel its current runner step during the grace period.

However, depends on how child process launched in the step, those process might still exit by itself if they dont' `trap '' INT TERM`.

Try to improve: https://github.com/actions/runner/issues/3308

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -37,6 +37,8 @@ namespace GitHub.Runner.Listener
         private readonly object _authMigrationTelemetryLock = new();
         private IRunnerServer _runnerServer;
         private CancellationTokenSource _authMigrationTelemetryTokenSource = new();
+        private bool _runnerExiting = false;
+        private bool _hasTerminationGracePeriod = false;
 
         // <summary>
         // Helps avoid excessive calls to Run Service when encountering non-retriable errors from /acquirejob.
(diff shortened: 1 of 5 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -430,9 +452,13 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
                     bool autoUpdateInProgress = false;
                     Task<bool> selfUpdateTask = null;
                     bool runOnceJobReceived = false;
-                    jobDispatcher = HostContext.CreateService<IJobDispatcher>();
+                    jobDispatcher = HostContext.GetService<IJobDispatcher>();
```
TingluoHuang, line 455: Change to `GetService`, so i can fetch the same instance in the event handler.
We never need more than 1 jobDispatcher in the listener, so switching to Get should be safe.

src/Runner.Listener/Runner.cs
```diff
@@ -37,6 +37,8 @@ public sealed class Runner : RunnerService, IRunner
         private readonly object _authMigrationTelemetryLock = new();
         private IRunnerServer _runnerServer;
         private CancellationTokenSource _authMigrationTelemetryTokenSource = new();
+        private bool _runnerExiting = false;
+        private bool _hasTerminationGracePeriod = false;
```
TingluoHuang, line 41: i use `_hasTerminationGracePeriod` as the feature flag, so unless the new ENV is set, no new code will get executed.

PR #3846, opened 2025-05-07
Make sure the token's claims are match as expected.

TingluoHuang:
https://github.com/github/actions-fusion/issues/2436

### DIFF
--- a/src/Runner.Listener/Runner.cs
+++ b/src/Runner.Listener/Runner.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
+using System.Security.Claims;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -15,7 +16,9 @@ using GitHub.Runner.Common.Util;
 using GitHub.Runner.Listener.Check;
 using GitHub.Runner.Listener.Configuration;
 using GitHub.Runner.Sdk;
+using GitHub.Services.OAuth;
 using GitHub.Services.WebApi;
+using GitHub.Services.WebApi.Jwt;
 using Pipelines = GitHub.DistributedTask.Pipelines;
 
 namespace GitHub.Runner.Listener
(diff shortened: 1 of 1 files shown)


src/Runner.Listener/Runner.cs
```diff
@@ -756,6 +763,131 @@ private void HandleAuthMigrationChanged(object sender, AuthMigrationEventArgs e)
...
+                                {
+                                    // if baseline has the claim, v2 should have it too with exactly same value.
+                                    if (baselineClaims.FirstOrDefault(c => c.Type.ToLowerInvariant() == claim) is Claim baselineClaim &&
+                                        !string.IsNullOrEmpty(baselineClaim?.Value))
+                                    {
+                                        var v2Claim = v2Claims.FirstOrDefault(c => c.Type.ToLowerInvariant() == claim);
+                                        if (v2Claim?.Value != baselineClaim.Value)
+                                        {
+                                            Trace.Info($"Token Claim mismatch between two issuers. Expected: {baselineClaim.Type}:{baselineClaim.Value}. Actual: {v2Claim?.Type ?? "Empty"}:{v2Claim?.Value ?? "Empty"}");
+                                            HostContext.DeferAuthMigration(TimeSpan.FromMinutes(60), $"Expected claim {baselineClaim.Type}:{baselineClaim.Value} does not match {v2Claim?.Type ?? "Empty"}:{v2Claim?.Value ?? "Empty"}");
```
ericsciple, line 869: Will this trigger any telemetry? Wondering how we know whether this is happening.

ericsciple: I noticed the while loop sleeps for 100m yet this for loop defers 60m. Is that discrepancy OK?

src/Runner.Listener/Runner.cs
```diff
@@ -756,6 +763,131 @@ private void HandleAuthMigrationChanged(object sender, AuthMigrationEventArgs e)
...
+            ];
+
+            try
+            {
+                var credMgr = HostContext.GetService<ICredentialManager>();
+                while (!token.IsCancellationRequested)
+                {
+                    try
+                    {
+                        await HostContext.Delay(TimeSpan.FromMinutes(100), token);
```
ericsciple, line 796: Should this only delay after the first iteration? Or first successful iteration? Or does it matter?

ericsciple, line 796: Does it matter than an intermittent network IO error may cause the logic further below not be skipped for more than multiple 100m delays?

ericsciple (approved):

src/Runner.Listener/Runner.cs
TingluoHuang: Any `DeferAuthMigration()` should publish telemetry back to service.

TingluoHuang: it's on purposely to not lines up so different things won't trigger at the same time.

src/Runner.Listener/Runner.cs
TingluoHuang: i don't want the runner to get multiple tokens during start up, try to avoid too much load on the service.

TingluoHuang: We should not see any claim diff from the token.
We will use Ring 0 to fully test this out and make sure we never see any failure.

PR #3853, opened 2025-05-12
Prefer _migrated config on startup

lokesh755:


src/Runner.Listener/Runner.cs
```diff
@@ -406,15 +406,65 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
+                    migratedSettings = configManager.LoadMigratedSettings();
+                    Trace.Info("Loaded migrated settings from .runner_migrated file");
+                    Trace.Info(migratedSettings);
                 }
-                else if (createSessionResult == CreateSessionResult.Failure)
+                catch (Exception ex)
                 {
-                    return Constants.Runner.ReturnCode.TerminatedError;
+                    // If migrated settings file doesn't exist or can't be loaded, we'll use the provided settings
+                    Trace.Info($"Failed to load migrated settings: {ex.Message}");
```
TingluoHuang, line 423: do we want do something during the catch in case the `migratedSettings` is empty?

src/Runner.Listener/Runner.cs
```diff
@@ -733,10 +788,17 @@ await configUpdater.UpdateRunnerConfigAsync(
...
                         configManager.DeleteLocalRunnerConfig();
                     }
                 }
+
+                // After cleanup, check if we need to restart the session
+                if (restartSession)
+                {
+                    Trace.Info("Restarting runner session after config update...");
+                    RunnerSettings newSettings = configManager.LoadSettings();
+                    return await RunAsync(newSettings, runOnce);
```
TingluoHuang, line 800: i am not sure we should do this recursive call, we never did this in the runner before, might be risky.

src/Runner.Listener/Runner.cs
```diff
@@ -406,15 +406,65 @@ private async Task<int> RunAsync(RunnerSettings settings, bool runOnce = false)
...
+                
+                if (migratedSettings != null)
+                {
+                    // Try to create session with migrated settings first
+                    Trace.Info("Attempting to create session using migrated settings");
+                    _listener = GetMessageListener(migratedSettings, isMigratedSettings: true);
+                    
+                    try
+                    {
+                        CreateSessionResult createSessionResult = await _listener.CreateSessionAsync(HostContext.RunnerShutdownToken);
```
TingluoHuang, line 436: do we want to report telemetry back to service in case of any failure, so we know customers are having issue with migration.

src/Runner.Listener/Runner.cs
```diff
@@ -679,6 +730,10 @@ await configUpdater.UpdateRunnerConfigAsync(
                                     configType: runnerRefreshConfigMessage.ConfigType,
                                     serviceType: runnerRefreshConfigMessage.ServiceType,
                                     configRefreshUrl: runnerRefreshConfigMessage.ConfigRefreshUrl);
+
+                                // Set flag to restart session and break out of the message loop
+                                restartSession = true;
```
TingluoHuang, line 735: we don't need to restart if we refresh the auth config.

src/Runner.Common/Constants.cs
```diff
@@ -155,6 +155,7 @@ public static class ReturnCode
                 public const int RunnerUpdating = 3;
                 public const int RunOnceRunnerUpdating = 4;
                 public const int SessionConflict = 5;
+                public const int RunnerConfigurationRefreshed = 6;
```
TingluoHuang, line 158: can we add some comment to indicates we can remove this return code later, and the service script will never receive this return code?

PR #3874, opened 2025-05-28
show helpful error message when resolving actions directly with launch

aiqiaoy:
Related: https://github.com/github/actions-runtime/issues/5135

Before:
Screenshot 2025-05-27 at 22 45 04

vs. After:
Screenshot 2025-05-27 at 22 45 17

actions-dotnet implementation: https://github.com/github/actions-dotnet/blob/f9322e925c9f23d04fb974295604864fd4a195cc/Launch/Sdk/Server/LaunchApi.cs#L87-L98

### DIFF
--- a/src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
+++ b/src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Net.Http.Formatting;
 using System.Net.Http.Headers;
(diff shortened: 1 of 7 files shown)


src/Sdk/WebApi/WebApi/LaunchHttpClient.cs
```diff
@@ -32,11 +33,49 @@ public LaunchHttpClient(
...
+                else
+                {
+                    combinedErrorMessage = responseError;
+                }
+
+                throw new UnresolvableActionDownloadInfoException(combinedErrorMessage);
+            }
+            else if (response.StatusCode == HttpStatusCode.TooManyRequests)
+            {
+                throw new NonRetryableActionDownloadInfoException(responseError + " (GitHub has reached an internal rate limit, please try again later)");
```
TingluoHuang, line 70: should we retry on 429?
is this a new behavior or existing behavior with actions-dotnet?

ericsciple: As far as I understand, 429 should be retried

aiqiaoy: Added

TingluoHuang (approved):

PR #3910, opened 2025-06-14
fix: bad kill mode in template

### LINKED ISSUE #3909: Self-hosted runners launched through systemd should have different defaults
**Describe the bug**
When following the steps to [configure a self-hosted runner service](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/configuring-the-self-hosted-runner-application-as-a-service#customizing-the-self-hosted-runner-service) the recommended way is to use the template, [found here](https://github.com/actions/runner/blob/097ada9374c9bde944aa9fa3de59ae2e656e79cf/src/Misc/layoutbin/actions.runner.service.template). 

The file content is, 

```service
[Unit]
Description={{Description}}
After=network.target

[Service]
ExecStart={{RunnerRoot}}/runsvc.sh
User={{User}}
WorkingDirectory={{RunnerRoot}}
KillMode=process
KillSignal=SIGTERM
TimeoutStopSec=5min

[Install]
WantedBy=multi-user.target
```
When used in tandem with the `Restart=always` option or when manually restarting the service, child processes are still hanging around. Which causes multiple processes to pickup a job under the same runner and it will crash.

The flow is:

```txt
service -> calls runsvc.sh -> calls RunnerService.js -> spawns children processes
```
As seen in the `RunnerService.js` file,

```js
        console.log("Starting Runner listener with startup type: service");
        listener = childProcess.spawn(
          listenerExePath,
          ["run", "--startuptype", "service"],
          { env: process.env }
        );
      }
```

**To Reproduce**
1. Start a runner with a service
2. Restart the service
3. Witness multiple children processes still hanging around from the last process.

**Expected behavior**
No orphan processes hanging around after one of the (grand-)parent processes has been killed.

Since we're using systemd for lifecycle management of the processes, it should be handling all processes spawned by the service.

v2.325.0

RHEL8

## Fix
We found that simply setting `KillMode=control-group` was enough to fix the issue, it is also what is recommended in [the systemd docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html).

Nathan-Furnal:
Fixes #3909

### DIFF
--- a/src/Misc/layoutbin/actions.runner.service.template
+++ b/src/Misc/layoutbin/actions.runner.service.template
@@ -6,7 +6,7 @@ After=network.target
 ExecStart={{RunnerRoot}}/runsvc.sh
 User={{User}}
 WorkingDirectory={{RunnerRoot}}
-KillMode=process
+KillMode=control-group
 KillSignal=SIGTERM
 TimeoutStopSec=5min
 


Ujang360:
+1, ran into this on a fleet of ~20 self-hosted runners and this one-liner fixed it.

One thing worth adding to the orphan angle: because the cgroup never gets torn
down on restart, its `memory.current` also just keeps climbing and a
`systemctl restart` won't bring it back. On a runner doing heavy Rust builds ours
sat at `~3.2G`, but only `~35M` of that was actual process memory — the rest was page
cache pinned to a cgroup that had been alive `~17h` across a bunch of restarts. The
thing keeping it non-empty was a stray `ssh-agent` reparented to pid 1 (exactly the
orphan case described in #3909).

With this change + a restart while idle it drops to ~1M and the orphan's gone. Same
result across all 20 runners.

It's not a leak — the cache is reclaimable — but it makes per-runner memory
monitoring (and any `MemoryMax` you set) basically useless. Only gotcha: `control-group`
will `SIGKILL` a running job if you restart mid-job. Auto-update is fine (SelfUpdater
waits for the job before restarting), so it's really only manual / `Restart=always`
restarts you'd want to drain first.

In our case, we `restart` the service after the runner stopped via hook.

Anyway — one-line change, open for a year. Any chance of a look?

Issue #3915, opened 2025-06-18
Runner registration API has been deprecated and needs to be updated


Baibhav168:
**Bug**
API to register a new self-hosted-runner has been updated, and corresponding change in new runner registration is missing.
Earlier it used : 
https://github.com/actions/runner/blob/main/scripts/create-latest-svc.sh#L132
`export RUNNER_TOKEN=$(curl -s -X POST ${base_api_url}/${orgs_or_repos}/${runner_scope}/actions/runners/registration-token -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" | jq -r '.token')
`
to generate a token .

However new API looks like this: 
`curl -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${RUNNER_CFG_PAT}" -H "X-GitHub-Api-Version: 2022-11-28"  https://${base_api_url}/${orgs_or_repos}/${runner_scope}/actions/runners/registration-token
`

**To Fix**
https://github.com/actions/runner/blob/main/scripts/create-latest-svc.sh#L132 needs to be updated with the new curl cmd

Baibhav168:
Raised PR for the fix: https://github.com/actions/runner/pull/3916

yurii-shylov:
Our self-hosted runners suddenly stopped working today, and while investigating the issue I stumbled upon this thread. When was the API deprecated? Could it affect existing runners? We haven't touched our configs in months

Baibhav168:
[Yurii Shylov](https://github.com/yurii-shylov): 
Litmus test: If your existing runner is showing `idle/active` in the github repo settings -> ... -> Runners. Then its good, otherwise if it is showing `offline`, you need to use the updated API to re-authenticate your self-hosted runner.

yurii-shylov:
Our issue was caused by a different thing. Apparently GitHub recently increased JIT configs for the runners to exceed 4 Kb, and it stopped fitting into SSM parameters. It was already fixed here: https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4613

PR #3916, opened 2025-06-18
Update for Deprecated API

Baibhav168:
https://github.com/actions/runner/issues/3915

**Bug**
API to register a new self-hosted-runner has been updated, and corresponding change in new runner registration is missing.
Earlier it used :
https://github.com/actions/runner/blob/main/scripts/create-latest-svc.sh#L132
export RUNNER_TOKEN=$(curl -s -X POST ${base_api_url}/${orgs_or_repos}/${runner_scope}/actions/runners/registration-token -H "accept: application/vnd.github.everest-preview+json" -H "authorization: token ${RUNNER_CFG_PAT}" | jq -r '.token') 
to generate a token .
Corresponding Bug: https://github.com/actions/runner/issues/3915

However new API looks like this:
curl -L -X POST -H "Accept: application/vnd.github+json" -H "Authorization: Bearer ${RUNNER_CFG_PAT}" -H "X-GitHub-Api-Version: 2022-11-28"  https://${base_api_url}/${orgs_or_repos}/${runner_scope}/actions/runners/registration-token 

To Fix
https://github.com/actions/runner/blob/main/scripts/create-latest-svc.sh#L132 needs to be updated with the new curl cmd

### DIFF
--- a/scripts/create-latest-svc.sh
+++ b/scripts/create-latest-svc.sh
@@ -129,7 +129,7 @@ if [[ "$runner_scope" == *\/* ]]; then
     orgs_or_repos="repos"
 fi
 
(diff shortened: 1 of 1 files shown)


Baibhav168:
Corresponding Bug: https://github.com/actions/runner/issues/3915

PR #3928, opened 2025-07-03
Add loongarch64 partial

chenguohui:
As now the nodejs and dotnet-sdk of LoongArch64 are still not published officially, this PR aims to add LoongArch64 constants and RID information only. 

* [LA] Add LoongArch64 to Constants
* [LA] Add linux-loongarch64 to RuntimeIdentifiers in csprojs
* [LA] Add LOONGARCH64 to DefineConstants conditionally

The runner I built locally with the unofficially built nodejs and dotnet-sdk can work under `self-hosted` mode normally.

### DIFF
--- a/src/Runner.Common/Runner.Common.csproj
+++ b/src/Runner.Common/Runner.Common.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
     <OutputType>Library</OutputType>
-    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64;osx-arm64;win-arm64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64;osx-arm64;win-arm64;linux-loongarch64</RuntimeIdentifiers>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
     <NoWarn>NU1701;NU1603;SYSLIB0050;SYSLIB0051</NoWarn>
     <Version>$(Version)</Version>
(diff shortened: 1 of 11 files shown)


chenguohui:
@shushanhf

src/Runner.Common/Runner.Common.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
     <OutputType>Library</OutputType>
-    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64;osx-arm64;win-arm64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;win-x86;linux-x64;linux-arm64;linux-arm;osx-x64;osx-arm64;win-arm64;linux-loongarch64</RuntimeIdentifiers>
```
alexrp: Unfortunately, as I found out for RISC-V, adding the RIDs here will not work when building on Microsoft-supported platforms like `linux-x64` because the SDK will fail to find `linux-loongarch64` apphost packages. So you'll have to maintain this part of the patch downstream.

The rest of the changes look uncontroversial to me, however, and I plan to propose equivalent changes for `ppc64le`, `riscv64`, and `s390x`.

chenguohui: Thanks, this is still a good news to me! I revert the change in `src/Runner.Common/Runner.Common.csproj`.

alexrp: Sorry, I should have clarified: This will be the case for all project files, unfortunately, not just this one. :slightly_frowning_face:

chenguohui: All right, I revert the change of all project files.

xen0n: For the record: the related dotnet upstream issue regarding `linux-loongarch64` AppHost packages is https://github.com/dotnet/sdk/issues/42248.

driver1998: @alexrp If AppHost is all it needs, one can disable the use of AppHost with `UseAppHost=false`. And maybe replace the AppHost with a script that executes `dotnet app.dll` so that `./app` still works.

If we need other platform-dependent packages from NuGet, then yes it is blocked by https://github.com/dotnet/sdk/issues/42248.

PR #3946, opened 2025-07-17
Fix null reference exception in user agent handling

salmanmkc:
Ensure that user agents are loaded and checked for null values to prevent null reference exceptions when processing orchestration IDs.

### DIFF
--- a/src/Runner.Worker/JobRunner.cs
+++ b/src/Runner.Worker/JobRunner.cs
@@ -50,12 +50,10 @@ namespace GitHub.Runner.Worker
             if (message.Variables.TryGetValue(Constants.Variables.System.OrchestrationId, out VariableValue orchestrationId) &&
                 !string.IsNullOrEmpty(orchestrationId.Value))
             {
-                HostContext.LoadDefaultUserAgents();
-                
                 if (!HostContext.UserAgents.Any(x => string.Equals(x.Product?.Name, "OrchestrationId", StringComparison.OrdinalIgnoreCase)))
                 {
                     // make the orchestration id the first item in the user-agent header to avoid get truncated in server log.
-                    HostContext.UserAgents?.Insert(0, new ProductInfoHeaderValue("OrchestrationId", orchestrationId.Value));
+                    HostContext.UserAgents.Insert(0, new ProductInfoHeaderValue("OrchestrationId", orchestrationId.Value));
                 }
 
                 // make sure orchestration id is in the user-agent header.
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/JobRunner.cs
```diff
@@ -50,10 +50,12 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
             if (message.Variables.TryGetValue(Constants.Variables.System.OrchestrationId, out VariableValue orchestrationId) &&
                 !string.IsNullOrEmpty(orchestrationId.Value))
             {
-                if (!HostContext.UserAgents.Any(x => string.Equals(x.Product.Name, "OrchestrationId", StringComparison.OrdinalIgnoreCase)))
+                HostContext.LoadDefaultUserAgents();
+                
+                if (HostContext.UserAgents != null && !HostContext.UserAgents.Any(x => string.Equals(x.Product?.Name, "OrchestrationId", StringComparison.OrdinalIgnoreCase)))
```
TingluoHuang: we don't need to check `HostContext.UserAgents != null` it's always initialized

salmanmkc: done

src/Runner.Worker/JobRunner.cs
```diff
@@ -50,10 +50,12 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
             if (message.Variables.TryGetValue(Constants.Variables.System.OrchestrationId, out VariableValue orchestrationId) &&
                 !string.IsNullOrEmpty(orchestrationId.Value))
             {
-                if (!HostContext.UserAgents.Any(x => string.Equals(x.Product.Name, "OrchestrationId", StringComparison.OrdinalIgnoreCase)))
+                HostContext.LoadDefaultUserAgents();
```
TingluoHuang: we don't need this, `LoadDefaultUserAgents` is called when we create the HostContext.

src/Runner.Worker/JobRunner.cs
```diff
@@ -50,10 +50,12 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                
+                if (!HostContext.UserAgents.Any(x => string.Equals(x.Product?.Name, "OrchestrationId", StringComparison.OrdinalIgnoreCase)))
                 {
                     // make the orchestration id the first item in the user-agent header to avoid get truncated in server log.
-                    HostContext.UserAgents.Insert(0, new ProductInfoHeaderValue("OrchestrationId", orchestrationId.Value));
+                    HostContext.UserAgents?.Insert(0, new ProductInfoHeaderValue("OrchestrationId", orchestrationId.Value));
```
TingluoHuang: we also don't need the `?.` since it's never empty.

TingluoHuang (approved):

lopez19940503-tech:
fix-null-ref-exception

St0nersdash (approved):

PR #3948, opened 2025-07-21
Node 20 -> Node 24 migration feature flagging, opt-in and opt-out environment variables

salmanmkc:
# Node 20 to Node 24 Migration Strategy

This is a proposal for the migration strategy, nothing here is 100% confirmed yet.

## Overview

This PR implements the migration strategy for GitHub Actions JavaScript/TypeScript actions from Node 20 to Node 24, preparing for Node 20's end of life. The approach uses feature flags for a phased rollout, similar to the previous Node 16 to Node 20 migration.

## Key Features

- Phased migration with clear transition points
- User controls via environment variables
- Feature flags for controlled rollout
- Compatibility handling for Linux ARM32
- Informative messaging to guide users through the transition

### Phase 1: Node 20 Default

Initially Node 20 remains the default:
- Default: Node 20
- Opt-in to Node 24: Set `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true`

### Phase 2: Node 24 Default

When the `actions.runner.usenode24bydefault` flag is enabled:
- Default: Node 24
- Temporary fallback to Node 20: Set `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`
- Informational message appears explaining Node 20 deprecation and opt-out options

### Phase 3: Node 24 Required

When the `actions.runner.requirenode24` flag is enabled:
- All actions use Node 24
- No opt-out options available

## Environment Variables

The migration uses two environment variables:

- `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true`: Opt into Node 24 during Phase 1
- `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`: Fall back to Node 20 during Phase 2

These can be set at either the workflow level (in YAML) or runner level (system environment). Workflow settings override runner settings to ensure job-specific control.

### Core Components

The migration strategy includes:

1. **Centralized Constants**
   - Version identifiers: `Node20` and `Node24`
   - Environment variable definitions
   - Feature flag definitions

2. **Version Selection Logic**
   - Logic that factors in:
     - Current migration phase
     - Environment variables (workflow and runner)
     - Action metadata
     - Platform compatibility

3. **User Feedback**
   - Warning when conflicting settings are detected
   - Helpful messages during Phase 2 explaining the Node version change
   - Clear instructions for temporarily using Node 20 when needed

4. **Platform Compatibility**
   - Special handling for Linux ARM32 where Node 24 isn't supported
   - Automatic fallback to Node 20 with explanatory message
   - Multiple safeguards to ensure consistent behavior

## Testing & Compatibility

Tests cover:
- Various feature flag and environment variable combinations
- All migration phases
- Environment variable precedence scenarios
- Platform-specific behavior (including ARM32)
- Conflict detection

Backward compatibility is maintained through:
- Zero-change workflow execution during migration
- Clear warning messages for configuration issues
- Automatic fallbacks on unsupported platforms
- Feature-flagged rollout for controlled deployment

### Variable Precedence

- **Workflow-level variables** always take precedence over **System-level variables** for the same variable
- Each variable (`FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` and `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION`) is evaluated independently
- Variables from different sources can be combined in the final decision

### Phase 1: Node 20 Default (No Feature Flags)

| ForceNode24 (Workflow) | AllowUnsecure (Workflow) | ForceNode24 (System) | AllowUnsecure (System) | Resulting Node Version | Notes |
|------------------------|--------------------------|----------------------|-----------------------|------------------------|-------|
| true                   | true                     | -                    | -                     | Node 20                | Warning: conflicting settings |
| true                   | false                    | -                    | -                     | Node 24                | |
| true                   | -                        | -                    | -                     | Node 24                | |
| false                  | true                     | -                    | -                     | Node 20                | |
| false                  | false                    | -                    | -                     | Node 20                | |
| false                  | -                        | -                    | -                     | Node 20                | |
| -                      | true                     | -                    | -                     | Node 20                | |
| -                      | false                    | -                    | -                     | Node 20                | |
| -                      | -                        | true                 | true                  | Node 20                | Warning: conflicting settings |
| -                      | -                        | true                 | false                 | Node 24                | |
| -                      | -                        | true                 | -                     | Node 24                | |
| -                      | -                        | false                | true                  | Node 20                | |
| -                      | -                        | false                | false                 | Node 20                | |
| -                      | -                        | false                | -                     | Node 20                | |
| -                      | -                        | -                    | true                  | Node 20                | |
| -                      | -                        | -                    | false                 | Node 20                | |
| -                      | -                        | -                    | -                     | Node 20                | Default |

### Phase 2: Node 24 Default (useNode24ByDefault=true)

| ForceNode24 (Workflow) | AllowUnsecure (Workflow) | ForceNode24 (System) | AllowUnsecure (System) | Resulting Node Version | Notes |
|------------------------|--------------------------|----------------------|-----------------------|------------------------|-------|
| true                   | true                     | -                    | -                     | Node 20                | Warning: conflicting settings |
| true                   | false                    | -                    | -                     | Node 24                | |
| true                   | -                        | -                    | -                     | Node 24                | |
| false                  | true                     | -                    | -                     | Node 20                | |
| false                  | false                    | -                    | -                     | Node 24                | |
| false                  | -                        | -                    | -                     | Node 24                | |
| -                      | true                     | -                    | -                     | Node 20                | |
| -                      | false                    | -                    | -                     | Node 24                | |
| -                      | -                        | true                 | true                  | Node 20                | Warning: conflicting settings |
| -                      | -                        | true                 | false                 | Node 24                | |
| -                      | -                        | true                 | -                     | Node 24                | |
| -                      | -                        | false                | true                  | Node 20                | |
| -                      | -                        | false                | false                 | Node 24                | |
| -                      | -                        | false                | -                     | Node 24                | |
| -                      | -                        | -                    | true                  | Node 20                | |
| -                      | -                        | -                    | false                 | Node 24                | |
| -                      | -                        | -                    | -                     | Node 24                | Default |
| true                   | -                        | -                    | true                  | Node 20                | Important edge case: AllowUnsecure=true always takes precedence in Phase 2 |
| -                      | true                     | true                 | -                     | Node 20                | Important edge case: AllowUnsecure=true always takes precedence in Phase 2 |

### Phase 3: Node 24 Required (requireNode24=true)

In Phase 3, the environment variables are ignored and Node 24 is always used.

#### 1. Mixed-Source Variables (Workflow vs System)

When variables come from different sources, each is evaluated independently:

**Example:** ForceNode24=false (Workflow), AllowUnsecure=true (System)
- The workflow explicitly opts out of Node 24 by setting ForceNode24=false
- The system has AllowUnsecure=true
- Result: Node 20 is used (workflow's opt-out from Node 24 is respected)

**Example:** AllowUnsecure=false (Workflow), ForceNode24=true (System)
- The workflow explicitly disallows Node 20 by setting AllowUnsecure=false
- The system has ForceNode24=true
- Result: Node 24 is used (system's ForceNode24 is respected since workflow doesn't override it)

**Key Edge Case in Phase 2:** When both variables are true (from any sources)
- If ForceNode24=true (Workflow) and AllowUnsecure=true (System)
- Or ForceNode24=true (System) and AllowUnsecure=true (Workflow)
- Result: Node 20 is used because AllowUnsecure=true takes precedence in Phase 2
- This is logical since in Phase 2, Node 24 is already the default, so the request to use Node 20 is more important

#### 2. Conflicting Settings

When both variables are set to true from the same source (both from workflow or both from system):

- Warning message is displayed
- Default version for the current phase is used
- Workflow level conflicts trigger the same behavior as system level conflicts

#### 3. ARM32 Linux Platform Compatibility

For Linux ARM32 platforms, Node 24 is not supported:

- If Node 24 would normally be selected, it automatically falls back to Node 20
- Warning message is displayed explaining the compatibility issue
- This occurs regardless of environment variables or feature flags

#### 4. Phase 2 User Notification

In Phase 2, when Node 24 is selected:
- Users see an informational message explaining that Node 20 is deprecated
- Instructions are provided for temporarily using Node 20 if needed

# Environment Variable Examples

## Opt-in to Node 24 using workflow environment variable

You can opt in to Node 24 during Phase 1 by adding the environment variable to your workflow:

```yaml
env:
  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true

jobs:
  build:
    runs-on: node24
    
    steps:
      - uses: actions/checkout@v4
```

This forces actions like `actions/checkout@v4` (which normally uses Node 20) to run with Node 24 instead:

Opt in using FORCE_JAVASCRIPT_ACTIONS_TO_NODE24 workflow env variable

## Opt-in to Node 24 using system environment variable

You can also opt in at the runner level by setting the system environment variable:

Using system variable for opt in

## This is how it behaves when node24 is default
Node 24 default on non-node 20 actions

## When node24 is default, actions can be opted out using the `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` environment variable

`ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION`

For example:

```name: node20 but default node24 opt out

on:
  workflow_dispatch:

env:
  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true

jobs:
  build:
    runs-on: node24
    
    steps:
      - uses: actions/checkout@v4

```
Opt out of node 24 default

## Environment Variable Precedence

When both system-level and workflow-level environment variables are set:

1. **Workflow-level variables** always take precedence over **system-level variables** for the same variable
2. Variables are evaluated independently based on their source
3. In Phase 1 (Node 20 default):
   - If `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true`, Node 24 is used
   - Otherwise, Node 20 is used
4. In Phase 2 (Node 24 default):
   - If `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`, Node 20 is used
   - Otherwise, Node 24 is used


salilsub:
I think that if someone sets both of the variables to true, we should default to the same logic that is used when neither is set. Setting both should only happen with user error, I think it's more intuitive that we pick the default in that case.
Looking at it again, I also think that if someone sets both, we probably should add an annotation to say both variables are being set and what version is being defaulted to.

salmanmkc:
> I think that if someone sets both of the variables to true, we should default to the same logic that is used when neither is set. Setting both should only happen with user error, I think it's more intuitive that we pick the default in that case. Looking at it again, I also think that if someone sets both, we probably should add an annotation to say both variables are being set and what version is being defaulted to.

That makes sense, modified it

src/Runner.Worker/FeatureManager.cs
```diff
@@ -11,5 +11,20 @@ public static bool IsContainerHooksEnabled(Variables variables)
             var isContainerHooksPathSet = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.Hooks.ContainerHooksPath));
             return isContainerHookFeatureFlagSet && isContainerHooksPathSet;
         }
+        
+        public static bool IsFeatureEnabled(Variables variables, string featureFlag)
+        {
+            return variables?.GetBoolean(featureFlag) ?? false;
+        }
+        
+        public static bool IsUseNode24ByDefaultEnabled(Variables variables)
```
TingluoHuang: is this `IsRequireNode24Enabled` and `IsUseNode24ByDefaultEnabled` used anywhere?

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -26,14 +107,34 @@ public static string GetInternalNodeVersion()
...
+        /// <param name="variableName">The name of the environment variable</param>
+        /// <param name="workflowEnvironment">Optional dictionary containing workflow-level environment variables</param>
+        /// <returns>True if the variable is set to "true" in either environment</returns>
+        private static bool IsEnvironmentVariableTrue(string variableName, IDictionary<string, string> workflowEnvironment)
+        {
+            // Workflow environment variables take precedence over system environment variables
+            // If the workflow explicitly sets the value (even to false), we respect that over the system value
+            if (workflowEnvironment != null && workflowEnvironment.TryGetValue(variableName, out string workflowValue))
+            {
+                return string.Equals(workflowValue, "true", StringComparison.OrdinalIgnoreCase);
```
TingluoHuang: use `StringUtil.ConvertToBoolean` for all string -> bool conversion.

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -18,6 +19,86 @@ public static string GetInternalNodeVersion()
...
+                return (Constants.Runner.NodeMigration.Node24, null);
+            }
+
+            // Check if both flags are set from the same source
+            bool bothFromWorkflow = false;
+            bool bothFromSystem = false;
+
+            if (workflowEnvironment != null)
+            {
+                bool workflowHasForce = workflowEnvironment.TryGetValue(Constants.Runner.NodeMigration.ForceNode24Variable, out string forceValue) &&
```
TingluoHuang: we already kind of checked the `workflowEnvironment` inside `IsEnvironmentVariableTrue`, should we flat the logic a little bit to make sure we only check those condition once?

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -18,6 +19,86 @@ public static string GetInternalNodeVersion()
...
+                bool workflowHasAllow = workflowEnvironment.TryGetValue(Constants.Runner.NodeMigration.AllowUnsecureNodeVersionVariable, out string allowValue) &&
+                                      !string.IsNullOrEmpty(allowValue);
+
+                bothFromWorkflow = workflowHasForce && workflowHasAllow &&
+                                  string.Equals(forceValue, "true", StringComparison.OrdinalIgnoreCase) &&
+                                  string.Equals(allowValue, "true", StringComparison.OrdinalIgnoreCase);
+            }
+
+            // Check if both are set in system and neither is overridden by workflow
+            string sysForce = Environment.GetEnvironmentVariable(Constants.Runner.NodeMigration.ForceNode24Variable);
```
TingluoHuang: same for these, we kind of did this already in `IsEnvironmentVariableTrue`.

src/Runner.Common/Constants.cs
```diff
@@ -170,6 +170,22 @@ public static class Features
...
+                public static readonly string Node20 = "node20";
+                public static readonly string Node24 = "node24";
+                
+                // Environment variables for controlling node version selection
+                public static readonly string ForceNode24Variable = "FORCE_JAVASCRIPT_ACTIONS_TO_NODE24";
+                public static readonly string AllowUnsecureNodeVersionVariable = "ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION";
+                
+                // Feature flags for controlling the migration phases
+                public static readonly string UseNode24ByDefaultFlag = "actions.runner.usenode24bydefault";
+                public static readonly string RequireNode24Flag = "actions.runner.requirenode24";
```
TingluoHuang, line 187: do we need this `actions.runner.requirenode24`?

would that be same as we remove node20 from the runner package once its LTS end?

src/Runner.Worker/FeatureManager.cs
salmanmkc: oh ya I realised I didn't update it to use that, I've updated it now to use them

src/Runner.Common/Util/NodeUtil.cs
salmanmkc: thanks!

src/Runner.Common/Util/NodeUtil.cs
salmanmkc: sure, I've done that now

salmanmkc: + updated to include source information, etc.

src/Runner.Common/Util/NodeUtil.cs
salmanmkc: 👍

src/Runner.Common/Constants.cs
salmanmkc: Yeah basically, but since it's feature flagged, we can turn it on/off, not sure about controlled roll out strategy here if it's possible, to do it via a gradual roll out with the feature flag or is this a global level one? Feature flagging though allows us to easily just turn if off, rather than having to time a release, and prepare a new PR, the future PR will just be a cleanup of this, also might be useful to have this logic be re-used for future node versions if it's refactored in the future?

src/Runner.Worker/FeatureManager.cs
TingluoHuang: i think we can remove `IsFeatureEnabled` since it's not used by anything outside and the method only has 1 line. 😄

salmanmkc: cool ya, removed those extra methods too, I think it's okay

src/Test/L0/Worker/FeatureManagerL0.cs.new
TingluoHuang, line 1: this file is named `.cs.new`. 😄

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -58,10 +58,41 @@ public IHandler Create(
...
+                        executionContext.Warning(configWarningMessage);
+                    }
+
+                    if (!string.IsNullOrEmpty(platformWarningMessage))
+                    {
+                        executionContext.Warning(platformWarningMessage);
+                    }
+
+                    // Show information about Node 24 migration in Phase 2
+                    if (useNode24ByDefault && !requireNode24 && string.Equals(finalNodeVersion, Constants.Runner.NodeMigration.Node24, StringComparison.OrdinalIgnoreCase))
```
TingluoHuang, line 90: do we need to make sure we also check `nodeData.NodeVersion`, so if the customers already use node24, we don't printout the message.

src/Runner.Common/Constants.cs
TingluoHuang: i am fine with having the feature flag, just want to point out the final state would be no node20 in the runner release package.

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -1,8 +1,36 @@
...
+
+        /// <summary>
+        /// Gets or sets whether the value came from the system environment
+        /// </summary>
+        public bool FromSystem { get; set; }
+
+        /// <summary>
+        /// Gets or sets the raw string value of the environment variable
+        /// </summary>
+        public string RawValue { get; set; }
```
TingluoHuang: do we use the `RawValue`?

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -1,8 +1,36 @@
...
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using GitHub.Runner.Sdk;
 
 namespace GitHub.Runner.Common.Util
 {
+    /// <summary>
+    /// Represents details about an environment variable, including its value and source
+    /// </summary>
+    public class EnvironmentVariableInfo
```
TingluoHuang: make this a private class if we don't plan to expose it right now.

src/Runner.Common/Util/NodeUtil.cs
```diff
@@ -18,6 +46,70 @@ public static string GetInternalNodeVersion()
...
+                Constants.Runner.NodeMigration.ForceNode24Variable, workflowEnvironment);
+
+            var allowUnsecureNodeDetails = GetEnvironmentVariableDetails(
+                Constants.Runner.NodeMigration.AllowUnsecureNodeVersionVariable, workflowEnvironment);
+
+            bool forceNode24 = forceNode24Details.IsTrue;
+            bool allowUnsecureNode = allowUnsecureNodeDetails.IsTrue;
+            string warningMessage = null;
+
+            // Phase 3: Always use Node 24 regardless of environment variables
```
TingluoHuang: i think you can move the phase3 up in this method, since we don't even need to check any env if `requreNode24` is true from the caller.

src/Runner.Worker/Handlers/HandlerFactory.cs
TingluoHuang: nvm, we already under the if condition that the customer provides node20

src/Test/L0/Worker/FeatureManagerL0.cs.new
salmanmkc: oh woops I meant to delete that

TingluoHuang (approved):

src/Runner.Worker/FeatureManager.cs
Aggerphine: Thanks

Issue #3955, opened 2025-07-25
self hosted runner broken update to 2.327.0


terickson:
**Describe the bug**
Self Hosted runner breaking upgrading to 2.327.0 of runner with error running  _update.sh file.  Here is the error:
/runner/_work/_update.sh: line 220: syntax error: unexpected end of file

It looks like not all the if conditionals in the shell script are getting closed.  After line 148 you should have a fi to close the if conditional.

**To Reproduce**
1. setup a self hosted runner that is not on 2.327.0.  We are using summerwind/actions-runner-dind as our base docker image for our runners.
2. Let it start up and you will see the runner try to upgrade to 2.327.0 and it will break with the error during the upgrade.
3. See error: /runner/_work/_update.sh: line 220: syntax error: unexpected end of file

**Expected behavior**
We expect it to upgrade with no error.

## Runner Version and Platform
Version of your runner? Trying to upgrade to: 2.327.0. 

OS of the machine running the runner? docker container: summerwind/actions-runner-dind 

## What's not working?
Current runner version: '2.316.1'
2025-07-25 13:55:27Z: Listening for Jobs
Runner update in progress, do not shutdown runner.
Downloading 2.327.0 runner
Waiting for current job finish running.
Generate and execute update script.
Runner will exit shortly for update, should be back online within 10 seconds.
Runner update process finished.
Runner listener exit because of updating, re-launch runner after successful update
/runner/_work/_update.sh: line 220: syntax error: unexpected end of file

<img width="1125" height="281" alt="Image" src="https://github.com/user-attachments/assets/2056c074-24bd-4d91-9e2d-7bf5fda23302" />

TingluoHuang:
Thanks for reporting this, we are going fix this and prepare a patch release. 🙇

salmanmkc:
The rollback has been completed, so the error shouldn't occur anymore. Next week we plan to release the above fix as a patch version. Thanks for alerting us!

patrickvinograd:
Hi, I saw the notification for patch release 2.327.1. 

Can somebody from GH clarify re: the policy of only supporting two most recent versions of `runner`, if that applies to patch releases as well as minor versions. I.e. is releasing 2.327.1 going to cause 2.326.0 to no longer work? I'd rather not have to take a brand new release on a Friday afternoon, but also given this issue it's unclear if 2.327.0 is functional.

TingluoHuang:
[...]

PR #3962, opened 2025-07-27
Add tests for template file syntax errors

salmanmkc:
Introduce tests to ensure that template files do not contain syntax errors. This enhances code reliability and maintainability.

Using shellcheck for syntax checking for sh template files https://github.com/koalaman/shellcheck

Aims to avoid issues like this https://github.com/actions/runner/issues/3955

### DIFF
--- a/src/Test/L0/Listener/ShellScriptSyntaxL0.cs
+++ b/src/Test/L0/Listener/ShellScriptSyntaxL0.cs
@@ -337,7 +337,11 @@ namespace GitHub.Runner.Common.Tests.Listener
                     // These flags help validate the syntax without fully executing the script
                     var process = new Process();
                     process.StartInfo.FileName = "cmd.exe";
-                    process.StartInfo.Arguments = $"/c /v:on /f:off /e:on \"{tempUpdatePath}\" echo SyntaxCheckOnly && exit /b 0";
+                    // Use a temporary batch file to check syntax without execution
+                    string tempBatchFile = Path.Combine(tempDir, "syntax_check.cmd");
+                    File.WriteAllText(tempBatchFile, $"@echo off\r\necho SyntaxCheckOnly\r\nexit /b 0");
+                    
+                    process.StartInfo.Arguments = $"/c \"{tempUpdatePath}\" \"{tempBatchFile}\"";
                     process.StartInfo.RedirectStandardError = true;
                     process.StartInfo.RedirectStandardOutput = true;
                     process.StartInfo.UseShellExecute = false;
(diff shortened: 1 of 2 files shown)


src/Test/L0/Listener/ShellScriptSyntaxL0.cs
```diff
@@ -0,0 +1,828 @@
...
+                return;
+            }
+            
+            // Create a test template with complex shell scripting patterns
+            string tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            Directory.CreateDirectory(tempDir);
+            string templatePath = Path.Combine(tempDir, "complex_shell.sh.template");
+
+            // Write a sample template with various shell features
+            string template = @"#!/bin/bash
```
TingluoHuang, line 348: What are we testing with these? 🤔

src/Test/L0/Listener/ShellScriptSyntaxL0.cs
TingluoHuang, line 1: Should we just replace all the placeholder strings in the script and then run `bash -n` against it?
So we don't need to add any 3rd party dependency.

Firas2515: Fr

salmanmkc: With just bash -n it wasn't able to catch all the syntax errors, I tried making various syntax errors and they would not get detected, missing brackets for example would not get detected at all

salmanmkc: For example, tests pass with missing bracket with `bash -n` and no shellcheck
test pass with missing bracket

salmanmkc: With shell check it gives an error 
syntax error

src/Test/L0/Listener/ShellScriptSyntaxL0.cs
salmanmkc: This test is valuable to keep because it tests complex shell syntax patterns that might not be present in our actual template files. It serves as validation for our syntax checking framework itself, not just our current templates.

The test covers important shell script features like:
- Nested quotes and complex quoting patterns
- Function definitions with parameter handling
- Command substitution and string manipulation
- Here documents

These patterns might not all exist in our current templates, but having this test ensures our validation system can handle them if we ever add them. It's basically testing the validator itself rather than just the templates.

I'd vote to keep it as a solid regression test for our validation framework, in-case we modify it going forward.

src/Test/L0/Listener/ShellScriptSyntaxL0.cs
salmanmkc: I have it so that if the dependency isn't there, it just uses `bash -n` still though

TingluoHuang: Now i am wondering whether we should just run the script for testing.
We want the test to be reliable, given the facts that `bash -n` might not catch expression error, and `shellcheck` may not exist on the CI machine, actually execute the script might be our best way to always test the script's syntax errors.

PR #4003, opened 2025-08-30
chore: migrate Husky config from v8 to v9 format

salmanmkc:
- Replace `husky install` with `husky` in prepare script
- Remove shebang and husky.sh import from pre-commit hook
- Maintain existing

V9 is a non-breaking change is is backwards compatible so we don't need to upgrade this in any rush even though v9 is now in main.


CosmicJesterX (commented):
well well well

Issue #3937, opened 2025-07-08
Fedora41 RPM dependency issue


volga629-1:
**Describe the bug**
I am building RPM package for Fedora Server 41.
Github action agent binaries are compiled with unsupported  liblttng-ust library version.

```
Problem: conflicting requests
  - nothing provides liblttng-ust.so.0()(64bit) needed by github-actions-agent-2.325.0-3.fc41.x86_64 from myrepo-F41-Server
  - nothing provides liblttng-ust.so.0()(64bit) needed by github-actions-agent-2.325.0-4.fc41.x86_64 from myrepo-F41-Server

```

RPM build process 

```

+ /usr/lib/rpm/redhat/brp-python-hardlink
+ /usr/bin/add-determinism --brp -j4 /home/volga629/rpmbuild/BUILD/github-actions-agent-2.325.0-build/BUILDROOT
Scanned 12 directories and 234 files,
               processed 0 inodes,
               0 modified (0 replaced + 0 rewritten),
               0 unsupported format, 0 errors
Reading /home/volga629/rpmbuild/BUILD/github-actions-agent-2.325.0-build/SPECPARTS/rpm-debuginfo.specpart
Processing files: github-actions-agent-2.325.0-4.fc41.x86_64
Provides: config(github-actions-agent) = 2.325.0-4.fc41 github-actions-agent = 2.325.0-4.fc41 github-actions-agent(x86-64) = 2.325.0-4.fc41 libSystem.Globalization.Native.so()(64bit) libSystem.IO.Compression.Native.so()(64bit) libSystem.IO.Compression.Native.so(V1.0)(64bit) libSystem.Native.so()(64bit) libSystem.Net.Security.Native.so()(64bit) libSystem.Security.Cryptography.Native.OpenSsl.so()(64bit) libclrgc.so()(64bit) libclrjit.so()(64bit) libclrjit.so(V1.0)(64bit) libcoreclr.so()(64bit) libcoreclr.so(V1.0)(64bit) libcoreclrtraceptprovider.so()(64bit) libhostfxr.so()(64bit) libhostfxr.so(V1.0)(64bit) libhostpolicy.so()(64bit) libhostpolicy.so(V1.0)(64bit) libmscordaccore.so()(64bit) libmscordaccore.so(V1.0)(64bit) libmscordbi.so()(64bit) libmscordbi.so(V1.0)(64bit)
Requires(interp): /bin/sh /bin/sh /bin/sh /bin/sh
[...]

PR #4005, opened 2025-08-30
Refactor package installation logic with fallbacks

salmanmkc:
Issue: https://github.com/actions/runner/issues/3937

Implement a fallback mechanism for package installation across different operating systems to enhance reliability during dependency installation. Used for liblttng-ust1.


salmanmkc:
Realised this probably wouldn't work

PR #4011, opened 2025-09-03
Map RUNNER_TEMP for container action

ericsciple:
Maps `RUNNER_TEMP` to `/github/runner_temp` for container actions.

Feature flag:
- `actions_container_action_runner_temp`

Below is a screenshot of the file structure. While testing, I created the file `foo` once inside the container action and confirmed it appeared twice in file structure (`touch /github/runner_temp/_runner_file_commands/foo`).

image

### DIFF
--- a/src/Runner.Worker/Handlers/ContainerActionHandler.cs
+++ b/src/Runner.Worker/Handlers/ContainerActionHandler.cs
@@ -191,11 +191,19 @@ namespace GitHub.Runner.Worker.Handlers
             ArgUtil.Directory(tempWorkflowDirectory, nameof(tempWorkflowDirectory));
 
             container.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
+            if (FeatureManager.IsContainerActionRunnerTempEnabled(ExecutionContext.Global.Variables))
+            {
+                container.MountVolumes.Add(new MountVolume(tempDirectory, "/github/runner_temp"));
+            }
             container.MountVolumes.Add(new MountVolume(tempHomeDirectory, "/github/home"));
             container.MountVolumes.Add(new MountVolume(tempWorkflowDirectory, "/github/workflow"));
             container.MountVolumes.Add(new MountVolume(tempFileCommandDirectory, "/github/file_commands"));
             container.MountVolumes.Add(new MountVolume(defaultWorkingDirectory, "/github/workspace"));
 
+            if (FeatureManager.IsContainerActionRunnerTempEnabled(ExecutionContext.Global.Variables))
+            {
+                container.AddPathTranslateMapping(tempDirectory, "/github/runner_temp");
+            }
             container.AddPathTranslateMapping(tempHomeDirectory, "/github/home");
             container.AddPathTranslateMapping(tempWorkflowDirectory, "/github/workflow");
             container.AddPathTranslateMapping(tempFileCommandDirectory, "/github/file_commands");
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -191,11 +191,19 @@ public async Task RunAsync(ActionRunStage stage)
             ArgUtil.Directory(tempWorkflowDirectory, nameof(tempWorkflowDirectory));
 
             container.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
+            if (FeatureManager.IsContainerActionRunnerTempEnabled(ExecutionContext.Global.Variables))
+            {
+                container.MountVolumes.Add(new MountVolume(tempDirectory, "/github/runner_temp"));
+            }
```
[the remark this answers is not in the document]
ericsciple: This is temporary and will be cleaned up soon.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -191,11 +191,19 @@ public async Task RunAsync(ActionRunStage stage)
             ArgUtil.Directory(tempWorkflowDirectory, nameof(tempWorkflowDirectory));
 
             container.MountVolumes.Add(new MountVolume("/var/run/docker.sock", "/var/run/docker.sock"));
+            if (FeatureManager.IsContainerActionRunnerTempEnabled(ExecutionContext.Global.Variables))
```
salmanmkc, line 194: Should we do this also with environment variables? So that we don't have to rely on just feature flags? Same comment for the check below too

Also nit: do we need the helper method since it's a one line helper method?

src/Runner.Worker/Handlers/ContainerActionHandler.cs
```diff
@@ -191,11 +191,19 @@ public async Task RunAsync(ActionRunStage stage)
...
+            {
+                container.MountVolumes.Add(new MountVolume(tempDirectory, "/github/runner_temp"));
+            }
             container.MountVolumes.Add(new MountVolume(tempHomeDirectory, "/github/home"));
             container.MountVolumes.Add(new MountVolume(tempWorkflowDirectory, "/github/workflow"));
             container.MountVolumes.Add(new MountVolume(tempFileCommandDirectory, "/github/file_commands"));
             container.MountVolumes.Add(new MountVolume(defaultWorkingDirectory, "/github/workspace"));
 
+            if (FeatureManager.IsContainerActionRunnerTempEnabled(ExecutionContext.Global.Variables))
```
salmanmkc, line 203: nit : Could we reorder this so that there's just one feature flag check call?  The also might help reduce the need for a helper method. Or does the order of operations here matter a lot? E.g. putting the mount of the new temp directory in the feature flag call at the end of the mounting section and then path translate at the beginning of the mapping section? 

Sorry quite a picky nit, feel free to ignore this, not sure on whether it makes a difference to the behavior 

Edit: also this would then make it easier to do that check plus an env variable check if added, since we'd be able to just check them all once.

Alternatively we  can just store the ff value above and then use it.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
ericsciple: The feature flag is temporary for safe rollout only. It will be cleaned up in the following runner release.

I didn't have a reason to add an environment variable check in this case, so I kept it clean and under server control. Environment variables help when customers are involved to help try out a pre-release behavior and confirm it works for their scenarios.

The only reason I added the helper method is to make this file read consistently wrt feature flag checks.

src/Runner.Worker/Handlers/ContainerActionHandler.cs
ericsciple: I optimized the code, to make the diff easier to read when the feature flag is removed (future PR).

I prefer checking the feature flag twice rather than introduce a local variable. Rule of thumb: more local variables -\> worse readability. Checking the feature flag isn't expensive, it's a dictionary lookup.

PR #4013, opened 2025-09-03
Break UseV2Flow into UseV2Flow and UseRunnerAdminFlow.

TingluoHuang:
https://github.com/github/actions-runner-admin/issues/1898


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -400,13 +400,26 @@ public async Task ConfigureAsync(CommandSettings command)
             }
             else
             {
-
                 throw new NotSupportedException("Message queue listen OAuth token.");
             }
 
+            // allow the server to override the serverUrlV2 and useV2Flow
```
TingluoHuang, line 406: these got moved up since i want to check them in line 422

src/Runner.Listener/Configuration/CredentialManager.cs
```diff
@@ -89,7 +89,7 @@ public sealed class GitHubAuthResult
         public string Token { get; set; }
 
         [DataMember(Name = "use_v2_flow")]
-        public bool UseV2Flow { get; set; }
+        public bool UseRunnerAdminFlow { get; set; }
```
TingluoHuang, line 92: this is returned from dotcom to indicates whether we should use runner-admin flow.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -211,7 +211,7 @@ public async Task ConfigureAsync(CommandSettings command)
             string poolName = null;
             TaskAgentPool agentPool = null;
             List<TaskAgentPool> agentPools;
-            if (runnerSettings.UseV2Flow)
+            if (runnerSettings.UseRunnerAdminFlow)
```
TingluoHuang, line 214: check the `UseRunnerAdminFlow` before calling any `_dotcomServer`

src/Runner.Listener/Configuration/CredentialManager.cs
TingluoHuang: i keep the datamember name the same so we don't need to make a deployment to the service.

src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -400,13 +400,26 @@ public async Task ConfigureAsync(CommandSettings command)
...
+                !string.IsNullOrEmpty(serverUrlV2))
+            {
+                Trace.Info($"Service enforced serverUrlV2: {serverUrlV2}");
+                runnerSettings.ServerUrlV2 = serverUrlV2;
+            }
+
+            if (agent.Properties.TryGetValue("UseV2Flow", out bool useV2Flow) && useV2Flow)
+            {
+                Trace.Info($"Service enforced useV2Flow: {useV2Flow}");
+                runnerSettings.UseV2Flow = useV2Flow;
```
sean-burgess, line 417: Is it worth renaming the remaining `UseV2Flow` cases as well?  maybe `UseBrokerMessageListenerFlow`?  Just to be more explicit.

TingluoHuang: for existing self-hosted runner, we already have file on disk with that name, so we can't change it anymore.

sean-burgess (approved):

PR #4014, opened 2025-09-03
Add libicu76 to support Debian 13

recmanj:
This PR adds `libicu76` to the list of available libicu packages to `installdependencies.sh` script.

Debian stable repositories bumped libicu to version 76 and that version is the only one available in Debian 13.

### DIFF
--- a/src/Misc/layoutbin/installdependencies.sh
+++ b/src/Misc/layoutbin/installdependencies.sh
@@ -110,7 +110,7 @@ then
             exit 1
         fi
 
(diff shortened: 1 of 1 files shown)


clopez:
This is neeeded on Debian 13, otherwise the install script fails and it blocks everything. Even if you try to install by hand libicu76 then it will not be enough and `config.sh` will keep telling you to run that `install.sh` script again. However with this patch it will work.

Please someone with commit rights, land this.

Issue #324, opened 2020-02-08
Expose GITHUB_JOB_ID as an environment variable


MarshallOfSound:
**Describe the enhancement**
Expose `GITHUB_JOB_ID` as an environment variable to the running steps.  We need this in order for a job to be able to self-identify itself and potentially send it's own ID to a different service that would then use the GitHub API to perform actions (like check logs for instance).

**Code Snippet**
It looks like we need to add `job_id` to the whitelist here --> https://github.com/actions/runner/blob/master/src/Runner.Worker/GitHubContext.cs

Not sure though, hence raising this issue

ericsciple:
Note, the job ID in yaml is the key in the `jobs` mapping (terminology aligns with step `id` keyword in yaml). So if we expose this via the context or env var, we should consider the terminology conflict.

ericsciple:
@MarshallOfSound whitelist wont solve it. The job ID isn't in the github context.

```yaml
on: push
jobs:
  one:
    runs-on: ubuntu-16.04
    steps:
      - name: Dump GitHub context
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
        run: echo "$GITHUB_CONTEXT"
```

It sounds like for this scenario you need the number ([this job_id](https://developer.github.com/v3/actions/workflow_jobs/#get-a-workflow-job)), correct?

MarshallOfSound:
@ericsciple Yes we'd need the `job_id` number that the GitHub API takes in.  If that's not in the context how can we get that exposed?

ericsciple:
@TingluoHuang @juliobbv might be able to help drive an ADR for this change. IIRC logs are best effort (although lots of retries, so likely there) and uploaded asyncronously. It's unclear to me whether adding the value would actually enable the desired scenario. Ting or Julio can help figure out.

Also it doesn't look like the runner has the job ID at all. I dumped the worker log and confirmed it's not in the job message received from the server.

ericsciple:
@chrispat fyi can help prioritize or redirect if there is a better place for the request

chrispat:
@MarshallOfSound CodeCov does something similar but they just use the run_id which is in the environment and context.  Is there any specific reason you need to have the job_id?

MarshallOfSound:
@chrispat It looks like the only way to get the logs of a job is with the `job_id`? Unless I'm missing something https://developer.github.com/v3/actions/workflow_jobs/#list-workflow-job-logs

Not sure how you can pull logs without it

[...]

PR #4053, opened 2025-09-22
Add JOB_ env vars

jenseng:
Add env vars for top-level job context primitives, i.e.

  `job.check_run_id` => `$JOB_CHECK_RUN_ID`
  `job.status` => `$JOB_STATUS`
  `job.workflow_ref` => `$JOB_WORKFLOW_REF`
  `job.workflow_sha` => `$JOB_WORKFLOW_SHA`
  `job.workflow_repository` => `$JOB_WORKFLOW_REPOSITORY`
  `job.workflow_file_path` => `$JOB_WORKFLOW_FILE_PATH`

References
* https://github.com/actions/runner/issues/324
* https://github.com/orgs/community/discussions/129314
* https://github.com/orgs/community/discussions/8945#discussioncomment-14376552
* https://github.com/orgs/community/discussions/38659

### DIFF
--- a/src/Runner.Worker/JobContext.cs
+++ b/src/Runner.Worker/JobContext.cs
@@ -1,10 +1,12 @@
-using GitHub.DistributedTask.Pipelines.ContextData;
+﻿using GitHub.DistributedTask.Pipelines.ContextData;
 using GitHub.Runner.Common.Util;
 using GitHub.Runner.Common;
+using System;
+using System.Collections.Generic;
 
 namespace GitHub.Runner.Worker
 {
-    public sealed class JobContext : DictionaryContextData
+    public sealed class JobContext : DictionaryContextData, IEnvironmentContextData
     {
         public ActionResult? Status
         {
@@ -144,6 +146,34 @@ namespace GitHub.Runner.Worker
             set
             {
                 this["workflow_file_path"] = value != null ? new StringContextData(value) : null;
+            }
+        }
+
+        private readonly HashSet<string> _contextEnvAllowlist = new(StringComparer.OrdinalIgnoreCase)
+        {
+            "check_run_id",
+            "status",
+            "workflow_ref",
+            "workflow_sha",
+            "workflow_repository",
+            "workflow_file_path",
+        };
+
+        public IEnumerable<KeyValuePair<string, string>> GetRuntimeEnvironmentVariables()
+        {
+            foreach (var data in this)
+            {
+                if (_contextEnvAllowlist.Contains(data.Key))
+                {
+                    if (data.Value is StringContextData value)
+                    {
+                        yield return new KeyValuePair<string, string>($"JOB_{data.Key.ToUpperInvariant()}", value.ToString());
+                    }
+                    else if (data.Value is NumberContextData numberValue)
+                    {
+                        yield return new KeyValuePair<string, string>($"JOB_{data.Key.ToUpperInvariant()}", numberValue.ToString());
+                    }
+                }
             }
         }
     }
--- a/src/Test/L0/Worker/JobContextL0.cs
+++ b/src/Test/L0/Worker/JobContextL0.cs
@@ -1,4 +1,5 @@
-using System;
+﻿using System;
+using System.Collections.Generic;
 using GitHub.DistributedTask.Pipelines.ContextData;
 using GitHub.Runner.Worker;
 using Xunit;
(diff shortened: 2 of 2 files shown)


src/Runner.Worker/JobContext.cs
```diff
@@ -82,5 +84,29 @@ public double? CheckRunId
...
+
+        public IEnumerable<KeyValuePair<string, string>> GetRuntimeEnvironmentVariables()
+        {
+            foreach (var data in this)
+            {
+                if (_contextEnvAllowlist.Contains(data.Key))
+                {
+                    if (data.Value is StringContextData value)
+                    {
+                        yield return new KeyValuePair<string, string>($"JOB_{data.Key.ToUpperInvariant()}", value);
```
qoomon: I would go for a `GITHUB_JOB_` prefix instead of just `JOB_` to make it more consistent to other environment variables and to point out that these variables are github context variables.

qoomon (changes requested):

src/Runner.Worker/JobContext.cs
jenseng: That's not a bad idea, I'm curious to see what the maintainers think. I had originally considered that, but:

- It makes it **_less_** consistent with how other context -> env vars work, i.e. `github.` -> `GITHUB_` and `runner.` -> `RUNNER_`
- There's potentially a risk of collisions if any `github.job_...` fields are added in the future, since those would also have a `GITHUB_JOB_` prefix

Personally my preference would have been to name the context field `github.job_id`, in which case the env var could just be `GITHUB_JOB_ID`, but I think that ship has sailed 🫤

cc @ericsciple

qoomon: @jenseng I think you are right your name pattern makes sense then

alexevanczuk:
Hey @jenseng wondering if you're still hoping to get this merged 🙏🏼 excited about it!

jenseng:
> Hey @jenseng wondering if you're still hoping to get this merged 🙏🏼 excited about it!

@alexevanczuk absolutely! Just waiting for the GitHub folks to review it🤞

alexevanczuk:
@TingluoHuang @ericsciple @pje I've noticed you've all created or reviewed PRs related to this file. Is this something you'd be able to take a look at and help merge? It would help folks use this new excellent flag in a way that is conventional with how they are checking other similar properties (using environment variables). Would be a huge boon for our ability to deep link into jobs and the change looks smallish so would love it if you could take a look and help get this merged if possible 🙏🏼

qoomon:
?

jenseng:
rebased and updated to account for https://github.com/actions/runner/pull/4335

src/Runner.Worker/JobContext.cs
```diff
@@ -146,5 +148,33 @@ public string WorkflowFilePath
                 this["workflow_file_path"] = value != null ? new StringContextData(value) : null;
             }
         }
+
+        private readonly HashSet<string> _contextEnvAllowlist = new(StringComparer.OrdinalIgnoreCase)
```
ericsciple, line 152: Thanks @jenseng! I want to collect some more thoughts internally.

I'm thinking about namespace conflict (if customers already use JOB_ variables). I'm waiting on one of the individuals to get back from vacation.

jenseng: Yeah that's a fair point and definitely worth considering. Another possibility could be to have a `GITHUB_JOB_` prefix, though that has some [downsides of its own](https://github.com/actions/runner/pull/4053#discussion_r2376260679) (i.e. inconsistent with how `runner`/`github` variables are named, and it could conflict with future `github.job_...` variables)

ericsciple: Good callout on RUNNER_ variables 👍. That naming precedent helps.

We can likely add JOB_ but may need some advance announcement before it rolls out. I will discuss more with folks internally.

jenseng: 👋 hey @ericsciple, any updates here?

thinking more about potential conflicts with (existing) `JOB_` variables people might set, perhaps this could be done slightly more gradually with a feature flag...

before the feature is enabled, these new variables wouldn't get set. but if someone tries to set one of these reserved variable names (via `env` or `$GITHUB_ENV`), actions/runner would emit a warning that this will stop working (e.g. `JOB_CHECK_RUN_ID is a reserved variable name, and you will no longer be able to set it as of $DATE. Please switch to a different variable name`)

then once the feature flag is enabled, the new variables would get set automatically. then if someone tries to set one of them explicitly, the warning text would be slightly different (e.g. `JOB_CHECK_RUN_ID is a reserved variable name, so the value you've set has no effect. Please switch to a different variable name`)

that could be more trouble than it's worth though, so perhaps an advance announcement would suffice? 🤷‍♂️

it could also be good to add telemetry around the env variable names people are setting in their workflows/actions, which would give you better insights into how disruptive these kinds of changes might be

proski:
Will the variables be available in the hooks (ACTIONS_RUNNER_HOOK_JOB_STARTED and ACTIONS_RUNNER_HOOK_JOB_COMPLETED)?

jenseng:
> Will the variables be available in the hooks (ACTIONS_RUNNER_HOOK_JOB_STARTED and ACTIONS_RUNNER_HOOK_JOB_COMPLETED)?

Yes, those hooks are implemented as steps internally, and all steps will have access to these context variables (just like the existing `GITHUB_*` and `RUNNER_*` variables)

rdhar:
Seems like a natural follow-up to last week's [GitHub Actions: Early September 2026 updates](https://github.blog/changelog/2026-09-03-github-actions-early-september-2026-updates/) blog post.

PR #4066, opened 2025-10-03
Make sure runner-admin has both auth_url and auth_url_v2.

TingluoHuang:


src/Runner.Listener/Configuration/ConfigurationManager.cs
```diff
@@ -284,13 +284,21 @@ public async Task ConfigureAsync(CommandSettings command)
                             {
                                 var runner = await _dotcomServer.ReplaceRunnerAsync(runnerSettings.PoolId, agent, runnerSettings.GitHubUrl, registerToken, publicKeyXML);
                                 runnerSettings.ServerUrlV2 = runner.RunnerAuthorization.ServerUrl;
+                                runnerSettings.UseV2Flow = true; // if we are using runner admin, we also need to hit broker
```
TingluoHuang, line 287: we forgot to do this for the update-runner flow, the same code exists in the add-runner flow.

PR #4068, opened 2025-10-03
Set Restart=always on systemd service for added resilience

ksacry-ft:
This update ensures that the runner will restart if crashed.
We've experienced that under certain network conditions, or github server errors, the runner can crash.
With this change, the service will be started back up and can try again.


FredeJ:
Came here to add this exact change.

The only things that could be up for discussion if there should also be a `RestartSec=5s` just to not spin in a too fast loop if the same thing is still wrong. Similarly, it could also be changed `Restart=on-failure`.

But these are details. I'd love to get this merged!

PR #4083, opened 2025-10-20
fix(dockerfile): set more lenient permissions on /home/runner

caxu-rh:
Currently, the permissions of /home/runner are 750. In some container runtimes and Kubernetes distributions (including OpenShift), a different uid/gid (not `runner`) may be used when running the image.

The runner expects to be able to read and execute scripts within the home directory, and it will also write ephemeral files, diagnostic data, etc. into the directory as well. Therefore, to support the ability to use the runner as a user apart from `runner`, full 777 permissions are needed.

A longer-term change to consider which may improve the security posture here would be to separate the executable portions of the application (scripts, etc.) from the places where temporary data is written, and control the permissions of these separately.

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -54,14 +54,15 @@ RUN add-apt-repository ppa:git-core/ppa \
     && apt install -y git \
     && rm -rf /var/lib/apt/lists/*
 
+WORKDIR /home/runner
+
 RUN adduser --disabled-password --gecos "" --uid 1001 runner \
     && groupadd docker --gid 123 \
     && usermod -aG sudo runner \
     && usermod -aG docker runner \
     && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers \
-    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers
-
-WORKDIR /home/runner
+    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers \
+    && chmod 777 /home/runner
 
 COPY --chown=runner:docker --from=build /actions-runner .
 COPY --from=build /usr/local/lib/docker/cli-plugins/docker-buildx /usr/local/lib/docker/cli-plugins/docker-buildx


acornett21:
Hi @nikola-jokic This is what we were talking about last week when we met, is this something that you can take a look at? Thanks!

nikola-jokic (approved):
Thanks!

images/Dockerfile
```diff
@@ -59,7 +59,8 @@ RUN adduser --disabled-password --gecos "" --uid 1001 runner \
     && usermod -aG sudo runner \
     && usermod -aG docker runner \
     && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers \
-    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers
+    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers \
+    && chmod 777 /home/runner
```
TingluoHuang, line 65: will `chmod 777 /home/runner` fail since we haven't create the folder?

nikola-jokic: Yeah, just tested it, we should move `WORKDIR` above

caxu-rh:
Thanks, moved `WORKDIR` above the `RUN`. I was expecting the `adduser` to also create the home dir but guess I was wrong.

nikola-jokic:
True @caxu-rh, my bad, I tested it partially, but it is still better at least from my point of view to have `WORKDIR` above. Thank you!

TingluoHuang (approved):

images/Dockerfile
```diff
@@ -54,14 +54,15 @@ RUN add-apt-repository ppa:git-core/ppa \
...
 RUN adduser --disabled-password --gecos "" --uid 1001 runner \
     && groupadd docker --gid 123 \
     && usermod -aG sudo runner \
     && usermod -aG docker runner \
     && echo "%sudo   ALL=(ALL:ALL) NOPASSWD:ALL" > /etc/sudoers \
-    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers
-
-WORKDIR /home/runner
+    && echo "Defaults env_keep += \"DEBIAN_FRONTEND\"" >> /etc/sudoers \
+    && chmod 777 /home/runner
```
anuragstar, line 65: @TingluoHuang @nikola-jokic  This is a breaking change after upgrading the GitHub runner from 2.329.0 to 2.330.0. We are now seeing the error “failed to initialize build cache.” We need to revert back to 2.329.0. Could you please fix this so we don’t need separate handling on our side?. 

failed to initialize build cache at /home/runner/.cache/go-build: mkdir /home/runner/.cache/go-build: permission denied
  Error: Command failed: /opt/hostedtoolcache/go/1.25.1/x64/bin/go env
  failed to initialize build cache at /home/runner/.cache/go-build: mkdir /home/runner/.cache/go-build: permission denied

diogotorres97:
```
  ➤ YN0001: │ foobar@https://github.com/foo.git#commit=9434354e7bf8f19d229fecabbef9dedc678f5c47: Failed cloning the repository
  ➤ YN0001: │   Repository URL: https://github.com/foo.git
  ➤ YN0001: │   Error: invalid key:  core.autocrlf
  ➤ YN0001: │   Fatal Error: unable to write parameters to config file
  ➤ YN0001: │   Exit Code: 128
```

issues cloning private repositories

caxu-rh: @nikola-jokic - one difference I'm seeing is in the owner uid/gid of the `/home/runner` directory.

- In 2.329.0: `Access: (0750/drwxr-x---)  Uid: ( 1001/  runner)   Gid: ( 1001/  runner)`
- In 2.330.0: `Access: (0777/drwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)`

I think the ownership change is due to moving `WORKDIR` above. I don't quite understand how this ownership change is causing the observed behavior, though.

images/Dockerfile
hicksjacobp: Doesn't `adduser` create `/home/runner`?

PR #4099, opened 2025-11-10
🧹 Remove unnecessary EvaluateJobSnapshot

lawrencegripper:
Actions will have already expanded the expressions in the snapshot message, so this is no longer required.

- https://github.com/github/actions-larger-runners/issues/3566


src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -240,7 +240,7 @@ public ActionsEnvironmentReference ActionsEnvironment
         }
 
         [DataMember(EmitDefaultValue = false)]
-        public TemplateToken Snapshot
+        public Snapshot Snapshot
```
ericsciple, line 243: i think we might need a new field to avoid a compatibility problem between the runner version and the server.

PR #4100, opened 2025-11-11
test: parametrize ubuntu bases

yanksyoon:
parametrize the bulid process for all ubuntu bases to catch any specific base errors.


yanksyoon:
Wrong base, sorry!

PR #4109, opened 2025-11-13
feat: update base image to ubuntu 24.04

frezbo:
Update base image to be Ubuntu 24.04 LTS.


StamenchoBog:
This is also causing trouble with the upgrade process that I am trying to do. Commenting for reach.

TingluoHuang:
Addressed in https://github.com/actions/runner/pull/4018

PR #4138, opened 2025-11-28
Update safe_sleep.sh: use native `sleep`

op30mmd:
The previous method is actually dangerous, i guess this addresses the problem.


thomasskk:
it should use the while loop as a fallback if sleep isnt defined

op30mmd:
Thanks, i updated the code to use while loop when sleep is missing.

fingolfin:
Duplicate of PR #3143 (from February 2024 😅)

inthesquarehole:
Just use something like this for Windows: https://blog.bearcats.nl/perfect-sleep-function/.

PR #4145, opened 2025-12-10
Implement safe_sleep function for improved sleep handling

Amineharrabi:
Refactored sleep logic to use safe_sleep function with 3 fallback for sleeping

### DIFF
--- a/src/Misc/layoutroot/safe_sleep.sh
+++ b/src/Misc/layoutroot/safe_sleep.sh
@@ -1,6 +1,27 @@
 #!/bin/bash
 
-SECONDS=0
-while [[ $SECONDS -lt $1 ]]; do
-    :
-done
+safe_sleep() {
+    local t="$1"
+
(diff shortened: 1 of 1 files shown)


src/Misc/layoutroot/safe_sleep.sh
```diff
@@ -1,6 +1,27 @@
...
+    fi
+
+    if command -v perl >/dev/null 2>&1; then
+        perl -e "select(undef, undef, undef, $t);" && return 0
+    fi
+
+    local end
+    end=$(printf "%.0f" "$(echo "$(date +%s.%3N) + $t" | bc)")
+    while (( $(date +%s) < end )); do
+        sleep 0.05
```
frozenspider, line 21: How's that supposed to work if we already established there's no `sleep` command?
Also, there's already a bunch of PRs addressing this, most notably #3143

PR #4147, opened 2025-12-11
Add support for case function

AllanGuigou:


src/Sdk/DTObjectTemplating/ObjectTemplating/Tokens/TemplateToken.cs
```diff
@@ -57,7 +57,7 @@ protected StringToken EvaluateStringToken(
             var originalBytes = context.Memory.CurrentBytes;
             try
             {
-                var tree = new ExpressionParser().CreateTree(expression, null, context.GetExpressionNamedValues(), context.ExpressionFunctions);
+                var tree = new ExpressionParser().CreateTree(expression, null, context.GetExpressionNamedValues(), context.ExpressionFunctions, allowCaseFunction: context.AllowCaseFunction);
```
ericsciple, line 60: is this for the evaluator? Do we hardcode to true for the evaluator?

Basically I just want to make sure we set to true for evaluation. If evaluation works E2E then 👍

AllanGuigou: Yup, this is defaulted to true but will allow the manifest manager code paths to disable it.

Issue #3792, opened 2025-04-07
safe_sleep.sh rarely hangs indefinitely


AlekseiNikiforovIBM:
**Describe the bug**
Very rarely on update of github actions runner `safe_sleep.sh` hangs forever:

```
$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
...
actions+   72298 96.3  0.0   4372  3328 ?        R    Apr02 1988:54 /bin/bash /home/actions-runner/safe_sleep.sh 1
...
```

I suspect it may happen sometimes if machine runs in cloud and is overloaded and/or overcommitted.

**To Reproduce**
1. Download runner of version prior to version, for example, 2.322.
2. Register and run runner.
3. Runner updates itself. It can also take a task to complete in meantime.

**Expected behavior**
Update should not hang infinitely.

## Runner Version and Platform
Runner 2.322
OS: Linux

mlugg:
[script in question](https://github.com/actions/runner/blob/0ebdf9e83d51b0ab8c168a4869954466f3e2e88f/src/Misc/layoutroot/safe_sleep.sh)

The bug in this "safe sleep" script is obvious from looking at it: if the process is not scheduled for the one-second interval in which the loop would return (due to `$SECONDS` having the correct value), then it simply spins forever. That can easily happen on a CI machine under extreme load. When this happens, it's pretty bad: it completely breaks a runner until manual intervention. On Zig's CI runner machines, we observed multiple of these processes which had been running for hundreds of hours, silently taking down two runner services for weeks.

I don't understand how we got here. Even ignoring the pretty clear bug, what makes this Bash script "safer" than calling into the POSIX standard `sleep` utility? It doesn't seem to solve any problem; meanwhile, it's less portable and needlessly eats CPU time by busy-waiting.

---

The sloppy coding which is evident here, as well as the inaction on core Actions bugs (in line with the decay in quality of almost every part of GitHub's product), is forcing the Zig project to strongly consider moving away from GitHub Actions entirely. With this bug, and many others (severe workflow scheduling issues resulting in dozens of timeouts; logs randomly becoming inaccessible; random job cancellations without details; perpetually "pending" jobs), we can no longer trust that Actions can be used to implement reliable CI infrastructure. I personally would seriously encourage other projects, particularly any using self-hosted runners, to look carefully at the stability of Actions and ask themselves whether it is a solution worth sticking with long-term when compared with alternatives.

[...]

PR #4150, opened 2025-12-12
Remove unneeded safe_sleep

vtrenton:
Hey Github Runner team!

I felt the need to open this PR as to at least have a discussion around this. safe_sleep has created serious problems in the past with wasting CPU cycles using #3792 as reference. The codebase keeps getting bigger with hacks like using ping as a timer... Which is great except it relies heavily on the assumption that ping on all systems will never change it send frequency (unlikely - but there is a potential).

There is also an existing concern about the latest fix being vulnerable to injection attacks. Although not likely exploitable - it's new risk introduced into the codebase: https://github.com/actions/runner/issues/3792#issuecomment-3639241280

There is no need for this as almost zero UNIX-Like systems including containers and Android/Embedded systems don't have sleep. Sleep is a standard POSIX utility on all UNIX systems.

To top it off these scripts require BASH which is installed on less systems than sleep. Let alone hardcoding the shebang path to /bin/bash which systems such as Nix and BSD do not use. BASH is much larger compatibility risk than sleep.

if you're making bash (a much less deployed tool) a hard requirement than sleep should be.

All of this is to say - we should remove safe_sleep from the codebase and just use/assume a POSIX sleep implementation is available. This is technical debt and the implementation costs much more than its worth. 

I would like to challenge anyone to give me an example of a UNIX-Like system that uses BASH that doesn't have sleep. Very open to being challenged on this. I seriously for the life of me can think of a single one. Would love to hear any feedback on this!

Thank you for your consideration!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
     # Upload runner package tar.gz/zip as artifact
     - name: Publish Artifact
       if: github.event_name != 'pull_request'
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: runner-package-${{ matrix.runtime }}
         path: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -118,7 +118,7 @@ jobs:
     # Upload runner package tar.gz/zip as artifact.
     - name: Publish Artifact
       if: github.event_name != 'pull_request'
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: runner-packages-${{ matrix.runtime }}
         path: |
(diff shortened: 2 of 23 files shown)


Elyytscha:
At the very least, maybe safe_sleep should prefer the system sleep binary when it exists (I think thats how it was designed initially), and only fall back to this bespoke shell-loop masterpiece when it doesn’t.

Much easier to meter runner minutes when your sleep actually wakes up

vtrenton:
> At the very least, maybe safe_sleep should prefer the system sleep binary when it exists (I think thats how it was designed initially), and only fall back to this bespoke shell-loop masterpiece when it doesn’t.
> 
> Much easier to meter runner minutes when your sleep actually wakes up

PRs have since been merged to reimplement this approach:
https://github.com/actions/runner/blob/c96dcd472907e514274cdb4af281116adf7aad18/src/Misc/layoutroot/safe_sleep.sh#L3

I agree it's much better than just the while loop. But my intention of this PR is to remove safe_sleep altogether. As I feel this is overengineering for a problem that doesn't exist. This design has created much more of a problem than it's worth.

If we want to go down the compatibility rabbit hole - we can. Does bash exist? does command exist? should we write a safe_bash and safe_command? when does it stop.

PR #4220, opened 2026-02-02
Fix parser comparison mismatches

ericsciple:
Fixes telemetry noise from parser comparison mismatches between legacy and new workflow parsers.

## Fixes

- **ActionManifestMismatch: Load** - Fix `ConvertToLegacySteps()` to use explicit property mapping instead of broken JSON round-trip
- **EvaluateJobContainer / EvaluateJobServiceContainers** - Only emit error for empty container image during early validation, not at runtime
- **TemplateEvaluatorMismatch (fromJSON)** - Treat JSON parse errors as matching even when error messages differ
- **Cancellation race conditions** - Skip mismatch recording when cancellation state changes during the evaluation window

### DIFF
--- a/src/Runner.Worker/PipelineTemplateEvaluatorWrapper.cs
+++ b/src/Runner.Worker/PipelineTemplateEvaluatorWrapper.cs
@@ -216,7 +216,7 @@ namespace GitHub.Runner.Worker
             }
         }
 
-        private TLegacy EvaluateAndCompare<TLegacy, TNew>(
+        internal TLegacy EvaluateAndCompare<TLegacy, TNew>(
             string methodName,
             Func<TLegacy> legacyEvaluator,
             Func<TNew> newEvaluator,
@@ -668,18 +668,14 @@ namespace GitHub.Runner.Worker
         /// </summary>
         private bool IsKnownEquivalentErrorPattern(Exception legacyException, Exception newException)
         {
-            // Get all messages in the exception chain
-            var legacyMessages = string.Join(" | ", GetExceptionMessages(legacyException));
-            var newMessages = string.Join(" | ", GetExceptionMessages(newException));
-
             // fromJSON('') - both parsers fail when parsing empty string as JSON
             // The error messages differ but both indicate JSON parsing failure.
             // Legacy throws raw JsonReaderException: "Error reading JToken from JsonReader..."
             // New wraps it: "Error parsing fromJson" with inner JsonReaderException
             // Both may be wrapped in TemplateValidationException: "The template is not valid..."
-            if (IsJsonParseError(legacyMessages) && IsJsonParseError(newMessages))
+            if (HasJsonExceptionType(legacyException) && HasJsonExceptionType(newException))
             {
-                _trace.Info("CompareExceptions - both exceptions are JSON parse errors (semantically equivalent)");
+                _trace.Info("CompareExceptions - both exceptions are JSON parse errors, treating as matched");
                 return true;
             }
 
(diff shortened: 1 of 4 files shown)


src/Runner.Worker/PipelineTemplateEvaluatorWrapper.cs
```diff
@@ -222,6 +222,9 @@ private TLegacy EvaluateAndCompare<TLegacy, TNew>(
             Func<TNew> newEvaluator,
             Func<TLegacy, TNew, bool> resultComparer)
         {
+            // Capture cancellation state before evaluation
```
ericsciple, line 225: This helps eliminate noise due to a cancellation race condition — i.e. cancellation fires after legacy evaluation, before new evaluation

src/Test/L0/Worker/ActionManifestParserComparisonL0.cs
```diff
@@ -0,0 +1,343 @@
...
+                var contextData = new DictionaryContextData();
+                var expressionFunctions = new List<LegacyExpressions.IFunctionInfo>();
+
+                // Evaluation without cancellation - should work normally and not record mismatch for matching results
+                var result = wrapper.EvaluateStepDisplayName(token, contextData, expressionFunctions);
+                Assert.Equal("test-value", result);
+
+                // Since both parsers return the same result, no mismatch should be recorded
+                Assert.False(_ec.Object.Global.HasTemplateEvaluatorMismatch);
+            }
```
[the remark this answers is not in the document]
ericsciple: fixed

src/Runner.Worker/PipelineTemplateEvaluatorWrapper.cs
```diff
@@ -634,6 +662,41 @@ private bool CompareExceptions(Exception legacyException, Exception newException
...
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks if the exception message chain indicates a JSON parsing error.
+        /// </summary>
+        private bool IsJsonParseError(string messages)
```
salmanmkc: are we also able to check the type of the exception?

ericsciple: ~I think at this point we just have the user facing error message~

ericsciple: I can look into that. For comparison purposes error message should be good enough.

src/Test/L0/Worker/ActionManifestParserComparisonL0.cs
```diff
@@ -0,0 +1,343 @@
...
+            finally
+            {
+                Teardown();
+            }
+        }
+
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public void EvaluateAndCompare_SkipsMismatchRecording_WhenCancellationOccursDuringEvaluation()
```
salmanmkc: Could be wrong here, but I'm not sure this exercises the skip path. Cancel() is called before the second evaluation, so wouldn't cancellationRequestedBefore already be true by the time we enter EvaluateAndCompare? That would make !cancellationRequestedBefore && cancellationRequestedAfter false and the skip wouldn't fire. Might just be passing because both parsers agree on "test-value"?

If that's the case, to test the actual race-skip I think cancellation would need to fire between the legacy and new evaluator calls. Wdyt?

src/Runner.Worker/PipelineTemplateEvaluatorWrapper.cs
ericsciple: I pushed a commit which uses ExceptionType instead. That approach should be more robust. Also I added the same check for ActionManifestManagerWrapper (missing before)

src/Test/L0/Worker/ActionManifestParserComparisonL0.cs
ericsciple: 👍 i pushed a commit to fix this

salmanmkc (approved):

PR #4221, opened 2026-02-02
Add telemetry tracking for deprecated set-output and save-state commands

ericsciple:
- Add HasDeprecatedSetOutput and HasDeprecatedSaveState flags to GlobalContext
- Emit JobTelemetry once per job when deprecated commands are used
- Add unit tests verifying once-per-job telemetry behavior


ericsciple:
Tested E2E

PR #4233, opened 2026-02-10
Support return job result as exitcode in hosted runner.

TingluoHuang:
https://github.com/github/hosted-compute/issues/3923


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -24,7 +24,7 @@ namespace GitHub.Runner.Listener
     public interface IJobDispatcher : IRunnerService
     {
         bool Busy { get; }
-        TaskCompletionSource<bool> RunOnceJobCompleted { get; }
```
TingluoHuang, line 27: We didn't use the bool other than set it to `true` before.
So i am expose the job result via it.

Issue #2009, opened 2022-07-20
Composite: Step Outputs are not available when in nested composite actions


antdking:
**Describe the bug**
When a workflow uses a composite action, the post steps of actions within the Composite have their inputs evaluated with the Steps context.

However when the same Composite Action is nested in another Composite Action, the Steps context is lost for the Post steps.

Our usecase for wanting this is to cache dependencies effectively without polluting our Workflows.

**To Reproduce**
workflow uses a composite that uses a composite.

Reproduced here: https://github.com/antdking/composite-post-step-outputs-demo, specifically [this](https://github.com/antdking/composite-post-step-outputs-demo/blob/main/.github/workflows/demo-nested-composites-break.yml) workflow
This run includes debug output: https://github.com/antdking/composite-post-step-outputs-demo/actions/runs/2710027823

**Expected behavior**
Able to use Post steps for Composite Actions that are used by other Composite Actions.

## Runner Version and Platform
Hosted runner, affects all Platforms

## What's not working?
Please include error messages and screenshots.

When executing the Post Run for the wrapping composite:
```
 ##[debug]Evaluating condition for step: 'Post wrapperComposite'
##[debug]Evaluating: always()
##[debug]Evaluating always:
##[debug]=> true
##[debug]Result: true
##[debug]Starting: Post wrapperComposite
##[debug]Loading inputs
##[debug]Loading env
Post job cleanup.
##[debug]Evaluating condition for step: 'run'
##[debug]Evaluating: always()
##[debug]Evaluating always:
##[debug]=> true
##[debug]Result: true
##[debug]Starting: run
##[debug]Loading inputs
##[debug]Evaluating: steps.setter.outputs.some-val
##[debug]Evaluating Index:
##[debug]..Evaluating Index:
##[debug]....Evaluating Index:
##[debug]......Evaluating steps:
##[debug]......=> Object
##[debug]......Evaluating String:
##[debug]......=> 'setter'
##[debug]....=> null
##[debug]..=> null
##[debug]=> null
##[debug]Result: null
##[debug]Loading env
Post job cleanup.
value received: 
Error: value: '' is not 'foo'
##[debug]Node Action run completed with exit code 1
##[debug]Finished: run
##[debug]Finishing: Post wrapperComposite
```

the `setter` Step seems to be missing when evaluating the post steps.

## Runner and Worker's Diagnostic Logs

n/a

Sytten:
I am also experiencing this issue.

[...]

PR #4243, opened 2026-02-12
Composite Action Step Markers

ericsciple:
## Summary

Emit `##[start-action]` / `##[end-action]` markers in the log stream around each nested step inside a composite action. The UI parses these markers to render collapsible regions, giving users visibility into individual steps that were previously hidden in a single opaque log blob.

## Design

The runner writes `##[` markers directly to the log stream via `ExecutionContext.Output()`, bypassing the logging command pipeline. No new `::` logging command is registered. Users cannot emit these markers from scripts.

### Marker format

```
##[start-action display=<step-display-name>;id=<step-id>]

... step output ...

##[end-action id=<step-id>;outcome=<result>;conclusion=<result>;duration_ms=<ms>]
```

- **id** — the step's `ContextName` (from YAML `id:` or auto-generated with `__` prefix)
- **outcome** — raw result before `continue-on-error` is applied
- **conclusion** — final result after `continue-on-error`
- **duration_ms** — wall-clock milliseconds (0 for skipped steps)

Nested composites use dot-separated IDs (e.g. `outer.inner-step`) to keep each step globally unique.

### Feature flag

Gated behind `actions_runner_emit_composite_markers` (job message variable) with `ACTIONS_RUNNER_EMIT_COMPOSITE_MARKERS` env var fallback (for internal testing only).

### Injection prevention

`OutputManager.OnDataReceived` replaces `##[start-action` and `##[end-action` from user process stdout/stderr with `##[\start-action` and `##[\end-action`. This preventing users from injecting fake markers. The runner's own markers bypass `OutputManager` entirely since they're written via `ExecutionContext.Output()` directly.

### DIFF
--- a/src/Runner.Worker/Handlers/CompositeActionHandler.cs
+++ b/src/Runner.Worker/Handlers/CompositeActionHandler.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -226,6 +227,11 @@ namespace GitHub.Runner.Worker.Handlers
         {
             ArgUtil.NotNull(embeddedSteps, nameof(embeddedSteps));
 
+            bool emitCompositeMarkers =
+                (ExecutionContext.Global.Variables.GetBoolean(Constants.Runner.Features.EmitCompositeMarkers) ?? false)
+                || StringUtil.ConvertToBoolean(
+                    System.Environment.GetEnvironmentVariable(Constants.Variables.Agent.EmitCompositeMarkers));
+
             foreach (IStep step in embeddedSteps)
             {
                 Trace.Info($"Processing embedded step: DisplayName='{step.DisplayName}'");
@@ -295,6 +301,20 @@ namespace GitHub.Runner.Worker.Handlers
                     Trace.Info("Caught exception from expression for embedded step.env");
                     step.ExecutionContext.Error(ex);
                     SetStepConclusion(step, TaskResult.Failed);
+                }
+
+                // Marker ID uses the step's fully qualified context name (ScopeName.ContextName),
+                // which encodes the full composite nesting chain at any depth.
+                var markerId = emitCompositeMarkers ? step.ExecutionContext.GetFullyQualifiedContextName() : null;
+                var stepStopwatch = default(Stopwatch);
+                var endMarkerEmitted = false;
+
+                // Emit start marker after full context setup so display name expressions resolve correctly
+                if (emitCompositeMarkers)
+                {
+                    step.TryUpdateDisplayName(out _);
+                    ExecutionContext.Output($"##[start-action display={EscapeProperty(SanitizeDisplayName(step.DisplayName))};id={EscapeProperty(markerId)}]");
+                    stepStopwatch = Stopwatch.StartNew();
                 }
 
                 // Register Callback
@@ -381,6 +401,14 @@ namespace GitHub.Runner.Worker.Handlers
                         // Condition is false
                         Trace.Info("Skipping step due to condition evaluation.");
                         SetStepConclusion(step, TaskResult.Skipped);
+
+                        if (emitCompositeMarkers)
+                        {
+                            stepStopwatch.Stop();
+                            ExecutionContext.Output($"##[end-action id={EscapeProperty(markerId)};outcome=skipped;conclusion=skipped;duration_ms=0]");
+                            endMarkerEmitted = true;
+                        }
+
                         continue;
                     }
                     else if (conditionEvaluateError != null)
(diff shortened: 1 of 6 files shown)


src/Runner.Worker/Handlers/CompositeActionHandler.cs
```diff
@@ -226,10 +226,34 @@ private async Task RunStepsAsync(List<IStep> embeddedSteps, ActionRunStage stage
...
 
+                // Build step ID from ContextName for markers
+                var contextName = (step as IActionRunner)?.Action?.ContextName ?? $"__{embeddedSteps.IndexOf(step)}";
+                var stepId = string.IsNullOrEmpty(idPrefix) ? contextName : $"{idPrefix}.{contextName}";
+                var stepStopwatch = new System.Diagnostics.Stopwatch();
+
+                // Emit start marker before condition evaluation
+                if (emitCompositeMarkers)
+                {
+                    ExecutionContext.Output($"##[start-action display={EscapeProperty(SanitizeDisplayName(step.DisplayName))};id={EscapeProperty(stepId)}]");
```
ChristopherHX: display Name is wrong in composite even if not using expressions, this value is updated later again as this may change if this depends on a changing context variable:

(Logs generated with custom cli tool + env variable)
<Details>
<Summary>Currently looks like in the log the displayname is not the expected one</Summary>

```
[.github/workflows/test.yml / _] Running: Run ./
| Prepare all required actions
| ##[group]Run ./
| ##[endgroup]
| ##[start-action display=run;id=__run]
| ##[group]Run echo "Hello, World"
| echo "Hello, World"
| shell: /bin/bash --noprofile --norc -e -o pipefail {0}
| ##[endgroup]
| Hello, World
| ##[end-action id=__run;outcome=success;conclusion=success;duration_ms=44]
| ##[start-action display=run;id=__run_2]
| ##[group]Run echo "Hello, World"
| echo "Hello, World"
| shell: /bin/bash --noprofile --norc -e -o pipefail {0}
| ##[endgroup]
| Hello, World
| ##[end-action id=__run_2;outcome=success;conclusion=success;duration_ms=37]
[.github/workflows/test.yml / _] Succeeded: Run ./
```

</Details>


<Details><Summary>Action</Summary>

```yaml
runs:
  using: composite
  steps:
  - name: Hello World
    run: echo "Hello, World"
    shell: bash
  - run: echo "Hello, World"
    shell: bash
```

</Details>


Minimally this is required for this to be acceptable for me as an actions user.


```suggestion
                    step.TryUpdateDisplayName(out _);
                    ExecutionContext.Output($"##[start-action display={EscapeProperty(SanitizeDisplayName(step.DisplayName))};id={EscapeProperty(stepId)}]");
```


<Details>
<Summary>After Constant Hello World is the display name or the fallback</Summary>

```
[.github/workflows/test.yml / _] Running: Run ./
| Prepare all required actions
| ##[group]Run ./
| ##[endgroup]
| ##[start-action display=Hello World;id=__run]
| ##[group]Run echo "Hello, World"
| echo "Hello, World"
| shell: /bin/bash --noprofile --norc -e -o pipefail {0}
| ##[endgroup]
| Hello, World
| ##[end-action id=__run;outcome=success;conclusion=success;duration_ms=35]
| ##[start-action display=Run echo "Hello, World";id=__run_2]
| ##[group]Run echo "Hello, World"
| echo "Hello, World"
| shell: /bin/bash --noprofile --norc -e -o pipefail {0}
| ##[endgroup]
| Hello, World
| ##[end-action id=__run_2;outcome=success;conclusion=success;duration_ms=27]
[.github/workflows/test.yml / _] Succeeded: Run ./
```

</Details>

---

Don't you need to expand the proposal with an update displayname command additionally to start-action or postpone issueing this log info.
https://github.com/actions/runner/blob/66800900843747f37591b077091dd2c8cf2c1796/src/Runner.Worker/StepsRunner.cs#L195-L199

ericsciple: I haven't tested yet 😄 

Will look into this during E2E testing (needs server changes to render)

ericsciple: addressed with https://github.com/actions/runner/pull/4243/commits/2a1588d3f8b7bc7085220050f3f3b65d8bcf9337

ericsciple: about to push one more commit to make it more correct...

ericsciple: btw thanks @ChristopherHX good callout and your comment helped focus on this aspect to make sure the details are right.

ChristopherHX:
You need to take a look at actions with post steps within composite actions, like actions/cache@v4.

- The step id of root level becomes a uuid
  - probably irrelevant, since the end marker use the same id
- the display name is the initially invalid one again namely "run" instead of `Run actions/cache@v4`
- ~~tested both local and remote composite actions (look equally broken)~~, **EDIT** [^1]
- tested both skipped post step and executed post step (no difference in display name)

```yaml
| ##[start-action display=Run actions/cache@v4;id=__self.__actions_cache]
| ##[group]Run actions/cache@v4
| with:
|   enableCrossOsArchive: false
|   fail-on-cache-miss: false
|   lookup-only: false
|   save-always: false
| ##[endgroup]
| ##[error]Input required and not supplied: key
| ##[end-action id=__self.__actions_cache;outcome=failure;conclusion=failure;duration_ms=148]
[.github/workflows/test.yml / _] Failed: Run ./
[.github/workflows/test.yml / _] Running: Post Run ./
| Post job cleanup.
| ##[start-action display=run;id=__ca4435ec-3068-4fdf-9051-7de2dd4c7686.__actions_cache]
| ##[end-action id=__ca4435ec-3068-4fdf-9051-7de2dd4c7686.__actions_cache;outcome=skipped;conclusion=skipped;duration_ms=0]
```

---

Pre and Main step markers look good now.

[^1]: local composite action show issue #2009 in id field (depending on the usage it might not show an issue) test case [composite-step-debug.zip](https://github.com/user-attachments/files/25459188/composite-step-debug.zip), `##[start-action display=run;id=__39393486-87cd-4be4-8557-065c226e76c1.level2_2]` id has a missing id component level1_2 for composite actions `./`. Had my unmerged fix for 2009 applied on my version.

ericsciple:
Thanks @ChristopherHX, good catch!

> The step id of root level becomes a uuid
probably irrelevant, since the end marker use the same id

I just tested E2E with the UI changes and as you said, the marker IDs match so rendering is fine.

> the display name is the initially invalid one again namely "run" instead of Run actions/cache@v4

Since it's scoped to post-steps only, I will likely address in a follow-up PR. That will keep this diff easier for reviewers. I'm not sure yet how messy the fix will be.

PR #4260, opened 2026-02-23
Symlink actions cache

paveliak:
Presently we do bunch of IOPS to install 1st-party actions from archive that could be expensive for small VM SKU runners:

- Tarball is copied into the staging directory
- Tarball is unpacked there
- Staging directory is copied into the workspace

With this change unpacked tarballs (if available) are symlinked into the workspace upon starting up the job.

NOTE: this change requires reactive work in the image

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -830,11 +830,7 @@ namespace GitHub.Runner.Worker
                                     executionContext.Debug($"Symlink '{nestedDirectories[0].Name}' to '{destDirectory}'");
                                     // make sure we get a clean folder ready to use.
                                     IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
-                                    // create directory chain
-                                    Directory.CreateDirectory(destDirectory);
-                                    // delete leaf directory
-                                    Directory.Delete(destDirectory);
-                                    Directory.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
+                                    IOUtil.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
                                 }
                                                         
                                 executionContext.Debug($"Created symlink from cached directory '{cacheDirectory}' to '{destDirectory}'");
--- a/src/Runner.Sdk/Util/IOUtil.cs
+++ b/src/Runner.Sdk/Util/IOUtil.cs
@@ -93,6 +93,16 @@ namespace GitHub.Runner.Sdk
             }
         }
 
+        public static FileSystemInfo CreateSymbolicLink(string destDirectory, string srcDirectory)
+        {
+            // ensure directory chain exists
+            Directory.CreateDirectory(destDirectory);
+            // delete leaf directory
+            Directory.Delete(destDirectory);
+            // create symlink for the leaf directory
+            return Directory.CreateSymbolicLink(destDirectory, srcDirectory);
+        }
+
         public static void Delete(string path, CancellationToken cancellationToken)
         {
             DeleteDirectory(path, cancellationToken);


src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                                        
+                                // repository archive from github always contains a nested folder
+                                var nestedDirectories = new DirectoryInfo(cacheDirectory).GetDirectories();
+                                if (nestedDirectories.Length != 1)
+                                {
+                                    throw new InvalidOperationException($"'{cacheDirectory}' contains '{nestedDirectories.Length}' directories");
+                                }
+                                else
+                                {
+                                    executionContext.Debug($"Symlink '{nestedDirectories[0].Name}' to '{destDirectory}'");
```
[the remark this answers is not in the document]
TingluoHuang: @paveliak is this true or copilot is lying?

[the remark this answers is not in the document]
paveliak: Good catch

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                    Directory.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
```
TingluoHuang: Will this fail if i use the same actions twice in a job?

```
- uses: actions/checkout@v5
- uses: actions/checkout@v5
```

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                }
+                                                        
+                                executionContext.Debug($"Created symlink from cached directory '{cacheDirectory}' to '{destDirectory}'");
+                                executionContext.Global.JobTelemetry.Add(new JobTelemetry()
+                                {
+                                    Type = JobTelemetryType.General,
+                                    Message = $"Action directory cache usage: {downloadInfo.ResolvedNameWithOwner}@{downloadInfo.ResolvedSha} use symlink from cache"
```
TingluoHuang: i think we want to keep the message format so we can render chart to telemetry easier.

                    Message = $"Action archive cache usage: {downloadInfo.ResolvedNameWithOwner}@{downloadInfo.ResolvedSha} use cache {useActionArchiveCache} has cache {hasActionArchiveCache} via symlink {true/false}"

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                });
+
```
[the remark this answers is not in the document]
paveliak: Cached directory could be located on a read-only file system. There is no guarantee that it would be writable.

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
                 if (!string.IsNullOrEmpty(actionArchiveCacheDir) &&
                     Directory.Exists(actionArchiveCacheDir))
                 {
+                    var symlinkCachedActions = executionContext.Global.Variables.GetBoolean(Constants.Runner.Features.SymlinkCachedActions) ?? false;
```
TingluoHuang: do we want to also check an ENV that only set by hosted runner? I am to make sure the new code path won't get used by self-hosted runner.

src/Runner.Worker/ActionManager.cs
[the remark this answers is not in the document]
paveliak: @TingluoHuang Copilot never lies. Sometimes it just gets carried away 😁

src/Runner.Worker/ActionManager.cs
paveliak: Not sure if it worked before but it works now. I added L0 test to check for that

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                var nestedDirectories = new DirectoryInfo(cacheDirectory).GetDirectories();
+                                if (nestedDirectories.Length != 1)
+                                {
+                                    throw new InvalidOperationException($"'{cacheDirectory}' contains '{nestedDirectories.Length}' directories");
+                                }
+                                else
+                                {
+                                    executionContext.Debug($"Symlink '{nestedDirectories[0].Name}' to '{destDirectory}'");
+                                    Directory.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
+                                }
```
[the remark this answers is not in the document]
paveliak: This will be taken care of in the image by making sure that cached directory is read-only.

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +811,48 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
                 if (!string.IsNullOrEmpty(actionArchiveCacheDir) &&
                     Directory.Exists(actionArchiveCacheDir))
                 {
+                    var symlinkCachedActions = executionContext.Global.Variables.GetBoolean(Constants.Runner.Features.SymlinkCachedActions) ?? false;
+                    if (symlinkCachedActions)
+                    {
```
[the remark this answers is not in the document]
paveliak: Done

src/Runner.Worker/ActionManager.cs
paveliak: @TingluoHuang What are your concerns here? That something gets accidentally broken or it is security related (cache poisoning, etc)? Normally self-hosted runner would not see those directories and will execute the original code path

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +807,54 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                else
+                                {
+                                    executionContext.Debug($"Symlink '{nestedDirectories[0].Name}' to '{destDirectory}'");
+                                    // make sure we get a clean folder ready to use.
+                                    IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken);
+                                    // create directory chain
+                                    Directory.CreateDirectory(destDirectory);
+                                    // delete leaf directory
+                                    Directory.Delete(destDirectory);
+                                    Directory.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
```
TingluoHuang: it feels odd that we have to:
- delete-dir
- create-dir
- delete-dir
- create-symlink

is it possible to only have the 
- delete-dir
- create-symlink

src/Runner.Worker/ActionManager.cs
```diff
@@ -811,6 +807,54 @@ private async Task DownloadRepositoryActionAsync(IExecutionContext executionCont
...
+                                }
+                                                        
+                                executionContext.Debug($"Created symlink from cached directory '{cacheDirectory}' to '{destDirectory}'");
+                                executionContext.Global.JobTelemetry.Add(new JobTelemetry()
+                                {
+                                    Type = JobTelemetryType.General,
+                                    Message = $"Action archive cache usage: {downloadInfo.ResolvedNameWithOwner}@{downloadInfo.ResolvedSha} use cache {useActionArchiveCache} has cache {hasActionArchiveCache} via symlink"
+                                });
+                                
+                                Trace.Info("Finished getting action repository.");
```
TingluoHuang, line 843: do we want to write watermark file before returning?

```C#
Trace.Verbose("Create watermark file indicate action cache symlink succeed.");
File.WriteAllText(watermarkFile, DateTime.UtcNow.ToString());
```

src/Runner.Worker/ActionManager.cs
TingluoHuang: if you do the FF via a ENV lookup, you can control the rollout via image or provisioner rollout.

src/Runner.Worker/ActionManager.cs
paveliak: Unfortunately not. Creating symlink requires all the chain of directories be pre-created. So ideally what we want:
- Ensure chain of directories exist
- Delete top level directory (if any)
- Create a symlink

I could rewrite it as this. Would it make it better? 
```
Directory.CreateDirectory(destDirectory); // Ensure directory chain. Noop, if directory exists
IOUtil.DeleteDirectory(destDirectory, executionContext.CancellationToken); // Delete all the pre-existing stuff
Directory.CreateSymbolicLink(destDirectory, nestedDirectories[0].FullName);
```

paveliak: I am bit concerned though that `Directory.CreateDirectory` might fail in some edge cases like orphaned symbolic links. So I would rather keep all these lines, juts in case

src/Runner.Worker/ActionManager.cs
paveliak: Where would I put it? `destDirectory` is a symlink to the source directory in the cache which is on the read-only file system. Attempt to write watermark file would fail. I do not think rerunning this code in case an actions is used multiple times would hurt performance because we pretty much will delete/recreate symlink

src/Runner.Worker/ActionManager.cs
paveliak: @TingluoHuang Added method into `IOUtil` class to encapsulate this behavior

src/Runner.Worker/ActionManager.cs
paveliak: Changed to use Env variable

src/Runner.Worker/ActionManager.cs
TingluoHuang: 🤔 we will print out multiple lines for:
`executionContext.Output($"Download action repository '{downloadInfo.NameWithOwner}@{downloadInfo.Ref}' (SHA:{downloadInfo.ResolvedSha})");`

maybe that's ok when we are in this mode.

TingluoHuang (approved):

FixBo:
Should not watermark file also created in action directory? There are duplicate "Download action repository" in log.

PR #4273, opened 2026-03-02
Fix parser comparison mismatches

ericsciple:
Fixes telemetry noise from parser comparison mismatches between legacy and new workflow parsers.

Fix: Port schema and converter changes from actions-workflow-parser PR 577. Relaxes container-mapping.image and services-container from non-empty-string to string so expressions that evaluate to empty at runtime return null (no container) instead of a validation error.

### DIFF
--- a/src/Sdk/WorkflowParser/Conversion/WorkflowTemplateConverter.cs
+++ b/src/Sdk/WorkflowParser/Conversion/WorkflowTemplateConverter.cs
@@ -1079,7 +1079,8 @@ namespace GitHub.Actions.WorkflowParser.Conversion
         internal static JobContainer ConvertToJobContainer(
             TemplateContext context,
             TemplateToken value,
-            bool isEarlyValidation = false)
+            bool isEarlyValidation = false,
+            bool isServiceContainer = false)
         {
             var result = new JobContainer();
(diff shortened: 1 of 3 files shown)


src/Sdk/WorkflowParser/Conversion/WorkflowTemplateConverter.cs
```diff
@@ -1089,11 +1090,34 @@ internal static JobContainer ConvertToJobContainer(
 
             if (value is StringToken containerLiteral)
             {
-                if (String.IsNullOrEmpty(containerLiteral.Value))
+                // Trim "docker://"
+                var trimmedImage = containerLiteral.Value;
+                var hasDockerPrefix = containerLiteral.Value.StartsWith(WorkflowTemplateConstants.DockerUriPrefix, StringComparison.Ordinal);
```
TingluoHuang, line 1095: do you want ignorecase?

TingluoHuang: the old code doesn't have it, so i guess it's ok.

ericsciple: i'm not opposed to ignore case, but better to wait until post-cutover. The goal with this PR is parity (working through compare diffs)

PR #4285, opened 2026-03-06
Exit with specified exit code when runner is outdated

nikola-jokic:
Exit with exit code 7 when `ACTIONS_RUNNER_RETURN_VERSION_DEPRECATED_EXIT_CODE=1`

### DIFF
--- a/src/Runner.Listener/Program.cs
+++ b/src/Runner.Listener/Program.cs
@@ -161,8 +161,7 @@ namespace GitHub.Runner.Listener
 
         private static int GetRunnerVersionDeprecatedExitCode()
         {
-            var envValue = Environment.GetEnvironmentVariable(Constants.Variables.Actions.ReturnVersionDeprecatedExitCode);
-            if (envValue == "1")
+            if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.ReturnVersionDeprecatedExitCode)))
             {
                 return Constants.Runner.ReturnCode.RunnerVersionDeprecated;
             }


src/Runner.Common/Constants.cs
```diff
@@ -278,6 +278,7 @@ public static class Actions
                 public static readonly string AllowUnsupportedStopCommandTokens = "ACTIONS_ALLOW_UNSECURE_STOPCOMMAND_TOKENS";
                 public static readonly string RequireJobContainer = "ACTIONS_RUNNER_REQUIRE_JOB_CONTAINER";
                 public static readonly string RunnerDebug = "ACTIONS_RUNNER_DEBUG";
+                public static readonly string RunnerOutdatedExitCode = "ACTIONS_RUNNER_OUTDATED_EXIT_CODE";
```
TingluoHuang: should we just define an exitcode ourselves and allow customers to get it when they set some enablement ENV?

ex:
```
export ACTIONS_RUNNER_RETURN_VERSION_DEPRECATED_EXIT_CODE=1
run.sh will return 6 (or whatever code we are free to use next)
export ACTIONS_RUNNER_RETURN_VERSION_DEPRECATED_EXIT_CODE=0
run.sh will return 1 (or whatever code we return today)
```

nikola-jokic: That sounds good to me!

src/Runner.Listener/Program.cs
```diff
@@ -159,6 +159,17 @@ private async static Task<int> MainAsync(IHostContext context, string[] args)
             }
         }
 
+        private static int GetRunnerVersionDeprecatedExitCode()
+        {
+            var envValue = Environment.GetEnvironmentVariable(Constants.Variables.Actions.ReturnVersionDeprecatedExitCode);
```
TingluoHuang: there is a stringutil you can use to get a bool.
```
if (StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Actions.ReturnVersionDeprecatedExitCode)))
{
}
```

Issue #4197, opened 2026-01-14
Segmentation fault in actions-runner 2.331.0 on Rocky Linux 9.7


daniel-widrick:
**Describe the bug**
When attempting to run ephemeral GitHub Actions runners using the ghcr.io/actions/actions-runner:2.331.0 image on Rocky Linux 9.7 hosts, the runner immediately crashes with a segmentation fault (core dumped). Reverting to 2.330.0 resolves the issue.

**To Reproduce**
1) Deploy gha-runner-scale-set-controller:0.13.1 in a Kubernetes cluster.
2) Launch ephemeral runner pods using ghcr.io/actions/actions-runner:2.331.0.
3) Observe runner pod logs:

```
[RUNNER 2026-01-14 16:09:02Z INFO RunnerServer] Establish connection with 100 seconds timeout.
[RUNNER 2026-01-14 16:09:02Z INFO GitHubActionsService] Starting operation Location.GetConnectionData
/home/runner/run-helper.sh: line 36:    14 Segmentation fault      (core dumped) "$DIR"/bin/Runner.Listener run $*
Exiting with unknown error code: 139
Exiting runner...
```

**Expected behavior**
Runner should start and connect to GitHub without crashing

## Runner Version and Platform
2.331.0

Host OS: Rocky Linux 9.7

Kernel flags: fpu de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 syscall nx lm constant_tsc nopl xtopology cpuid tsc_known_freq pni ssse3 cx16 sse4_1 sse4_2 x2apic popcnt aes hypervisor lahf_lm cpuid_fault pti

## What's not working?

dmesg from host:
```
[  261.559979] .NET TP Worker[46774]: segfault at 20 ip 00007f97827e5b5c sp 00007f9709294110 error 4 in libssl.so.3[7f97827cc000+62000] likely on CPU 2
[  261.585173] .NET TP Worker[46792]: segfault at 20 ip 00007f5cd640fb5c sp 00007f5c56ff9110 error 4 in libssl.so.3[7f5cd63f6000+62000] likely on CPU 0
```

TingluoHuang:
We have to upgrade to ubuntu:24.04 as our base image, my guess is that's not compat with your Kubernetes
https://github.com/actions/runner/pull/4018

atmcarmo:
Can confirm that I'm having the same issue with a host based on Amazon Linux 2023.

[...]

PR #4288, opened 2026-03-09
Net10 upgrade

paveliak:
This PR upgrades runner to using .Net 10 (specifically `10.0.103`):
- Source generator is added for the `BuildContants.cs` file
- `RunnerService` project is migrated to SDK style to avoid C# Dev Kit warnings
- C# DevKit extension is added to the Codespace
- Code and tests is refactored to allow developing/testing in VStudio/VSCode without  prior running of `dev.sh build/layout`
- On Windows VStudio 2026  is needed to develop for `10.0.103`

### DIFF
--- a/src/Test/L0/ProcessExtensionL0.cs
+++ b/src/Test/L0/ProcessExtensionL0.cs
@@ -24,7 +24,7 @@ namespace GitHub.Runner.Common.Tests
                 try
                 {
 #if OS_WINDOWS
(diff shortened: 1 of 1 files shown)


LockedThread:
FYI it seems like .NET 11 will be GA in September 2026. Upgrading to 11 would close https://github.com/actions/runner/issues/4197
https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-11/overview

Issue #4232, opened 2026-02-09
Warning: Failed to download action ... Error: Response status code does not indicate success: 429 (Too Many Requests).


jsoref:
https://github.com/actions/runner/blob/1b5486aa8f65a03b733617cfb4eb5832c0b23766/src/Runner.Worker/ActionManager.cs#L1191

I'm not sure where the other half comes from, my guess is:
https://github.com/dotnet/runtime/blob/4a0b333e80a29ea29a8256285f8a01a54283b5ef/src/libraries/System.Net.Http/src/Resources/Strings.resx#L211

This error happens periodically (there are tickets here that reference it; we tripped on it during one of the many actions outages today).

The message isn't great. The response is to try again later, which isn't entirely wrong, but the message could use some love on all sides.

PR #4296, opened 2026-03-12
Batch and deduplicate action resolution across composite depths

### LINKED ISSUE #3731: `Download action repository` called repeatedly for actions that have to be the same
**Describe the bug**
If an action is used multiple times and the owner or repository name varies in case, then the action will be downloaded multiple times.

As GitHub asserts that owners and repository names are case-insensitive, this is wasteful.

(Note: paths within repositories may be case-sensitive, that's outside the scope of this report.)

**To Reproduce**
1. Create a workflow that uses an action multiple times where the owner/repository name differ only by case -- https://github.com/check-spelling-sandbox/animated-umbrella/actions/runs/13579366005/workflow
2. Trigger the workflow
3. Look at the `Set up job` section of the workflow run for `Getting action download info` and `Download action repository`...

**Expected behavior**
* Possibly a warning to the job log (`::notice ::action requested by multiple cases <...>`)
* The action should only be downloaded once

Version of your runner? 2.322.0

OS of the machine running the runner? Linux

## What's not working?

The runner is needlessly downloading multiple copies of an action that it can know to be identical.

## Job Log Output

```
Getting action download info
Download action repository 'actions/checkout@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Download action repository 'actIONS/checkout@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Download action repository 'actions/CHECKout@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Download action repository 'aCTions/checkout@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Download action repository 'actions/checKOut@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Download action repository 'actIONS/cHECkout@v4' (SHA:11bd71901bbe5b1630ceea73d27597364c9af683)
Complete job name: checkout
```

https://github.com/check-spelling-sandbox/animated-umbrella/actions/runs/13579439266/job/37962625594#step:1:25

---

For comparison, here's a workflow run where the same action is used multiple times, each time w/ the same case. Note that it only downloads the action repository once:
https://github.com/check-spelling-sandbox/animated-umbrella/actions/runs/13579366005/job/37962422085

stefanpenner:
## Summary

- Cache resolved actions across recursion depths — same action resolved once
- Batch-resolve sibling composites' sub-actions in one API call instead of N
- Defer pre/post step registration until after recursion (fixes HasPre/HasPost correctness)

Internal workflow with ~30 composites: ~20 resolve API calls → 3-4. Also reduces 429s (#4232).

Fixes #3731

### Other possibilities

* https://github.com/actions/runner/pull/4297 performs even better (~2x faster in my tests) when there are more nested downloads, but it's a larger refactor. Which may or may not be appealing to the maintainers.

## Smoke test results

Tested with a self-hosted runner built from this branch vs stock (`main`), using a [50-action composite tree across 6 repos with 5 depth levels](https://github.com/stefanpenner/resolution-test).

**Action graph:**
```
workflow → 10× L1 composites + 5 leaves (15 top-level steps)
  L1 (×10) → 1 leaf + 3 L2 composites (heavy cross-referencing)
  L2 (×10) → 1 leaf + 3 L3 composites
  L3 (×10) → 2 leaves + 2 L4 composites (cross-repo refs to 5 shard repos)
  L4 (×10) → 3 leaves (cross-repo refs)
```

### Results ([run](https://github.com/stefanpenner/resolution-test/actions/runs/23030453177))

| | Stock runner (`main`) | This PR |
|---|---|---|
| **Resolve API calls** | **311** | **4** |
| **Resolution time** | **70.6s** | **8.2s** |

- **78× fewer API calls**, **8.6× faster** action resolution
- Output parity confirmed (3845 leaf-action log lines in both runners)
- All composites resolved and executed at all 5 depth levels
- Cross-repo references resolved correctly
- Dedup verified: same `owner/repo@ref` never re-resolved across depths
- Batching verified: distinct repos at the same depth batched into 1 call

### Earlier run (single-repo, same-ref tree)

[Run](https://github.com/stefanpenner/resolution-test/actions/runs/23030300306): 311 → 1 resolve calls (all 50 actions shared the same `owner/repo@ref` lookup key, so the cache eliminated every call after the first).

## Test plan

- [x] 4 new L0 tests covering batching, cross-depth dedup, multi-top-level, and nested containers
- [x] Smoke test with self-hosted runners: 50-action/5-depth/6-repo composite tree (see above)


src/Runner.Worker/ActionManager.cs
```diff
@@ -79,6 +80,9 @@ public sealed class ActionManager : RunnerService, IActionManager
                 PreStepTracker = new Dictionary<Guid, IActionRunner>()
             };
             var containerSetupSteps = new List<JobExtensionRunner>();
+            // Stack-local cache: same action (owner/repo@ref) is resolved only once,
+            // even if it appears at multiple depths in a composite tree.
+            var resolvedDownloadInfos = new Dictionary<string, WebApi.ActionDownloadInfo>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
stefanpenner: I don't believe there should be any relevant observable difference, other then performance

src/Runner.Worker/ActionManager.cs
```diff
@@ -105,7 +108,7 @@ public sealed class ActionManager : RunnerService, IActionManager
             PrepareActionsState result = new PrepareActionsState();
             try
             {
-                result = await PrepareActionsRecursiveAsync(executionContext, state, actions, depth, rootStepId);
+                result = await PrepareActionsRecursiveAsync(executionContext, state, actions, resolvedDownloadInfos, depth, rootStepId);
```
ericsciple: Can you add a feature flag? If messy, prefer duplicating the methods with suffix _Legacy vs _New

Runner feature flags are sent from the server in the job message as variables. In the runner code, you read them via `context.Global.Variables.GetBoolean(...)`.

**Step 1: Define the flag constant** in [`Constants.Runner.Features`](https://github.com/actions/runner/blob/9728019b24400dd2d99b1ad5e5622a218d588360/src/Runner.Common/Constants.cs#L165-L180):

```csharp
public static readonly string BatchActionResolution = "actions_batch_action_resolution";
```

**Step 2: Check the flag** at the top of your new codepath, falling back to the old behavior when it's off:

```csharp
var batchActionResolution = executionContext.Global.Variables.GetBoolean(Constants.Runner.Features.BatchActionResolution) ?? false
    || StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("ACTIONS_BATCH_ACTION_RESOLUTION"));

if (batchActionResolution)
{
    // new batched/deduped resolution
}
else
{
    // original resolution logic
}
```

**Step 3: Env var fallback for local testing.** The `|| Environment.GetEnvironmentVariable(...)` pattern lets you test locally with a self-hosted runner by setting an env var, without needing the server to send the flag. This is an existing pattern — see [`ExecutionContext.cs` L1422-1426](https://github.com/actions/runner/blob/9728019b24400dd2d99b1ad5e5622a218d588360/src/Runner.Worker/ExecutionContext.cs#L1422-L1426) for an example:

```csharp
var allowServiceContainerCommand = (context.Global.Variables.GetBoolean(Constants.Runner.Features.ServiceContainerCommand) ?? false)
    || StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("ACTIONS_SERVICE_CONTAINER_COMMAND"));

if ((context.Global.Variables.GetBoolean(Constants.Runner.Features.CompareWorkflowParser) ?? false)
    || StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable("ACTIONS_RUNNER_COMPARE_WORKFLOW_PARSER")))
```

stefanpenner:
@ericsciple any interest in https://github.com/actions/runner/pull/4297 or is that too invasive?

src/Runner.Worker/ActionManager.cs
stefanpenner: onit.

stefanpenner: updated, LMK

ericsciple: Thanks! I will finish reviewing today or tomorrow (extremely busy week)

ericsciple (approved):

ericsciple:
@stefanpenner I am comfortable with the changes. I am checking with @TingluoHuang if he has any concerns before merging.

stefanpenner:
@ericsciple keep keep me posted, happy to jump in and make changes as needed. 

Any interest in https://github.com/actions/runner/pull/4297 or is that (as I assumed) too invasive?

ericsciple:
> Any interest in #4297 or is that (as I assumed) too invasive?

Regarding #4297,  unbounded `Task.WhenAll` on tarball downloads could slam our infrastructure if a workflow has a large number of unique actions at one depth. I'd be more comfortable with a conservative concurrency limit. Maybe `SemaphoreSlim(2)`?

@stefanpenner let's through this PR first, then revisit #4297.

TingluoHuang (approved):

stefanpenner:
We finally enabled this at work, it's functioning as expected (at scale, not just ad-hoc testing)...

image

@ericsciple what are the next steps to graduate this from feature-flag to on-by-default?

stefanpenner:
> Regarding https://github.com/actions/runner/pull/4297, unbounded Task.WhenAll on tarball downloads could slam our infrastructure if a workflow has a large number of unique actions at one depth. I'd be more comfortable with a conservative concurrency limit. Maybe SemaphoreSlim(2)?

@ericsciple sounds good, if I make that change, is there appetite? The existing improvement is already quite impressive, but I do have a nagging feeling of leaving this all buttoned up. But will defer to your expertise, as the remaining last few improvements do add disproportionally more complexity.

PR #4298, opened 2026-03-13
Add DAP server

rentziass:
This adds a DAP server to the runner to build debugging functionalities. The whole DAP integration is gated by the new `EnableDebugger` flag on the job message (feature flagged at the API level).

When a job starts, after the job setup we will start the DAP server and allow users to step over to every step in the job, and:

- inspect the scope of the runner
- test evaluating expressions
- run shell commands as if they were `run` steps in their jobs (full job context, supporting expression expansion, etc.)

Here's an example of what this looks like connecting to the runner from nvim-dap:

CleanShot 2026-03-13 at 15 43 05@2x


src/Runner.Worker/Dap/DapMessages.cs
```diff
@@ -0,0 +1,1231 @@
+﻿using System.Collections.Generic;
```
rentziass, line 1: This file contains types from the [protocol](https://microsoft.github.io/debug-adapter-protocol//).

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -253,6 +253,13 @@ public String BillingOwnerId
             set;
         }
 
+        [DataMember(EmitDefaultValue = false)]
```
rentziass, line 256: Start here: this is a new field on the job message, and if true we're going to start the new DAP server.

src/Runner.Worker/JobRunner.cs
```diff
@@ -121,6 +125,36 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 jobContext.Start();
                 jobContext.Debug($"Starting: {message.JobDisplayName}");
 
+                if (jobContext.Global.EnableDebugger)
```
rentziass: If `EnableDebugger` is true we start the DAP server (default on `:4711`). These changes and those in `StepsRunner` are the most important ones in the PR as they plug DAP into existing flow, all the rest is feature flagged/new code.

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
+﻿using System;
```
rentziass, line 1: Similar to `DapMessages` this file is the handler for all DAP protocol requests, more worth looking into than types themselves.

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+                        "next" => HandleNext(request),
+                        "setBreakpoints" => HandleSetBreakpoints(request),
+                        "setExceptionBreakpoints" => HandleSetExceptionBreakpoints(request),
+                        "completions" => HandleCompletions(request),
+                        "stepIn" => CreateResponse(request, false, "Step In is not supported. Actions jobs debug at the step level — use 'next' to advance to the next step.", body: null),
+                        "stepOut" => CreateResponse(request, false, "Step Out is not supported. Actions jobs debug at the step level — use 'continue' to resume.", body: null),
+                        "stepBack" => CreateResponse(request, false, "Step Back is not yet supported.", body: null),
+                        "reverseContinue" => CreateResponse(request, false, "Reverse Continue is not yet supported.", body: null),
+                        "pause" => CreateResponse(request, false, "Pause is not supported. The debugger pauses automatically at step boundaries.", body: null),
+                        _ => CreateResponse(request, false, $"Unsupported command: {request.Command}", body: null)
```
rentziass, line 328: Think of this as our DAP "routes".

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+                        Id = JobThreadId,
+                        Name = threadName
+                    }
+                }
+            };
+
+            return CreateResponse(request, true, body: body);
+        }
+
+        private Response HandleStackTrace(Request request)
```
rentziass, line 858: We're currently using the stack trace to show where we are in the job:

Image

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+            var body = new StackTraceResponseBody
+            {
+                StackFrames = frames,
+                TotalFrames = frames.Count
+            };
+
+            return CreateResponse(request, true, body: body);
+        }
+
+        private Response HandleScopes(Request request)
```
rentziass, line 924: These next two handlers power scope inspection, e.g.

Image

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+            }
+
+            var variables = _variableProvider.GetVariables(context, variablesRef);
+            return CreateResponse(request, true, body: new VariablesResponseBody
+            {
+                Variables = variables
+            });
+        }
+
+        private async Task<Response> HandleEvaluateAsync(Request request, CancellationToken cancellationToken)
```
rentziass, line 966: `evaluate` requests can both be requests to evaluate expressions: 

Image

or REPL commands (coming from a console):

Image

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+                });
+            }
+
+            return CreateResponse(request, true, body: new CompletionsResponseBody
+            {
+                Targets = items
+            });
+        }
+
+        private Response HandleContinue(Request request)
```
rentziass, line 1102: `continue` == execute till the end (until we add breakpoint support)

src/Runner.Worker/Dap/DapReplExecutor.cs
```diff
@@ -0,0 +1,377 @@
...
+    /// Executes <see cref="RunCommand"/> objects in the job's runtime context.
+    ///
+    /// Mirrors the behavior of a normal workflow <c>run:</c> step as closely
+    /// as possible by reusing the runner's existing shell-resolution logic,
+    /// script fixup helpers, and process execution infrastructure.
+    ///
+    /// Output is streamed to the debugger via DAP <c>output</c> events with
+    /// secrets masked before emission.
+    /// </summary>
+    internal sealed class DapReplExecutor
```
rentziass, line 26: Enables `run` like commands through REPL:

Image

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Maps runner execution context data to DAP scopes and variables.
+    ///
+    /// This is the single point where runner context values are materialized
+    /// for the debugger. All string values pass through the runner's existing
+    /// <see cref="GitHub.DistributedTask.Logging.ISecretMasker"/> so the DAP
```
rentziass, line 15: Re-reading this I'm pretty sure we need to mask non string values too 😄

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+            _state = DapSessionState.Initializing;
+
+            // Build capabilities — MVP only supports configurationDone
+            var capabilities = new Capabilities
+            {
+                SupportsConfigurationDoneRequest = true,
+                // All other capabilities are false for MVP
+                SupportsFunctionBreakpoints = false,
+                SupportsConditionalBreakpoints = false,
+                SupportsEvaluateForHovers = true,
```
rentziass: Not ALL others are false 😅

src/Runner.Worker/Dap/DapDebugSession.cs
```diff
@@ -0,0 +1,906 @@
...
+        }
+
+        private string MaskUserVisibleText(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value ?? string.Empty;
+            }
+
+            return _variableProvider?.MaskSecrets(value)
```
rentziass: Do we need both here since variable provider uses the HostContext.SecretMasker?

src/Runner.Worker/JobRunner.cs
TingluoHuang: should we start the DAP server right before `// Get the job extension.`?

src/Runner.Worker/JobRunner.cs
```diff
@@ -121,6 +125,36 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                {
+                    Trace.Info("Debugger enabled for this job run");
+
+                    try
+                    {
+                        var port = 4711;
+                        var portEnv = Environment.GetEnvironmentVariable("ACTIONS_DAP_PORT");
```
TingluoHuang: `ACTIONS_RUNNER_DAP_PORT` maybe?

src/Runner.Worker/JobRunner.cs
```diff
@@ -121,6 +125,36 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        if (!string.IsNullOrEmpty(portEnv) && int.TryParse(portEnv, out var customPort))
```
TingluoHuang:
```suggestion
                        if (!string.IsNullOrEmpty(portEnv) && int.TryParse(portEnv, out var customPort) && customPort > 1024)
```

do we have any limitation on which port you can use?

src/Runner.Worker/JobRunner.cs
```diff
@@ -121,6 +125,36 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        {
+                            port = customPort;
+                        }
+
+                        dapServer = HostContext.GetService<IDapServer>();
+                        debugSession = HostContext.GetService<IDapDebugSession>();
+
+                        dapServer.SetSession(debugSession);
+                        debugSession.SetDapServer(dapServer);
```
TingluoHuang: it's little bit odd that server set session, and session set server. 😄 

should the jobrunner care about the session+server, or it should only care about a debugger?

src/Runner.Worker/Dap/DapReplParser.cs
```diff
@@ -0,0 +1,409 @@
...
+            }
+
+            // Not a DSL command
+            return null;
+        }
+
+        internal static string GetGeneralHelp()
+        {
+            var sb = new StringBuilder();
+            sb.AppendLine("Actions Debug Console");
```
salmanmkc: could potentially just make a big string here instead of string building since we're just returning a string at the end

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +253,41 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        await debugSession.WaitForHandshakeAsync(jobRequestCancellationToken);
+                        Trace.Info("DAP handshake complete.");
+
+                        dapCancellationRegistration = jobRequestCancellationToken.Register(() =>
+                        {
+                            Trace.Info("Job cancellation requested, cancelling debug session.");
+                            debugSession.CancelSession();
+                        });
+                    }
+                    catch (OperationCanceledException) when (jobRequestCancellationToken.IsCancellationRequested)
```
rentziass, line 254: This is where we will do autocancellation if users don't connect to the DAP server

src/Runner.Worker/JobRunner.cs
```diff
@@ -121,6 +125,36 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        debugSession.SetDapServer(dapServer);
+
+                        await dapServer.StartAsync(port, jobRequestCancellationToken);
+                        Trace.Info($"DAP server started on port {port}, listening for debugger client");
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Warning($"Failed to start DAP server: {ex.Message}. Job will continue without debugging.");
+                        dapServer = null;
+                        debugSession = null;
```
TingluoHuang: should we not set these to `null`, so we always run the cleanup in the final, in case the server/session holds up some resources?

src/Runner.Worker/JobRunner.cs
TingluoHuang, line 254: when we don't get anything connect or timeout, do we still want to run the job?

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +253,41 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        dapCancellationRegistration = jobRequestCancellationToken.Register(() =>
+                        {
+                            Trace.Info("Job cancellation requested, cancelling debug session.");
+                            debugSession.CancelSession();
+                        });
+                    }
+                    catch (OperationCanceledException) when (jobRequestCancellationToken.IsCancellationRequested)
+                    {
+                        Trace.Info("Job was cancelled before debugger client connected. Continuing without debugger.");
+                        try { await dapServer.StopAsync(); } catch { }
```
TingluoHuang:
```suggestion
                        try
                        {
                             await dapServer.StopAsync();
                         }
                         catch (Exception ex)
                         {
                             Trace.Error("Fail to stop debugger server")
                             Trace.Error(ex); 
                         }
```


same for the other online try-catch.

TingluoHuang: also, do we need to have 2 try-catch if the only thing different is a `trace`?

src/Runner.Worker/JobRunner.cs
```diff
@@ -259,6 +328,25 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                if (dapServer != null)
+                {
+                    try
+                    {
+                        Trace.Info("Stopping DAP server");
+                        await dapServer.StopAsync();
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Warning($"Error stopping DAP server: {ex.Message}");
```
TingluoHuang:
```suggestion
                        Trace.Error($"Error stopping DAP server");
                        Trace.Error(ex);
```

src/Runner.Worker/JobRunner.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/4bf2b29367a46377042b12a4b7d05ba5d4c76c17

src/Runner.Worker/JobRunner.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/4bf2b29367a46377042b12a4b7d05ba5d4c76c17

src/Runner.Worker/JobRunner.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/4bf2b29367a46377042b12a4b7d05ba5d4c76c17

src/Runner.Worker/JobRunner.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/4bf2b29367a46377042b12a4b7d05ba5d4c76c17

src/Runner.Worker/JobRunner.cs
rentziass: Addressed in https://github.com/actions/runner/commit/4bf2b29367a46377042b12a4b7d05ba5d4c76c17, we're cleaning up in place now too.

src/Runner.Worker/JobRunner.cs
rentziass: Moved to StopAsync definition

src/Runner.Worker/JobRunner.cs
rentziass: Nope, https://github.com/actions/runner/pull/4298/commits/437e20db7d7aec949a014685e918f9dc0e4cde13 now fails the job.

rentziass: Added in https://github.com/actions/runner/pull/4298/commits/437e20db7d7aec949a014685e918f9dc0e4cde13

src/Runner.Worker/Dap/DapReplParser.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/f9919b2ba4cb91a8710729e46c43a32da06b00e6

src/Runner.Worker/Dap/DapVariableProvider.cs
rentziass: Addressed in https://github.com/actions/runner/pull/4298/commits/f9919b2ba4cb91a8710729e46c43a32da06b00e6

src/Runner.Worker/JobRunner.cs
```diff
@@ -178,6 +180,28 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                        await dapDebugger.StartAsync(jobRequestCancellationToken);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Error($"Failed to start DAP debugger: {ex.Message}");
+                        if (dapDebugger != null)
+                        {
+                            try { await dapDebugger.StopAsync(); } catch { }
+                        }
+                        dapDebugger = null;
```
TingluoHuang: do we want to failed the job here? and rely on `final` for the `dapDebugger.StopAsync()?

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +243,37 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                {
+                    try
+                    {
+                        await dapDebugger.WaitUntilReadyAsync(jobRequestCancellationToken);
+                    }
+                    catch (OperationCanceledException) when (jobRequestCancellationToken.IsCancellationRequested)
+                    {
+                        Trace.Info("Job was cancelled before debugger client connected.");
+                        try { await dapDebugger.StopAsync(); } catch { }
+                        dapDebugger = null;
```
TingluoHuang: same here, do we want to always just rely on the `finally` to stop the debugger?

src/Runner.Worker/JobRunner.cs
```diff
@@ -178,6 +180,28 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                    {
+                        dapDebugger = HostContext.GetService<IDapDebugger>();
+                        await dapDebugger.StartAsync(jobRequestCancellationToken);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Error($"Failed to start DAP debugger: {ex.Message}");
+                        if (dapDebugger != null)
+                        {
+                            try { await dapDebugger.StopAsync(); } catch { }
```
TingluoHuang: also, `StopAsync()` seems doesn't throw exception.

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +243,37 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                // so the job page shows the setup step before we block on the debugger
+                if (dapDebugger != null)
+                {
+                    try
+                    {
+                        await dapDebugger.WaitUntilReadyAsync(jobRequestCancellationToken);
+                    }
+                    catch (OperationCanceledException) when (jobRequestCancellationToken.IsCancellationRequested)
+                    {
+                        Trace.Info("Job was cancelled before debugger client connected.");
```
TingluoHuang, line 256: do we want to `jobContext.Error("Debugger is cancelled due to job cancellation")` or something similar?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -50,6 +51,16 @@ public async Task RunAsync(IExecutionContext jobContext)
...
             var scopeInputs = new Dictionary<string, PipelineContextData>(StringComparer.OrdinalIgnoreCase);
             bool checkPostJobActions = false;
+            IDapDebugger dapDebugger = null;
+            try
+            {
+                dapDebugger = HostContext.GetService<IDapDebugger>();
+            }
+            catch
+            {
+                // Debugger not available — continue without debugging
```
TingluoHuang: ` dapDebugger = HostContext.GetService<IDapDebugger>();` would never throw, i think. So we don't need to have try-catch.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -255,6 +278,11 @@ public async Task RunAsync(IExecutionContext jobContext)
 
                 Trace.Info($"Current state: job state = '{jobContext.Result}'");
             }
+
+            if (dapDebugger != null)
+            {
+                dapDebugger.OnJobCompleted();
```
TingluoHuang: does the `OnJobCompleted` need to be part of some finally, also do we want to merge it with `StopAsync()`?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -226,9 +237,21 @@ public async Task RunAsync(IExecutionContext jobContext)
                         }
                         else
                         {
+                            // Pause for DAP debugger before step execution
+                            if (dapDebugger != null)
+                            {
+                                await dapDebugger.OnStepStartingAsync(step, jobContext, isFirstStep, jobContext.CancellationToken);
```
TingluoHuang: should the `isFirstStep` just be tracked within `dapDebugger`?

TingluoHuang: also, since the GetService<> should always return something, we can remove the null check.
we can also just do `await dapDebugger?.OnStepStartingAsync(step, jobContext, isFirstStep, jobContext.CancellationToken);`

src/Runner.Worker/StepsRunner.cs
```diff
@@ -226,9 +237,21 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                                isFirstStep = false;
+                            }
+
                             // Run the step
                             await RunStepAsync(step, jobContext.CancellationToken);
                             CompleteStep(step);
+
+                            if (dapDebugger != null)
```
TingluoHuang: same about the null check.

src/Runner.Worker/Dap/IDapDebugger.cs
TingluoHuang, line 1: what's the reason for having the debugger and session?
is it possible to merge them?
would we ever support multiple sessions under the same debugger?

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +243,37 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                     await Task.WhenAny(_jobServerQueue.JobRecordUpdated.Task, Task.Delay(1000));
                 }
 
+                // Wait for DAP debugger client connection and handshake after "Set up job"
+                // so the job page shows the setup step before we block on the debugger
+                if (dapDebugger != null)
+                {
+                    try
+                    {
+                        await dapDebugger.WaitUntilReadyAsync(jobRequestCancellationToken);
```
TingluoHuang: do you want to report telemetry on how many job is able to connect?

src/Runner.Worker/Dap/IDapDebugger.cs
rentziass: If you mean the server the only reason was separating TCP logic from "actual" functionality but yeah I can look at merging them.

rentziass: > would we ever support multiple sessions under the same debugger?

this is supported by the protocol but I don't think we'd be doing that anytime soon.

rentziass: Ignore me, I've merged debugger and session in https://github.com/actions/runner/pull/4298/commits/da35402572cf0df3471edc8ac73e7313c166d375, let me know if that's what you had in mind!

src/Runner.Worker/JobRunner.cs
rentziass: Added in https://github.com/actions/runner/pull/4298/commits/fa298ebe45fc85297926fea7f88ba59211d6eb8c

src/Runner.Worker/JobRunner.cs
```diff
@@ -178,6 +181,26 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+
+                    try
+                    {
+                        dapDebugger = HostContext.GetService<IDapDebugger>();
+                        await dapDebugger.StartAsync(jobRequestCancellationToken);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Error($"Failed to start DAP debugger: {ex.Message}");
+                        AddDebuggerConnectionTelemetry(jobContext, "Failed");
```
TingluoHuang:
```suggestion
                        AddDebuggerConnectionTelemetry(jobContext, $"Failed: {ex.Message}");
```

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +242,34 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+                    {
+                        Trace.Info("Job was cancelled before debugger client connected.");
+                        AddDebuggerConnectionTelemetry(jobContext, "Canceled");
+                        jobContext.Error("Job was cancelled before debugger client connected.");
+                        return await CompleteJobAsync(server, jobContext, message, TaskResult.Canceled);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Error($"DAP debugger failed to become ready: {ex.Message}");
+                        AddDebuggerConnectionTelemetry(jobContext, "Failed");
```
TingluoHuang:
```suggestion
                        AddDebuggerConnectionTelemetry(jobContext, $"Failed: {ex.Message}");
```

src/Runner.Worker/JobRunner.cs
```diff
@@ -219,6 +242,34 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
...
+
+                        // If debugging was requested but the debugger is not available, fail the job
+                        var errorMessage = "The debugger failed to start or no debugger client connected in time.";
+                        jobContext.Error(errorMessage);
```
TingluoHuang:
```suggestion
                        jobContext.Error("The debugger failed to start or no debugger client connected in time.");
```

src/Runner.Worker/Dap/IDapDebugger.cs
TingluoHuang: do we want to keep merge this with `IDapServer`?
Try to understand what's the benefit for having these interfaces.

src/Runner.Worker/Dap/IDapDebugger.cs
```diff
@@ -0,0 +1,29 @@
...
+        Ready,
+        Paused,
+        Running,
+        Terminated
+    }
+
+    [ServiceLocator(Default = typeof(DapDebugger))]
+    public interface IDapDebugger : IRunnerService
+    {
+        bool IsActive { get; }
```
TingluoHuang: `IsActive` is not used externally, do we need it on the interface?

src/Runner.Worker/Dap/IDapDebugger.cs
```diff
@@ -0,0 +1,29 @@
...
+        Task StartAsync(CancellationToken cancellationToken);
+        Task WaitUntilReadyAsync(CancellationToken cancellationToken);
+        Task StopAsync();
+        void CancelSession();
```
TingluoHuang: same to `CancelSession()`

src/Runner.Worker/Dap/IDapDebugger.cs
```diff
@@ -0,0 +1,29 @@
...
+        Terminated
+    }
+
+    [ServiceLocator(Default = typeof(DapDebugger))]
+    public interface IDapDebugger : IRunnerService
+    {
+        bool IsActive { get; }
+        Task StartAsync(CancellationToken cancellationToken);
+        Task WaitUntilReadyAsync(CancellationToken cancellationToken);
+        Task StopAsync();
```
TingluoHuang: same to `StopAsync()`

src/Runner.Worker/JobRunner.cs
```diff
@@ -178,6 +181,26 @@ public async Task<TaskResult> RunAsync(AgentJobRequestMessage message, Cancellat
                 _tempDirectoryManager = HostContext.GetService<ITempDirectoryManager>();
                 _tempDirectoryManager.InitializeTempDirectory(jobContext);
 
+                // Setup the debugger
+                if (jobContext.Global.EnableDebugger)
+                {
+                    Trace.Info("Debugger enabled for this job run");
```
ChristopherHX, line 187: Write this to job log as well? So you can actually see why the job does not continue with the next step?

Only the error `Failed to start debugger` can be seen there.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -255,6 +262,7 @@ public async Task RunAsync(IExecutionContext jobContext)
 
                 Trace.Info($"Current state: job state = '{jobContext.Result}'");
             }
+
```
ChristopherHX, line 265: I usually find it useful to be able to inspect variables and run scripts after the last step of a job

Image

```suggestion

            // Pause for DAP debugger before sending job completion
            await dapDebugger?.OnJobCompleteStartingAsync(jobContext.CancellationToken);
```

src/Runner.Worker/Dap/IDapDebugger.cs
```diff
@@ -0,0 +1,29 @@
...
+        void CancelSession();
+        Task OnStepStartingAsync(IStep step, IExecutionContext jobContext, CancellationToken cancellationToken);
+        void OnStepCompleted(IStep step);
+        Task OnJobCompletedAsync();
```
ChristopherHX, line 24: RE https://github.com/actions/runner/pull/4298/changes#r2949684913
```suggestion
        Task OnJobCompleteStartingAsync(CancellationToken cancellationToken);
        Task OnJobCompletedAsync();
```

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1085 @@
...
+                throw new TimeoutException($"No debugger client connected within {timeoutMinutes} minutes.");
+            }
+
+            _cancellationRegistration = cancellationToken.Register(() =>
+            {
+                Trace.Info("Job cancellation requested, cancelling debug session.");
+                CancelSession();
+            });
+        }
+
```
ChristopherHX, line 157: RE https://github.com/actions/runner/pull/4298/changes#r2949684913
```suggestion

        public async Task OnJobCompleteStartingAsync(CancellationToken cancellationToken)
        {
            if (_started)
            {
                try
                {
                    bool pauseOnNextStep;
                    lock (_stateLock)
                    {
                        if (_state == DapSessionState.Terminated)
                        {
                            Trace.Info("Session already terminated, skipping OnJobCompleted events");
                            return;
                        }
                        pauseOnNextStep = _pauseOnNextStep;
                    }

                    if (pauseOnNextStep)
                    {
                        var reason = "step";
                        var description = "Stopped at job completion";

                        Trace.Info("Job completed with debugger pause");

                        // Send stopped event to debugger (only if client is connected)
                        SendStoppedEvent(reason, description);

                        // Wait for debugger command
                        await WaitForCommandAsync(cancellationToken);
                    }
                }
                catch (Exception ex)
                {
                    Trace.Warning($"DAP OnJobCompleteStartingAsync error: {ex.Message}");
                }
            }
        }

```

ChristopherHX (commented):
I like it 👍 , works pretty well in vscode.

- Only the Callstack did not auto focus the active callframe for me so e.g. watch displayed the error (`(no execution context available)`) until selecting the step where the debugger stopped by hand
- I used this vscode extension stub to attach: [vscode-ext-to-attach.zip](https://github.com/user-attachments/files/26069187/vscode-ext-to-attach.zip) (generated by AI)
- Expression debug logs end up in the uploaded job log (not debugger log) if you evaluate expressions (ACTIONS_STEP_DEBUG=true)

src/Runner.Worker/JobRunner.cs
rentziass: Thank you that's a great point, I'm going to add a new "artificial" step after job setup to better show users that the debugger is enabled and we're waiting on them to connect ❤️

src/Runner.Worker/StepsRunner.cs
rentziass: I totally agree with you but we need to figure out the right way for doing this, and it's probably going to come in the next PRs since this one is huge as is already 😄 

We need to find a good way to both:
- give users a good way to setup breakpoints looking at source code
- allow users to setup breakpoints at times that can't be referred to from normal source code (pre/post actions steps and job completion come to mind here).

That is to say, we're quite a bit away still from fully releasing this but this is most definitely on our radar!

src/Runner.Worker/JobRunner.cs
rentziass: Got a first draft working locally but it's a sizeable change, going to move it to its own PR. Adding a pre/post step might also be a good way to let us inspect state after the last actual job step completed too.

ChristopherHX: Indeed much more visible than just a log message in a Job setup and reads really like a much bigger change.

> Adding a pre/post step might also be a good way to let us inspect state after the last actual job step completed too.

Sounds like an interesting solution.

src/Runner.Worker/StepsRunner.cs
ChristopherHX: > * allow users to setup breakpoints at times that can't be referred to from normal source code [...]

> job completion

Idk how this works for a DAP Adapter, but some vscode extension debugger have a checkbox for unhandled exception if this would allow a free text JobCompletion could be such a predefined enable/disable breakpoint.
E.g. in my experiment, only step over is a possible way to pause before job completion.

The initialize request can do this
```
					"exceptionBreakpointFilters": []map[string]any{
						{"filter": "jobexit", "label": "Job Exit", "default": true},
					},
```
Then you get a exception filter event on every toggle.

> pre/post actions steps

Conditional breakpoints come into my mind: E.g. you make a breakpoint that would by default pause before pre, main and post. Optionally the condition could have a variable that can filter the stage e.g. condition `stage == 'main'` (using existing actions expressions with custom global variable from debugger)

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+        {
+            base.Initialize(hostContext);
+            _variableProvider = new DapVariableProvider(hostContext);
+            _replExecutor = new DapReplExecutor(hostContext, SendOutput);
+            Trace.Info("DapDebugger initialized");
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            ResetSessionState();
```
TingluoHuang: do we need to have `ResetSessionState`? will we ever plan to call `StartAsync()` twice?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+                Success = success,
+                Message = success ? null : message,
+                Body = body
+            };
+        }
+
+        internal int ResolvePort()
+        {
+            var portEnv = Environment.GetEnvironmentVariable(PortEnvironmentVariable);
+            if (!string.IsNullOrEmpty(portEnv) && int.TryParse(portEnv, out var customPort) && customPort > 0 && customPort <= 65535)
```
TingluoHuang: do we want to dis-allow system port usage for <1024?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            ResetSessionState();
+            var port = ResolvePort();
+
+            Trace.Info($"Starting DAP debugger on port {port}");
+
+            _cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+            _connectionTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
TingluoHuang: 🤷 not sure what is RunContinuationsAsynchronously doing.

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+            {
+                try
+                {
+                    Trace.Info("Waiting for debug client connection...");
+
+                    using (cancellationToken.Register(() =>
+                    {
+                        try { _listener?.Stop(); }
+                        catch { /* listener already stopped */ }
+                    }))
```
TingluoHuang: this is odd, it means we are keep creating new Register() per loop.

We should just have one `Register()`, i think.

src/Runner.Worker/Dap/IDapDebugger.cs
```diff
@@ -0,0 +1,26 @@
...
+        Ready,
+        Paused,
+        Running,
+        Terminated
+    }
+
+    [ServiceLocator(Default = typeof(DapDebugger))]
+    public interface IDapDebugger : IRunnerService
+    {
+        Task StartAsync(CancellationToken cancellationToken);
```
TingluoHuang: would it make the code cleaner if we pass the jobContext into StartAsync, and store it?
so we don't have to pass in `cancellationToken` to these method?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+
+            // Intentionally do NOT release the command TCS here.
+            // The session stays paused, waiting for a client to reconnect.
+            // The debugger's connection loop will accept a new client and
+            // call HandleClientConnected, which re-sends the stopped event.
+        }
+
+        private async Task ConnectionLoopAsync(CancellationToken cancellationToken)
+        {
+            while (_acceptConnections && !cancellationToken.IsCancellationRequested)
```
TingluoHuang: is the `_acceptConnections` kind of redundant to `_started`

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+                Trace.Info("Message processing cancelled");
+            }
+            catch (IOException ex)
+            {
+                Trace.Info($"Connection closed ({ex.GetType().Name})");
+            }
+            catch (Exception ex)
+            {
+                Trace.Error($"Error in message loop ({ex.GetType().Name})");
+            }
```
TingluoHuang, line 492: not sure if we need to catch different exception type if we just want to log to runner diag

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+                    CleanupConnection();
+                }
+                catch (ObjectDisposedException) when (cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }
+                catch (SocketException) when (cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }
```
TingluoHuang: catch (Exception) when (cancellationToken.IsCancellationRequested)
{
   break;
}

Should we just break when we got token cancelled on any exception?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+                    CleanupConnection();
+
+                    if (!_acceptConnections || cancellationToken.IsCancellationRequested)
+                    {
+                        break;
+                    }
+
+                    try
+                    {
+                        await Task.Delay(100, cancellationToken);
```
TingluoHuang: would 500ms feel safer?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1433 @@
...
+                {
+                    break;
+                }
+                catch (SocketException) when (cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    Trace.Warning($"Connection error ({ex.GetType().Name})");
```
TingluoHuang: we can just log the full error to diag log.
```
Trace.Error("Debugger connection error");
Trace.Error(ex);
```

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1410 @@
...
+            {
+                EventType = "exited",
+                Body = new ExitedEventBody
+                {
+                    ExitCode = exitCode
+                }
+            });
+        }
+
+        private static TaskCompletionSource<bool> CreateHandshakeCompletionSource()
```
TingluoHuang: do we need this method that only has 1 line? 😄 :copilot:

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -0,0 +1,1410 @@
...
+                SupportsRunInTerminalRequest = false,
+                SupportsCancelRequest = false,
+                SupportsExceptionOptions = false,
+                SupportsValueFormattingOptions = false,
+                SupportsExceptionInfoRequest = false,
+            };
+
+            // Send initialized event after a brief delay to ensure the
+            // response is delivered first (DAP spec requirement)
+            _ = Task.Run(async () =>
```
TingluoHuang: We don't do `Task.Run` in the runner, i think.

src/Runner.Worker/Dap/DapDebugger.cs
rentziass: Doesn't it make sense though to distinguish between cancel/disconnects and actual errors?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+        // Scope references occupy the range [1, ScopeReferenceMax].
+        private const int ScopeReferenceBase = 1;
+        private const int ScopeReferenceMax = 100;
+
+        // Dynamic (nested) variable references start above the scope range.
+        private const int DynamicReferenceBase = 101;
+
+        internal const string RedactedValue = "***";
+
+        private readonly IHostContext _hostContext;
```
TingluoHuang: we can just take an `ISecretMasker` instead of `IHostContext`

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,387 @@
...
+            return variables;
+        }
+
+        /// <summary>
+        /// Applies the runner's secret masker to the given value.
+        /// This is the single masking entry-point for all DAP-visible strings
+        /// and is intentionally public so future DAP features (evaluate, REPL)
+        /// can reuse it without duplicating policy.
+        /// </summary>
+        public string MaskSecrets(string value)
```
TingluoHuang: does this need to be a public method, or even need the method?

src/Runner.Worker/Dap/DapReplParser.cs
```diff
@@ -0,0 +1,411 @@
...
+                  run("script")         Execute a script (like a workflow run step)
+
+                Anything else is evaluated as a GitHub Actions expression.
+                  Example: github.repository
+                  Example: ${{ github.event_name }}
+
+                """;
+        }
+
+        internal static string GetRunHelp()
```
TingluoHuang, line 102: this feels like a parsed json?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// value is replaced with a constant redaction marker.
+    ///
+    /// Designed to be reusable by future DAP features (evaluate, hover, REPL)
+    /// so that masking policy is never duplicated.
+    /// </summary>
+    internal sealed class DapVariableProvider
+    {
+        // Well-known scope names that map to top-level expression contexts.
+        // Order matters: the index determines the stable variablesReference ID.
+        internal static readonly string[] ScopeNames =
```
TingluoHuang: `private`

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+        };
+
+        // Scope references occupy the range [1, ScopeReferenceMax].
+        private const int ScopeReferenceBase = 1;
+        private const int ScopeReferenceMax = 100;
+
+        // Dynamic (nested) variable references start above the scope range.
+        private const int DynamicReferenceBase = 101;
+
+        internal const string RedactedValue = "***";
```
TingluoHuang: `private`

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+        // Well-known scope names that map to top-level expression contexts.
+        // Order matters: the index determines the stable variablesReference ID.
+        internal static readonly string[] ScopeNames =
+        {
+            "github", "env", "runner", "job", "steps",
+            "secrets", "inputs", "vars", "matrix", "needs"
+        };
+
+        // Scope references occupy the range [1, ScopeReferenceMax].
+        private const int ScopeReferenceBase = 1;
```
TingluoHuang: all private var is named as `_xxxx` in runner, i think.

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+            var childPath = string.IsNullOrEmpty(basePath) ? name : $"{basePath}.{name}";
+            var variable = new Variable
+            {
+                Name = name,
+                EvaluateName = $"${{{{ {childPath} }}}}"
+            };
+
+            if (value == null)
+            {
+                variable.Value = isSecretsScope ? RedactedValue : "null";
```
TingluoHuang: do we need to check `isSecretsScope` here?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+                    break;
+
+                case NumberContextData num:
+                    variable.Value = _secretMasker.MaskSecrets(num.Value.ToString("G15", System.Globalization.CultureInfo.InvariantCulture));
+                    variable.Type = "number";
+                    variable.VariablesReference = 0;
+                    break;
+
+                case BooleanContextData boolVal:
+                    variable.Value = _secretMasker.MaskSecrets(boolVal.Value ? "true" : "false");
```
TingluoHuang: do we need secret mask true/false?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+            switch (value)
+            {
+                case StringContextData str:
+                    variable.Value = _secretMasker.MaskSecrets(str.Value);
+                    variable.Type = "string";
+                    variable.VariablesReference = 0;
+                    break;
+
+                case NumberContextData num:
+                    variable.Value = _secretMasker.MaskSecrets(num.Value.ToString("G15", System.Globalization.CultureInfo.InvariantCulture));
```
TingluoHuang: do we need to mask the number?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+        /// <summary>
+        /// Infers a simple DAP type hint from the string representation of a result.
+        /// </summary>
+        internal static string InferResultType(string value)
+        {
+            if (value == null || value == "null")
+                return "null";
+            if (value == "true" || value == "false")
+                return "boolean";
+            if (double.TryParse(value, System.Globalization.NumberStyles.Any,
```
TingluoHuang: Move the `System.Globalization` to using on top of the file

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+                };
+            }
+        }
+
+        /// <summary>
+        /// Infers a simple DAP type hint from the string representation of a result.
+        /// </summary>
+        internal static string InferResultType(string value)
+        {
+            if (value == null || value == "null")
```
TingluoHuang, line 234: do you need to `tolower()` before comparing string?

src/Runner.Worker/Dap/DapVariableProvider.cs
```diff
@@ -0,0 +1,372 @@
...
+                    Type = "string",
+                    VariablesReference = 0
+                };
+            }
+        }
+
+        /// <summary>
+        /// Infers a simple DAP type hint from the string representation of a result.
+        /// </summary>
+        internal static string InferResultType(string value)
```
TingluoHuang, line 231: `private` maybe better?

src/Runner.Worker/Dap/DapReplParser.cs
TingluoHuang: or the workflow `- run` step?

src/Runner.Worker/Dap/DapVariableProvider.cs
rentziass: We used to mask only strings at the beginning, I agree that bool might be a bit much but numbers could be used as secrets?

src/Runner.Worker/Dap/DapReplParser.cs
TingluoHuang: run: {"command": "echo $foo", env: {....}}

rentziass: Let's move this to a bigger conversation but good call, something recognizable like `run()` + JSON as the arg would save us a lot of parsing work. I liked the ruby-ness of it in the console but truth we'll probably have tools that make these calls on behalf of users.

ChristopherHX: I believe you could also handle additionally action yml snippets (vscode shift enter for multiline input,  not unlikely nvim-dap can as well)
```
run
- run: |
    echo Test
```

- Runner.Worker has a steps yaml parser for this
- Can be detected by `run` followed by newline
- Similar to the lldb `script` command that can be followed by multiline python code

A single line fallback is always good, but for some advanced users is copy paste actual actions workflow yaml possibly a big usability win

Single line could look like this if we omit the list `-` or `[]`
```
run { run: echo Test, env: { a: b }}
```
Multiline without list
```
run
run: |
   echo Test
```

TingluoHuang (approved):

gordolara789-tech (commented):
Éxito

PR #4303, opened 2026-03-16
Node 24 enforcement + Linux ARM32 deprecation support

salmanmkc:
When we flip `UseNode24ByDefault`, actions targeting node20 get forced to node24 and we show an annotation saying which ones were forced over. On linux-arm32 we can't do that since there's no node24 binary, so instead we show a deprecation warning and keep them on node20 for now.

Added two new feature flags for the arm32 side:
- `actions_runner_deprecate_linux_arm32` - warns that arm32 is going away, keeps running on node20
- `actions_runner_kill_linux_arm32` - for when we actually want to stop supporting it, fails the step

Also split the tracking so upgraded-to-node24 actions and still-on-node20 actions get separate annotations instead of lumping them together.

Date for arm32 EOL is TBD, placeholder is October 2026, need to align on this still


src/Runner.Common/Constants.cs
```diff
@@ -195,8 +195,18 @@ public static class NodeMigration
...
 
+                // Feature flags for Linux ARM32 deprecation
+                public static readonly string DeprecateLinuxArm32Flag = "actions_runner_deprecate_linux_arm32";
+                public static readonly string KillLinuxArm32Flag = "actions_runner_kill_linux_arm32";
+
                 // Blog post URL for Node 20 deprecation
                 public static readonly string Node20DeprecationUrl = "https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/";
+
+                // Node 20 migration dates
+                public static readonly string Node24DefaultDate = "June 2nd, 2026";
```
ericsciple, line 206: Have you considered sending the date in the job message, in the variables section? Not critical, but might give us more flexibility if needed.

ericsciple: Will defer to your judgement, may not be necessary. But something to consider for future deprecations.

src/Runner.Worker/Handlers/HandlerFactory.cs
```diff
@@ -100,6 +101,26 @@ public IHandler Create(
...
 
+                    // Track actions based on their final node version
+                    if (!string.IsNullOrEmpty(actionName))
+                    {
+                        if (string.Equals(finalNodeVersion, Constants.Runner.NodeMigration.Node24, StringComparison.OrdinalIgnoreCase))
+                        {
+                            // Action was upgraded from node20 to node24
+                            executionContext.Global.UpgradedToNode24Actions?.Add(actionName);
+                        }
+                        else if (deprecateArm32 && string.Equals(finalNodeVersion, Constants.Runner.NodeMigration.Node20, StringComparison.OrdinalIgnoreCase))
```
ericsciple: 🤖 **Copilot Review**

Bug: this branch checks `deprecateArm32` (the feature flag) but not the actual platform. Since the server sends `actions_runner_deprecate_linux_arm32` to **all** runners regardless of architecture, this will misclassify actions on non-ARM32 runners.

**Scenario:** On an x64 runner with `deprecateArm32 == true`, if a user opts out of node24 (via `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION`), `finalNodeVersion` will be `"node20"` and this branch fires — adding the action to `Arm32Node20Actions`. The job annotation will say *"running on Node.js 20 because Node.js 24 is not available on Linux ARM32"* on an x64 machine.

**Suggested fix:** Gate on the platform outcome, e.g.:
```suggestion
                        else if (!string.IsNullOrEmpty(platformWarningMessage) && string.Equals(finalNodeVersion, Constants.Runner.NodeMigration.Node20, StringComparison.OrdinalIgnoreCase))
```
`platformWarningMessage` is only non-null when `CheckNodeVersionForLinuxArm32` actually detected ARM32, so this naturally scopes the tracking to the correct platform.

ericsciple (commented):
Review from GitHub Copilot (via CLI)

src/Runner.Worker/Handlers/HandlerFactory.cs
ericsciple: This may not be true. I think the only way to reach here is when we're stuck on node20 because node24 doesn't exist for arm32.

ericsciple: 🤖 **Copilot Review**

There is a path where this fires on x64. Traced through `DetermineActionsNodeVersion`:

1. Server sends `useNode24ByDefault = true` and `deprecateArm32 = true`
2. On an **x64** runner, user sets `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` to opt out
3. `DetermineActionsNodeVersion` returns `node20` ([NodeUtil.cs line ~99](https://github.com/actions/runner/blob/3f201cb0/src/Runner.Common/Util/NodeUtil.cs#L99): `if (allowUnsecureNode) return (Node20, null)`)
4. `CheckNodeVersionForLinuxArm32("node20", deprecateArm32: true)` on x64 → early returns `("node20", null)` since `!isArm32Linux`
5. `finalNodeVersion = "node20"`, `deprecateArm32 = true` → enters this branch

The action gets added to `Arm32Node20Actions` and the job annotation says "because Node.js 24 is not available on Linux ARM32" on an x64 machine.

PR #4325, opened 2026-04-06
Add `ACTIONS_RUNTIME_TOKEN` and runtime env vars to `ScriptHandler`

0b01:
`ScriptHandler` was missing the Actions Runtime environment variables (`ACTIONS_RUNTIME_URL`, `ACTIONS_RUNTIME_TOKEN`, `ACTIONS_CACHE_URL`, `ACTIONS_RESULTS_URL`) that `NodeScriptActionHandler` and `ContainerActionHandler` already expose. This meant run: steps and hook scripts couldn't interact with Actions runtime services (such as caching, artifacts, etc.) the same way Node/container actions(such as `@actions/upload-artifact`) can.

Added the same runtime env var block from `NodeScriptActionHandler` into `ScriptHandler`, placed before the existing `ACTIONS_ID_TOKEN_REQUEST_*` setup similar to `NodeScriptActionHandler`.

### DIFF
--- a/src/Runner.Worker/Handlers/ScriptHandler.cs
+++ b/src/Runner.Worker/Handlers/ScriptHandler.cs
@@ -311,11 +311,26 @@ namespace GitHub.Runner.Worker.Handlers
                 fileName = node;
             }
 #endif
+            // Add Actions Runtime server info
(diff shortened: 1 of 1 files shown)


fclante:
@0b01 Good Idea - Something I could really use - But I think you should implement the tests that are missing according to @copilot's review above and rebase the branch. I wonder why no `real person` has reacted to this yet?

PR #4328, opened 2026-04-08
Add WS bridge over DAP TCP server

rentziass:
This adds a bridge converting messages to/from TCP <-> WS so that we can connect to the DAP server through `wss://` directly and using GitHub credentials out of the box with Dev Tunnels.

https://github.com/github/c2c-actions/issues/9831

### DIFF
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -114,6 +114,7 @@ namespace GitHub.Runner.Worker.Dap
 
         internal DapSessionState State => _state;
         internal int InternalDapPort => (_listener?.LocalEndpoint as IPEndPoint)?.Port ?? 0;
+        internal int BridgeListenPort => _webSocketBridge?.ListenPort ?? 0;
 
         public override void Initialize(IHostContext hostContext)
         {
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -149,10 +149,9 @@ namespace GitHub.Runner.Worker.Dap
             else
             {
                 Trace.Info($"Internal DAP debugger listening on {_listener.LocalEndpoint}");
-                _webSocketBridge = new WebSocketDapBridge(
-                    HostContext.GetTrace("DapWebSocketBridge"),
-                    debuggerConfig.Tunnel.Port,
-                    InternalDapPort);
+                _webSocketBridge = new WebSocketDapBridge();
+                _webSocketBridge.Initialize(HostContext);
+                _webSocketBridge.Configure(debuggerConfig.Tunnel.Port, InternalDapPort);
                 _webSocketBridge.Start();
             }
 
@@ -296,10 +295,10 @@ namespace GitHub.Runner.Worker.Dap
                 if (_webSocketBridge != null)
                 {
                     Trace.Info("Stopping WebSocket DAP bridge");
-                    var disposeTask = _webSocketBridge.DisposeAsync().AsTask();
-                    if (await Task.WhenAny(disposeTask, Task.Delay(5_000)) != disposeTask)
+                    var shutdownTask = _webSocketBridge.ShutdownAsync();
+                    if (await Task.WhenAny(shutdownTask, Task.Delay(5_000)) != shutdownTask)
                     {
-                        Trace.Warning("WebSocket DAP bridge dispose timed out after 5s");
+                        Trace.Warning("WebSocket DAP bridge shutdown timed out after 5s");
                     }
                     else
                     {
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -114,7 +114,6 @@ namespace GitHub.Runner.Worker.Dap
 
         internal DapSessionState State => _state;
         internal int InternalDapPort => (_listener?.LocalEndpoint as IPEndPoint)?.Port ?? 0;
-        internal int BridgeListenPort => _webSocketBridge?.ListenPort ?? 0;
 
         public override void Initialize(IHostContext hostContext)
         {
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -66,7 +66,7 @@ namespace GitHub.Runner.Worker.Dap
 
         // Dev Tunnel relay host for remote debugging
         private TunnelRelayTunnelHost _tunnelRelayHost;
-        private WebSocketDapBridge _webSocketBridge;
+        private IWebSocketDapBridge _webSocketBridge;
 
         // Cancellation source for the connection loop, cancelled in StopAsync
         // so AcceptTcpClientAsync unblocks cleanly without relying on listener disposal.
@@ -149,10 +149,8 @@ namespace GitHub.Runner.Worker.Dap
             else
             {
                 Trace.Info($"Internal DAP debugger listening on {_listener.LocalEndpoint}");
-                _webSocketBridge = new WebSocketDapBridge();
-                _webSocketBridge.Initialize(HostContext);
-                _webSocketBridge.Configure(debuggerConfig.Tunnel.Port, InternalDapPort);
-                _webSocketBridge.Start();
+                _webSocketBridge = HostContext.CreateService<IWebSocketDapBridge>();
+                _webSocketBridge.Start(debuggerConfig.Tunnel.Port, InternalDapPort);
             }
 
             // Start Dev Tunnel relay so remote clients reach the local DAP port.
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -299,6 +299,9 @@ namespace GitHub.Runner.Worker.Dap
                     if (await Task.WhenAny(shutdownTask, Task.Delay(5_000)) != shutdownTask)
                     {
                         Trace.Warning("WebSocket DAP bridge shutdown timed out after 5s");
+                        _ = shutdownTask.ContinueWith(
+                            t => Trace.Error($"WebSocket DAP bridge shutdown faulted: {t.Exception?.GetBaseException().Message}"),
+                            TaskContinuationOptions.OnlyOnFaulted);
                     }
                     else
                     {
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -66,6 +66,7 @@ namespace GitHub.Runner.Worker.Dap
 
         // Dev Tunnel relay host for remote debugging
         private TunnelRelayTunnelHost _tunnelRelayHost;
+        private WebSocketDapBridge _webSocketBridge;
 
         // Cancellation source for the connection loop, cancelled in StopAsync
         // so AcceptTcpClientAsync unblocks cleanly without relying on listener disposal.
@@ -73,6 +74,10 @@ namespace GitHub.Runner.Worker.Dap
 
         // When true, skip tunnel relay startup (unit tests only)
         internal bool SkipTunnelRelay { get; set; }
+
+        // When true, skip the public websocket bridge and expose the raw DAP
+        // listener directly on the configured tunnel port (unit tests only).
+        internal bool SkipWebSocketBridge { get; set; }
 
         // Synchronization for step execution
         private TaskCompletionSource<DapCommand> _commandTcs;
@@ -108,6 +113,7 @@ namespace GitHub.Runner.Worker.Dap
             _state == DapSessionState.Running;
 
         internal DapSessionState State => _state;
+        internal int InternalDapPort => (_listener?.LocalEndpoint as IPEndPoint)?.Port ?? 0;
 
         public override void Initialize(IHostContext hostContext)
         {
@@ -133,9 +139,22 @@ namespace GitHub.Runner.Worker.Dap
             _jobContext = jobContext;
             _readyTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
-            _listener = new TcpListener(IPAddress.Loopback, debuggerConfig.Tunnel.Port);
+            var dapPort = SkipWebSocketBridge ? debuggerConfig.Tunnel.Port : 0;
+            _listener = new TcpListener(IPAddress.Loopback, dapPort);
             _listener.Start();
-            Trace.Info($"DAP debugger listening on {_listener.LocalEndpoint}");
+            if (SkipWebSocketBridge)
+            {
+                Trace.Info($"DAP debugger listening on {_listener.LocalEndpoint}");
+            }
+            else
+            {
+                Trace.Info($"Internal DAP debugger listening on {_listener.LocalEndpoint}");
+                _webSocketBridge = new WebSocketDapBridge(
+                    HostContext.GetTrace("DapWebSocketBridge"),
+                    debuggerConfig.Tunnel.Port,
+                    InternalDapPort);
+                _webSocketBridge.Start();
+            }
 
             // Start Dev Tunnel relay so remote clients reach the local DAP port.
             // The relay is torn down explicitly in StopAsync (after the DAP session
@@ -274,6 +293,22 @@ namespace GitHub.Runner.Worker.Dap
                     _tunnelRelayHost = null;
                 }
 
+                if (_webSocketBridge != null)
+                {
+                    Trace.Info("Stopping WebSocket DAP bridge");
+                    var disposeTask = _webSocketBridge.DisposeAsync().AsTask();
+                    if (await Task.WhenAny(disposeTask, Task.Delay(5_000)) != disposeTask)
+                    {
+                        Trace.Warning("WebSocket DAP bridge dispose timed out after 5s");
+                    }
+                    else
+                    {
+                        Trace.Info("WebSocket DAP bridge stopped");
+                    }
+
+                    _webSocketBridge = null;
+                }
+
                 CleanupConnection();
 
                 // Cancel the connection loop first so AcceptTcpClientAsync unblocks
@@ -315,6 +350,7 @@ namespace GitHub.Runner.Worker.Dap
             _connectionLoopTask = null;
             _loopCts?.Dispose();
             _loopCts = null;
+            _webSocketBridge = null;
         }
 
         public async Task OnStepStartingAsync(IStep step)
--- a/src/Runner.Worker/Dap/WebSocketDapBridge.cs
+++ b/src/Runner.Worker/Dap/WebSocketDapBridge.cs
@@ -13,7 +13,7 @@ using GitHub.Runner.Common;
 
 namespace GitHub.Runner.Worker.Dap
 {
-    internal sealed class WebSocketDapBridge : RunnerService
+    internal sealed class WebSocketDapBridge : RunnerService, IWebSocketDapBridge
     {
         internal enum IncomingStreamPrefixKind
         {
@@ -37,7 +37,6 @@ namespace GitHub.Runner.Worker.Dap
 
         private int _listenPort;
         private int _targetPort;
-        private bool _configured;
 
         private TcpListener _listener;
         private CancellationTokenSource _loopCts;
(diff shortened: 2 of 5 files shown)


src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+using System.Net.WebSockets;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.Runner.Common;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    internal sealed class WebSocketDapBridge : IAsyncDisposable
```
TingluoHuang: do you want to wrap this as `IRunnerService`, so you don't have to pass in a tracer.

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+            _acceptLoopTask = AcceptLoopAsync(_loopCts.Token);
+
+            _trace.Info($"WebSocket DAP bridge listening on {_listener.LocalEndpoint} -> 127.0.0.1:{_targetPort}");
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            try
+            {
+                _loopCts?.Cancel();
```
TingluoHuang: we probably don't need to try-catch for `Cancel()`

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+
+            _listener = new TcpListener(IPAddress.Loopback, _listenPort);
+            _listener.Start();
+            _loopCts = new CancellationTokenSource();
+            _acceptLoopTask = AcceptLoopAsync(_loopCts.Token);
+
+            _trace.Info($"WebSocket DAP bridge listening on {_listener.LocalEndpoint} -> 127.0.0.1:{_targetPort}");
+        }
+
+        public async ValueTask DisposeAsync()
```
TingluoHuang: do you want to explicit shutdown the `WebSocketDapBridge` from the caller?
ex: we can just make a method call `public async Task Shutdown()`

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+                // best effort during shutdown
+            }
+
+            try
+            {
+                _listener?.Stop();
+            }
+            catch
+            {
+                // best effort during shutdown
```
TingluoHuang: do we want to log error for these?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+                {
+                    client = await _listener.AcceptTcpClientAsync(cancellationToken);
+                    client.NoDelay = true;
+                    await HandleClientAsync(client, cancellationToken);
+                }
+                catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+                {
+                    break;
+                }
+                catch (ObjectDisposedException) when (cancellationToken.IsCancellationRequested)
```
TingluoHuang: we shouldn't get any ObjectDisposedException

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+                {
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    client?.Dispose();
```
TingluoHuang, line 121: i would move client?.Dispose() into a `final{}` after all the catch

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+                    _trace.Warning($"WebSocket DAP bridge connection error ({ex.GetType().Name})");
+                    _trace.Error(ex);
```
TingluoHuang:
```suggestion
                    _trace.Error($"WebSocket DAP bridge connection error");
                    _trace.Error(ex);
```

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,812 @@
...
+                }
+            }
+
+            _trace.Info("WebSocket DAP bridge accept loop ended");
+        }
+
+        private async Task HandleClientAsync(TcpClient incomingClient, CancellationToken cancellationToken)
+        {
+            using (incomingClient)
```
TingluoHuang: we should let the caller dispose the object.

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,828 @@
...
+
+            if (!headers.TryGetValue("Sec-WebSocket-Version", out var webSocketVersion) ||
+                !string.Equals(webSocketVersion.Trim(), "13", StringComparison.Ordinal))
+            {
+                Trace.Warning("Rejected WebSocket request with unsupported version");
+                await WriteHttpErrorAsync(stream, (HttpStatusCode)426, "Unsupported WebSocket version. Expected: 13.", cancellationToken);
+                return null;
+            }
+
+            var webSocketKey = headers["Sec-WebSocket-Key"];
```
salmanmkc, line 277: `Sec-WebSocket-Key` like this is just for now right, cause at the moment its' untrusted right?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,828 @@
...
+                        {
+                            break;
+                        }
+
+                        if (result.Count > 0)
+                        {
+                            if (messageStream.Length + result.Count > MaxInboundMessageSize)
+                            {
+                                Trace.Warning($"WebSocket message exceeds maximum allowed size of {MaxInboundMessageSize} bytes, closing connection");
+                                await source.CloseAsync(
```
salmanmkc, line 328: Could we cancel `sessionCts` *before* initiating the close, or signal the session teardown path in `HandleClientAsync` to handle the graceful close instead?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,828 @@
...
+        private async Task<WebSocket> AcceptWebSocketAsync(NetworkStream stream, CancellationToken cancellationToken)
+        {
+            var initialBytes = await ReadInitialBytesAsync(stream, cancellationToken);
+            if (initialBytes == null || initialBytes.Length == 0)
+            {
+                return null;
+            }
+
+            var prefixKind = ClassifyIncomingStreamPrefix(initialBytes);
+            if (prefixKind == IncomingStreamPrefixKind.PreUpgradedWebSocket)
```
salmanmkc, line 197: this probs for dev tunnels right? jsut thinking something could easily skip e.g. by sending like `0x81 0x85 ...`, maybe something like a flag `allowPreUpgraded: true` to `Configure()`, and the
bridge only accepts that byte-sniff path when the flag, that still has a limitation though if the relay is active but a bit more secure at least. I think tho if doing secrets or diff ports in the future it would address this anyway

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,828 @@
...
+                        var proxyToken = sessionCts.Token;
+                        var wsToTcpTask = PumpWebSocketToTcpAsync(webSocket, dapStream, proxyToken);
+                        var tcpToWsTask = PumpTcpToWebSocketAsync(dapStream, webSocket, proxyToken);
+
+                        var completedTask = await Task.WhenAny(wsToTcpTask, tcpToWsTask);
+                        sessionCts.Cancel();
+
+                        try
+                        {
+                            await completedTask;
```
salmanmkc: can we skip the re-await of `completedTask` and go straight to `Task.WhenAll` after `sessionCts.Cancel()`, currently if `completedTask` faults with something other than `OperationCanceledException` (e.g., `IOException` from a peer disconnect), this rethrows and we never reach the `Task.WhenAll` below?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,828 @@
...
+                    var valueStart = line.IndexOf(':') + 1;
+                    if (int.TryParse(line.Substring(valueStart).Trim(), out var parsedLength))
+                    {
+                        contentLength = parsedLength;
+                        break;
+                    }
+                }
+            }
+
+            if (contentLength < 0)
```
salmanmkc, line 413: in DAP isn't there a big error if the content-length is wrong, so should we not clear the entire buffer so the next message starts clean, or (probably better) treat this as a fatal protocol error and tear down the session?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
rentziass: We're actually authed at the devtunnel layer, this is just meant to be protocol-compliant, that's part of the initial handshake.

src/Test/L0/Worker/WebSocketDapBridgeL0.cs
```diff
@@ -0,0 +1,275 @@
...
+        [Fact]
+        [Trait("Level", "L0")]
+        [Trait("Category", "Worker")]
+        public async Task BridgeRejectsNonWebSocketRequests()
+        {
+            using var hc = CreateTestContext();
+            var bridgePort = GetFreePort();
+
+            var bridge = new WebSocketDapBridge();
+            bridge.Initialize(hc);
```
salmanmkc, line 139: should we try catch the configure start and start?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
rentziass: I got this all in place while I was failing to figure out connectivity going through dev tunnels yeah 😄 I think this is also covered by auth at dev tunnel layer?

src/Runner.Worker/Dap/WebSocketDapBridge.cs
salmanmkc: yeah that makes sense, tunnel layer handling auth is good. just thinking since we're going for protocol compliance anyway, rfc 6455 expects the key to be a base64 encoded 16 byte nonce so a quick format check might be nice to have. would also guard against any weird edge cases with cr/lf in the value messing with the response framing

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,819 @@
...
+            var requestLine = await ReadLineAsync(handshakeStream, cancellationToken);
+            if (string.IsNullOrEmpty(requestLine))
+            {
+                return null;
+            }
+
+            var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            while (true)
+            {
+                var line = await ReadLineAsync(handshakeStream, cancellationToken);
```
salmanmkc, line 230: individual header lines are capped at 8 KB but theres no cap on the count, so someone could just spam thousands of small headers within the handshake timeout. maybe a `maxHeaderCount` (64ish?) to be safe?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -274,6 +292,22 @@ public async Task StopAsync()
                     _tunnelRelayHost = null;
                 }
 
+                if (_webSocketBridge != null)
+                {
+                    Trace.Info("Stopping WebSocket DAP bridge");
+                    var shutdownTask = _webSocketBridge.ShutdownAsync();
+                    if (await Task.WhenAny(shutdownTask, Task.Delay(5_000)) != shutdownTask)
```
salmanmkc, line 297: small thing, if the 5s timeout fires `shutdownTask` is still running but we drop the reference and null the bridge. if it eventually faults that goes unobserved. might be nice to have a `ContinueWith` to log if it fails or something, just so we dont silently lose errors

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,819 @@
...
+                    await destination.SendAsync(
+                        new ArraySegment<byte>(messageBody),
+                        WebSocketMessageType.Text,
+                        endOfMessage: true,
+                        cancellationToken);
+                }
+            }
+        }
+
+        private static bool TryParseDapMessage(List<byte> buffer, out byte[] messageBody)
```
salmanmkc, line 386: nit: this gets allocated on every call to `TryParseDapMessage`, could just be a `private static readonly byte[]` field

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -133,9 +139,21 @@ public async Task StartAsync(IExecutionContext jobContext)
...
             _listener.Start();
-            Trace.Info($"DAP debugger listening on {_listener.LocalEndpoint}");
+            if (SkipWebSocketBridge)
+            {
+                Trace.Info($"DAP debugger listening on {_listener.LocalEndpoint}");
+            }
+            else
+            {
+                Trace.Info($"Internal DAP debugger listening on {_listener.LocalEndpoint}");
+                _webSocketBridge = new WebSocketDapBridge();
```
TingluoHuang: you can do `HostContext.CreateService<IWebSocketDapBridge>()` which will create you new instance everyone instead of returned the signleton created via `HostContext.GetService`

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,855 @@
...
+                            // peer disconnected while unwinding
+                        }
+                        catch (WebSocketException)
+                        {
+                            // peer disconnected while unwinding
+                        }
+                        catch (InvalidOperationException ex)
+                        {
+                            Trace.Warning($"DAP protocol error: {ex.Message}");
+                        }
```
TingluoHuang:
```suggestion
                        catch (Exception ex)
                        {
                            Trace.Warning($"DAP protocol error: {ex}");
                        }
```

src/Runner.Worker/Dap/WebSocketDapBridge.cs
TingluoHuang, line 1: is there any existing library doing this?
it feels like we are writing too much code. 🤔

TingluoHuang: Here is the reply i got from copilot. the `watsonwebserver` and `System.IO.Pipelines` are interesting points.

Here's a thorough analysis of the `WebSocketDapBridge.cs` implementation and what public libraries could replace parts of it.

---

## What the File Does

[`WebSocketDapBridge.cs`](https://github.com/actions/runner/pull/4328) implements a **WebSocket ↔ TCP proxy** specifically for the Debug Adapter Protocol (DAP). It:

1. **Listens** on a TCP port for incoming connections
2. **Performs the WebSocket HTTP Upgrade handshake** manually (RFC 6455)
3. **Detects pre-upgraded WebSocket streams** (e.g. from Dev Tunnels that already stripped the HTTP layer)
4. **Pumps frames bidirectionally**: WS frames → DAP `Content-Length`-framed TCP messages, and vice versa

---

## What Could Be Replaced by Public Libraries

### ✅ **WebSocket Accept Handshake** (the largest custom section)

The most substantial custom code is the HTTP Upgrade handshake (~150 lines): header parsing, `Sec-WebSocket-Accept` computation, rejection responses.

**This can be fully replaced by [`Microsoft.AspNetCore.WebSockets`](https://www.nuget.org/packages/Microsoft.AspNetCore.WebSockets/)**, specifically via a minimal Kestrel host or even via `HttpListener`-based upgrade:

```csharp
// With a Kestrel/minimal API host — the handshake becomes 1 line:
await context.WebSockets.AcceptWebSocketAsync();
```

However, **the project intentionally avoids ASP.NET Core** (this is `Runner.Worker`, a lean console process), and `WebSocket.CreateFromStream` (which they already use at the bottom of the handshake) is BCL, not ASP.NET. So the handshake must be done manually — or via a library.

**Best drop-in alternative: [`WatsonWebsocket`](https://www.nuget.org/packages/WatsonWebsocket/)** or **[`Fleck`](https://www.nuget.org/packages/Fleck/)** — both handle the server-side WS handshake without requiring ASP.NET Core.

---

### ✅ **`ReplayableStream`** (prefix-replay buffer)

This is a small custom `Stream` subclass that replays the first 4 bytes after they've been peeked to detect the protocol. This pattern is sometimes called a **"peekable" or "pre-read" stream**.

**Libraries with this built-in:**
- [`System.IO.Pipelines`](https://www.nuget.org/packages/System.IO.Pipelines/) (BCL, .NET 6+) — `PipeReader` supports peeking/examining bytes before consuming, eliminating the need for a custom `ReplayableStream` entirely. This is likely the most idiomatic .NET replacement.

---

### ✅ **DAP `Content-Length` Message Framing** (`TryParseDapMessage`)

The DAP header parsing loop is straightforward but custom. There is no widely-used dedicated NuGet package for DAP framing, but:

- **[`Microsoft.VisualStudio.Shared.VSCodeDebugProtocol`](https://www.nuget.org/packages/Microsoft.VisualStudio.Shared.VSCodeDebugProtocol/)** or **[`OmniSharp/omnisharp-roslyn`](https://github.com/OmniSharp/omnisharp-roslyn)** include DAP protocol serializers, though not low-level TCP framing.
- [`System.IO.Pipelines`](https://www.nuget.org/packages/System.IO.Pipelines/) again — `SequenceReader<byte>` makes the header scanning loop cleaner and allocation-free.

---

### ✅ **WS→TCP Pumping** (`PumpWebSocketToTcpAsync` / `PumpTcpToWebSocketAsync`)

This is generic bidirectional stream proxying. The current implementation uses `List<byte>` as a buffer, which is allocation-heavy.

**Better alternatives:**
- [`System.IO.Pipelines`](https://www.nuget.org/packages/System.IO.Pipelines/) — avoids `List<byte>` and `ToArray()` allocations entirely; this is what ASP.NET Core itself uses internally for WebSocket proxying.
- [`YARP` (Yet Another Reverse Proxy)](https://www.nuget.org/packages/Yarp.ReverseProxy/) — overkill here, but it does contain robust WebSocket forwarding middleware.

---

## Summary Table

| Section of `WebSocketDapBridge.cs` | Publicly Available Alternative |
|---|---|
| HTTP Upgrade handshake parsing | [`Fleck`](https://www.nuget.org/packages/Fleck/), [`WatsonWebsocket`](https://www.nuget.org/packages/WatsonWebsocket/), or `Microsoft.AspNetCore.WebSockets` (needs Kestrel) |
| `WebSocket.CreateFromStream` | ✅ Already uses BCL (built into .NET 6+, no library needed) |
| `ReplayableStream` (prefix peek) | [`System.IO.Pipelines`](https://www.nuget.org/packages/System.IO.Pipelines/) `PipeReader.AdvanceTo` |
| DAP `Content-Length` framing | `System.IO.Pipelines` + `SequenceReader<byte>` |
| Bidirectional WS↔TCP pump | `System.IO.Pipelines`, or `YARP` for a full proxy scenario |
| Protocol sniffing (`ClassifyIncomingStreamPrefix`) | No public library — **keep as-is**, this is custom to the Dev Tunnels use case |

---

## Practical Recommendation

The most impactful and well-scoped replacement would be:

1. **Replace `ReplayableStream` + the pump's `List<byte>` buffer** with [`System.IO.Pipelines`](https://www.nuget.org/packages/System.IO.Pipelines/) — it's a BCL package, zero external dependency risk, and would make the DAP framing and WS pumping both faster and allocation-free.

2. **Keep the handshake manual** OR adopt `Fleck`/`WatsonWebsocket` — but adopting a third-party WS library adds a dependency to a security-sensitive component (a runner handling GitHub credentials), so the manual implementation is arguably the safer call here.

3. **The protocol sniffing logic** (`ClassifyIncomingStreamPrefix`) is unique to the Dev Tunnels pre-upgraded stream scenario and has **no public library equivalent** — it should stay custom.

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,839 @@
...
+
+            var handshakeStream = new ReplayableStream(stream, initialBytes);
+            var requestLine = await ReadLineAsync(handshakeStream, cancellationToken);
+            if (string.IsNullOrEmpty(requestLine))
+            {
+                return null;
+            }
+
+            var headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            while (true)
```
TingluoHuang:
```suggestion
            while (!cancellationToken.IsCancellationRequested)
```

src/Runner.Worker/Dap/WebSocketDapBridge.cs
```diff
@@ -0,0 +1,839 @@
...
+                                    WebSocketCloseStatus.MessageTooBig,
+                                    $"Message exceeds {MaxInboundMessageSize} byte limit",
+                                    CancellationToken.None);
+                                return;
+                            }
+
+                            messageStream.Write(buffer, 0, result.Count);
+                        }
+                    }
+                    while (!result.EndOfMessage);
```
TingluoHuang:
```suggestion
                    while (!result.EndOfMessage && !cancellationToken.IsCancellationRequested);
```

TingluoHuang (approved):

PR #4347, opened 2026-04-14
fix: expand commit hash regex to support SHA-256 (64-char) hashes

yaananth:
Fixes https://github.com/github/actions-runtime/issues/5475
Fixes https://github.com/github/actions-runtime/issues/5483
Fixes https://github.com/github/actions-runtime/issues/5484

## Summary
Expand runner commit-hash handling to support both SHA-1 (40-char) and SHA-256 (64-char) commit hashes, preparing the runner for Git's SHA-256 transition.

## Changes
- **`WellKnownRegularExpressions.cs`**: Update the commit-hash regex from `\b[0-9a-f]{40}\b` to `\b(?:[0-9a-f]{40}|[0-9a-f]{64})\b`; add `CommitHash` as the canonical key while keeping `SHA1` as a backward-compatible alias.
- **`RepositoryPlugin.cs` (v1.0/v1.1)**: Update repository plugin callers to use `WellKnownRegularExpressions.CommitHash` instead of the legacy `SHA1` key; remove the unused v1.0 `_validSha1` field.
- **`WellKnownRegularExpressionsL0.cs`**: Add coverage for 40-char SHA-1, 64-char SHA-256, 63/65-char boundaries, mixed-case matching, alias compatibility, and unknown keys.
- **`ConstantGenerationL0.cs`**: Fold in the former #4348 assertion update so generated `CommitHash` may be either 40-char SHA-1 or 64-char SHA-256 and is hex-only.
- **`ActionManagerL0.cs`**: Add validation for 64-character resolved SHAs in action archive cache and symlink cache paths.

## Validation
- Focused local verification passed:
  - `dotnet test src/Test/Test.csproj --filter "FullyQualifiedName~WellKnownRegularExpressionsL0|FullyQualifiedName~ConstantGenerationL0" --no-restore`
  - `dotnet test src/Test/Test.csproj --filter "FullyQualifiedName~ActionManagerL0" --no-restore --logger "console;verbosity=minimal"`
- CI is green on the latest commit.

## PR consolidation
- Supersedes and includes the change from #4348, which has been closed.

## Context
- Parent epic: SHA-256 Support for Git Repositories
- Related: [SHA-256 Migration Guide](https://docs.google.com/document/d/1Ei2NUABeY9hIZdA9SaaapIW9KXsAay5mW6gZjVANUDg)

---
*Automated by `project-board-agents-plugin` via `/project-board-agents:lead`*
*Board: https://github.com/orgs/github/projects/24272/views/1*
*Item: [P0] Fix core SHA-1 regex in actions/runner WellKnownRegularExpressions.cs*
*Run: sha256-p0-runner-00050*


src/Sdk/DTPipelines/Pipelines/Expressions/WellKnownRegularExpressions.cs
```diff
@@ -46,9 +48,9 @@ public static Lazy<Regex> GetRegex(String regexType)
             )
         );
 
-        // 40 hex characters
-        private static readonly Lazy<Regex> s_validSha1 = new Lazy<Regex>(() => new Regex(
-            @"\b[0-9a-f]{40}\b",
+        // 40 or 64 hex characters (SHA-1 or SHA-256 commit hash)
+        private static readonly Lazy<Regex> s_validCommitHash = new Lazy<Regex>(() => new Regex(
```
salmanmkc, line 52: The regex change looks correct, but `WellKnownRegularExpressions` currently has no dedicated unit tests.

Would be good if we could add a  test class (e.g. `src/Test/L0/Sdk/WellKnownRegularExpressionsL0.cs` alongside the existing `ExpressionParserL0.cs`) with these minimum:
- 40-char hex → matches (regression guard)
- 64-char hex → matches (the new case this PR adds)
- 63-char and 65-char hex → do NOT match (boundary)
- Mixed-case 64-char → matches (`IgnoreCase` still works)
>

Would be good to avoid a regression in the future, in-case a future change breaks 64-char support to avoid it going undetected since nothing in CI directly tests it.

src/Sdk/DTPipelines/Pipelines/Expressions/WellKnownRegularExpressions.cs
```diff
@@ -24,7 +25,8 @@ public static Lazy<Regex> GetRegex(String regexType)
                 case IPv4Address:
                     return s_validIPv4Address;
                 case SHA1:
```
TingluoHuang, line 27: do we need to remove the SHA1 and update all the caller to use the CommitHash?

salmanmkc (approved):

gordolara789-tech (commented):
Exit

PR #4354, opened 2026-04-17
Pin all GitHub Actions versions using `pinact`

borchero:
# Motivation

It is generally good practice (see e.g. [this](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/) post) to pin actions to their commit SHAs to prevent the execution of malicious code.

This change would not only make the use of actions more secure in this repo, but also allow to more easily maintain forks in organizations where SHA pinning is enforced.

# Changes

- Pin all actions using [pinact](https://github.com/suzuki-shunsuke/pinact)
- Replace `master` reference for `actions/create-release` with the latest release tag

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,7 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v6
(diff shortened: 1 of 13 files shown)


salmanmkc:
Much appreciated, however I think we'll stick to floating tags here. You're right that it's best practice as called out in our documentation, in this instance though since we want to essentially be dogfooding our own actions, this will help catch issues earlier in our workflows.

PR #4383, opened 2026-04-28
Update setup job starting logs

GitPaulo:
# Overview

We are cleaning up setup job logs and this is a first step to restore parity to LHR logs and improve on it further in the future.

Issue: https://github.com/github/hosted-compute/issues/3955

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -76,27 +76,26 @@ namespace GitHub.Runner.Worker
                     context.Output($"Current runner version: '{BuildConstants.RunnerPackage.Version}'");
 
                     var setting = HostContext.GetService<IConfigurationStore>().GetSettings();
-                    // print runner info for lhr and self-hosted runners, standard uses PoolId 0
-                    if (setting.PoolId > 0)
+                    var credFile = HostContext.GetConfigFile(WellKnownConfigFile.Credentials);
+                    var credData = File.Exists(credFile) ? IOUtil.LoadObject<CredentialData>(credFile) : null;
+                    // self-hosted runner is the only runner type using OAuth, can be identified via clientId
+                    if (credData != null &&
+                        credData.Data.TryGetValue("clientId", out _))
                     {
                         context.Output($"Runner name: '{setting.AgentName}'");
-                        if (!string.IsNullOrEmpty(setting.PoolName))
+                        // use system variable for group name since self-hosted runners can be renamed
+                        if (message.Variables.TryGetValue("system.runnerGroupName", out VariableValue runnerGroupName))
                         {
-                            context.Output($"Runner group name: '{setting.PoolName}'");
+                            context.Output($"Runner group name: '{runnerGroupName.Value}'");
                         }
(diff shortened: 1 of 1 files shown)


src/Runner.Worker/JobExtension.cs
```diff
@@ -76,19 +76,22 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     context.Output($"Current runner version: '{BuildConstants.RunnerPackage.Version}'");
 
                     var setting = HostContext.GetService<IConfigurationStore>().GetSettings();
+                    // print runner info for lhr and self-hosted runners, standard uses PoolId 0
+                    if (setting.PoolId > 0)
+                    {
+                        context.Output($"Runner name: '{setting.AgentName}'");
+                        context.Output($"Runner group name: '{setting.PoolName}'");
```
GitPaulo: Question, should we use variables system here? 
`message.Variables["system.runnerGroupName"]` was used before but now PoolName will be available, can we trust it to be up to date?

src/Runner.Worker/JobExtension.cs
```diff
@@ -76,19 +76,25 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     context.Output($"Current runner version: '{BuildConstants.RunnerPackage.Version}'");
 
                     var setting = HostContext.GetService<IConfigurationStore>().GetSettings();
+                    // print runner info for lhr and self-hosted runners, standard uses PoolId 0
+                    if (setting.PoolId > 0)
+                    {
+                        context.Output($"Runner name: '{setting.AgentName}'");
+                        if (!string.IsNullOrEmpty(setting.PoolName))
+                        {
+                            context.Output($"Runner group name: '{setting.PoolName}'");
```
TingluoHuang: we probably still want to use `system.runnerGroupName` for self-hosted runner since you can rename the runner group.

src/Runner.Worker/JobExtension.cs
```diff
@@ -76,19 +76,25 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                        }
+                    }
+
                     var credFile = HostContext.GetConfigFile(WellKnownConfigFile.Credentials);
                     if (File.Exists(credFile))
                     {
                         var credData = IOUtil.LoadObject<CredentialData>(credFile);
```
TingluoHuang: i would suggest just add the following after the `if (credData != null &&  credData.Data.TryGetValue("clientId", out var clientId))`
```
else if  (setting.PoolId > 0 && !string.IsNullOrEmpty(setting.PoolName) && !string.IsNullOrEmpty(setting.AgentName))
{
    context.Output($"Runner name: '{setting.AgentName}'");
    context.Output($"Runner group name: '{setting.PoolName}'");
}
```

GitPaulo:
Note: L0 tests are flaky, I can't re-run them as I'm committing from a fork. It should be 🆗 to ignore.

TingluoHuang (approved):

gordolara789-tech (commented):
Exito

PR #4403, opened 2026-05-05
Move dap setup to setup job step

rentziass:
This adds the DAP server setup to the Setup job step, to both help with messaging for users and keep the step as in progress to communicate we're waiting on a debugger to connect.

We also add a final pause during the Complete job step to allow one final inspection after all steps have completed.
CleanShot 2026-05-05 at 16 21 37@2x
CleanShot 2026-05-05 at 16 22 27@2x

https://github.com/github/c2c-actions/issues/9828

### DIFF
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -1320,6 +1320,13 @@ namespace GitHub.Runner.Worker.Dap
                 _commandTcs = new TaskCompletionSource<DapCommand>(TaskCreationOptions.RunContinuationsAsynchronously);
             }
 
+            // If cancellation already fired before we created the new TCS,
+            // the registration callback targeted the old one. Unblock now.
+            if (cancellationToken.IsCancellationRequested)
+            {
+                _commandTcs.TrySetResult(DapCommand.Disconnect);
+            }
+
             Trace.Info("Waiting for debugger command...");
 
             var command = await _commandTcs.Task;
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -243,6 +243,24 @@ namespace GitHub.Runner.Worker.Dap
         {
             if (_state != DapSessionState.NotStarted)
             {
+                // Pause so the user can inspect final job state before we tear down.
+                if (IsActive && _pauseOnNextStep)
+                {
+                    try
+                    {
+                        var cancellationToken = _jobContext?.CancellationToken ?? CancellationToken.None;
+
+                        Trace.Info("Job completed — pausing for inspection");
+                        SendStoppedEvent("completed", "Job completed — inspect variables before the session ends.");
+
+                        await WaitForCommandAsync(cancellationToken);
+                    }
+                    catch (Exception ex)
+                    {
+                        Trace.Warning($"DAP job-completed pause error: {ex.Message}");
+                    }
+                }
+
                 try
                 {
                     OnJobCompleted();
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -243,17 +243,21 @@ namespace GitHub.Runner.Worker.Dap
         {
             if (_state != DapSessionState.NotStarted)
             {
-                // Pause so the user can inspect final job state before we tear down.
+                // Pause so the user can inspect final job state before we tear down,
+                // but only if the user was stepping through (not if they hit continue).
                 if (IsActive && _pauseOnNextStep)
                 {
                     try
                     {
-                        var cancellationToken = _jobContext?.CancellationToken ?? CancellationToken.None;
+                        if (_jobContext != null)
+                        {
+                            var cancellationToken = _jobContext.CancellationToken;
 
-                        Trace.Info("Job completed — pausing for inspection");
-                        SendStoppedEvent("completed", "Job completed — inspect variables before the session ends.");
+                            Trace.Info("Job completed — pausing for inspection");
+                            SendStoppedEvent("completed", "Job completed — inspect variables before the session ends.");
 
-                        await WaitForCommandAsync(cancellationToken);
+                            await WaitForCommandAsync(cancellationToken);
+                        }
                     }
                     catch (Exception ex)
                     {
--- a/src/Runner.Worker/Dap/DapDebugger.cs
+++ b/src/Runner.Worker/Dap/DapDebugger.cs
@@ -251,12 +251,10 @@ namespace GitHub.Runner.Worker.Dap
                     {
                         if (_jobContext != null)
                         {
-                            var cancellationToken = _jobContext.CancellationToken;
-
                             Trace.Info("Job completed — pausing for inspection");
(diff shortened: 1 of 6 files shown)


src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -243,6 +243,24 @@ public async Task OnJobCompletedAsync()
         {
             if (_state != DapSessionState.NotStarted)
             {
+                // Pause so the user can inspect final job state before we tear down.
+                if (IsActive && _pauseOnNextStep)
+                {
+                    try
+                    {
+                        var cancellationToken = _jobContext?.CancellationToken ?? CancellationToken.None;
```
salmanmkc: this relies on `IsActive` always implying `_jobContext != null` which is true right now but feels fragile. if that invariant breaks later this silently hangs the runner. can we just bail early?

```suggestion
                        if (_jobContext == null)
                        {
                            Trace.Warning("No job context available, skipping completion pause.");
                            break;
                        }
                        var cancellationToken = _jobContext.CancellationToken;
```

src/Runner.Worker/JobExtension.cs
```diff
@@ -782,6 +855,22 @@ public async Task FinalizeJob(IExecutionContext jobContext, Pipelines.AgentJobRe
                 }
                 finally
                 {
+                    // Pause for debugger inspection, then tear down the DAP session.
+                    // OnJobCompletedAsync pauses first, then sends terminated/exited
+                    // events and stops the transport.
+                    if (_dapDebugger != null)
```
salmanmkc, line 850: nit: init-failure path nulls `_dapDebugger` after cleanup but this doesn't. doesn't matter since `JobExtension` is per-job and `StopAsync()` handles the real teardown, just looks inconsistent

src/Runner.Worker/JobExtension.cs
```diff
@@ -481,6 +483,53 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+
+                        try
+                        {
+                            _dapDebugger = HostContext.GetService<IDapDebugger>();
+                            await _dapDebugger.StartAsync(jobContext);
+                        }
+                        catch (Exception ex)
+                        {
+                            Trace.Error($"Failed to start DAP debugger: {ex.Message}");
+                            AddDebuggerConnectionTelemetry(jobContext, $"Failed: {ex.Message}");
```
salmanmkc: `ex.Message` going into `JobTelemetry`, probably fine since the host token goes in auth headers not URLs, but I don't see `HostToken` registered with `SecretMasker` either. safer to just log the type:

```suggestion
                            AddDebuggerConnectionTelemetry(jobContext, $"Failed: {ex.GetType().Name}");
```

full message is already in `Trace.Error`

salmanmkc (commented):
couple tests that'd be nice to have:
- cancellation during `WaitForCommandAsync` (not just pre-cancelled)
- exception in `OnJobCompletedAsync` during FinalizeJob

src/Runner.Worker/Dap/DapDebugger.cs
rentziass: `break` doesn't compile here but adding the check 👍

salmanmkc (dismissed):
Looks good to me, OnJobCompletedAsync exception test as a nice follow-up if you want, but non-blocking
edit: ignore this actually, i'm wrong about the test not being there, lgtm!

src/Runner.Worker/JobExtension.cs
```diff
@@ -481,6 +483,53 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                    // "Set up job" so the step stays in-progress while we wait.
+                    if (jobContext.Global.Debugger?.Enabled == true)
+                    {
+                        Trace.Info("Debugger enabled — starting inside Set up job");
+                        context.Output("Starting debugger…");
+
+                        try
+                        {
+                            _dapDebugger = HostContext.GetService<IDapDebugger>();
+                            await _dapDebugger.StartAsync(jobContext);
```
TingluoHuang, line 497: should we pass in the step level context instead of the `jobContext`?

src/Runner.Worker/JobExtension.cs
```diff
@@ -501,12 +550,36 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                 }
                 finally
                 {
+                    // If InitializeJob failed after the debugger was started,
+                    // clean it up here since FinalizeJob won't run.
+                    if (context.Result != null && _dapDebugger != null)
+                    {
+                        try
+                        {
+                            await _dapDebugger.OnJobCompletedAsync();
```
TingluoHuang: it's a little bit odd we are calling `OnJobCompletedAsync` during setup job.

also, depends on `context.Result != null` might be fragile.

src/Runner.Worker/JobExtension.cs
```diff
@@ -481,6 +483,53 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                        {
+                            _dapDebugger = HostContext.GetService<IDapDebugger>();
+                            await _dapDebugger.StartAsync(jobContext);
+                        }
+                        catch (Exception ex)
+                        {
+                            Trace.Error($"Failed to start DAP debugger: {ex.Message}");
+                            AddDebuggerConnectionTelemetry(jobContext, $"Failed: {ex.GetType().Name}");
+                            context.Error("Failed to start debugger.");
+                            context.Result = TaskResult.Failed;
```
TingluoHuang: `context.Result = TaskResult.Failed;`

you don't really need this since the `catch()` from outside will set the result to failed.

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -243,6 +243,28 @@ public async Task OnJobCompletedAsync()
...
             {
+                // Pause so the user can inspect final job state before we tear down,
+                // but only if the user was stepping through (not if they hit continue).
+                if (IsActive && _pauseOnNextStep)
+                {
+                    try
+                    {
+                        if (_jobContext != null)
+                        {
+                            var cancellationToken = _jobContext.CancellationToken;
```
TingluoHuang: do we need the extra `var cancellationToken = _jobContext.CancellationToken;`?

src/Runner.Worker/JobExtension.cs
rentziass: The debugger needs jobs cancellation, Global.Debugger and variable provider from the job context, plus I think it needs a context that outlives the setup step? This isn't new btw, it just moved into the step.

TingluoHuang: the `Global` object is shared across all context (step + job).
you were using jobcontext since the code in jobrunner.cs doesn't have a step context.

rentziass: But also we hook up to jobcontext cancelation [in the debugger](https://github.com/actions/runner/blob/89b6b32e3e0cefb241222a377586cdb9f4449ee2/src/Runner.Worker/Dap/DapDebugger.cs?plain=1#L169-L174), if we used the step context wouldn't that be completed when Setup job completes?

TingluoHuang (approved):

PR #4406, opened 2026-05-06
Normalize Windows workspace path casing to NTFS canonical form

salmanmkc:
## Problem

On Windows self-hosted runners, the path casing for `GITHUB_WORKSPACE`, `RUNNER_TEMP`, and other derived paths depends on how the runner process was started. If a scheduled task or script uses a lowercase drive letter (e.g. `c:\actions-runner\run.cmd` instead of `C:\actions-runner\run.cmd`), the runner propagates that casing into all environment variables.

However, tools like git resolve paths to their NTFS canonical casing. This creates mismatches that break case-sensitive operations — most notably `includeIf.gitdir:` in `actions/checkout@v6`, which does exact string matching and fails to load credentials when the casing differs.

Tracked in https://github.com/actions/checkout/issues/2345

## Root cause

Windows is case-insensitive at the filesystem level, so `c:\actions-runner` and `C:\actions-runner` refer to the same directory. But the runner uses whatever casing the parent process provides, while git queries NTFS for the canonical casing. When these differ, any tool doing exact path comparison sees a mismatch.

GitHub-hosted runners aren't affected because their workspace paths already use canonical casing.

## Fix

Add `PathUtil.GetCanonicalPath()` which uses the Win32 `GetFinalPathNameByHandle` API to resolve a directory path to its NTFS canonical casing. This is called when resolving the runner root directory (`WellKnownDirectory.Root`), so all derived paths (workspace, temp, actions, tools, etc.) inherit the correct casing.

On non-Windows platforms, `GetCanonicalPath` is a no-op that returns the input unchanged.

## Why fix here instead of in individual actions?

Individual actions like `actions/checkout` can work around the issue (e.g. using `includeIf.gitdir/i:` for case-insensitive matching), but the root cause is the runner providing inconsistently-cased paths. Fixing it at the runner level prevents the entire class of problems for all actions.

## Changes

- `src/Runner.Sdk/Util/PathUtil.cs` — Add `GetCanonicalPath()` with P/Invoke to `GetFinalPathNameByHandle` (Windows only)
- `src/Runner.Common/HostContext.cs` — Normalize the root directory path through `GetCanonicalPath()`
- `src/Test/L0/Util/PathUtilL0.cs` — Tests for null/empty/nonexistent paths and drive letter normalization

## Testing

- L0 unit tests for `GetCanonicalPath` covering null, empty, nonexistent, and existing directory cases
- Windows-specific test verifying drive letter normalization

### DIFF
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -392,6 +392,7 @@ namespace GitHub.Runner.Common
 
                 case WellKnownDirectory.Root:
                     path = new DirectoryInfo(GetDirectory(WellKnownDirectory.Bin)).Parent.FullName;
+                    path = PathUtil.GetCanonicalPath(path);
                     break;
 
                 case WellKnownDirectory.Temp:
--- a/src/Runner.Common/HostContext.cs
+++ b/src/Runner.Common/HostContext.cs
@@ -64,6 +64,7 @@ namespace GitHub.Runner.Common
         private readonly List<ProductInfoHeaderValue> _userAgents = new() { new ProductInfoHeaderValue($"GitHubActionsRunner-{BuildConstants.RunnerPackage.PackageName}", BuildConstants.RunnerPackage.Version) };
         private CancellationTokenSource _runnerShutdownTokenSource = new();
         private object _perfLock = new();
+        private string _canonicalRootDirectory;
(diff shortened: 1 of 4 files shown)


salmanmkc:
This is not going to be released for some time.

PR #4416, opened 2026-05-12
Add infrastructure for background step support: thread safety, SDK models, and results plumbing

lokesh755:
## Add infrastructure for background step support

This PR lays the groundwork for background steps by adding the necessary infrastructure without changing step execution behavior.

### Thread safety
Adds `CollectionLock` to `GlobalContext` and wraps all concurrent access to shared collections (`JobTelemetry`, `DeprecatedNode20Actions`, etc.) with locks, preparing for multiple steps executing simultaneously.

### SDK step models
Introduces `WaitStep`, `WaitAllStep`, and `CancelStep` pipeline types alongside `StepType` enum extensions and `StepConverter` deserialization support.

### Timeline & results plumbing
Extends `TimelineRecord` with `IsBackground`, `StepType`, `WaitStepIds`, `CancelStepId`, and `ParallelGroupId` fields. Wires them through `ResultsHttpClient` and `JobServerQueue` merge logic.

### StepsContext
Refactors locking to use `CollectionLock` for thread-safe outcome/conclusion/output writes from concurrent background steps.

Related to 

https://github.com/github/actions-runtime/issues/5420
https://github.com/github/actions-runtime/issues/5421

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -85,6 +85,12 @@ namespace GitHub.Runner.Worker
         IExecutionContext CreateChild(Guid recordId, string displayName, string refName, string scopeName, string contextName, ActionRunStage stage, Dictionary<string, string> intraActionState = null, int? recordOrder = null, IPagingLogger logger = null, bool isEmbedded = false, List<Issue> embeddedIssueCollector = null, CancellationTokenSource cancellationTokenSource = null, Guid embeddedId = default(Guid), string siblingScopeName = null, TimeSpan? timeout = null);
         IExecutionContext CreateEmbeddedChild(string scopeName, string contextName, Guid embeddedId, ActionRunStage stage, Dictionary<string, string> intraActionState = null, string siblingScopeName = null);
 
+        // Background step deferral properties
+        Dictionary<string, string> DeferredOutputs { get; set; }
+        Dictionary<string, string> DeferredEnvironmentVariables { get; set; }
+        List<string> DeferredPrependPath { get; set; }
+        bool DeferOutcomeConclusion { get; set; }
+
         // logging
         long Write(string tag, string message);
         void QueueAttachFile(string type, string name, string filePath);
@@ -100,11 +106,18 @@ namespace GitHub.Runner.Worker
         void SetGitHubContext(string name, string value);
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
+
+        // Background step deferral flush methods
+        void FlushDeferredOutputs();
+        void FlushDeferredEnvironment();
+        void FlushDeferredOutcomeConclusion();
+
         void AddIssue(Issue issue, ExecutionContextLogOptions logOptions);
         void Progress(int percentage, string currentOperation = null);
         void UpdateDetailTimelineRecord(TimelineRecord record);
 
         void UpdateTimelineRecordDisplayName(string displayName);
+        void SetBackgroundStepMetadata(bool isBackground = false, string backgroundControlType = null, string[] backgroundControlStepIds = null, string parallelGroupId = null);
 
         // matchers
         void Add(OnMatcherChanged handler);
@@ -278,6 +291,12 @@ namespace GitHub.Runner.Worker
         }
 
         public List<string> StepEnvironmentOverrides { get; } = new List<string>();
+
+        // Background step deferral properties
+        public Dictionary<string, string> DeferredOutputs { get; set; }
+        public Dictionary<string, string> DeferredEnvironmentVariables { get; set; }
+        public List<string> DeferredPrependPath { get; set; }
+        public bool DeferOutcomeConclusion { get; set; }
 
         public override void Initialize(IHostContext hostContext)
         {
@@ -516,6 +535,11 @@ namespace GitHub.Runner.Worker
                     Annotations = new List<Annotation>()
                 };
 
+                // Populate background step metadata from timeline record fields
+                stepResult.IsBackground = _record.IsBackground;
+                stepResult.BackgroundControlType = _record.BackgroundControlType;
+                stepResult.BackgroundControlStepIds = _record.BackgroundControlStepIds;
+
                 _record.Issues?.ForEach(issue =>
                 {
                     var annotation = issue.ToAnnotation();
@@ -541,7 +565,10 @@ namespace GitHub.Runner.Worker
                         var annotation = issue.ToAnnotation();
                         if (annotation != null)
                         {
-                            Global.JobAnnotations.Add(annotation.Value);
+                            lock (Global.CollectionLock)
+                            {
+                                Global.JobAnnotations.Add(annotation.Value);
+                            }
                             if (annotation.Value.IsInfrastructureIssue && string.IsNullOrEmpty(Global.InfrastructureFailureCategory))
                             {
                                 Global.InfrastructureFailureCategory = issue.Category;
@@ -559,9 +586,20 @@ namespace GitHub.Runner.Worker
 
             _logger.End();
 
-            UpdateGlobalStepsContext();
+            if (!DeferOutcomeConclusion)
+            {
+                UpdateGlobalStepsContext();
+            }
 
             return Result.Value;
+        }
+
+        public void FlushDeferredOutcomeConclusion()
+        {
+            if (DeferOutcomeConclusion)
+            {
+                UpdateGlobalStepsContext();
+            }
         }
 
         public void UpdateGlobalStepsContext()
@@ -637,6 +675,40 @@ namespace GitHub.Runner.Worker
             // todo: restrict multiline?
 
             Global.StepsContext.SetOutput(ScopeName, ContextName, name, value, out reference);
+        }
+
+        public void FlushDeferredOutputs()
+        {
+            if (DeferredOutputs == null || DeferredOutputs.Count == 0)
+            {
+                return;
+            }
+
+            foreach (var kvp in DeferredOutputs)
+            {
+                Global.StepsContext.SetOutput(ScopeName, ContextName, kvp.Key, kvp.Value, out _);
+            }
+        }
+
+        public void FlushDeferredEnvironment()
+        {
+            if (DeferredEnvironmentVariables != null)
+            {
+                foreach (var kvp in DeferredEnvironmentVariables)
+                {
+                    Global.EnvironmentVariables[kvp.Key] = kvp.Value;
+                    SetEnvContext(kvp.Key, kvp.Value);
+                }
+            }
+
+            if (DeferredPrependPath != null)
+            {
+                foreach (var path in DeferredPrependPath)
+                {
+                    Global.PrependPath.RemoveAll(x => string.Equals(x, path, StringComparison.CurrentCulture));
+                    Global.PrependPath.Add(path);
+                }
+            }
         }
 
         public void SetTimeout(TimeSpan? timeout)
@@ -809,6 +881,15 @@ namespace GitHub.Runner.Worker
         {
             ArgUtil.NotNull(displayName, nameof(displayName));
             _record.Name = displayName;
+            _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
+        }
+
+        public void SetBackgroundStepMetadata(bool isBackground = false, string backgroundControlType = null, string[] backgroundControlStepIds = null, string parallelGroupId = null)
+        {
+            _record.IsBackground = isBackground;
+            _record.BackgroundControlType = backgroundControlType;
+            _record.BackgroundControlStepIds = backgroundControlStepIds;
+            _record.ParallelGroupId = parallelGroupId;
             _jobServerQueue.QueueTimelineRecordUpdate(_mainTimelineId, _record);
         }
 
@@ -1196,7 +1277,10 @@ namespace GitHub.Runner.Worker
                     }
 
                     Trace.Info($"Publish step telemetry for current step {StringUtil.ConvertToJson(StepTelemetry)}.");
-                    Global.StepsTelemetry.Add(StepTelemetry);
+                    lock (Global.CollectionLock)
+                    {
+                        Global.StepsTelemetry.Add(StepTelemetry);
+                    }
                     _stepTelemetryPublished = true;
                 }
             }
@@ -1335,7 +1419,10 @@ namespace GitHub.Runner.Worker
                 Trace.Info($"Updated step result (continue on error)");
             }
 
-            UpdateGlobalStepsContext();
+            if (!DeferOutcomeConclusion)
+            {
+                UpdateGlobalStepsContext();
+            }
         }
 
         internal IPipelineTemplateEvaluator ToPipelineTemplateEvaluatorInternal(bool allowServiceContainerCommand, ObjectTemplating.ITraceWriter traceWriter = null)
--- a/src/Runner.Worker/FileCommandManager.cs
+++ b/src/Runner.Worker/FileCommandManager.cs
@@ -122,8 +122,18 @@ namespace GitHub.Runner.Worker
                     {
                         continue;
                     }
-                    context.Global.PrependPath.RemoveAll(x => string.Equals(x, line, StringComparison.CurrentCulture));
-                    context.Global.PrependPath.Add(line);
+                    if (context.DeferredPrependPath != null)
+                    {
+                        // Background step: buffer path additions until wait/wait-all
+                        context.DeferredPrependPath.RemoveAll(x => string.Equals(x, line, StringComparison.CurrentCulture));
+                        context.DeferredPrependPath.Add(line);
+                        context.Debug($"Deferred prepend path '{line}' for background step");
+                    }
+                    else
+                    {
+                        context.Global.PrependPath.RemoveAll(x => string.Equals(x, line, StringComparison.CurrentCulture));
+                        context.Global.PrependPath.Add(line);
+                    }
                 }
             }
         }
@@ -172,8 +182,17 @@ namespace GitHub.Runner.Worker
             string name,
             string value)
         {
-            context.Global.EnvironmentVariables[name] = value;
-            context.SetEnvContext(name, value);
+            if (context.DeferredEnvironmentVariables != null)
+            {
+                // Background step: buffer env changes until wait/wait-all
+                context.DeferredEnvironmentVariables[name] = value;
+                context.Debug($"Deferred env '{name}' for background step");
+            }
+            else
+            {
+                context.Global.EnvironmentVariables[name] = value;
+                context.SetEnvContext(name, value);
+            }
             context.Debug($"{name}='{value}'");
         }
 
@@ -302,7 +321,16 @@ namespace GitHub.Runner.Worker
             var pairs = new EnvFileKeyValuePairs(context, filePath);
             foreach (var pair in pairs)
             {
-                context.SetOutput(pair.Key, pair.Value, out var reference);
+                if (context.DeferredOutputs != null)
+                {
+                    // Background step: buffer outputs until wait/wait-all
+                    context.DeferredOutputs[pair.Key] = pair.Value;
+                    context.Debug($"Deferred output '{pair.Key}' for background step");
+                }
+                else
+                {
+                    context.SetOutput(pair.Key, pair.Value, out var reference);
+                }
                 context.Debug($"Set output {pair.Key} = {pair.Value}");
             }
         }
--- a/src/Runner.Worker/GlobalContext.cs
+++ b/src/Runner.Worker/GlobalContext.cs
@@ -12,6 +12,9 @@ namespace GitHub.Runner.Worker
 {
     public sealed class GlobalContext
     {
+        // Lock for thread-safe access to shared collections during concurrent background step execution
+        public readonly object CollectionLock = new object();
+
         public ContainerInfo Container { get; set; }
         public List<ServiceEndpoint> Endpoints { get; set; }
         public IDictionary<String, String> EnvironmentVariables { get; set; }
--- a/src/Runner.Worker/StepsContext.cs
+++ b/src/Runner.Worker/StepsContext.cs
@@ -18,6 +18,7 @@ namespace GitHub.Runner.Worker
     {
         private static readonly Regex _propertyRegex = new("^[a-zA-Z_][a-zA-Z0-9_]*$", RegexOptions.Compiled);
         private readonly DictionaryContextData _contextData = new();
+        private readonly object _lock = new();
 
         /// <summary>
         /// Clears memory for a composite action's isolated "steps" context, after the action
@@ -25,9 +26,12 @@ namespace GitHub.Runner.Worker
         /// </summary>
         public void ClearScope(string scopeName)
         {
-            if (_contextData.TryGetValue(scopeName, out _))
+            lock (_lock)
             {
-                _contextData[scopeName] = new DictionaryContextData();
+                if (_contextData.TryGetValue(scopeName, out _))
+                {
+                    _contextData[scopeName] = new DictionaryContextData();
+                }
             }
         }
 
@@ -41,23 +45,26 @@ namespace GitHub.Runner.Worker
         /// </summary>
         public DictionaryContextData GetScope(string scopeName)
         {
-            if (scopeName == null)
+            lock (_lock)
             {
-                scopeName = string.Empty;
-            }
+                if (scopeName == null)
+                {
+                    scopeName = string.Empty;
+                }
 
-            var scope = default(DictionaryContextData);
-            if (_contextData.TryGetValue(scopeName, out var scopeValue))
-            {
-                scope = scopeValue.AssertDictionary("scope");
-            }
-            else
-            {
-                scope = new DictionaryContextData();
-                _contextData.Add(scopeName, scope);
-            }
+                var scope = default(DictionaryContextData);
+                if (_contextData.TryGetValue(scopeName, out var scopeValue))
+                {
+                    scope = scopeValue.AssertDictionary("scope");
+                }
+                else
+                {
+                    scope = new DictionaryContextData();
+                    _contextData.Add(scopeName, scope);
+                }
 
-            return scope;
+                return scope;
+            }
         }
 
         public void SetOutput(
@@ -67,16 +74,19 @@ namespace GitHub.Runner.Worker
             string value,
             out string reference)
         {
-            var step = GetStep(scopeName, stepName);
-            var outputs = step["outputs"].AssertDictionary("outputs");
-            outputs[outputName] = new StringContextData(value);
-            if (_propertyRegex.IsMatch(outputName))
+            lock (_lock)
             {
-                reference = $"steps.{stepName}.outputs.{outputName}";
-            }
-            else
-            {
-                reference = $"steps['{stepName}']['outputs']['{outputName}']";
+                var step = GetStep(scopeName, stepName);
+                var outputs = step["outputs"].AssertDictionary("outputs");
+                outputs[outputName] = new StringContextData(value);
+                if (_propertyRegex.IsMatch(outputName))
+                {
+                    reference = $"steps.{stepName}.outputs.{outputName}";
+                }
+                else
+                {
+                    reference = $"steps['{stepName}']['outputs']['{outputName}']";
+                }
             }
         }
 
@@ -85,8 +95,11 @@ namespace GitHub.Runner.Worker
             string stepName,
             ActionResult conclusion)
         {
-            var step = GetStep(scopeName, stepName);
-            step["conclusion"] = new StringContextData(conclusion.ToString().ToLowerInvariant());
+            lock (_lock)
+            {
+                var step = GetStep(scopeName, stepName);
+                step["conclusion"] = new StringContextData(conclusion.ToString().ToLowerInvariant());
+            }
         }
 
         public void SetOutcome(
@@ -94,8 +107,11 @@ namespace GitHub.Runner.Worker
             string stepName,
             ActionResult outcome)
         {
-            var step = GetStep(scopeName, stepName);
-            step["outcome"] = new StringContextData(outcome.ToString().ToLowerInvariant());
+            lock (_lock)
+            {
+                var step = GetStep(scopeName, stepName);
+                step["outcome"] = new StringContextData(outcome.ToString().ToLowerInvariant());
+            }
         }
 
         private DictionaryContextData GetStep(string scopeName, string stepName)
--- a/src/Sdk/DTPipelines/Pipelines/BackgroundStepControl.cs
+++ b/src/Sdk/DTPipelines/Pipelines/BackgroundStepControl.cs
@@ -1,0 +1,89 @@
+using System.ComponentModel;
+using System.Runtime.Serialization;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using Newtonsoft.Json;
+
+namespace GitHub.DistributedTask.Pipelines
+{
+    /// <summary>
+    /// Known control-flow types for background step control steps.
+    /// Wire values must match run-service constants (wait, wait-all, cancel).
+    /// </summary>
+    public static class BackgroundControlTypes
+    {
+        public const string Wait = "wait";
+        public const string WaitAll = "wait-all";
+        public const string Cancel = "cancel";
+    }
+
+    /// <summary>
+    /// Nested data for background step control, matching the run-service JSON shape.
+    /// </summary>
+    public class BackgroundStepControlData
+    {
+        [JsonProperty("controlType")]
+        public string ControlType { get; set; }
+
+        [JsonProperty("stepIds")]
+        public string[] StepIds { get; set; }
+
+        [JsonProperty("parallelGroupId")]
+        public string ParallelGroupId { get; set; }
+    }
+
+    /// <summary>
+    /// Represents a unified background step control-flow step (wait, wait-all, cancel).
+    /// Replaces the separate WaitStep, WaitAllStep, and CancelStep types.
+    /// </summary>
+    [DataContract]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class BackgroundStepControl : JobStep
+    {
+        [JsonConstructor]
+        public BackgroundStepControl()
+        {
+        }
+
+        private BackgroundStepControl(BackgroundStepControl stepToClone)
+            : base(stepToClone)
+        {
+            this.Data = stepToClone.Data != null ? new BackgroundStepControlData
+            {
+                ControlType = stepToClone.Data.ControlType,
+                StepIds = stepToClone.Data.StepIds != null
+                    ? (string[])stepToClone.Data.StepIds.Clone()
+                    : null,
+                ParallelGroupId = stepToClone.Data.ParallelGroupId,
+            } : null;
+            this.DisplayNameToken = stepToClone.DisplayNameToken?.Clone();
+        }
+
+        public override StepType Type => StepType.BackgroundStepControl;
+
+        /// <summary>
+        /// Nested control data, deserialized from the "backgroundStepControl" JSON property.
+        /// </summary>
+        [JsonProperty("backgroundStepControl")]
+        public BackgroundStepControlData Data { get; set; }
+
+        /// <summary>
+        /// Convenience accessors that delegate to Data.
+        /// </summary>
+        [JsonIgnore]
+        public string ControlType => Data?.ControlType;
+
+        [JsonIgnore]
+        public string[] StepIds => Data?.StepIds;
+
+        [JsonIgnore]
+        public string ParallelGroupId => Data?.ParallelGroupId;
+
+        [DataMember(EmitDefaultValue = false)]
+        public TemplateToken DisplayNameToken { get; set; }
+
+        public override Step Clone()
+        {
+            return new BackgroundStepControl(this);
+        }
+    }
+}
--- a/src/Sdk/DTPipelines/Pipelines/BackgroundStepControl.cs
+++ b/src/Sdk/DTPipelines/Pipelines/BackgroundStepControl.cs
@@ -33,7 +33,6 @@ namespace GitHub.DistributedTask.Pipelines
 
     /// <summary>
     /// Represents a unified background step control-flow step (wait, wait-all, cancel).
-    /// Replaces the separate WaitStep, WaitAllStep, and CancelStep types.
     /// </summary>
     [DataContract]
     [EditorBrowsable(EditorBrowsableState.Never)]
--- a/src/Sdk/DTPipelines/Pipelines/Step.cs
+++ b/src/Sdk/DTPipelines/Pipelines/Step.cs
@@ -7,6 +7,7 @@ namespace GitHub.DistributedTask.Pipelines
 {
     [DataContract]
     [KnownType(typeof(ActionStep))]
+    [KnownType(typeof(BackgroundStepControl))]
     [JsonConverter(typeof(StepConverter))]
     [EditorBrowsable(EditorBrowsableState.Never)]
     public abstract class Step
@@ -68,5 +69,7 @@ namespace GitHub.DistributedTask.Pipelines
     {
         [DataMember]
         Action = 4,
+        [DataMember]
+        BackgroundStepControl = 5,
     }
 }
--- a/src/Sdk/DTPipelines/Pipelines/StepConverter.cs
+++ b/src/Sdk/DTPipelines/Pipelines/StepConverter.cs
@@ -51,6 +51,9 @@ namespace GitHub.DistributedTask.Pipelines
                     case StepType.Action:
                         stepObject = new ActionStep();
                         break;
+                    case StepType.BackgroundStepControl:
+                        stepObject = new BackgroundStepControl();
+                        break;
                 }
 
                 using (var objectReader = value.CreateReader())
--- a/src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
+++ b/src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
@@ -43,6 +43,10 @@ namespace GitHub.DistributedTask.WebApi
                 this.WarningCount = recordToBeCloned.WarningCount;
                 this.NoticeCount = recordToBeCloned.NoticeCount;
                 this.AgentPlatform = recordToBeCloned.AgentPlatform;
+                this.IsBackground = recordToBeCloned.IsBackground;
+                this.BackgroundControlType = recordToBeCloned.BackgroundControlType;
+                this.BackgroundControlStepIds = recordToBeCloned.BackgroundControlStepIds;
+                this.ParallelGroupId = recordToBeCloned.ParallelGroupId;
 
                 if (recordToBeCloned.Log != null)
                 {
@@ -284,6 +288,34 @@ namespace GitHub.DistributedTask.WebApi
 
         [DataMember(Order = 132, EmitDefaultValue = false)]
         public string AgentPlatform
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 140, EmitDefaultValue = false)]
+        public bool IsBackground
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 141, EmitDefaultValue = false)]
+        public string BackgroundControlType
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 142, EmitDefaultValue = false)]
+        public string[] BackgroundControlStepIds
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 144, EmitDefaultValue = false)]
+        public string ParallelGroupId
         {
             get;
             set;
--- a/src/Sdk/RSWebApi/Contracts/StepResult.cs
+++ b/src/Sdk/RSWebApi/Contracts/StepResult.cs
@@ -50,5 +50,14 @@ namespace GitHub.Actions.RunService.WebApi
 
         [DataMember(Name = "annotations", EmitDefaultValue = false)]
         public List<Annotation> Annotations { get; set; }
+
+        [DataMember(Name = "is_background", EmitDefaultValue = false)]
+        public bool IsBackground { get; set; }
+
+        [DataMember(Name = "background_control_type", EmitDefaultValue = false)]
+        public string BackgroundControlType { get; set; }
+
+        [DataMember(Name = "background_control_step_ids", EmitDefaultValue = false)]
+        public string[] BackgroundControlStepIds { get; set; }
     }
 }
--- a/src/Sdk/WebApi/WebApi/Contracts.cs
+++ b/src/Sdk/WebApi/WebApi/Contracts.cs
@@ -179,6 +179,17 @@ namespace GitHub.Services.Results.Contracts
         public string CompletedAt;
         [DataMember]
         public Conclusion Conclusion;
+        [DataMember(EmitDefaultValue = false)]
+        public bool IsBackground;
+        [DataMember(EmitDefaultValue = false)]
+        [JsonProperty("backgroundControlType")]
+        public string BackgroundControlType;
+        [DataMember(EmitDefaultValue = false)]
+        [JsonProperty("backgroundControlStepIds")]
+        public string[] BackgroundControlStepIds;
+        [DataMember(EmitDefaultValue = false)]
+        [JsonProperty("parallelGroupId")]
+        public string ParallelGroupId;
     }
 
     public enum Status
--- a/src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
+++ b/src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
@@ -596,13 +596,6 @@ namespace GitHub.Services.Results.Client
             var stepUpdateEndpoint = new Uri(m_resultsServiceUrl, Constants.WorkflowStepsUpdate);
             foreach (var request in stepUpdateRequests)
             {
-                // DEBUG: capture request JSON
-                try
-                {
-                    var debugJson = Newtonsoft.Json.JsonConvert.SerializeObject(request, Newtonsoft.Json.Formatting.Indented);
-                    System.IO.File.AppendAllText("/tmp/runner-step-update-debug.json", $"\n---{DateTime.UtcNow}---\n{debugJson}\n");
-                }
-                catch { }
                 await SendRequest<StepsUpdateRequest>(stepUpdateEndpoint, cancellationToken, request, timestamp);
             }
         }
--- a/src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
+++ b/src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
@@ -514,7 +514,7 @@ namespace GitHub.Services.Results.Client
 
         private Step ConvertTimelineRecordToStep(TimelineRecord r)
         {
-            return new Step()
+            var step = new Step()
             {
                 ExternalId = r.Id.ToString(),
                 Number = r.Order.GetValueOrDefault(),
@@ -522,8 +522,25 @@ namespace GitHub.Services.Results.Client
                 Status = ConvertStateToStatus(r.State.GetValueOrDefault()),
                 StartedAt = r.StartTime?.ToString(Constants.TimestampFormat, CultureInfo.InvariantCulture),
                 CompletedAt = r.FinishTime?.ToString(Constants.TimestampFormat, CultureInfo.InvariantCulture),
-                Conclusion = ConvertResultToConclusion(r.Result)
+                Conclusion = ConvertResultToConclusion(r.Result),
+                IsBackground = r.IsBackground,
             };
+
+            // Set background control type directly (no enum mapping needed)
+            if (!string.IsNullOrEmpty(r.BackgroundControlType))
+            {
+                step.BackgroundControlType = r.BackgroundControlType;
+            }
+            if (r.BackgroundControlStepIds != null)
+            {
+                step.BackgroundControlStepIds = r.BackgroundControlStepIds;
+            }
+            if (!string.IsNullOrEmpty(r.ParallelGroupId))
+            {
+                step.ParallelGroupId = r.ParallelGroupId;
+            }
+
+            return step;
         }
 
         private Status ConvertStateToStatus(TimelineRecordState s)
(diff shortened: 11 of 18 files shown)


src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,6 +37,11 @@ public interface IStepsRunner : IRunnerService
 
     public sealed class StepsRunner : RunnerService, IStepsRunner
     {
+        // Track active background steps
+        private readonly ConcurrentDictionary<string, BackgroundStepContext> _backgroundSteps = new();
```
ericsciple: Does this need to be a ConcurrentDictionary?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -57,6 +64,31 @@ public async Task RunAsync(IExecutionContext jobContext)
                 if (jobContext.JobSteps.Count == 0 && !checkPostJobActions)
                 {
                     checkPostJobActions = true;
+
+                    // Implicit wait-all before post-job hooks:
+                    // If any background steps haven't been waited on, inject a visible wait-all step.
+                    if (_backgroundSteps.Count > 0)
```
ericsciple: I'm wondering if this is something that should be done in "Set up job" instead.

That's where all the pre/post/etc stuff is registered today.

src/Runner.Worker/JobExtension.cs
```diff
@@ -407,6 +456,42 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                             ArgUtil.NotNull(actionStep, step.DisplayName);
                             intraActionStates.TryGetValue(actionStep.Action.Id, out var intraActionState);
                             actionStep.ExecutionContext = jobContext.CreateChild(actionStep.Action.Id, actionStep.DisplayName, actionStep.Action.Name, null, actionStep.Action.ContextName, ActionRunStage.Main, intraActionState);
+
+                            // Store background step metadata on the timeline record for results service
+                            if (actionStep.Action?.Background == true)
```
ericsciple: I'm wondering whether pre/post should not follow `background`

That is:
- pre-stage would be sequential
- main-stage could have background
- implicit wait at the end of main-stage
- post-stage is sequential

Otherwise might need implicit waits at the end of each stage, which might be fine too.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +443,317 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+            _backgroundSteps[stepId] = bgContext;
+            Trace.Info($"Background step '{stepId}' started.");
+        }
+
+        private void SetWaitStepIdsTimelineVariable(IExecutionContext executionContext, IEnumerable<string> logicalStepIds)
+        {
+            var externalIds = GetBackgroundExternalIds(logicalStepIds);
+            if (externalIds.Count > 0)
+            {
+                executionContext.SetTimelineRecordVariable("wait_step_ids", string.Join(",", externalIds));
```
ericsciple: Per copilot:
> The same key (wait_step_ids / cancel_step_id) is written twice with different value semantics:
>
>  - Setup-time write: comma-separated logical ids (from waitStep.WaitStepIds).
>  - Runtime overwrite: comma-separated external GUIDs (from BackgroundStepContext.ExternalId).
>
> Both are queued to the results service. The consumer sees the logical-ids version first, then sees them replaced with GUIDs later. That's confusing and probably unintentional.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -68,12 +100,72 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                    continue;
+                }
+
                 ArgUtil.NotNull(step.ExecutionContext, nameof(step.ExecutionContext));
                 ArgUtil.NotNull(step.ExecutionContext.Global, nameof(step.ExecutionContext.Global));
                 ArgUtil.NotNull(step.ExecutionContext.Global.Variables, nameof(step.ExecutionContext.Global.Variables));
 
-                // Start
-                step.ExecutionContext.Start();
+                bool isBackground = false;
```
ericsciple: nit: consider:

`var isBackground = (step as IActionRunner)?.Action?.Background || false;`

src/Runner.Worker/StepsRunner.cs
ericsciple: I'm wondering if all the timeline record variable stuff can be contained to "Set up job". Per copilot, all the external IDs are known during "set up job".

ericsciple: Also curious @TingluoHuang's thoughts on timeline record variable approach in general.

Probably makes sense, but not something I normally think deeply about.

src/Runner.Worker/BackgroundStepContext.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    internal sealed class BackgroundStepContext
+    {
+        public string StepId { get; }
+        public IStep Step { get; }
+        public Task ExecutionTask { get; set; }
+        public CancellationTokenSource Cts { get; set; }
+        public GitHub.DistributedTask.WebApi.TaskResult? Result { get; set; }
+        public bool IsCompleted => ExecutionTask?.IsCompleted ?? false;
+        public string ExternalId => Step.ExecutionContext == null || Step.ExecutionContext.Id == Guid.Empty ? null : Step.ExecutionContext.Id.ToString("N");
```
ericsciple: I'm wondering if Step.ExecutionContext.Id is ever null or empty?

iiuc external IDs are known during "Set up job"

src/Runner.Worker/JobExtension.cs
```diff
@@ -315,8 +315,10 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
                     }
 
                     // Add action steps
+                    var stepOrder = 0;
                     foreach (var step in message.Steps)
                     {
+                        stepOrder++;
```
ericsciple: is this used?

src/Runner.Worker/StepsContext.cs
```diff
@@ -67,16 +68,19 @@ public void SetOutput(
             string value,
             out string reference)
         {
-            var step = GetStep(scopeName, stepName);
-            var outputs = step["outputs"].AssertDictionary("outputs");
-            outputs[outputName] = new StringContextData(value);
-            if (_propertyRegex.IsMatch(outputName))
+            lock (_lock)
```
ericsciple, line 77: I'm thinking background step outputs shouldn't be available to foreground steps before the `wait` synchronization point

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -579,6 +632,16 @@ public async Task UpdateWorkflowStepsAsync(Guid planId, IEnumerable<TimelineReco
             var stepUpdateEndpoint = new Uri(m_resultsServiceUrl, Constants.WorkflowStepsUpdate);
             foreach (var request in stepUpdateRequests)
             {
+                // DEBUG: Serialize and log the JSON payload
```
ericsciple: it looks like this should be deleted

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -567,7 +609,18 @@ private Conclusion ConvertResultToConclusion(TaskResult? r)
         public async Task UpdateWorkflowStepsAsync(Guid planId, IEnumerable<TimelineRecord> records, CancellationToken cancellationToken)
         {
             var timestamp = DateTime.UtcNow.ToString(Constants.TimestampFormat, CultureInfo.InvariantCulture);
-            var stepRecords = records.Where(r => String.Equals(r.RecordType, "Task", StringComparison.Ordinal));
+            var stepRecords = records.Where(r => String.Equals(r.RecordType, "Task", StringComparison.Ordinal)).ToList();
+
+            try
+            {
+                System.IO.File.AppendAllText("/tmp/bg-steps-debug.log", $"[BG-DEBUG] UpdateWorkflowStepsAsync: {stepRecords.Count} task records\n");
```
ericsciple: should be deleted?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -514,7 +514,19 @@ public async Task UploadResultsDiagnosticLogsAsync(string planId, string jobId,
...
-            return new Step()
+            // DEBUG: Log all variables on this timeline record to a file
+            try
+            {
+                var debugLine = $"[BG-DEBUG] ConvertTimelineRecordToStep: name={r.Name}, id={r.Id}, variableCount={r.Variables.Count}";
+                foreach (var kvp in r.Variables)
+                {
+                    debugLine += $"\n  Variable: {kvp.Key}={kvp.Value.Value}";
+                }
+                System.IO.File.AppendAllText("/tmp/bg-steps-debug.log", debugLine + "\n");
```
ericsciple: delete

src/Runner.Worker/StepsRunner.cs
ericsciple: I'm wondering whether we should have first class properties instead of timeline record variables, which iiuc we don't otherwise use today.

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,30 @@ public class Step
...
+        public string StepType;
+        [DataMember(EmitDefaultValue = false)]
+        public WaitControlDto Wait;
+        [DataMember(EmitDefaultValue = false)]
+        public CancelControlDto Cancel;
+    }
+
+    [DataContract]
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class WaitControlDto
```
ericsciple: What is "Dto" ?

gordolara789-tech (approved):

gordolara789-tech (approved):

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -514,7 +514,19 @@ public async Task UploadResultsDiagnosticLogsAsync(string planId, string jobId,
...
+            }
+            catch { }
```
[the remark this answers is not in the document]
lokesh755: removed

src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,6 +37,11 @@ public interface IStepsRunner : IRunnerService
...
     public sealed class StepsRunner : RunnerService, IStepsRunner
     {
+        // Track active background steps
+        private readonly ConcurrentDictionary<string, BackgroundStepContext> _backgroundSteps = new();
+        private readonly HashSet<string> _waitedStepIds = new();
+        private readonly SemaphoreSlim _backgroundSlotSemaphore = new(10); // max 10 concurrent background steps
+
         // StepsRunner should never throw exception to caller
         public async Task RunAsync(IExecutionContext jobContext)
         {
```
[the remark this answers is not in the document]
lokesh755: updated

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +443,317 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+
+            var stillRunning = activeSteps.Where(bg => !bg.IsCompleted).ToList();
+            if (stillRunning.Count > 0)
+            {
+                Trace.Warning($"{stillRunning.Count} background step(s) did not terminate gracefully.");
+            }
+
+            // Final wait for all tasks to complete
+            await Task.WhenAll(allTasks);
+        }
```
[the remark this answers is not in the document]
lokesh755: Done.. updated with 30 seconds termination

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +443,317 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+                    await Task.WhenAny(bgCtx.ExecutionTask, Task.Delay(gracePeriod));
+
+                    if (!bgCtx.IsCompleted)
+                    {
+                        Trace.Warning($"Background step '{cancelStep.CancelStepId}' did not terminate gracefully after {gracePeriod.TotalSeconds}s.");
+                    }
+                }
+
+                await bgCtx.ExecutionTask;
+                Trace.Info($"Background step '{cancelStep.CancelStepId}' cancelled/completed.");
```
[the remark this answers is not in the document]
lokesh755: updated

src/Runner.Worker/StepsRunner.cs
lokesh755: not really..updated.

src/Runner.Worker/StepsRunner.cs
lokesh755: updated to do this in "Set up job"

src/Runner.Worker/StepsRunner.cs
lokesh755: done

src/Runner.Worker/BackgroundStepContext.cs
lokesh755: updated.. it's never empty or null.

src/Runner.Worker/JobExtension.cs
lokesh755: removed

lokesh755: it was some intermediate approach.. not fully cleaned.

src/Runner.Worker/StepsContext.cs
lokesh755: updated. Basically storing outputs temporarily in different object until corresponding wait step or cancel step is executed.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: deleted

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: deleted

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: deleted

src/Sdk/WebApi/WebApi/Contracts.cs
lokesh755: updated to just `WaitControl`

src/Runner.Worker/StepsRunner.cs
lokesh755: Moved all timeline record variable setup to JobExtension

lokesh755: fixed. removed the double writes.

src/Runner.Worker/JobExtension.cs
lokesh755: updated such that only main stage runs in background.

src/Runner.Worker/JobExtension.cs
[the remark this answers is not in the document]
lokesh755: removed

src/Runner.Worker/JobExtension.cs
```diff
@@ -407,6 +456,42 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                        else if (step is WaitStepRunner waitRunner)
+                        {
+                            waitRunner.ExecutionContext = jobContext.CreateChild(
+                                waitRunner.StepId, waitRunner.DisplayName, waitRunner.StepName,
+                                null, waitRunner.StepName, ActionRunStage.Main);
+                            waitRunner.ExecutionContext.SetTimelineRecordVariable("step_type", "wait");
+                            if (waitRunner.StepIds != null && waitRunner.StepIds.Length > 0)
+                            {
+                                waitRunner.ExecutionContext.SetTimelineRecordVariable("wait_step_ids", string.Join(",", waitRunner.StepIds));
+                            }
```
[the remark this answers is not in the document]
lokesh755: updated to use only external ids

src/Sdk/DTPipelines/Pipelines/WaitStep.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.ComponentModel;
+using System.Runtime.Serialization;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
lokesh755: removed

src/Sdk/DTPipelines/Pipelines/WaitAllStep.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using System.ComponentModel;
+using System.Runtime.Serialization;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
lokesh755: removed

src/Sdk/DTPipelines/Pipelines/CancelStep.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.ComponentModel;
+using System.Runtime.Serialization;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using Newtonsoft.Json;
```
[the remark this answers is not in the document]
lokesh755: removed

src/Runner.Worker/BackgroundStepContext.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Concurrent;
```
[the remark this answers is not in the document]
lokesh755: removed

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +443,317 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+            }
+        }
+
+        private void SetCancelStepIdTimelineVariable(IExecutionContext executionContext, string logicalStepId)
+        {
+            var externalId = GetBackgroundExternalId(logicalStepId);
+            if (!string.IsNullOrEmpty(externalId))
+            {
+                executionContext.SetTimelineRecordVariable("cancel_step_id", externalId);
+            }
```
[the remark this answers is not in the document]
lokesh755: Timeline variable setup (wait_step_ids, cancel_step_id, is_background, step_type) is now done entirely in JobExtension during step creation, where we have the full contextNameToExternalId mapping and can resolve logical IDs → external GUIDs reliably. This eliminates the stale-value problem since the variables are set once at creation time, not conditionally updated at runtime.

src/Test/L0/Worker/BackgroundStepsL0.cs
```diff
@@ -0,0 +1,527 @@
...
+                {
+                    try
+                    {
+                        await Task.Delay(TimeSpan.FromSeconds(30), bgCts.Token);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        throw;
+                    }
+                });
```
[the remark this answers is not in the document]
lokesh755: Fixed. The test was reworked, it now wires CancelToken() to a local CancellationTokenSource and uses a 5s delay with that token

src/Runner.Worker/StepsContext.cs
```diff
@@ -67,16 +68,19 @@ public void SetOutput(
...
+                outputs[outputName] = new StringContextData(value);
+                if (_propertyRegex.IsMatch(outputName))
+                {
+                    reference = $"steps.{stepName}.outputs.{outputName}";
+                }
+                else
+                {
+                    reference = $"steps['{stepName}']['outputs']['{outputName}']";
+                }
             }
```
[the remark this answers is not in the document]
lokesh755: addressed.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -1440,7 +1463,6 @@ public static IPipelineTemplateEvaluator ToPipelineTemplateEvaluator(this IExecu
             return new PipelineTemplateEvaluator(traceWriter, schema, context.Global.FileTable)
             {
                 MaxErrorMessageLength = int.MaxValue, // Don't truncate error messages otherwise we might not scrub secrets correctly
```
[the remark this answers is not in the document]
lokesh755: reverted that change.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,306 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+            var bgContext = new BackgroundStepContext(stepId, step)
+            {
+                Cts = bgCts,
+            };
+
+            bgContext.ExecutionTask = Task.Run(async () =>
+            {
+                try
+                {
+                    await step.RunAsync();
```
ericsciple: Are `add-path` and `set-env` buffered until the `wait` synchronization point?

lokesh755: Not currently.. I can change it's behavior as well similar to outputs.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,306 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+
+                    // Apply continue-on-error: if the step failed and has continue-on-error: true,
+                    // this changes ExecutionContext.Result to Succeeded (Outcome keeps the raw failure).
+                    // We sync bgContext.Result so that wait/wait-all see the adjusted result.
+                    step.ExecutionContext.Result = bgContext.Result;
+                    step.ExecutionContext.ApplyContinueOnError(step.ContinueOnError);
+                    bgContext.Result = step.ExecutionContext.Result;
+
+                    // Update steps context with outcome/conclusion
+                    step.ExecutionContext.Complete(bgContext.Result);
```
ericsciple: Does this make the step outcome/conclusion available in the steps context before the `wait` synchronization point?

src/Runner.Worker/StepsContext.cs
```diff
@@ -18,16 +18,20 @@ public sealed class StepsContext
     {
         private static readonly Regex _propertyRegex = new("^[a-zA-Z_][a-zA-Z0-9_]*$", RegexOptions.Compiled);
         private readonly DictionaryContextData _contextData = new();
+        private readonly object _lock = new();
```
ericsciple, line 21: Are all of the following accessed in a threadsafe way:

Global.JobAnnotations, Global.JobTelemetry, Global.StepsTelemetry, Global.DeprecatedNode20Actions, Global.UpgradedToNode24Actions, Global.Arm32Node20Action

src/Runner.Worker/StepsRunner.cs
lokesh755: yes. But if we want strict consistency. we can defer conclusion as well.

src/Runner.Worker/StepsContext.cs
lokesh755: not as of now.. these might have low impact. we can still add lock or convert these two concurrent data structures.

src/Runner.Worker/StepsRunner.cs
lokesh755: updated to same behavior as outputs

src/Runner.Worker/StepsRunner.cs
lokesh755: updated to same behavior as outputs

src/Runner.Worker/StepsContext.cs
lokesh755: Added lock. ideally should be less contention.

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,30 @@ public class Step
...
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class WaitControl
+    {
+        [DataMember]
+        public string[] StepIds;
+    }
+
+    [DataContract]
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class CancelControl
```
TingluoHuang: do we need a nested struct with only 1 field?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -524,6 +524,35 @@ private Step ConvertTimelineRecordToStep(TimelineRecord r)
                 CompletedAt = r.FinishTime?.ToString(Constants.TimestampFormat, CultureInfo.InvariantCulture),
                 Conclusion = ConvertResultToConclusion(r.Result)
             };
+
+            // Populate background step metadata from TimelineRecord.Variables
+            if (r.Variables.TryGetValue("is_background", out var bgVar) && bgVar.Value == "true")
```
TingluoHuang: 🤔 should we stay away from `r.Variables` since that might cause problem when the latest runner connect to old GHES?

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -567,7 +596,8 @@ private Conclusion ConvertResultToConclusion(TaskResult? r)
         public async Task UpdateWorkflowStepsAsync(Guid planId, IEnumerable<TimelineRecord> records, CancellationToken cancellationToken)
         {
             var timestamp = DateTime.UtcNow.ToString(Constants.TimestampFormat, CultureInfo.InvariantCulture);
-            var stepRecords = records.Where(r => String.Equals(r.RecordType, "Task", StringComparison.Ordinal));
+            var stepRecords = records.Where(r => String.Equals(r.RecordType, "Task", StringComparison.Ordinal)).ToList();
+
```
TingluoHuang:
```suggestion
```

src/Runner.Worker/GlobalContext.cs
```diff
@@ -12,6 +12,9 @@ namespace GitHub.Runner.Worker
 {
     public sealed class GlobalContext
     {
+        // Lock for thread-safe access to shared collections during concurrent background step execution
+        public readonly object CollectionLock = new object();
```
TingluoHuang, line 16: instead of using lock, should we change everything to be thread-safe, ex: using concurrecnybag/dictionary/queue, etc.
So we won't have problem for any future code that forgot to add `lock`

src/Runner.Worker/StepsContext.cs
TingluoHuang, line 21: same question here, should we make `_contextData` thread-safe, so we don't need the lock.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,12 +36,27 @@ public interface IStepsRunner : IRunnerService
 
     public sealed class StepsRunner : RunnerService, IStepsRunner
     {
+        // Track active background steps
+        private const int DefaultMaxBackgroundSteps = 10;
```
TingluoHuang: do we still need a default if we always sends it from the service?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,315 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+                var timeout = TimeSpan.FromMinutes(timeoutMinutes);
+                step.ExecutionContext.SetTimeout(timeout);
+            }
+
+            var bgContext = new BackgroundStepContext(stepId, step)
+            {
+                Cts = bgCts,
+            };
+
+            bgContext.ExecutionTask = Task.Run(async () =>
```
TingluoHuang: we don't have to use `Task.Run(async () =>` since we are in an async task method.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -511,6 +526,24 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
                     Annotations = new List<Annotation>()
                 };
 
+                // Populate background step metadata from timeline record variables
+                if (_record.Variables.TryGetValue("is_background", out var bgVar) && bgVar.Value == "true")
```
ericsciple: Are timeline record variables used internally to pass info around within the runner?

I'm wondering if this information is available in a first class way internally (e.g. instance step instance type), rather than grabbing from a property bag here.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -100,11 +100,26 @@ public interface IExecutionContext : IRunnerService
         void SetGitHubContext(string name, string value);
         void SetOutput(string name, string value, out string reference);
         void SetTimeout(TimeSpan? timeout);
+
+        // Background step output deferral
+        Dictionary<string, string> DeferredOutputs { get; set; }
```
ericsciple: Properties are sorted before methods.

This interface is huge. I'm wondering if it would be good to organize all the new "Deferred" properties together as one visual block, and all the new methods "Flush" methods together as one visual block.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -618,6 +662,15 @@ public string GetGitHubContext(string name)
             }
         }
 
+        public Dictionary<string, string> DeferredOutputs { get; set; }
```
ericsciple: Are all the other properties sorted above methods?

src/Runner.Worker/StepsRunner.cs
lokesh755: Having some default value, in case we didn't receive from the service.

src/Runner.Worker/StepsRunner.cs
lokesh755: updated.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: removed

src/Runner.Worker/ExecutionContext.cs
lokesh755: grouped all together.

src/Runner.Worker/ExecutionContext.cs
lokesh755: They weren't, fixed now. All Deferred* properties are moved above methods in both the interface and the class.

src/Sdk/WebApi/WebApi/Contracts.cs
lokesh755: They match the server-side contract structure where wait/cancel are grouped as sub-objects in the JSON payload. Happy to flatten them if the results receiver deserializes them as flat proto fields,  let me know if you'd prefer that.

src/Runner.Worker/StepsContext.cs
lokesh755: We might still need lock since we do multiple operations in GetScope method.

src/Runner.Worker/GlobalContext.cs
lokesh755: Good point, I looked into this. The main challenge is that List<T> uses .AddRange() in a couple of places and there's no ConcurrentHashSet in .NET, so swapping types ends up touching ~20 call sites across 8 files. Feels too invasive for this PR,  happy to do it as a follow-up.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: Agreed, moved the metadata to a runner-internal dictionary instead. No longer relies on timeline record variables.

src/Runner.Worker/ExecutionContext.cs
lokesh755: Good call. Moved the metadata to a runner-internal dictionary instead

src/Runner.Worker/GlobalContext.cs
TingluoHuang: we can front load the change in its own PR, and change all the consumers.
i am pretty sure we will forget calling lock on this in the future and cause bugs.

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
```diff
@@ -512,9 +513,27 @@ public async Task UploadResultsDiagnosticLogsAsync(string planId, string jobId,
             await UploadLogFile(file, true, true, uploadUrlResponse.DiagLogsURL, uploadUrlResponse.BlobStorageType, cancellationToken);
         }
 
+        /// <summary>
+        /// Register background/control-flow step metadata so ConvertTimelineRecordToStep
+        /// can populate the Background step info of Steps
+        /// </summary>
+        public void RegisterStepMetadata(Guid recordId, BackgroundStepInfo info)
```
TingluoHuang: it's a little bit odd we store the data on the client.
It also mean, we will lost all the data if we ever need to recreate the client, ex: hitting network error force us recreate connection and client

lokesh755: this is very transient data. Also metadata in this is useful for rendering the UI.

TingluoHuang: i get what you are trying to do, but we don't have any existing http client code doing this kind of logic.
normally the client is just a simple wrapper around constructing url and payload for the endpoint.

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +181,42 @@ public class Step
...
+    {
+        [DataMember]
+        public string StepId;
```
TingluoHuang: i remember the original ADR contains cancel multiple background steps?
is that not supported now?

Seems odd that WaitControl and CancelControl are not matched.

src/Sdk/DTPipelines/Pipelines/StepConverter.cs
```diff
@@ -51,6 +51,15 @@ public override object ReadJson(
...
                         stepObject = new ActionStep();
                         break;
+                    case StepType.Wait:
+                        stepObject = new WaitStep();
+                        break;
+                    case StepType.WaitAll:
+                        stepObject = new WaitAllStep();
+                        break;
+                    case StepType.Cancel:
+                        stepObject = new CancelStep();
```
TingluoHuang: if we have wait and wait-all, why we don't have cancel and cancel-all?
I would expect these control method matches with each other?

src/Runner.Worker/WaitStepRunner.cs
```diff
@@ -0,0 +1,41 @@
...
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using GitHub.DistributedTask.Pipelines.ContextData;
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// A step that blocks until specific background step(s) complete.
+    /// Execution is handled by StepsRunner, not by RunAsync.
+    /// </summary>
+    public sealed class WaitStepRunner : IStep
```
TingluoHuang: it's odd that the `WaitStepRunner` and `WaitAllStepRunner` are so similar with each other, plus the class is almost empty to just make the interface happy. 🤔

lokesh755: Yeah.. they are similar.. with subtle difference of wait-all doesn't have any input steps.

src/Sdk/DTPipelines/Pipelines/StepConverter.cs
lokesh755: wait-all exists because "wait for everything" is a common pattern, cancel is intentional by nature, you pick what to stop. A cancel-all would basically be job cancellation. We could add it later if there is demand for `cancel-all`

src/Sdk/WebApi/WebApi/Contracts.cs
lokesh755: Makes sense. I started with single-target cancel since most use cases are "stop this specific service" after you're done with it. Happy to extend cancel to accept an array of step IDs as a follow-up.

src/Runner.Worker/FileCommandManager.cs
```diff
@@ -172,8 +181,16 @@ private static void SetEnvironmentVariable(
             string name,
             string value)
         {
-            context.Global.EnvironmentVariables[name] = value;
-            context.SetEnvContext(name, value);
+            if (context.DeferredEnvironmentVariables != null)
+            {
+                // Background step: buffer env changes until wait/wait-all
+                context.DeferredEnvironmentVariables[name] = value;
```
TingluoHuang, line 188: we might want to update the `context.Debug()` so we know which path we took?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -511,6 +529,24 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
                     Annotations = new List<Annotation>()
                 };
 
+                // Populate background step metadata from timeline record variables
+                if (_record.Variables.TryGetValue("is_background", out var bgVar) && bgVar.Value == "true")
```
TingluoHuang: i thought we no longer use record.variable for the background stuff?

lokesh755: should be bug.. will remove it.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -554,11 +590,22 @@ public TaskResult Complete(TaskResult? result = null, string currentOperation =
 
             _logger.End();
 
-            UpdateGlobalStepsContext();
+            if (!DeferOutcomeConclusion)
```
TingluoHuang, line 589: it's odd the we check `DeferredOutputs` inside of the `public void SetOutput`, but not use the same pattern for the other 2, should they match in behavior?

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -629,11 +676,56 @@ public void SetOutput(string name, string value, out string reference)
                 return;
             }
 
+            // For background steps, buffer outputs instead of writing to StepsContext.
+            // Outputs are flushed to StepsContext when a wait/wait-all step completes.
+            if (DeferredOutputs != null)
+            {
+                DeferredOutputs[name] = value;
+                reference = System.Text.RegularExpressions.Regex.IsMatch(name, "^[a-zA-Z_][a-zA-Z0-9_]*$")
```
TingluoHuang: do we do any regex before?

TingluoHuang: if yes, we might want to consider reuse the same regex.

lokesh755: Yeah.. there is.. will use a variable..

src/Runner.Worker/CancelStepRunner.cs
```diff
@@ -0,0 +1,41 @@
...
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using GitHub.DistributedTask.Pipelines.ContextData;
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// A step that cancels a specific background step.
+    /// Execution is handled by StepsRunner, not by RunAsync.
+    /// </summary>
+    public sealed class CancelStepRunner : IStep
```
TingluoHuang: do we need to merge the cancelsteprunner/watisteprunner/waitallsteprunner?
these 3 all looks very similar.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,12 +36,27 @@ public interface IStepsRunner : IRunnerService
...
             ArgUtil.NotNull(jobContext, nameof(jobContext));
             ArgUtil.NotNull(jobContext.JobSteps, nameof(jobContext.JobSteps));
 
+            // Reset per-job state (StepsRunner may be reused across jobs)
+            _backgroundSteps.Clear();
+            _waitedStepIds.Clear();
+
+            // Read max background steps from job payload (server-controlled), default to 10
+            var maxBgSteps = jobContext.Global.Variables.GetInt("system.runner.maxbackgroundsteps");
+            var maxConcurrent = (maxBgSteps.HasValue && maxBgSteps.Value > 0 && maxBgSteps.Value <= DefaultMaxBackgroundSteps) ? maxBgSteps.Value : DefaultMaxBackgroundSteps;
```
TingluoHuang: i don't think we want to force `maxBgSteps.Value <= DefaultMaxBackgroundSteps`?  in case we will allow config in the future, we probably want to use the default if the var is empty or non-positive int?

lokesh755: Yeah.. makes sense. Add this is a hard limit for now. Will make that change.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,321 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+
+        /// <summary>
+        /// Propagate any background step failures to the job result.
+        /// Called during implicit wait-all before post-hooks.
+        /// </summary>
+        private void PropagateBackgroundStepFailures(IExecutionContext jobContext)
+        {
+            foreach (var bgCtx in _backgroundSteps.Values)
+            {
+                if (bgCtx.Result == TaskResult.Failed)
```
TingluoHuang: what about other result, do we need to propagate back?
ex: cancel due to timeout?

src/Runner.Worker/JobExtension.cs
```diff
@@ -400,16 +448,119 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                            resultsServer.RegisterStepMetadata(cancelRunner.StepId, new BackgroundStepInfo
+                            {
+                                StepType = "cancel",
+                                CancelStepId = cancelExternalId
+                            });
                         }
                     }
 
+                    // Add implicit wait-all only if there are uncovered background steps
```
TingluoHuang: do we need a step or this can be part of `complete_job`?

src/Runner.Worker/JobExtension.cs
```diff
@@ -400,16 +448,119 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
                             actionStep.ExecutionContext = jobContext.CreateChild(actionStep.Action.Id, actionStep.DisplayName, actionStep.Action.Name, null, actionStep.Action.ContextName, ActionRunStage.Main, intraActionState);
+
+                            // Register background step metadata with the results service
+                            if (actionStep.Action?.Background == true)
+                            {
+                                hasBackgroundSteps = true;
+                                var externalId = actionStep.Action.Id.ToString("N");
+                                contextNameToExternalId[actionStep.Action.ContextName] = externalId;
+                                backgroundStepExternalIds.Add(externalId);
+                                resultsServer.RegisterStepMetadata(actionStep.Action.Id, new BackgroundStepInfo
```
TingluoHuang: maybe we should add whatever new fields into the timelinerecord, so you don't have to do these registerstepmetadata?

lokesh755: Yeah.. I am currently doing this change to move them to timelinerecord.. Didn't add originally since they are not part of original timeline record contract (atleast not yet :)

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,321 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+            if (_backgroundSteps.TryGetValue(stepId, out var bgCtx))
+            {
+                bgCtx.Step.ExecutionContext.FlushDeferredOutputs();
+                bgCtx.Step.ExecutionContext.FlushDeferredEnvironment();
+                bgCtx.Step.ExecutionContext.FlushDeferredOutcomeConclusion();
+                Trace.Info($"Flushed deferred state (outputs, env, path, outcome/conclusion) for background step '{stepId}'.");
+            }
+        }
+
+        private async Task HandleWaitAsync(WaitStepRunner waitStep, CancellationToken cancellationToken)
```
TingluoHuang: i am also wondering if we need to merge `HandleWaitAsync` and `HandleWaitAllAsync` since the only different is provide a list of step id?

src/Runner.Worker/StepsRunner.cs
```diff
@@ -342,5 +453,321 @@ private void CompleteStep(IStep step, TaskResult? result = null, string resultCo
...
+                    Trace.Info("Wait interrupted by job cancellation — cancelling waited background steps.");
+                    foreach (var stepId in stepIds)
+                    {
+                        if (_backgroundSteps.TryGetValue(stepId, out var bgCtx) && !bgCtx.IsCompleted)
+                        {
+                            bgCtx.Step.ExecutionContext.CancelToken();
+                            bgCtx.Cts.Cancel();
+                        }
+                    }
+                    await Task.WhenAny(Task.WhenAll(tasks), Task.Delay(TimeSpan.FromSeconds(7.5)));
```
TingluoHuang: it's a little bit odd we have multiple `WhenAny`, `WhenAll` in this method, there must be a better way to organize these. 🤔

src/Runner.Worker/JobExtension.cs
lokesh755: updated to use timelinerecord

src/Sdk/WebApi/WebApi/ResultsHttpClient.cs
lokesh755: updated to use timelinerecords.

src/Runner.Worker/ExecutionContext.cs
lokesh755: removed..

src/Runner.Worker/ExecutionContext.cs
lokesh755: moving other 2 to public method is not straightforward. instead moved the check of `DeferredOutputs` to the caller method for consistency.

src/Runner.Worker/StepsRunner.cs
lokesh755: updated.

src/Runner.Worker/ExecutionContext.cs
lokesh755: no longer needed.

src/Runner.Worker/JobExtension.cs
lokesh755: I think having a step will be more intuitive since wait-all step will surface on the UI.

src/Runner.Worker/StepsRunner.cs
lokesh755: timeout will result in fail as well. Cancel will be caused by either job getting canceled or canceled by cancel step. In both scenarios we need not propagate back result.

src/Runner.Worker/StepsRunner.cs
lokesh755: merged wait and wait-all steps processing into single method.

src/Runner.Worker/CancelStepRunner.cs
lokesh755: They're intentionally thin, just data carriers so the main loop can type-check them

src/Runner.Worker/StepsRunner.cs
lokesh755: refactored to make it cleaner

src/Runner.Worker/FileCommandManager.cs
lokesh755: added few dubug statements in codepaths of deferred outputs/env/path.

src/Sdk/WebApi/WebApi/Contracts.cs
TingluoHuang: we can't really change the contract once we shipped, if we think cancel multiple steps is a valid case we want to support, we might want to get it right in the first place.

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,30 @@ public class Step
         public string CompletedAt;
         [DataMember]
         public Conclusion Conclusion;
+        [DataMember(EmitDefaultValue = false)]
+        public bool IsBackground;
+        [DataMember(EmitDefaultValue = false)]
+        public string StepType;
```
TingluoHuang: why we need a step type?
what kinds of type do we have?
- runner
- action
- control

src/Sdk/RSWebApi/Contracts/StepResult.cs
```diff
@@ -50,5 +50,17 @@ public class StepResult
 
         [DataMember(Name = "annotations", EmitDefaultValue = false)]
         public List<Annotation> Annotations { get; set; }
+
+        [DataMember(Name = "is_background", EmitDefaultValue = false)]
+        public bool IsBackground { get; set; }
+
+        [DataMember(Name = "step_type", EmitDefaultValue = false)]
+        public string StepType { get; set; }
```
TingluoHuang: There is a `Type` in `StepResult` already, what are the difference?

src/Runner.Worker/WaitAllStepRunner.cs
```diff
@@ -0,0 +1,40 @@
...
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using GitHub.DistributedTask.Pipelines.ContextData;
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// A step that blocks until all prior background steps complete.
+    /// Execution is handled by StepsRunner, not by RunAsync.
+    /// </summary>
+    public sealed class WaitAllStepRunner : IStep
```
TingluoHuang: should these just use `JobExtensionRunner`?

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -289,6 +294,41 @@ public string AgentPlatform
...
+
+        [DataMember(Order = 142, EmitDefaultValue = false)]
+        public string[] WaitStepIds
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 143, EmitDefaultValue = false)]
+        public string CancelStepId
```
TingluoHuang: do we want to do the `string[] CancelStepIds`?

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -289,6 +294,41 @@ public string AgentPlatform
...
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 144, EmitDefaultValue = false)]
+        public string ParallelGroupId
```
TingluoHuang, line 318: What is `ParallelGroupId`, will you have multiple groups per job?

src/Sdk/RSWebApi/Contracts/StepResult.cs
TingluoHuang: ☝️ should we give a better name to this fied?

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
lokesh755: it's for adding support for paralle group. Yes, there could be multiple parallel groups

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,32 @@ public class Step
...
```
TingluoHuang: `BackgroundControlStepType`?

src/Sdk/DTPipelines/Pipelines/Step.cs
```diff
@@ -7,6 +7,9 @@ namespace GitHub.DistributedTask.Pipelines
 {
     [DataContract]
     [KnownType(typeof(ActionStep))]
+    [KnownType(typeof(WaitStep))]
+    [KnownType(typeof(WaitAllStep))]
+    [KnownType(typeof(CancelStep))]
```
TingluoHuang: BackgroundControlStep:

- DisplayName
- ControlType:
   - Cancel,
   - CancelAll
   - Wait,
   - WaitAll
- ParallelGroupId?
- StepIds?

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
```diff
@@ -289,6 +294,41 @@ public string AgentPlatform
...
 
+        [DataMember(Order = 140, EmitDefaultValue = false)]
+        public bool IsBackground
+        {
+            get;
+            set;
+        }
+
+        [DataMember(Order = 141, EmitDefaultValue = false)]
+        public string StepType
```
TingluoHuang: maybe also scoped to `BackgroundControlStepType`

src/Sdk/DTWebApi/WebApi/TimelineRecord.cs
TingluoHuang: i understand how it will be used now.

src/Sdk/RSWebApi/Contracts/StepResult.cs
TingluoHuang: `backgroundcontrolsteptype`

src/Sdk/RSWebApi/Contracts/StepResult.cs
```diff
@@ -50,5 +50,17 @@ public class StepResult
...
+
+        [DataMember(Name = "wait_step_ids", EmitDefaultValue = false)]
+        public string[] WaitStepIds { get; set; }
+
+        [DataMember(Name = "cancel_step_id", EmitDefaultValue = false)]
+        public string CancelStepId { get; set; }
```
TingluoHuang: public string[] CancelStepIds { get; set; }

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,32 @@ public class Step
...
+        [DataMember]
+        public string[] StepIds;
+    }
+
+    [DataContract]
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class CancelControl
+    {
+        [DataMember]
+        public string StepId;
```
TingluoHuang: public strings[] StepIds

src/Sdk/WebApi/WebApi/Contracts.cs
```diff
@@ -179,6 +179,17 @@ public class Step
         public string CompletedAt;
         [DataMember]
         public Conclusion Conclusion;
+        [DataMember(EmitDefaultValue = false)]
+        public bool IsBackground;
+        [DataMember(EmitDefaultValue = false)]
+        [JsonProperty("backgroundControlType")]
```
TingluoHuang, line 185: do we need to provide ` [JsonProperty("backgroundControlType")]`?

src/Sdk/DTPipelines/Pipelines/BackgroundStepControl.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// Convenience accessors that delegate to Data.
+        /// </summary>
+        [JsonIgnore]
+        public string ControlType => Data?.ControlType;
+
+        [JsonIgnore]
+        public string[] StepIds => Data?.StepIds;
+
+        [JsonIgnore]
+        public string ParallelGroupId => Data?.ParallelGroupId;
```
TingluoHuang, line 78: should we move it to `JobStep`?

PR #4419, opened 2026-05-14
Send welcome message in debugger console on connect

rentziass:
When a user connects to the DAP debugger, there's no indication of what the debug console can do. This PR sends a welcome message after the `configurationDone` handshake to draw the user's attention to the console and its capabilities.

## Approach

After the DAP `configurationDone` response is sent (following the same pattern as the post-`initialize` event), the debugger emits a console output event with a welcome message. By default the runner shows its built-in help text (the same content as typing `help` in the console).

The default can be overridden via the job message from run-service using two pieces:

- A feature flag `actions_runner_override_debugger_welcome_message` in the job variables -- when set to `true`, the runner takes its welcome message from the field below instead of using the default.
- A new `DebuggerWelcomeMessage` string field on the job message carrying the override content.

Behavior:

| `actions_runner_override_debugger_welcome_message` | `DebuggerWelcomeMessage`       | Result                       |
| -------------------------------------------------- | ------------------------------ | ---------------------------- |
| absent / `false`                                   | (ignored)                      | default built-in help text   |
| `true`                                             | non-empty                      | custom message shown as-is   |
| `true`                                             | null / empty                   | welcome message suppressed   |

This shape plays nicely with a Go producer using `omitempty`: the bool flag carries the intent and the string carries the content, so we no longer rely on preserving a null-vs-empty distinction across serialization boundaries.

A `_welcomeMessageSent` guard prevents duplicate messages if `configurationDone` is sent more than once. The guard is reset when a new client connects so that a reconnecting client (which has lost its console output) sees the welcome message again.

## Changes

- `Constants.Runner.Features`: new `OverrideDebuggerWelcomeMessage` feature flag key.
- `AgentJobRequestMessage`: new `DebuggerWelcomeMessage` string field.
- `DebuggerConfig`: new `OverrideWelcomeMessage` bool plus `WelcomeMessage` string; populated from the feature flag and job message field in `ExecutionContext`.
- `DapDebugger.SendWelcomeMessage()`: emits the console output event after `configurationDone`, honoring the override flag.
- Tests: DAP-level tests covering default, custom, empty-suppressed, null-suppressed, once-only, and reconnect; plus a job-message serialization round-trip test.


src/Test/L0/Worker/DapDebuggerL0.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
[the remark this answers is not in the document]
rentziass: The BOM is actually the repo convention -- all existing `.cs` files (e.g. `ExecutionContext.cs`, `JobExtension.cs`, `Constants.cs`) include a UTF-8 BOM. The files in this PR are consistent with that, so no change needed.

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -818,6 +824,35 @@ private void SendOutput(string category, string text)
...
         }
 
+        internal void SendWelcomeMessage()
+        {
+            if (_welcomeMessageSent)
+            {
+                return;
+            }
+            _welcomeMessageSent = true;
+
```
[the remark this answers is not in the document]
rentziass: Good point -- a reconnecting client loses console output and should see the welcome message again. Fixed by resetting `_welcomeMessageSent = false` in `HandleClientConnected()`. The duplicate guard still prevents multiple sends within a single connection's handshake (e.g. repeated `configurationDone` requests).

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -267,6 +267,18 @@ public DebuggerTunnelInfo DebuggerTunnel
             set;
         }
 
+        /// <summary>
+        /// Optional welcome message to show in the debugger console when a
```
TingluoHuang: is the comment about null/empty right?
after JSON deserialization, null and empty are the same, i think.

src/Runner.Worker/Dap/DebuggerConfig.cs
```diff
@@ -23,6 +24,12 @@ public DebuggerConfig(bool enabled, DebuggerTunnelInfo tunnel)
         /// </summary>
         public DebuggerTunnelInfo Tunnel { get; }
 
+        /// <summary>
+        /// Optional welcome message for the debugger console.
+        /// Null = show default help, empty = show nothing, non-empty = show as-is.
```
TingluoHuang: same question about the comment.

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
rentziass: It does work this way yeah, confirmed in tests 👀 but I think this won’t work on the backend with run-service because of Go serialisation, adding an explicit flag to override the welcome message.

rentziass: Good catch -- after going back and forth, I agree the original three-state design (null vs empty vs value) was fragile, especially given that the producer is a Go service where `omitempty` would collapse null and empty anyway. I've refactored to use an explicit `actions_runner_override_debugger_welcome_message` feature flag in job variables: when absent the runner uses the default help text, when set the (possibly empty) `DebuggerWelcomeMessage` is taken as-is. This makes the intent unambiguous on both sides of the wire.

src/Runner.Worker/Dap/DebuggerConfig.cs
rentziass: Same reasoning as above -- the comment was describing fragile semantics that depended on JSON producers preserving null vs empty. The model is now driven by an explicit `OverrideWelcomeMessage` bool on `DebuggerConfig`, populated from the `actions_runner_override_debugger_welcome_message` feature flag, so the doc comment is now unambiguous.

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
rentziass: lol Copilot was eager to also reply 😄

src/Sdk/DTPipelines/Pipelines/AgentJobRequestMessage.cs
```diff
@@ -267,6 +267,20 @@ public DebuggerTunnelInfo DebuggerTunnel
             set;
         }
 
+        /// <summary>
```
TingluoHuang, line 270: do we want to update the comment?

TingluoHuang (approved):

PR #4420, opened 2026-05-14
Execute debugger REPL commands inside job container

rentziass:
When a job uses a `container:` definition, the debugger REPL was always executing `run` commands on the runner host via `IProcessInvoker` directly. This meant commands like `cat /etc/os-release` would show the host OS (Ubuntu) instead of the container's OS (e.g. Alpine), making the debugger unreliable for inspecting container job state.

## Approach

Instead of duplicating `docker exec` logic, the REPL executor now reuses `IStepHost` -- the same abstraction that normal workflow `run:` steps use. The debugger determines the current step type and creates the appropriate host:

- **Action steps** (user-defined `run:`/`uses:`) get a `ContainerStepHost`, which routes execution through `docker exec` (or container hooks when enabled). Script paths and working directories are translated to container paths, and `PrependPath` is propagated to the container's `PATH`.
- **Infrastructure steps** (Set up job, Initialize containers, Stop containers, Complete job) get a `DefaultStepHost` and always execute on the runner host, since these steps operate outside the container lifecycle.

The guiding principle: running a command in the debug console should behave identically to having a `run:` step in your workflow file.

## Execution context banner

Since DAP doesn't support customizing the console prompt, the debugger now emits a banner each time it pauses at a step so the user knows where commands will run:

```
Commands will run on job container: alpine:3.20 (a1b2c3d4e5f6)
```
```
Commands will run on runner host
```

## Testing

Added unit tests for `CreateStepHost` covering:
- No container configured -- returns `DefaultStepHost`
- Container with ID + action step -- returns `ContainerStepHost`
- Container with ID + infrastructure step -- returns `DefaultStepHost`
- Container without ID (not yet started) -- falls back to `DefaultStepHost`


src/Runner.Worker/Dap/DapReplExecutor.cs
```diff
@@ -65,9 +67,17 @@ public async Task<EvaluateResponseBody> ExecuteRunCommandAsync(
...
+            bool isActionStep,
             CancellationToken cancellationToken)
         {
-            // 1. Resolve shell — same logic as ScriptHandler
+            // 1. Resolve step host — container or host, same as ActionRunner.
+            //    Only action steps (user-defined run:/uses:) execute inside the
+            //    container.  Infrastructure steps (Set up job, Initialize
+            //    containers, Complete job, etc.) always run on the host.
+            var stepHost = CreateStepHost(context, isActionStep);
+            var isContainerStepHost = stepHost is ContainerStepHost;
```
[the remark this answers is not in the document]
rentziass: Good catch -- changed both checks to use `IContainerStepHost` instead of the concrete type, so any alternative implementation registered in the service locator is handled correctly.

src/Runner.Worker/Dap/DapReplExecutor.cs
```diff
@@ -141,48 +170,49 @@ private async Task<EvaluateResponseBody> ExecuteScriptAsync(
...
                     {
-                        if (!string.IsNullOrEmpty(args.Data))
-                        {
-                            var masked = _hostContext.SecretMasker.MaskSecrets(args.Data);
-                            SendOutput("stdout", masked + "\n");
-                        }
-                    };
+                        var masked = _hostContext.SecretMasker.MaskSecrets(args.Data);
+                        SendOutput("stdout", masked + "\n");
+                    }
```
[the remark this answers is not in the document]
rentziass: Valid point. For now I've added a trace warning when hooks are enabled so it's visible in diagnostics. Fully wiring output streaming through the hook manager is a bigger change that we can tackle separately.

src/Test/L0/Worker/DapReplExecutorL0.cs
```diff
@@ -233,5 +237,71 @@ public void BuildEnvironment_NullReplEnv_ReturnsContextEnvOnly()
...
+            using (var hc = CreateTestContext())
+            {
+                hc.EnqueueInstance<IDefaultStepHost>(new DefaultStepHost());
+                // Container exists but hasn't been started yet (no ContainerId)
+                var container = new ContainerInfo();
+                var context = CreateMockContext(container: container);
+                var result = _executor.CreateStepHost(context.Object, isActionStep: true);
+
+                Assert.IsType<DefaultStepHost>(result);
+            }
```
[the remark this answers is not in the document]
rentziass: Added -- the new test sets both the `AllowRunnerContainerHooks` variable and the `ACTIONS_RUNNER_CONTAINER_HOOKS` env var, then verifies a container without `ContainerId` still returns an `IContainerStepHost`.

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -860,6 +860,9 @@ internal async Task OnStepStartingAsync(IStep step, bool isFirstStep)
             // Send stopped event to debugger (only if client is connected)
             SendStoppedEvent(reason, description);
 
+            // Emit a banner so the user knows where REPL commands will execute
+            SendExecutionContextBanner(step);
```
TingluoHuang: should we pass in `_currentStep` since that's the one captured under the lock?

TingluoHuang: or just use `_currentStep` and not pass in any arg?

src/Runner.Worker/Dap/DapDebugger.cs
```diff
@@ -1407,6 +1415,40 @@ private void SendStoppedEvent(string reason, string description)
...
+                return;
+            }
+
+            bool isActionStep = step is IActionRunner;
+            var container = _jobContext?.Global?.Container;
+
+            string target;
+            if (isActionStep && container != null &&
+                (!string.IsNullOrEmpty(container.ContainerId) ||
+                 FeatureManager.IsContainerHooksEnabled(_jobContext?.Global?.Variables)))
```
TingluoHuang, line 1435: the condition might have some issue?
for a containerhook, can you have containerhook enabled without a container job?
maybe yes.
in that case, the `container != null` will always filter out that case?

src/Runner.Worker/Dap/DapReplExecutor.cs
```diff
@@ -198,6 +237,43 @@ private async Task<EvaluateResponseBody> ExecuteScriptAsync(
...
+        internal IStepHost CreateStepHost(IExecutionContext context, bool isActionStep)
+        {
+            if (!isActionStep)
+            {
+                _trace.Info("Creating DefaultStepHost for REPL execution (infrastructure step)");
+                return _hostContext.CreateService<IDefaultStepHost>();
+            }
+
+            var container = context?.Global?.Container;
+            if (container != null)
```
TingluoHuang, line 264: kind of similar question, when containerhook enabled for container step, and there is no job container, what would we use?

src/Runner.Worker/Dap/DapReplExecutor.cs
```diff
@@ -111,25 +121,44 @@ private async Task<EvaluateResponseBody> ExecuteScriptAsync(
...
+                if (context.Global.PrependPath.Count > 0)
+                {
+                    if (stepHost is IContainerStepHost containerHost)
+                    {
+                        containerHost.PrependPath = prependPath;
+                    }
+                    else
+                    {
+                        string existingPath;
+                        environment.TryGetValue(Constants.PathVariable, out existingPath);
```
TingluoHuang: do we need to port everything?
```
string originalPath = RuntimeVariables.Get(Constants.PathVariable) ?? // Prefer a job variable.
                    taskEnvPATH ?? // Then a task-environment variable.
                    System.Environment.GetEnvironmentVariable(Constants.PathVariable) ?? // Then an environment variable.
                    string.Empty;
```
```

File: src/Runner.Worker/Dap/DapReplExecutor.cs [hunk 8]
```diff
@@ -141,48 +170,58 @@ private async Task<EvaluateResponseBody> ExecuteScriptAsync(
...
-                // 7. Execute via IProcessInvoker (same as DefaultStepHost)
-                int exitCode;
-                using (var processInvoker = _hostContext.CreateService<IProcessInvoker>())
+                // NOTE: When container hooks are enabled, ContainerStepHost routes
+                // execution through IContainerHookManager which does not raise
+                // OutputDataReceived/ErrorDataReceived events. Output will not be
+                // streamed to the debug console in that mode.
+                if (isContainerStepHost && FeatureManager.IsContainerHooksEnabled(context.Global?.Variables))
+                {
+                    _trace.Warning("Container hooks are enabled -- REPL output will not be streamed to the debug console");
```
TingluoHuang: where would end user see this?
should this be a warning to output in addition to diag log?

src/Runner.Worker/Dap/DapDebugger.cs
rentziass: It’s the same kind of check we do in https://github.com/actions/runner/blob/d4b902b533e3cc9c86dfc62c9e689c1ef5926c83/src/Runner.Worker/ActionRunner.cs#L156-L157; you can have containerhooks without a container job only if you have service containers, and in that case step host should always be the host.

src/Runner.Worker/Dap/DapReplExecutor.cs
rentziass: See above, I think this should be host. Stephost should be host during `Initialize containers` anyway, and after that it should still be host for the following steps if you only have service containers (like for normal steps).

src/Runner.Worker/Dap/DapDebugger.cs
rentziass: Good call. I went with your second suggestion and dropped the parameter entirely so the banner method reads `_currentStep` directly, matching the rest of the private helpers in this class. Fixed in ce5bdf5.

src/Runner.Worker/Dap/DapReplExecutor.cs
rentziass: You're right, the REPL `run` command accepts `env:` just like a real workflow step, so it should follow the same PATH priority chain. Ported the full `RuntimeVariables.Get(PATH)` -> task env -> system env order from `Handler.AddPrependPathToEnvironment` in ce5bdf5.

src/Runner.Worker/Dap/DapReplExecutor.cs
rentziass: Good catch - the user is sitting in the debug console and burying this in the diag log defeats the purpose. The warning now also goes out via `SendOutput("stderr", ...)` so it lands in the debug console alongside the `$ <command>` echo, right before the silent run. Fixed in ce5bdf5.

TingluoHuang (approved):

Issue #4423, opened 2026-05-17
Multiline runner stdout logs are extremely difficult to parse due to repeated line headers


nuclearpidgeon:
When running the Runner with stdout logs (via `ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1`), multiline log messages (e.g. JSON objects) for some reason have a header/timestamp appended for every single line of the log. This makes parsing these logs back to a multiline payload (or e.g. JSON object) nearly impossible, as you can't just scan lines until you see a new timestamp header to work out when the multiline message ends (which is how [Fluentbit](https://github.com/fluent/fluent-bit), as one example of a log forwarding tool, [handles multiline logs](https://docs.fluentbit.io/manual/data-pipeline/parsers/multiline-parsing) like Java stack traces).

Example output:
```
[RUNNER 2026-05-06 13:23:59Z INFO Runner] Using BrokerMessageListener
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener] Using provided settings
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener] {
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "AgentId": 2,
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "AgentName": "fishtin",
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "PoolId": 1,
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "PoolName": "Default",
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "ServerUrl": "https://pipelinesghubeus24.actions.githubusercontent.com/REDACTED/",
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "GitHubUrl": "https://github.com/nuclearpidgeon/actions-runner-testing",
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "WorkFolder": "_work",
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "UseV2Flow": true,
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener]   "ServerUrlV2": "https://broker.actions.githubusercontent.com/"
[RUNNER 2026-05-06 13:23:59Z INFO BrokerMessageListener] }
[...]

PR #4424, opened 2026-05-17
Add new env var to allow single-prefix multiline logs on stdout

nuclearpidgeon:
Addresses #4423.

Given the existing behaviour is that multiline messages have a prefix per line, I've designed the variable in a manner where it only _disables_ this behaviour if explicitly told to via the env var.

### DIFF
--- a/src/Runner.Common/Constants.cs
+++ b/src/Runner.Common/Constants.cs
@@ -308,7 +308,7 @@ namespace GitHub.Runner.Common
                 public static readonly string ForcedInternalNodeVersion = "ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION";
                 public static readonly string ForcedActionsNodeVersion = "ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION";
                 public static readonly string PrintLogToStdout = "ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT";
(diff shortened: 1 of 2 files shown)


src/Runner.Common/Constants.cs
```diff
@@ -307,6 +307,7 @@ public static class Agent
                 public static readonly string ForcedInternalNodeVersion = "ACTIONS_RUNNER_FORCED_INTERNAL_NODE_VERSION";
                 public static readonly string ForcedActionsNodeVersion = "ACTIONS_RUNNER_FORCE_ACTIONS_NODE_VERSION";
                 public static readonly string PrintLogToStdout = "ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT";
+                public static readonly string StdoutMultilineLogPrefixing = "ACTIONS_RUNNER_STDOUT_MULTILINE_LOG_PREFIXING";
```
TingluoHuang: I would suggest name the ENV as the disablexxxx.
ex:
`DisableStdoutMultilineLogPrefixing = "ACTIONS_RUNNER_DISABLE_STDOUT_MULTILINE_LOG_PREFIXING"`.

then, in `src/Runner.Common/StdoutTraceListener.cs`

`bool _disablePrefixMultilineLogs = StringUtil.ConvertToBoolean(Environment.GetEnvironmentVariable(Constants.Variables.Agent.DisableStdoutMultilineLogPrefixing, defaultValue: false)`

lokesh755 (approved):

PR #4417, opened 2026-05-13
DAP: add JobExecutionView model + renderer

rentziass:
This is part 1 of 2 of the work to serve the job "source" during debugging.

In this PR is all the code to build that representation: this is intended to be a 1:1 representation of how the runner sees the job it's executing rendered as YAML, it's _not_ the workflow file the job comes from; this allows us to consider pre and post actions steps as "lines" we can pause/add breakpoints (coming next) and show things exactly as the runner interprets them. We use YamlDotNet to render scalar values but we "hand-write" the source skeleton because we need to keep track of each step's line to be able to refer to it later (and to write comments but that's secondary at the moment).

<details>

<summary><strong>Example YAML served as source</strong></summary>

```yaml
# Job: build
# Runner execution plan — read-only.

setup:
  - step: Setup job

main:
  - step: Run actions/checkout@v6
    uses: actions/checkout@v6
    if: success()
  - step: Cache Primes
    id: cache-primes
    uses: actions/cache@v5
    if: success()
    with:
      path: prime-numbers
      key: ${{ runner.os }}-primes
  - step: Check prerequisites
    run: |
      echo banana
      echo "Checking system requirements..."
      echo "Node version: $(node --version 2>/dev/null || echo 'not installed')"
      echo "Python version: $(python3 --version 2>/dev/null || echo 'not installed')"
    if: success()
  - step: Install dependencies
    run: |
      cat > package.json << 'EOF'
      {
        "name": "demo",
        "scripts": {
          "build": "node build.js",
          "test": "node test.js"
        }
      }
      EOF
      cat > build.js << 'EOF'
      const fs = require('fs');
      const config = require('./config.json');
      fs.writeFileSync('dist/output.js', `// Built for ${config.env}`);
      EOF
    if: success()
  - step: Build
    run: |
      mkdir -p dist
      npm run build
    if: success()
  - step: Verify build
    run: |
      test -f dist/output.js && echo "Build successful!"
    if: success()

post:
  - step: Post Cache Primes
    action: actions/cache@v5
  - step: Post Run actions/checkout@v6
    action: actions/checkout@v6

cleanup:
  - step: Complete job
```

</details>

PR #4431, opened 2026-05-19
DAP: add YamlScalarFormatter (1/5)

rentziass:
> **Part 1 of 5** of the foundation for the DAP debugger's job source view. **No caller yet** — the renderer in part 2 will consume this.
>

## What this PR adds

A thin wrapper around YamlDotNet's `Emitter` that formats one string as a quote-safe YAML scalar — plain, single-quoted, or double-quoted as the emitter chooses — with no surrounding document markers or trailing newline. Safe to splice into a hand-emitted YAML document fragment.

## Why it exists

The upcoming execution-view renderer (part 2) hand-emits the YAML skeleton so it can track per-step line offsets (needed for `stackTrace` responses). Scalar values, however, must respect YAML's reserved chars, leading/trailing whitespace, and embedded `: `/`#` sequences. Doing this by hand is bug-prone, especially on edge cases (empty strings, expressions, multiline content), so we route scalars through the library.

## Notable details

- **LF line breaks forced** (`sw.NewLine = "\n"`). `StringWriter` otherwise picks up Windows's `Environment.NewLine` (CRLF), which corrupts the document-end stripping below.
- **Both `--- ` and `---\n` document-start prefixes are stripped.** The emitter elides them for most scalars but emits them for some edge cases (e.g. empty strings).

## Tests

Direct round-trip coverage: format a value → splice under a YAML key → parse with YamlDotNet's deserializer → assert the parsed value equals the input. Covers 15 special-character cases, empty strings, the LF guarantee, and document-marker stripping.

### DIFF
--- a/src/Runner.Worker/Dap/YamlScalarFormatter.cs
+++ b/src/Runner.Worker/Dap/YamlScalarFormatter.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Globalization;
+using System.IO;
+using GitHub.Runner.Sdk;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Formats a single string as a quote-safe YAML scalar by routing it
+    /// through YamlDotNet's <see cref="Emitter"/>. The returned text is
+    /// safe to splice into a hand-emitted YAML document fragment.
+    ///
+    /// Caller responsibility: this only handles the scalar value; it does
+    /// not emit a key, indent, or trailing newline.
+    /// </summary>
+    internal static class YamlScalarFormatter
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


rentziass:

- **run-service** will pre-render each step as a YAML list-item fragment and include it on the step contract sent to the runner.
- **The runner** will only stitch fragments together into the phase-keyed source document (setup / pre / main / post / cleanup). No `TemplateToken` → YAML, no `IStep` → entry translation, no scalar formatting on the runner side. Pre/Post sections reuse the corresponding Main fragment opaquely.

This makes the runner code substantially smaller, and — critically — moves the source format to a service that ships in minutes, so we can iterate the rendered view without waiting for runner releases.

This PR's code (`YamlScalarFormatter`) was the C# implementation of that scalar-formatting logic — it now lives in run-service / the workflow parser instead. Closing rather than retargeting since none of it survives.

PR #4432, opened 2026-05-19
DAP: add JobExecutionViewRenderer (2/5)

rentziass:
> **Part 2 of 5** of the foundation for the DAP debugger's job source view. Stacks on #4431.
>

## What this PR adds

The core rendering algorithm for the DAP execution view. Given a list of phase-tagged entries (`JobExecutionViewEntry`), produces a phase-keyed YAML document **plus a parallel array of 1-based line numbers** pointing at each entry's `- step:` key. The line numbers are what later powers the DAP `stackTrace` handler.

The rendered YAML is intended as a 1:1 representation of how the runner sees the job it's executing — _not_ the workflow file. That lets pre/post action steps appear as their own "lines" the user can pause/breakpoint on, and lets us show parameters exactly as the runner interprets them.

<details>
<summary><strong>Example output</strong></summary>

```yaml
# Job: build
# Runner execution plan — read-only.

setup:
  - step: Setup job

main:
  - step: Run actions/checkout@v6
    uses: actions/checkout@v6
    if: success()
  - step: Cache Primes
    id: cache-primes
    uses: actions/cache@v5
    if: success()
    with:
      path: prime-numbers
      key: ${{ runner.os }}-primes
  - step: Check prerequisites
    run: |
      echo "Checking system requirements..."
    if: success()

post:
  - step: Post Cache Primes
    action: actions/cache@v5
  - step: Post Run actions/checkout@v6
    action: actions/checkout@v6

cleanup:
  - step: Complete job
```

</details>

## Why hand-emit the skeleton?

Per-entry line offsets must be tracked **at emission time**. Using a generic YAML serializer would force a second pass to scan the output for `- step:` lines, which is fragile and breaks the moment indentation conventions shift. Scalar values still go through the library (via `YamlScalarFormatter` from part 1), so we don't carry quoting rules.

## What's _not_ in this PR

- The stateful container that grows the view over a job's lifetime (part 3).
- The bridge that turns runner `IStep` instances into `JobExecutionViewEntry` (parts 4 & 5).
- The DAP wiring that actually serves the YAML as a `source` (final integration PR).

## Tests

Worked-example round-trip against a verbatim expected document; phase-section emission order; per-entry line offsets; special-character round-trip via `YamlDotNet` deserialization; multiline `run:` rendering; LF guarantee.

### DIFF
--- a/src/Runner.Worker/Dap/JobExecutionViewRenderer.cs
+++ b/src/Runner.Worker/Dap/JobExecutionViewRenderer.cs
@@ -1,0 +1,336 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Phase a step occupies in the runner's flat execution sequence.
+    /// Setup and Cleanup are NOT modeled here — they are synthetic
+    /// boundaries hard-coded by <see cref="JobExecutionViewRenderer"/>
+    /// and cannot be constructed by callers.
+    /// </summary>
+    internal enum JobExecutionPhase
+    {
+        Pre,
+        Main,
+        Post,
+    }
+
+    /// <summary>
+    /// One step in the rendered execution view. Pure data; no link to
+    /// any worker type. Phase 2 will translate runner step objects
+    /// into instances of this record.
+    /// </summary>
+    internal sealed class JobExecutionViewEntry
+    {
+        public JobExecutionViewEntry(
+            JobExecutionPhase phase,
+            string displayName,
+            string uses = null,
+            string run = null,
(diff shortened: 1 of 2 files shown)


rentziass:

- **run-service** will pre-render each step as a YAML list-item fragment and include it on the step contract sent to the runner.
- **The runner** will only stitch fragments together into the phase-keyed source document. No per-entry YAML emission on the runner side — just byte-level concatenation of pre-rendered fragments plus the synthetic Setup/Cleanup headers. Pre/Post sections reuse the corresponding Main fragment opaquely.

The renderer in this PR (entry types + phase-keyed YAML emitter + per-entry line tracking) is replaced by a much smaller runner-side stitcher. Closing rather than retargeting since the emission half is gone; only the "track which line each fragment starts on" half survives — and that's small enough to live with the state container.

PR #4433, opened 2026-05-19
DAP: add JobExecutionView state container (3/5)

rentziass:
> **Part 3 of 5** of the foundation for the DAP debugger's job source view. Stacks on #4432.
>

## What this PR adds

A stateful, append-only wrapper around the renderer (part 2). Maintains the current entry list, caches the most recent rendered YAML, and provides O(1) `IStep -> line` lookup for the DAP `stackTrace` handler.

## Why a separate container?

The renderer is pure: list of entries in → YAML + line offsets out. But the debugger's view grows over the job's lifetime — post steps register lazily, and the integration layer needs fast IStep→line lookup at every pause. This container owns that mutable state.

## API surface

Each `Append` can register an entry in one of three modes:

| Mode | Use case |
|---|---|
| `stepIdentity` non-null | Real `IStep` known at append time — registers `IStep → line` mapping immediately. |
| `matchKey` non-null | Unclaimed placeholder. A later `TryClaim(matchKey, IStep)` binds it to the real step. Used when an entry is predicted before the runner materializes its `IStep` (e.g. a Post-step placeholder synthesized at job-init from an action's metadata). |
| Neither | Static informational entry that needs no line lookup. |

`Append` and `AppendRange` never remove or reorder existing entries. `TryClaim` does not re-render. The `IStep → line` mapping is rebuilt on every render, so lookups stay accurate even if a later `Append` happens to shift previously-emitted entries.

## Tests

Constructor render of empty view; single/multi-append flow; line-number stability; concurrent appends; `TryClaim` happy path / unknown key / already-claimed / null guards; `Append` precondition rejecting both `stepIdentity` and `matchKey` together.

### DIFF
--- a/src/Runner.Worker/Dap/JobExecutionView.cs
+++ b/src/Runner.Worker/Dap/JobExecutionView.cs
@@ -1,0 +1,276 @@
+﻿using System;
+using System.Collections.Generic;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Stateful, append-only container that wraps <see cref="JobExecutionViewRenderer"/>
+    /// for runtime use. Maintains a mutable list of entries, caches the rendered YAML,
+    /// and provides O(1) lookup from <see cref="IStep"/> identity to the current line
+    /// in the rendered YAML where that step's <c>- step:</c> key appears.
+    ///
+    /// Each <see cref="Append"/> can register the entry in one of three modes:
+    ///   - With a non-null <c>stepIdentity</c>: registers the IStep→line mapping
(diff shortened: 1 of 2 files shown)


rentziass:
Closing as part of a strategy pivot — see the redirect notes on #4431, #4432, #4434, #4435.

The state container itself (append-only IStep→line lookup, `TryClaim` for predicted placeholders) survives in the new design, but its only caller — the runner-side renderer — was just retired in favour of run-service-emitted YAML fragments. With no caller and a stale dependency on closed code, this PR is parked in a half-useful state.

The contract is small enough (~280 lines) to reintroduce cleanly once the new run-service / runner contract is settled.

PR #4434, opened 2026-05-19
DAP: add TemplateTokenYamlAdapter (4/5)

rentziass:
> **Part 4 of 5** of the foundation for the DAP debugger's job source view. Stacks on #4433 only for branch ordering — this piece is functionally independent of parts 1-3.
>

## What this PR adds

A YamlDotNet `IObjectWriter` adapter that lets the runner's existing `TemplateWriter.Write` drive a YamlDotNet `Emitter` — i.e. serialize a `TemplateToken` tree to YAML in a single call.

## Why we need it

A step's parameters (`with:`, `env:`, `if:`, ...) arrive at the runner as `TemplateToken` trees with `${{ ... }}` expressions still embedded. The DAP execution view (the source the debugger serves) must reflect those parameters **as the user authored them — pre evaluation, with expressions intact** — so what the user sees in their debugger matches their workflow file.

## Notable details

- **`BasicExpressionToken` is walked via `ToDisplayString()`** instead of `ToString()`. The parser tokenizes composite scalars like `${{ runner.os }}-primes` as a single `BasicExpressionToken` whose internal expression is `format('{0}-primes', runner.os)`. `ToString()` would emit the rewritten form verbatim; `ToDisplayString()` reverses it back to the authored form.
- **LF line breaks forced** for the same reason as part 1.
- **Both `--- ` and `---\n` document-start prefixes are stripped** — sometimes YamlDotNet emits the marker on its own line for collection roots.

## What's _not_ in this PR

- The translator that calls this adapter to render a step's `with:`/`env:` — that's part 5.

## Tests

Scalar/mapping/sequence/null serialization; `BasicExpressionToken` round-trip including the `format(...)` rewrite case; indentation; trailing-newline stripping; LF guarantee; document-marker stripping for the empty-string edge case.

### DIFF
--- a/src/Runner.Worker/Dap/TemplateTokenYamlAdapter.cs
+++ b/src/Runner.Worker/Dap/TemplateTokenYamlAdapter.cs
@@ -1,0 +1,223 @@
+﻿using System;
+using System.Globalization;
+using System.IO;
+using GitHub.DistributedTask.ObjectTemplating;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using GitHub.Runner.Sdk;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Adapts a YamlDotNet <see cref="IEmitter"/> as a DT
+    /// <see cref="IObjectWriter"/> so a <see cref="TemplateToken"/> DOM
+    /// can be serialized back to YAML preserving its pre-evaluation form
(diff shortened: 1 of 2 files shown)


rentziass:

- **run-service** will pre-render each step as a YAML list-item fragment — including `with:` / `env:` with `${{ }}` expressions preserved. That's exactly the responsibility this adapter served on the runner side.
- **The runner** receives the pre-rendered fragment as an opaque string and stitches sections.

So `TemplateToken` → YAML emission is no longer the runner's job. The Go equivalent will live in run-service or in the workflow parser. Closing rather than retargeting since none of this code survives runner-side.

PR #4435, opened 2026-05-19
DAP: add StepEntryTranslator (5/5)

rentziass:
> **Part 5 of 5** of the foundation for the DAP debugger's job source view. Stacks on #4434.
>

## What this PR adds

Bridges the runner's `IStep` / `IActionRunner` types to the renderer's `JobExecutionViewEntry` (from part 2). Given a runtime step, produces the data the renderer needs to emit one entry in the execution view.

## Behaviour

- **Phase** comes from `ActionRunStage` / `IStep` type.
- **Filters out** `JobExtensionRunner` and other non-`IActionRunner` steps — those represent runner-internal scaffolding, not user-visible steps.
- **Filters auto-generated step IDs** (regex against `^__\d+$` and GUID-shaped strings) so only explicit `id:` fields surface in the rendered view.
- **`with:` / `env:`** are serialized via `TemplateTokenYamlAdapter` (part 4) so `${{ ... }}` expressions are preserved verbatim.
- **`run:` / `shell:` / `working-directory:`** are extracted from a script step's `Inputs` map using the constants defined in `PipelineConstants.ScriptStepInputs` (the runner stores these as camelCase `workingDirectory`, **not** the kebab-case spelling from workflow YAML — see `ActionManifestManagerWrapper.cs:244`).

## What's _not_ in this PR

- The wiring that actually flows runner steps into the execution view at runtime (the DAP integration PR — coming next).

## Tests

`JobExtensionRunner` is filtered; auto-generated and GUID-shaped IDs are filtered; explicit IDs are preserved; run-step inputs are extracted via `ScriptStepInputs` constants; action-step `with:` is rendered via the template-token adapter; run-step internal keys are defensively filtered from `with:`.

### DIFF
--- a/src/Runner.Worker/Dap/StepEntryTranslator.cs
+++ b/src/Runner.Worker/Dap/StepEntryTranslator.cs
@@ -1,0 +1,240 @@
+﻿using System;
+using System.Collections.Generic;
+using GitHub.DistributedTask.ObjectTemplating.Tokens;
+using GitHub.DistributedTask.Pipelines;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Worker.Dap
+{
+    /// <summary>
+    /// Translates runner <see cref="IStep"/> instances into pure-data
+    /// <see cref="JobExecutionViewEntry"/> records used by the DAP debugger
+    /// execution view. Filters out runner-internal steps (e.g.
+    /// <see cref="JobExtensionRunner"/>) so the rendered view only shows
+    /// user-visible workflow steps.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


rentziass:

- **run-service** has the same parsed step data this translator was bridging (`Reference`, `Inputs`, `Env`, `If`, ...) and will pre-render each step as a YAML list-item fragment directly.
- **The runner** no longer needs an `IStep` → `JobExecutionViewEntry` bridge — the fragments arrive over the wire, ready to splice.

So this translator is fully redundant under the new contract. Closing rather than retargeting since none of it survives.

PR #4457, opened 2026-05-27
feat: add self-repository action reference syntax

nodeselector:
## `$/` — self-referencing action syntax

Adds `$/path` as a new `uses:` syntax meaning "this repo, at this SHA, at the given subpath." The containing YAML file determines which repo and SHA — so `$/lib/helper` inside `acme/toolkit@v1` resolves to `acme/toolkit` at that SHA with path `lib/helper`, not the caller's repo.

### Why

Composite actions that ship multiple sub-actions in one repo currently hardcode `owner/repo@ref` to reference siblings. This creates a circular versioning problem — you can't cut a release tag until the YAML already references that tag. `$/` breaks the cycle: siblings reference each other by path, and the ref is inherited from whatever version the caller pinned.

### Feature flag

`actions_dollar_self_reference` — gates parsing (Launch → parser), server-side resolution (ARS), and runner-side resolution. Unified name across all four repos.

### How it works

**Parsing.** Both template converters (`PipelineTemplateConverter` for action.yml, `WorkflowTemplateConverter` for workflow YAML) detect `$/` via `PipelineConstants.TryParseDollarSelfReference()`. The parsed reference gets `RepositoryType = "selfRepository"` and `Path = <subpath>` — no `Name` or `Ref` yet. Leading slashes are stripped and bare `$/` (no subpath) is rejected.

**Resolution.** `ActionManager.ResolveDollarSelfReferences()` rewrites `selfRepository` refs to concrete `GitHub`-type refs in-place before they reach any download logic:

| Depth | Source of truth | Why |
|-------|----------------|-----|
| 0 (workflow steps) | `job.WorkflowRepository` / `job.WorkflowSha` | Points to the repo containing the workflow file. Correct for both regular and reusable workflows. |
| > 0 (inside composites) | Parent action's `RepositoryPathReference.Name` / `.Ref` | The parent is already resolved. `$/` in a child always means "the repo that contains me." |

After resolution, the ref is indistinguishable from a normal `owner/repo/path@sha` reference.

**Both resolution paths supported.** Works on both the batch resolution path (`PrepareActionsRecursiveAsync`, gated on `actions_batch_action_resolution`) and the legacy serial path (`PrepareActionsRecursiveLegacyAsync`). The batch path is not yet broadly rolled out, so legacy support is required.

**Upfront resolution (not lazy).** All `$/` refs are resolved during `PrepareActionsAsync` (Setup Job phase) before any step executes. Pre/post step registration depends on the complete recursive walk having finished, so lazy resolution would break it.

**Runtime re-resolution.** `LoadAction` re-parses action.yml from disk at execution time, producing fresh `RepositoryPathReference` objects with `RepositoryType = "selfRepository"`. `ResolveDollarSelfReferences` runs again using the parent's already-resolved `Name`/`Ref`. No network calls — the tarball is already cached.

**Safety net.** `GetDownloadInfoLookupKey` throws `InvalidOperationException` if a `selfRepository` ref reaches it unresolved.

### Cross-repo `$/` context propagation

```
private-server-1/actions/crossrepo-chain/action.yml
  └─ uses: github/public-server/actions/leaf@sha
       └─ leaf/action.yml says: uses: $/actions/other-leaf
          → resolves to public-server/actions/other-leaf@sha  ✅
          (NOT private-server-1 — $/ means the repo containing THIS file)
```

This works because `childRepoName`/`childRepoRef` at each depth come from the parent action's resolved `RepositoryPathReference.Name/.Ref`, not from the workflow-level context.

### Tests (7 total)

- `ResolvesAtDepthZero` — batch path, top-level `$/` resolution
- `ResolvesAtDepthZero_LegacyPath` — legacy path equivalent
- `NotResolvedWhenFeatureFlagDisabled` — `$/` stays unresolved, hits safety net
- `ResolvesNestedInComposite` — depth-1 `$/` inside a composite (batch)
- `ResolvesNestedInComposite_LegacyPath` — legacy path equivalent
- `CrossRepoCompositeResolvesToParentRepo` — `$/` resolves to parent's repo, not caller's
- `MultiLevelChain` — 3-level composite chain with `$/` at each depth

### DIFF
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -427,23 +427,12 @@ namespace GitHub.Runner.Worker
         /// sub-actions individually, with no cross-depth deduplication.
         /// Used when the BatchActionResolution feature flag is disabled.
         /// </summary>
-        private async Task<PrepareActionsState> PrepareActionsRecursiveLegacyAsync(IExecutionContext executionContext, PrepareActionsState state, IEnumerable<Pipelines.ActionStep> actions, Int32 depth = 0, Guid parentStepId = default(Guid), string dollarSelfRepoName = null, string dollarSelfRepoRef = null)
+        private async Task<PrepareActionsState> PrepareActionsRecursiveLegacyAsync(IExecutionContext executionContext, PrepareActionsState state, IEnumerable<Pipelines.ActionStep> actions, Int32 depth = 0, Guid parentStepId = default(Guid))
         {
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             if (depth > Constants.CompositeActionsMaxDepth)
             {
                 throw new Exception($"Composite action depth exceeded max depth {Constants.CompositeActionsMaxDepth}");
-            }
-
-            // Resolve dollar-self ($/path) references before processing
-            if (executionContext.Global.Variables.GetBoolean(Constants.Runner.Features.DollarSelfReference) == true)
-            {
-                if (string.IsNullOrEmpty(dollarSelfRepoName))
-                {
-                    dollarSelfRepoName = executionContext.JobContext?.WorkflowRepository;
-                    dollarSelfRepoRef = executionContext.JobContext?.WorkflowSha;
-                }
-                ResolveDollarSelfReferences(executionContext, actions, dollarSelfRepoName, dollarSelfRepoRef);
             }
 
             var repositoryActions = new List<Pipelines.ActionStep>();
@@ -570,18 +559,7 @@ namespace GitHub.Runner.Worker
                     }
                     else if (setupInfo != null && setupInfo.Steps != null && setupInfo.Steps.Count > 0)
                     {
-                        // Determine dollar-self resolution context for child steps
-                        string childRepoName = dollarSelfRepoName;
-                        string childRepoRef = dollarSelfRepoRef;
-                        var parentRef = action.Reference as Pipelines.RepositoryPathReference;
-                        if (parentRef != null &&
-                            string.Equals(parentRef.RepositoryType, Pipelines.RepositoryTypes.GitHub, StringComparison.OrdinalIgnoreCase))
-                        {
-                            childRepoName = parentRef.Name;
-                            childRepoRef = parentRef.Ref;
-                        }
-
-                        state = await PrepareActionsRecursiveLegacyAsync(executionContext, state, setupInfo.Steps, depth + 1, action.Id, childRepoName, childRepoRef);
+                        state = await PrepareActionsRecursiveLegacyAsync(executionContext, state, setupInfo.Steps, depth + 1, action.Id);
                     }
                     var repoAction = action.Reference as Pipelines.RepositoryPathReference;
                     if (repoAction.RepositoryType != Pipelines.PipelineConstants.SelfAlias)
--- a/src/Runner.Worker/ActionManager.cs
+++ b/src/Runner.Worker/ActionManager.cs
@@ -175,7 +175,6 @@ namespace GitHub.Runner.Worker
                 }
 #endif
             }
-
             return new PrepareResult(containerSetupSteps, result.PreStepTracker);
         }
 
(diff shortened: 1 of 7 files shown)


src/Runner.Worker/ActionManager.cs
```diff
@@ -175,17 +175,33 @@ public sealed class ActionManager : RunnerService, IActionManager
...
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             if (depth > Constants.CompositeActionsMaxDepth)
             {
                 throw new Exception($"Composite action depth exceeded max depth {Constants.CompositeActionsMaxDepth}");
             }
 
+            // Resolve dollar-self ($/path) references before processing
+            if (executionContext.Global.Variables.GetBoolean(Constants.Runner.Features.DollarSelfReference) == true)
+            {
+                if (string.IsNullOrEmpty(dollarSelfRepoName))
```
ericsciple: nit: have you considered describing this as "selfRepoName" instead?

"$/" is the syntax; self repo is the concept

ericsciple: Also consider comments distinguishing the existing "self" type as "self workspace" or "self local workspace"

src/Sdk/WorkflowParser/WorkflowConstants.cs
```diff
@@ -30,6 +30,11 @@ public static class WorkflowConstants
         /// </summary>
```
ericsciple, line 30: > Alias for the self repository.

maybe something like: "Alias for the self local-workspace repository type."

src/Runner.Worker/ActionManager.cs
```diff
@@ -278,16 +294,41 @@ public sealed class ActionManager : RunnerService, IActionManager
                 // then recurse per parent (which hits the cache, not the API).
                 if (nextLevel.Count > 0)
                 {
+                    // Pre-compute parent context per group for self-reference
+                    // resolution and recursion. Self-reference refs in nextLevel
+                    // must be resolved BEFORE batch download-info resolution,
+                    // since GetDownloadInfoLookupKey requires concrete refs.
+                    var groups = nextLevel.GroupBy(x => x.parentId).Select(group =>
```
ericsciple: Should this be guarded by the feature flag?

I realize this affects the foreach further below (L287/R329). Have you considered duplicating the relevant code (L281 - L291), so the change can be made safely behind a feature flag.

src/Runner.Worker/ActionManager.cs
```diff
@@ -175,17 +175,33 @@ public sealed class ActionManager : RunnerService, IActionManager
...
-        private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionContext executionContext, PrepareActionsState state, IEnumerable<Pipelines.ActionStep> actions, Dictionary<string, WebApi.ActionDownloadInfo> resolvedDownloadInfos, Int32 depth = 0, Guid parentStepId = default(Guid))
+        private async Task<PrepareActionsState> PrepareActionsRecursiveAsync(IExecutionContext executionContext, PrepareActionsState state, IEnumerable<Pipelines.ActionStep> actions, Dictionary<string, WebApi.ActionDownloadInfo> resolvedDownloadInfos, Int32 depth = 0, Guid parentStepId = default(Guid), string selfRepoName = null, string selfRepoRef = null)
         {
             ArgUtil.NotNull(executionContext, nameof(executionContext));
             if (depth > Constants.CompositeActionsMaxDepth)
             {
                 throw new Exception($"Composite action depth exceeded max depth {Constants.CompositeActionsMaxDepth}");
             }
 
+            // Resolve self-reference ($/) references before processing
```
ericsciple: With this PR there are now two "self" concepts.

This is the "self repository" flavor. Instead of describing this as "self-reference", have you considered "self repository" ?

Likewise, it may good to distinguish existing "self" code with comments clarifying it's the "self workspace" or "self local workspace".

nodeselector: Thought I got these renames earlier, turns out I did not push. Thanks!

src/Runner.Worker/ActionManager.cs
nodeselector: absolutely! good catch

src/Runner.Worker/ActionManager.cs
nodeselector: added some comments.

ericsciple (approved):

PR #4463, opened 2026-05-28
Populate telemetry for non-action post-job steps

drielenr:
## Summary
Populate `StepTelemetry.Type` and `StepTelemetry.Action` for non-action post-job steps. 

This is needed for snapshot requests, which are added as `RegisterPostJobStep`. Currently, `action_execution_history` telemetry type/action show as empty
 
issue:
- https://github.com/github/hosted-compute-ims/issues/1639

## Changes

- Set `Type = "runner"` and derive `Action` from `DisplayName` for post-job steps

Before:
image

After:
image

## Testing

- Added `RegisterPostJobStep_JobExtensionRunner_DefaultsRunnerTelemetry` test to verify that runner-owned post-job steps get default telemetry values.
- Added `RegisterPostJobStep_ActionRunner_DoesNotOverrideTelemetry` test to ensure that action post-job steps do not have their telemetry values pre-populated, allowing the handler to set them as needed.

### DIFF
--- a/src/Runner.Worker/ExecutionContext.cs
+++ b/src/Runner.Worker/ExecutionContext.cs
@@ -337,6 +337,11 @@ namespace GitHub.Runner.Worker
             }
 
             step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState, siblingScopeName);
+            if (step is not IActionRunner)
+            {
+                step.ExecutionContext.StepTelemetry.Type = "runner";
+                step.ExecutionContext.StepTelemetry.Action = step.DisplayName.ToLowerInvariant().Replace(' ', '_');
+            }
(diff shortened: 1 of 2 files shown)


src/Runner.Worker/ExecutionContext.cs
```diff
@@ -337,6 +337,11 @@ public void RegisterPostJobStep(IStep step)
             }
 
             step.ExecutionContext = Root.CreatePostChild(step.DisplayName, IntraActionState, siblingScopeName);
+            if (step is not IActionRunner)
```
TingluoHuang:
```suggestion
            if (step is JobExtensionRunner)
```

TingluoHuang (approved):

PR #4470, opened 2026-06-03
Add job execution view model

rentziass:
**Part 1 of 2** of adding a job execution view for the DAP debugger. Followed by #4471.

This PR introduces a typed `JobExecutionView` model so we can define and review the rendered job view separately from the debugger plumbing that will serve it.

The reason for splitting it this way is that the view has its own behavior and invariants:
- it needs to render a stable synthetic source for the job
- it needs to preserve source lines so the debugger can point at the right step later
- it needs to model pre/main/post sections, predicted post steps, and the synthetic `Complete job` row consistently

Putting that behavior behind a dedicated type keeps the follow-up DAP integration PR much smaller and lets us validate the output shape directly in focused tests.

Example rendered job view:

```yaml
pre:
  - step: "Setup job"
  - step: "Pre cache"

main:
  - step: "Checkout"
  - step: "Build"

post:
  - step: "Post cache"
  - step: "Complete job"
```

### DIFF
--- a/src/Runner.Worker/InternalsVisibleTo.cs
+++ b/src/Runner.Worker/InternalsVisibleTo.cs
@@ -1,4 +1,4 @@
-using System.Runtime.CompilerServices;
+﻿using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Test")]
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
--- a/src/Runner.Worker/InternalsVisibleTo.cs
+++ b/src/Runner.Worker/InternalsVisibleTo.cs
@@ -1,4 +1,4 @@
-﻿using System.Runtime.CompilerServices;
+using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Test")]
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/InternalsVisibleTo.cs
TingluoHuang, line 1: what is changed?

src/Runner.Worker/InternalsVisibleTo.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.Runtime.CompilerServices;
```
salmanmkc, line 1: what's this for, is this some encoding thing?

src/Runner.Worker/InternalsVisibleTo.cs
rentziass: Copilot axed a BOM, restoring

src/Runner.Worker/InternalsVisibleTo.cs
rentziass: Copilot axed a BOM, restoring

AllanGuigou (approved):

PR #4476, opened 2026-06-04
Background steps execution engine

lokesh755:
## Summary

Adds the core background step execution engine to the runner, enabling concurrent step execution within a single job. Background steps (`background: true`) run asynchronously while the foreground step loop continues. Control-flow steps (`wait`, `wait-all`, `cancel`) coordinate synchronization and lifecycle management.

### New files
- **`BackgroundStepCoordinator.cs`** — Coordinates background step execution, waiting, cancellation, and deferred state flushing. Manages a concurrency semaphore (default max 10). 
- **`ControlFlowStepData.cs`** — Data class for wait/wait-all/cancel steps. 
- **`BackgroundStepsL0.cs`** — Unit tests for background step coordination, waiting, cancellation, and failure propagation.

### Modified files
- **`StepsRunner.cs`** — Integrates `BackgroundStepCoordinator`; background steps are queued without blocking the main loop; control-flow steps are dispatched directly; implicit wait-all runs before post-hooks as a safety net.
- **`JobExtension.cs`** — Initializes control-flow steps from `BackgroundStepControl` pipeline type; maps logical step IDs to external IDs for UI reference; creates implicit wait-all for uncovered background steps.
- **`ExecutionContext.cs`** — Adds background step metadata as optional parameters to `CreateChild`

## Key design decisions

- **Runner owns concurrency** — run-service remains agnostic to step execution order
- **Non-blocking semaphore** — foreground steps execute immediately even when all background slots are occupied
- **Implicit wait-all** — safety net before post-hooks waits for any uncovered background steps

### DIFF
--- a/src/Runner.Worker/BackgroundStepCoordinator.cs
+++ b/src/Runner.Worker/BackgroundStepCoordinator.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using GitHub.DistributedTask.Pipelines.ContextData;
 using GitHub.DistributedTask.WebApi;
 using GitHub.Runner.Common;
 using GitHub.Runner.Common.Util;
@@ -143,51 +142,25 @@ namespace GitHub.Runner.Worker
         // Wait
         // -----------------------------------------------------------------
 
-        /// <summary>
-        /// Wait for specific background steps by ID. Flushes deferred state and
-        /// returns the aggregate result.
-        /// </summary>
-        public async Task<TaskResult> WaitForStepsAsync(string[] stepIds, CancellationToken cancellationToken)
+        private async Task<TaskResult> WaitForStepsAsync(string[] stepIds, CancellationToken cancellationToken)
         {
             var ids = stepIds ?? Array.Empty<string>();
-
             await WaitForStepTasksAsync(ids, cancellationToken);
-
-            foreach (var id in ids)
-            {
-                FlushDeferredState(id);
-                _waitedStepIds.Add(id);
-            }
-
-            return GetWaitResult(ids);
+            return CompleteWaitedSteps(ids);
         }
 
-        /// <summary>
-        /// Wait for all not-yet-waited background steps. Returns the aggregate result.
-        /// </summary>
-        public async Task<TaskResult> WaitForAllAsync(CancellationToken cancellationToken)
+        private async Task<TaskResult> WaitForAllAsync(CancellationToken cancellationToken)
         {
             var remaining = _backgroundSteps.Keys.Where(id => !_waitedStepIds.Contains(id)).ToList();
-
             await WaitForStepTasksAsync(remaining, cancellationToken);
-
-            foreach (var id in remaining)
-            {
-                FlushDeferredState(id);
-                _waitedStepIds.Add(id);
-            }
-
-            return GetWaitResult(remaining);
+            return CompleteWaitedSteps(remaining);
         }
 
         // -----------------------------------------------------------------
         // Cancel
         // -----------------------------------------------------------------
 
-        /// <summary>
-        /// Cancel specific background steps by ID. Flushes deferred state after cancellation.
-        /// </summary>
-        public async Task CancelStepsAsync(string[] cancelStepIds)
+        private async Task CancelStepsAsync(string[] cancelStepIds)
         {
             if (cancelStepIds == null || cancelStepIds.Length == 0) return;
 
@@ -208,12 +181,9 @@ namespace GitHub.Runner.Worker
         }
 
         // -----------------------------------------------------------------
-        // Failure propagation
+        // Safety net
         // -----------------------------------------------------------------
 
-        /// <summary>
-        /// Safety net: wait for any unwaited background steps and propagate failures.
-        /// </summary>
         public async Task WaitForUnwaitedStepsAsync(IExecutionContext jobContext)
         {
             var unwaitedIds = _backgroundSteps.Keys.Where(id => !_waitedStepIds.Contains(id)).ToList();
@@ -224,14 +194,8 @@ namespace GitHub.Runner.Worker
 
             Trace.Info($"Safety net: {unwaitedIds.Count} unwaited background step(s) at post-job boundary: {string.Join(", ", unwaitedIds)}");
             await WaitForAllAsync(jobContext.CancellationToken);
-            PropagateFailures(jobContext);
-        }
 
-        /// <summary>
-        /// Propagate any unhandled background step failures to the job result.
-        /// </summary>
-        public void PropagateFailures(IExecutionContext jobContext)
-        {
+            // Propagate any background step failures to the job result
             foreach (var (_, (step, _, _)) in _backgroundSteps)
             {
                 if (step.ExecutionContext.Result == TaskResult.Failed)
@@ -347,19 +311,19 @@ namespace GitHub.Runner.Worker
             }
         }
 
-        private TaskResult GetWaitResult(IEnumerable<string> stepIds)
+        private TaskResult CompleteWaitedSteps(IEnumerable<string> stepIds)
         {
-            if (stepIds == null) return TaskResult.Succeeded;
-
-            foreach (var stepId in stepIds)
+            var result = TaskResult.Succeeded;
+            foreach (var id in stepIds)
             {
-                if (_backgroundSteps.TryGetValue(stepId, out var entry) && entry.Step.ExecutionContext.Result == TaskResult.Failed)
+                FlushDeferredState(id);
+                _waitedStepIds.Add(id);
+                if (_backgroundSteps.TryGetValue(id, out var entry) && entry.Step.ExecutionContext.Result == TaskResult.Failed)
                 {
-                    Trace.Info($"Background step '{stepId}' failed.");
-                    return TaskResult.Failed;
+                    result = TaskResult.Failed;
                 }
             }
-            return TaskResult.Succeeded;
+            return result;
         }
     }
 }
--- a/src/Runner.Worker/BackgroundStepCoordinator.cs
+++ b/src/Runner.Worker/BackgroundStepCoordinator.cs
@@ -1,0 +1,365 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using GitHub.DistributedTask.Pipelines.ContextData;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Common;
+using GitHub.Runner.Common.Util;
+using GitHub.Runner.Sdk;
+using Pipelines = GitHub.DistributedTask.Pipelines;
+
+namespace GitHub.Runner.Worker
+{
+    [ServiceLocator(Default = typeof(BackgroundStepCoordinator))]
+    public interface IBackgroundStepCoordinator : IRunnerService
+    {
+        void InitializeCoordinator(int maxConcurrent);
+        void StartBackgroundStep(IStep step, CancellationToken jobCancellationToken);
+        Task WaitForUnwaitedStepsAsync(IExecutionContext jobContext);
+        Task RunControlFlowAsync(IExecutionContext stepContext, object data);
+    }
+
+    /// <summary>
+    /// Coordinates background step execution, waiting, cancellation, and deferred state.
+    /// Extracted from StepsRunner so the main step loop stays clean.
+    /// </summary>
+    public sealed class BackgroundStepCoordinator : RunnerService, IBackgroundStepCoordinator
+    {
+        private const int DefaultMaxBackgroundSteps = 10;
+        private readonly Dictionary<string, (IStep Step, Task Task, CancellationTokenSource Cts)> _backgroundSteps = new();
+        private readonly HashSet<string> _waitedStepIds = new();
+        private SemaphoreSlim _backgroundSlotSemaphore = new SemaphoreSlim(DefaultMaxBackgroundSteps);
+
+        /// <summary>
+        /// Reset per-job state. Call at the start of each job.
+        /// </summary>
+        public void InitializeCoordinator(int maxConcurrent)
+        {
+            _backgroundSteps.Clear();
+            _waitedStepIds.Clear();
+            var max = maxConcurrent > 0 ? maxConcurrent : DefaultMaxBackgroundSteps;
+            _backgroundSlotSemaphore = new SemaphoreSlim(max);
+        }
+
+        // -----------------------------------------------------------------
+        // Starting background steps
+        // -----------------------------------------------------------------
+
+        /// <summary>
+        /// Prepare and launch a background step. Does not block the caller.
+        /// </summary>
+        public void StartBackgroundStep(IStep step, CancellationToken jobCancellationToken)
+        {
+            var stepId = step.ExecutionContext?.ContextName ?? step.DisplayName;
+
+            // Isolate GitHubContext so concurrent steps don't overwrite each other's GITHUB_OUTPUT paths
+            if (step.ExecutionContext.ExpressionValues.TryGetValue("github", out var ghCtx) && ghCtx is GitHubContext sharedGitHub)
+            {
+                step.ExecutionContext.ExpressionValues["github"] = sharedGitHub.ShallowCopy();
+            }
+
+            var bgCts = CancellationTokenSource.CreateLinkedTokenSource(jobCancellationToken);
+
+            // Evaluate timeout on the main thread (needs expression context)
+            var timeoutMinutes = 0;
+            try
+            {
+                var templateEvaluator = step.ExecutionContext.ToPipelineTemplateEvaluator();
+                timeoutMinutes = templateEvaluator.EvaluateStepTimeout(step.Timeout, step.ExecutionContext.ExpressionValues, step.ExecutionContext.ExpressionFunctions);
+            }
+            catch (Exception ex)
+            {
+                Trace.Info($"Error determining timeout for background step '{stepId}': {ex.Message}");
+            }
+
+            var task = ExecuteBackgroundStepCoreAsync(step, bgCts, stepId, timeoutMinutes);
+            _backgroundSteps[stepId] = (step, task, bgCts);
+            Trace.Info($"Background step '{stepId}' queued (slot will be acquired asynchronously).");
+        }
+
+        private async Task ExecuteBackgroundStepCoreAsync(
+            IStep step, CancellationTokenSource bgCts,
+            string stepId, int timeoutMinutes)
+        {
+            Trace.Info($"Background step '{stepId}' waiting for slot.");
+            await _backgroundSlotSemaphore.WaitAsync(bgCts.Token);
+            Trace.Info($"Background step '{stepId}' acquired slot.");
+
+            step.ExecutionContext.Start();
+
+            if (timeoutMinutes > 0)
+            {
+                step.ExecutionContext.SetTimeout(TimeSpan.FromMinutes(timeoutMinutes));
+            }
+
+            using var cancelReg = bgCts.Token.Register(() =>
+            {
+                Trace.Info($"Background step '{stepId}': cancellation signalled, sending CancelToken to process.");
+                step.ExecutionContext.CancelToken();
+            });
+
+            TaskResult? result = null;
+            try
+            {
+                await step.RunAsync();
+                result = step.ExecutionContext.Result ?? TaskResult.Succeeded;
+            }
+            catch (OperationCanceledException) when (bgCts.Token.IsCancellationRequested)
+            {
+                result = TaskResult.Canceled;
+            }
+            catch (OperationCanceledException) when (step.ExecutionContext.CancellationToken.IsCancellationRequested)
+            {
+                Trace.Info($"Background step '{stepId}' timed out after {timeoutMinutes} minutes.");
+                step.ExecutionContext.Error($"The background step '{step.DisplayName}' has timed out after {timeoutMinutes} minutes.");
+                result = TaskResult.Failed;
+            }
+            catch (Exception ex)
+            {
+                Trace.Info($"Background step '{stepId}' failed: {ex.Message}");
+                step.ExecutionContext.Error(ex);
+                result = TaskResult.Failed;
+            }
+            finally
+            {
+                _backgroundSlotSemaphore.Release();
+
+                if (step.ExecutionContext.CommandResult != null)
+                {
+                    result = TaskResultUtil.MergeTaskResults(result, step.ExecutionContext.CommandResult.Value);
+                }
+
+                step.ExecutionContext.Result = result;
+                step.ExecutionContext.ApplyContinueOnError(step.ContinueOnError);
+
+                step.ExecutionContext.Complete(step.ExecutionContext.Result);
+                Trace.Info($"Background step '{stepId}' completed with result: {step.ExecutionContext.Result}");
+            }
+        }
+
+        // -----------------------------------------------------------------
+        // Wait
+        // -----------------------------------------------------------------
+
+        /// <summary>
+        /// Wait for specific background steps by ID. Flushes deferred state and
+        /// returns the aggregate result.
+        /// </summary>
+        public async Task<TaskResult> WaitForStepsAsync(string[] stepIds, CancellationToken cancellationToken)
+        {
+            var ids = stepIds ?? Array.Empty<string>();
+
+            await WaitForStepTasksAsync(ids, cancellationToken);
+
+            foreach (var id in ids)
+            {
+                FlushDeferredState(id);
+                _waitedStepIds.Add(id);
+            }
+
+            return GetWaitResult(ids);
+        }
+
+        /// <summary>
+        /// Wait for all not-yet-waited background steps. Returns the aggregate result.
+        /// </summary>
+        public async Task<TaskResult> WaitForAllAsync(CancellationToken cancellationToken)
+        {
+            var remaining = _backgroundSteps.Keys.Where(id => !_waitedStepIds.Contains(id)).ToList();
+
+            await WaitForStepTasksAsync(remaining, cancellationToken);
+
+            foreach (var id in remaining)
+            {
+                FlushDeferredState(id);
+                _waitedStepIds.Add(id);
+            }
+
+            return GetWaitResult(remaining);
+        }
+
+        // -----------------------------------------------------------------
+        // Cancel
+        // -----------------------------------------------------------------
+
+        /// <summary>
+        /// Cancel specific background steps by ID. Flushes deferred state after cancellation.
+        /// </summary>
+        public async Task CancelStepsAsync(string[] cancelStepIds)
+        {
+            if (cancelStepIds == null || cancelStepIds.Length == 0) return;
+
+            var idsToCancel = cancelStepIds
+                .Where(id => _backgroundSteps.ContainsKey(id) && !_backgroundSteps[id].Task.IsCompleted)
+                .ToArray();
+
+            if (idsToCancel.Length > 0)
+            {
+                Trace.Info($"Cancelling {idsToCancel.Length} background step(s): {string.Join(", ", idsToCancel)}");
+                await CancelWithGracePeriodAsync(idsToCancel);
+            }
+
+            foreach (var id in cancelStepIds)
+            {
+                FlushDeferredState(id);
+            }
+        }
+
+        // -----------------------------------------------------------------
+        // Failure propagation
+        // -----------------------------------------------------------------
+
+        /// <summary>
+        /// Safety net: wait for any unwaited background steps and propagate failures.
+        /// </summary>
+        public async Task WaitForUnwaitedStepsAsync(IExecutionContext jobContext)
+        {
+            var unwaitedIds = _backgroundSteps.Keys.Where(id => !_waitedStepIds.Contains(id)).ToList();
+            if (unwaitedIds.Count == 0)
+            {
+                return;
+            }
+
+            Trace.Info($"Safety net: {unwaitedIds.Count} unwaited background step(s) at post-job boundary: {string.Join(", ", unwaitedIds)}");
+            await WaitForAllAsync(jobContext.CancellationToken);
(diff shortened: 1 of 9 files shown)


src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,338 @@
...
+using GitHub.Runner.Common.Util;
+using GitHub.Runner.Sdk;
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// Coordinates background step execution, waiting, cancellation, and deferred state.
+    /// Extracted from StepsRunner so the main step loop stays clean.
+    /// </summary>
+    public sealed class BackgroundStepCoordinator
```
TingluoHuang: this should use the `RunnerService`, ex: `public sealed class BackgroundStepCoordinator: RunnerService, IBackgroundStepCoordinator`.
So you can use `HostContext.Get<IBackgroundStepCoordinator>()` everywhere else.

src/Runner.Worker/JobRunner.cs
TingluoHuang, line 1: undo this changes.

src/Runner.Worker/JobExtension.cs
```diff
@@ -345,6 +345,33 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                        else if (step.Type == Pipelines.StepType.BackgroundStepControl)
+                        {
+                            var ctrl = step as Pipelines.BackgroundStepControl;
+                            Trace.Info($"Adding {ctrl.ControlType} step for: {string.Join(", ", ctrl.StepIds ?? Array.Empty<string>())}");
+                            var controlType = ctrl.ControlType switch
+                            {
+                                Pipelines.BackgroundControlTypes.Wait => ControlFlowType.Wait,
+                                Pipelines.BackgroundControlTypes.WaitAll => ControlFlowType.WaitAll,
+                                Pipelines.BackgroundControlTypes.Cancel => ControlFlowType.Cancel,
+                                _ => ControlFlowType.Wait,
```
TingluoHuang: should we ignore or error the step if the type is unexpected?

src/Runner.Worker/JobExtension.cs
```diff
@@ -345,6 +345,33 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                            var data = new ControlFlowStepData
+                            {
+                                Type = controlType,
+                                StepId = step.Id,
+                                StepName = step.Name,
+                                StepIds = ctrl.StepIds,
+                                ParallelGroupId = ctrl.ParallelGroupId,
+                            };
+                            jobSteps.Add(new JobExtensionRunner(
+                                runAsync: (ctx, _) => data.RunAsync(ctx),
```
TingluoHuang: we should provide a func that take the data as input, similar to how containerprovide was used.

src/Runner.Worker/JobExtension.cs
```diff
@@ -400,16 +427,108 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                            if (actionStep.Action?.Background == true)
+                            {
+                                hasBackgroundSteps = true;
+                                var externalId = actionStep.Action.Id.ToString("N");
+                                contextNameToExternalId[actionStep.Action.ContextName] = externalId;
+                                backgroundStepExternalIds.Add(externalId);
+                                actionStep.ExecutionContext.SetBackgroundStepMetadata(isBackground: true, parallelGroupId: actionStep.Action.ParallelGroupId);
+                            }
+                        }
+                        else if (step is JobExtensionRunner { Data: ControlFlowStepData cf })
```
TingluoHuang: what is this mean?
i would expect we have `step is JobExtensionRunner runnerStep`, and we base on the `runnerStep.Data` to decide it's the background control step?

src/Runner.Worker/JobExtension.cs
```diff
@@ -400,16 +427,108 @@ public async Task<List<IStep>> InitializeJob(IExecutionContext jobContext, Pipel
...
+                                    .Select(id => contextNameToExternalId[id])
+                                    .ToArray();
+                            }
+
+                            var controlTypeStr = cf.Type switch
+                            {
+                                ControlFlowType.Wait => Pipelines.BackgroundControlTypes.Wait,
+                                ControlFlowType.WaitAll => Pipelines.BackgroundControlTypes.WaitAll,
+                                ControlFlowType.Cancel => Pipelines.BackgroundControlTypes.Cancel,
+                                _ => Pipelines.BackgroundControlTypes.Wait,
```
TingluoHuang: same question, should we error/skip unexpected step?

src/Runner.Worker/BackgroundStepContext.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Tracks a background step's execution state.
+    /// </summary>
+    internal sealed class BackgroundStepContext
+    {
+        public string StepId { get; }
+        public IStep Step { get; }
+        public Task ExecutionTask { get; set; }
+        public CancellationTokenSource Cts { get; set; }
+        public GitHub.DistributedTask.WebApi.TaskResult? Result { get; set; }
```
TingluoHuang: you can also get Result from the `Step.ExecutionContext`

src/Runner.Worker/BackgroundStepContext.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// Tracks a background step's execution state.
+    /// </summary>
+    internal sealed class BackgroundStepContext
+    {
+        public string StepId { get; }
+        public IStep Step { get; }
```
TingluoHuang: it feels odd we expose IStep and things can be used under IStep.

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+using System.Threading.Tasks;
+using GitHub.DistributedTask.WebApi;
+using GitHub.Runner.Common.Util;
+
+namespace GitHub.Runner.Worker
+{
+    public enum ControlFlowType
```
TingluoHuang: any existing enum you can reuse?
i think you aded some recently?

src/Runner.Worker/FileCommandManager.cs
TingluoHuang, line 1: do we also need to change the old `IActionCommandManager` since it also have code for setting env, etc.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -100,11 +107,18 @@ public interface IExecutionContext : IRunnerService
...
+        void FlushDeferredOutputs();
+        void FlushDeferredEnvironment();
+        void FlushDeferredOutcomeConclusion();
+
         void AddIssue(Issue issue, ExecutionContextLogOptions logOptions);
         void Progress(int percentage, string currentOperation = null);
         void UpdateDetailTimelineRecord(TimelineRecord record);
 
         void UpdateTimelineRecordDisplayName(string displayName);
+        void SetBackgroundStepMetadata(bool isBackground = false, string backgroundControlType = null, string[] backgroundControlStepIds = null, string parallelGroupId = null);
```
TingluoHuang: should this just be part of `CreateChild`?

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,80 @@
...
+        Cancel,
+    }
+
+    /// <summary>
+    /// Data for control-flow steps (wait, wait-all, cancel).
+    /// Used with JobExtensionRunner instead of dedicated IStep implementations.
+    /// The Coordinator is set by StepsRunner before execution so the step can
+    /// interact with background step state.
+    /// </summary>
+    public sealed class ControlFlowStepData
```
TingluoHuang: it feels like this can somehow merged with BackgroundStepContext.

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,338 @@
...
+            var stepId = step.ExecutionContext?.ContextName ?? step.DisplayName;
+
+            // Isolate GitHubContext so concurrent steps don't overwrite each other's GITHUB_OUTPUT paths
+            if (step.ExecutionContext.ExpressionValues.TryGetValue("github", out var ghCtx) && ghCtx is GitHubContext sharedGitHub)
+            {
+                step.ExecutionContext.ExpressionValues["github"] = sharedGitHub.ShallowCopy();
+            }
+
+            // Defer outputs, env, path, outcome — flushed at wait/wait-all
+            step.ExecutionContext.DeferredOutputs = new Dictionary<string, string>();
```
TingluoHuang: should this initialized during executioncontext creation?

src/Runner.Worker/BackgroundStepContext.cs
lokesh755: removed

src/Runner.Worker/FileCommandManager.cs
lokesh755: updated in different PR

src/Runner.Worker/ExecutionContext.cs
lokesh755: changed..

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: removed backgroundStepContext

src/Runner.Worker/BackgroundStepContext.cs
lokesh755: removed this file

src/Runner.Worker/JobRunner.cs
lokesh755: done..

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: moved

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: updated to use enum under pipelines

src/Runner.Worker/JobExtension.cs
lokesh755: throwing error now

src/Runner.Worker/JobExtension.cs
lokesh755: updated to do explicit cast

src/Runner.Worker/JobExtension.cs
lokesh755: changed to use func instead of lambda.

src/Runner.Worker/JobExtension.cs
lokesh755: throwing error

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: updated to use interface.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,12 +35,20 @@ public interface IStepsRunner : IRunnerService
 
     public sealed class StepsRunner : RunnerService, IStepsRunner
     {
+        private IBackgroundStepCoordinator _bgCoordinator;
```
TingluoHuang: you don't need this anymore, can you have just do `var _bgCoordinator = HostContext.GetService<IBackgroundStepCoordinator>();`

src/Runner.Worker/StepsRunner.cs
```diff
@@ -35,12 +35,20 @@ public interface IStepsRunner : IRunnerService
...
         public async Task RunAsync(IExecutionContext jobContext)
         {
             ArgUtil.NotNull(jobContext, nameof(jobContext));
             ArgUtil.NotNull(jobContext.JobSteps, nameof(jobContext.JobSteps));
 
+            // Initialize background step coordinator
+            _bgCoordinator = HostContext.GetService<IBackgroundStepCoordinator>();
+            var maxBgSteps = jobContext.Global.Variables.GetInt("system.runner.maxbackgroundsteps");
+            var maxConcurrent = (maxBgSteps.HasValue && maxBgSteps.Value > 0) ? maxBgSteps.Value : 10;
+            _bgCoordinator.InitializeCoordinator(maxConcurrent);
```
TingluoHuang: i would suggest we move these initialization into jobextension after we realize we have any background step.

lokesh755: updated

src/Runner.Worker/StepsRunner.cs
lokesh755: updated

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,87 @@
...
+                    await Coordinator.CancelStepsAsync(StepIds);
+                    stepContext.Result = TaskResult.Succeeded;
+                    break;
+
+                default:
+                    throw new ArgumentException($"Unknown background step control type '{Type}'.");
+            }
+        }
+    }
+}
```
TingluoHuang: should these be part of Coordinator?

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,87 @@
...
+            if (Coordinator == null) return;
+
+            switch (Type)
+            {
+                case Pipelines.BackgroundControlTypes.Wait:
+                    var waitResult = await Coordinator.WaitForStepsAsync(StepIds, stepContext.CancellationToken);
+                    stepContext.Result = waitResult;
+                    if (waitResult == TaskResult.Failed && JobContext != null)
+                    {
+                        JobContext.Result = TaskResultUtil.MergeTaskResults(JobContext.Result, TaskResult.Failed);
```
TingluoHuang: we should not modify JobContext.Result in here.
If you throw exception here, i think the steprunner can properly set the job to fail as needed.

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,87 @@
...
+                    if (waitAllResult == TaskResult.Failed && JobContext != null)
+                    {
+                        JobContext.Result = TaskResultUtil.MergeTaskResults(JobContext.Result, TaskResult.Failed);
+                        JobContext.JobContext.Status = JobContext.Result?.ToActionResult();
+                    }
+                    break;
+
+                case Pipelines.BackgroundControlTypes.Cancel:
+                    await Coordinator.CancelStepsAsync(StepIds);
+                    stepContext.Result = TaskResult.Succeeded;
```
TingluoHuang: i think Result is default to Succeeded already.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -68,12 +74,32 @@ public async Task RunAsync(IExecutionContext jobContext)
...
 
                 Trace.Info($"Processing step: DisplayName='{step.DisplayName}'");
+
+                // Control-flow steps (wait, cancel, etc.) run their own logic directly.
+                // They don't need condition evaluation, expression setup, or timeout handling.
+                if (step is JobExtensionRunner runnerStep && runnerStep.Data is ControlFlowStepData cf)
+                {
+                    cf.Coordinator = _bgCoordinator;
+                    cf.JobContext = jobContext;
+                    step.ExecutionContext.Start();
```
TingluoHuang: why do we need to `step.ExecutionContext.Start();`? control step are not background, so line 101 should take care of this already.

src/Runner.Worker/StepsRunner.cs
```diff
@@ -68,12 +74,32 @@ public async Task RunAsync(IExecutionContext jobContext)
...
+                    await step.RunAsync();
```
TingluoHuang: 🤔 i am not sure whether we want to special case the control step like this.

src/Runner.Worker/ExecutionContext.cs
```diff
@@ -82,7 +82,7 @@ public interface IExecutionContext : IRunnerService
         // Initialize
         void InitializeJob(Pipelines.AgentJobRequestMessage message, CancellationToken token);
         void CancelToken();
-        IExecutionContext CreateChild(Guid recordId, string displayName, string refName, string scopeName, string contextName, ActionRunStage stage, Dictionary<string, string> intraActionState = null, int? recordOrder = null, IPagingLogger logger = null, bool isEmbedded = false, List<Issue> embeddedIssueCollector = null, CancellationTokenSource cancellationTokenSource = null, Guid embeddedId = default(Guid), string siblingScopeName = null, TimeSpan? timeout = null);
```
TingluoHuang, line 85: expose IsBackground as part of the ExecutionContext

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: moved to cordinator

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: updated stepsrunner to run control steps in normal flow.. no longer need to update jobContext.

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: Result is null by default. But it's updated to `Succeeded` when null.. keeping it just to be safe :)

src/Runner.Worker/StepsRunner.cs
lokesh755: removed this short circuit, executing control steps as regular steps flow.

src/Runner.Worker/StepsRunner.cs
lokesh755: removed..

src/Runner.Worker/ExecutionContext.cs
lokesh755: added isBackground as part of ExecutionContext

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,379 @@
...
+using Pipelines = GitHub.DistributedTask.Pipelines;
+
+namespace GitHub.Runner.Worker
+{
+    [ServiceLocator(Default = typeof(BackgroundStepCoordinator))]
+    public interface IBackgroundStepCoordinator : IRunnerService
+    {
+        bool HasActiveBackgroundSteps { get; }
+        bool HasUnwaitedSteps { get; }
+        IReadOnlyList<string> UnwaitedStepIds { get; }
```
TingluoHuang: i didn't see where these 3 is used outside of the `BackgroundStepCoordinator`, maybe we should remove them from the interface.

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,379 @@
...
+        void InitializeCoordinator(int maxConcurrent);
+        void StartBackgroundStep(IStep step, CancellationToken jobCancellationToken);
+        Task<TaskResult> WaitForStepsAsync(string[] stepIds, CancellationToken cancellationToken);
+        Task<TaskResult> WaitForAllAsync(CancellationToken cancellationToken);
+        Task CancelStepsAsync(string[] cancelStepIds);
+        void PropagateFailures(IExecutionContext jobContext);
+        Task WaitForUnwaitedStepsAsync(IExecutionContext jobContext);
```
TingluoHuang: same for these, do we need to expose them in the interface if nothing calling them.

lokesh755: cleaned up interface

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+        private readonly HashSet<string> _waitedStepIds = new();
+        private SemaphoreSlim _backgroundSlotSemaphore = new SemaphoreSlim(DefaultMaxBackgroundSteps);
+
+        /// <summary>
+        /// Reset per-job state. Call at the start of each job.
+        /// </summary>
+        public void InitializeCoordinator(int maxConcurrent)
+        {
+            _backgroundSteps.Clear();
+            _waitedStepIds.Clear();
```
TingluoHuang: is `_waitedStepIds` track the completed steps or uncompleted steps?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+            await WaitForStepTasksAsync(ids, cancellationToken);
+            return CompleteWaitedSteps(ids);
+        }
+
+        private async Task<TaskResult> WaitForAllAsync(CancellationToken cancellationToken)
+        {
+            var remaining = _backgroundSteps.Keys.Where(id => !_waitedStepIds.Contains(id)).ToList();
+            await WaitForStepTasksAsync(remaining, cancellationToken);
+            return CompleteWaitedSteps(remaining);
+        }
```
TingluoHuang: do we still want to have 2 extra private methods? it feels like we can just inline the code to where it used?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+            {
+                Trace.Info($"Error determining timeout for background step '{stepId}': {ex.Message}");
+            }
+
+            var task = ExecuteBackgroundStepCoreAsync(step, bgCts, stepId, timeoutMinutes);
+            _backgroundSteps[stepId] = (step, task, bgCts);
+            Trace.Info($"Background step '{stepId}' queued (slot will be acquired asynchronously).");
+        }
+
+        private async Task ExecuteBackgroundStepCoreAsync(
```
TingluoHuang, line 192: we can probably move all private method under the `Private helpers` section down.

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+            return CompleteWaitedSteps(remaining);
+        }
+
+        // -----------------------------------------------------------------
+        // Cancel
+        // -----------------------------------------------------------------
+
+        private async Task CancelStepsAsync(string[] cancelStepIds)
+        {
+            if (cancelStepIds == null || cancelStepIds.Length == 0) return;
```
TingluoHuang:
```suggestion
            if (cancelStepIds == null || cancelStepIds.Length == 0)
            {
                return;
            }
```

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            if (idsToCancel.Length > 0)
+            {
+                Trace.Info($"Cancelling {idsToCancel.Length} background step(s): {string.Join(", ", idsToCancel)}");
+                await CancelWithGracePeriodAsync(idsToCancel);
+            }
+
+            foreach (var id in cancelStepIds)
+            {
+                FlushDeferredState(id);
```
TingluoHuang: should we call CompleteWaitedSteps() instead?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+            Trace.Info($"Safety net: {unwaitedIds.Count} unwaited background step(s) at post-job boundary: {string.Join(", ", unwaitedIds)}");
+            await WaitForAllAsync(jobContext.CancellationToken);
+
+            // Propagate any background step failures to the job result
+            foreach (var (_, (step, _, _)) in _backgroundSteps)
+            {
+                if (step.ExecutionContext.Result == TaskResult.Failed)
+                {
+                    Trace.Info($"Propagating failure from background step '{step.ExecutionContext.ContextName}' to job result.");
+                    jobContext.Result = TaskResultUtil.MergeTaskResults(jobContext.Result, TaskResult.Failed);
```
TingluoHuang: can we defer the caller in steprunner to update jobcontext.result?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+        // -----------------------------------------------------------------
+        // Control-flow step dispatch
+        // -----------------------------------------------------------------
+
+        /// <summary>
+        /// Execute a control-flow step (wait, wait-all, cancel) and propagate results.
+        /// </summary>
+        public async Task RunControlFlowAsync(IExecutionContext stepContext, object data)
+        {
+            var controlFlow = data as ControlFlowStepData;
```
TingluoHuang: do we need to print anything to the output, so customer can know what we are doing under the step?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            if (tasks.Count > 0)
+            {
+                try
+                {
+                    await Task.WhenAll(tasks).WaitAsync(TimeSpan.FromSeconds(graceSeconds));
+                }
+                catch (TimeoutException)
+                {
+                    Trace.Info($"Some background steps did not terminate within {graceSeconds}s grace period.");
```
TingluoHuang, line 326: should we mark all steps that haven't finish as canceled?

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        private TaskResult CompleteWaitedSteps(IEnumerable<string> stepIds)
+        {
+            var result = TaskResult.Succeeded;
+            foreach (var id in stepIds)
+            {
+                FlushDeferredState(id);
+                _waitedStepIds.Add(id);
+                if (_backgroundSteps.TryGetValue(id, out var entry) && entry.Step.ExecutionContext.Result == TaskResult.Failed)
```
TingluoHuang: why we only care about TaskResult.Failed? what about canceled, etc.

src/Runner.Worker/ControlFlowStepData.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace GitHub.Runner.Worker
+{
+    /// <summary>
+    /// Pure data for control-flow steps (wait, wait-all, cancel).
+    /// Type uses Pipelines.BackgroundControlTypes string constants.
+    /// </summary>
+    public sealed class ControlFlowStepData
```
TingluoHuang: should we rename this `BackgroundStepControlFlowData`?

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: renamed to _completedStepIds for clarity.

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: calling them inline now..

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: done

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: done

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: updated

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: deferred to steps runner for updating jobcontext result

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: Good catch!

src/Runner.Worker/ControlFlowStepData.cs
lokesh755: updated

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: added few log statements

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,369 @@
...
+                    stepContext.Result = CompleteWaitedSteps(remaining);
+                    ReportCompletedSteps(stepContext, "Finished waiting for all background step(s).", remaining);
+                    break;
+                }
+
+                case Pipelines.BackgroundControlTypes.Cancel:
+                {
+                    var cancelIds = controlFlow.StepIds ?? Array.Empty<string>();
+                    stepContext.Output($"Cancelling background step(s): {DescribeSteps(cancelIds)}");
+                    await CancelStepsAsync(controlFlow.StepIds);
```
TingluoHuang, line 154: for the `Cancel` step, do we want to consider the cancel "failed" if not all steps are properly canceled within timeout?
i am fine either way, just want to bring this up.

src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -0,0 +1,369 @@
...
+                        {
+                            step.ExecutionContext.Result = TaskResult.Canceled;
+                            Trace.Info($"Background step '{stepId}' did not terminate within grace period; marking as canceled.");
+                        }
+                    }
+                }
+            }
+        }
+
+        private void FlushDeferredState(string stepId)
```
TingluoHuang: maybe just inline `FlushDeferredState` into `CompleteWaitedSteps` since that's the only place this code is needed?

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: I'd keep cancel itself as Succeeded as long as we issued the cancellation

src/Runner.Worker/BackgroundStepCoordinator.cs
lokesh755: updated

TingluoHuang (approved):

PR #4482, opened 2026-06-08
Canceled background steps should not impact job result

lokesh755:
## Summary

Canceled background steps should not impact the job result. A background step that is explicitly canceled by a `cancel` control step is *expected* to be canceled, so its `Canceled` result no longer influences the overall job result. A genuinely **failed** step still counts, even when it was targeted by a cancel (e.g. it failed before the cancellation took effect).

This also splits `BackgroundStepCoordinator.WaitForUnwaitedStepsAsync` into two members so each has a single responsibility:

- `WaitForUnwaitedStepsAsync` — drains any unwaited background steps at the post-job boundary.
- `GetAggregatedResult()` — folds the background step results into a single `TaskResult` for the caller.

## Changes

- Track explicitly-canceled step IDs (`_explicitlyCanceledStepIds`) so a `Canceled` outcome from a `cancel` step is skipped during result aggregation, while `Failed` outcomes still propagate.
- Split `WaitForUnwaitedStepsAsync` (drain) from `GetAggregatedResult()` (aggregation); updated the `StepsRunner` call site to call both in sequence.

### DIFF
--- a/src/Runner.Worker/BackgroundStepCoordinator.cs
+++ b/src/Runner.Worker/BackgroundStepCoordinator.cs
@@ -107,6 +107,8 @@ namespace GitHub.Runner.Worker
             var result = TaskResult.Succeeded;
             foreach (var (stepId, (step, _, _)) in _backgroundSteps)
             {
+                // A step that succeeded does not set a Result by default, so a missing
+                // value means the step succeeded and there is nothing to merge.
                 if (!step.ExecutionContext.Result.HasValue)
                 {
                     continue;
--- a/src/Runner.Worker/BackgroundStepCoordinator.cs
+++ b/src/Runner.Worker/BackgroundStepCoordinator.cs
@@ -16,8 +16,7 @@ namespace GitHub.Runner.Worker
     {
         void InitializeCoordinator(int maxConcurrent);
         void StartBackgroundStep(IStep step, CancellationToken jobCancellationToken);
-        Task WaitForUnwaitedStepsAsync(CancellationToken cancellationToken);
-        TaskResult GetAggregatedResult();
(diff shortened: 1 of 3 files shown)


src/Runner.Worker/BackgroundStepCoordinator.cs
```diff
@@ -94,15 +104,30 @@ public async Task<TaskResult> WaitForUnwaitedStepsAsync(CancellationToken cancel
...
+        // Merge the final results of all background steps into a single result for the caller to
+        // fold into the job result
+        public TaskResult GetAggregatedResult()
+        {
             var result = TaskResult.Succeeded;
-            foreach (var (_, (step, _, _)) in _backgroundSteps)
+            foreach (var (stepId, (step, _, _)) in _backgroundSteps)
             {
-                if (step.ExecutionContext.Result.HasValue)
+                if (!step.ExecutionContext.Result.HasValue)
```
TingluoHuang, line 112: we might want to add tiny comment to explain when `if (!step.ExecutionContext.Result.HasValue)` is true?

or merge all the the conditions?
```
if (step.ExecutionContext.Result.HasValue && !(_explicitlyCanceledStepIds.Contains(stepId) &&
                    step.ExecutionContext.Result.Value == TaskResult.Canceled))
{
result = TaskResultUtil.MergeTaskResults(result, step.ExecutionContext.Result.Value);
}
```

TingluoHuang (approved):

eeryngloriani-sudo:
Commit now

eeryngloriani-sudo:
Successful to Quic Request

eeryngloriani-sudo:
Change like datalog info 04/152026

eeryngloriani-sudo:
Exit to all bad error. Request now done true

PR #4526, opened 2026-06-25
Link config.sh and installdependencies.sh in docs

Wuodan:


Wuodan:
I had to search for `config.sh`.

PR #4537, opened 2026-07-01
Have Runner support sending context to firewall (if its running)

steiza:
With https://github.com/github-early-access/actions-native-egress-firewall/, sometimes Runner is running in a guest VM with a firewall process listening on a vsock on the host.

This change makes it so that Runner checks to see if it's running a Linux environment, if so if there's a vsock present, and if so send context to the vsock so it makes its way to the firewall (like what patterns of secrets should be masked from the log).

This is my first time writing C# code, so feedback is welcome! I was able to run this successfully as a self-hosted runner.


src/Runner.Common/HostContext.cs
```diff
@@ -60,7 +60,7 @@ public sealed class HostContext : EventListener, IObserver<DiagnosticListener>,
         private static int[] _vssHttpCredentialEventIds = new int[] { 11, 13, 14, 15, 16, 17, 18, 20, 21, 22, 27, 29 };
         private readonly ConcurrentDictionary<Type, object> _serviceInstances = new();
         private readonly ConcurrentDictionary<Type, Type> _serviceTypes = new();
-        private readonly ISecretMasker _secretMasker = new SecretMasker();
+        private readonly ISecretMasker _secretMasker = new SecretMasker(new RunnerSecretRegistrationNotifier());
```
TingluoHuang: there might be a better approach for doing this, we probably just want to send over the data after we get them from the job message, so we don't have to touch secretmasker since its job is to mask.

steiza: There's a couple of different secrets I'm worried about, which I think come from different places:

- Secrets GitHub provides as part of Actions (like `GITHUB_TOKEN`)
- Secrets a user provides for a workflow run (like things set at https://github.com/my-org/my-repo/settings/secrets/actions)
- Secrets set during the workflow run (e.g. `echo "::add-mask::$SOME_SECRET_VALUE"`)

I think we want to make sure we're getting context from all 3 sources, but also I'm very new to this codebase.

src/Runner.Common/RunnerFirewallNotifier.cs
```diff
@@ -0,0 +1,150 @@
...
+                SafeSocketHandle nativeSocket = NativeSocket((int)VsockAddressFamily, (int)SocketType.Stream, 0);
+                if (nativeSocket.IsInvalid)
+                {
+                    int error = Marshal.GetLastPInvokeError();
+                    nativeSocket.Dispose();
+                    throw new SocketException(error);
+                }
+
+                socket = new Socket(nativeSocket);
+                socket.Connect(new HostVsockEndPoint(2, 9999));
```
TingluoHuang, line 72: we probably want to use some input to provide the value, so it's not hardcoded.

src/Runner.Common/HostContext.cs
steiza: I pushed up changes reflecting what I think you were asking for - is this what you had in mind?

src/Runner.Common/RunnerFirewallNotifier.cs
```diff
@@ -0,0 +1,150 @@
...
+                    }
+                });
+
+                byte[] payloadBytes = Encoding.UTF8.GetBytes(jsonPayload);
+                byte[] lengthPrefix = BitConverter.GetBytes(IPAddress.HostToNetworkOrder(payloadBytes.Length));
+
+                lock (_vsockSendLock)
+                {
+                    SendAll(_vsock, lengthPrefix);
+                    SendAll(_vsock, payloadBytes);
```
TingluoHuang, line 49: why do we need to call `sendall` twice?
it the protocol require us doing this, or this can be combined into 1 Send call?

steiza: We could combine the byte arrays and do one call to `SendAll()`, as long as we send the 4 byte length field (network byte order) before we send the payload.

TingluoHuang:
replaced with https://github.com/actions/runner/pull/4565

PR #4550, opened 2026-07-09
Setup Job: reword locked-dependencies log line to use lockfile language

nodeselector:
Copy-only follow-up to #4546.

That PR shipped a **Set up job** log line announcing when a job runs with locked action dependencies. Based on follow-up feedback we're adjusting the wording only:

- A reviewer asked us to avoid an uppercase "Action" in user-facing copy.
- The product owner prefers **lockfile** wording over "pinning"/"pinned" language.

### Change
The line in `src/Runner.Worker/JobExtension.cs` now reads:

> Using locked actions versions from the workflow's lockfile

(previously "Running with locked dependencies")

This updates:
- the log string,
- the two matching assertions in `src/Test/L0/Worker/JobExtensionL0.cs` (positive + negative), and
- the preceding code comment (drops "pinned" in favor of lockfile language).

No behavior change — copy only.

Refs: github/actions-dispatch#567

### DIFF
--- a/src/Runner.Worker/JobExtension.cs
+++ b/src/Runner.Worker/JobExtension.cs
@@ -195,7 +195,7 @@ namespace GitHub.Runner.Worker
                     // versions from the workflow's lockfile.
                     if (message.ActionsDependencies != null && message.ActionsDependencies.Count > 0)
                     {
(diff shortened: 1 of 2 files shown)


thboop (approved):
lgtm!

ericsciple (approved):

Issue #4357, opened 2026-04-20
TempDirectoryManager race condition: cancelled Worker wipes _temp directory used by concurrently spawned new Worker


allanrogerr:
**Describe the bug**

When the runner receives a new job while a previous worker process is still running, it cancels the old worker and immediately spawns a new one. Both worker processes share the same `_temp` directory (`orgs_<org>_work/_temp`). The cancelled worker's `TempDirectoryManager` cleanup runs *after* the new worker has already created its `_runner_file_commands` pipes in that shared directory, deleting them out from under the active job. This causes the new job to fail with:

```
Missing file at path: .../_temp/_runner_file_commands/set_output_<uuid>
```

The root cause is that `JobDispatcher` spawns the new Worker process immediately upon receiving the new job request — it does not wait for the previous Worker process to fully exit and complete its `TempDirectoryManager` cleanup. This creates a window (17 seconds in our case) where two Worker PIDs are alive and operating on the same `_temp` directory.

**To Reproduce**

This is a race condition that requires two jobs to be dispatched to the same non-ephemeral self-hosted runner in quick succession. The exact sequence:

1. Runner is executing Job A (a long-running job, e.g. integration tests)
2. GitHub dispatches Job B to the same runner while Job A is still actively running and being renewed
3. Runner acknowledges Job B, logs `"We are not yet checking the state of jobrequest <Job A ID>... Cancel running worker right away."`
4. Runner sends cancellation to Job A's Worker and **immediately** spawns Job B's Worker — both PIDs are now alive
5. Job B's Worker initializes, creates `_runner_file_commands/set_output_<uuid>` and `step_summary_<uuid>` files in the shared `_temp` directory
6. Job B begins executing its first action step (e.g. `actions/checkout@v6`)
7. Job A's Worker finishes its cancellation teardown and calls `TempDirectoryManager: Cleaning runner temp folder: <shared _temp path>` — this **deletes the entire `_temp` directory contents**, including Job B's active file command pipes
8. Job B's action step fails because its `set_output` and `step_summary` files no longer exist
9. Job B exits with code 102 (runner infrastructure failure)

In our case, the gap between Job B starting (11:55:19Z) and Job A's cleanup running (11:55:36Z) was 17 seconds — plenty of time for Job B to have created and started using the file command pipes.

**Expected behavior**

[...]

PR #4553, opened 2026-07-13
Wait for worker to finish during cancel

TingluoHuang:
https://github.com/actions/runner/issues/4357

https://github.com/github/c2c-actions-support/issues/6455


src/Runner.Listener/JobDispatcher.cs
```diff
@@ -254,6 +254,8 @@ private async Task EnsureDispatchFinished(WorkerDispatcher jobDispatch, bool can
                 {
                     Trace.Error($"We are not yet checking the state of jobrequest {jobDispatch.JobId} status. Cancel running worker right away.");
                     jobDispatch.WorkerCancellationTokenSource.Cancel();
+                    // wait for worker process exit then return.
+                    await jobDispatch.WorkerDispatch;
```
TingluoHuang, line 258: this is the only place we didn't wait after `jobDispatch.WorkerCancellationTokenSource.Cancel();`

AllanGuigou (approved):

PR #4556, opened 2026-07-14
Recreate session on RunnerSessionInvalid from broker

luketomlinson:
## What

The service is changing so that when a runner polls (`GET /message`) or acknowledges with an invalid session, it returns **HTTP 400** with a structured broker error body:

```json
{"source":"actions-broker-listener","errorKind":"RunnerSessionInvalid","statusCode":400,"errorMessage":"Runner session is invalid"}
```

instead of a bare `403`. Today a bare `403` makes the runner throw a fatal, misleading `AccessDeniedException{ErrorCode=1}` ("runner deprecated") and stop.

This PR makes the broker path behave like legacy Azure Pipelines did: on an expired/invalid session the runner recreates its session and keeps polling.

This brings it into parity with the MessageListener.cs

## Changes

1. `BrokerErrorKind.cs` — add the `RunnerSessionInvalid` constant.
2. `BrokerHttpClient.GetRunnerMessageAsync` — map `RunnerSessionInvalid` to `TaskAgentSessionExpiredException`.
3. `BrokerMessageListener.GetNextMessageAsync` — mirror legacy `MessageListener`: on `TaskAgentSessionExpiredException` (and when `!SkipSessionRecover`), recreate the session and continue polling instead of throwing a non-retryable error.

Added an L0 test asserting that when `GetRunnerMessageAsync` throws `TaskAgentSessionExpiredException`, the listener recreates the session and continues polling.

## Rollout

Companion service PR: github/actions-broker-listener#2490. **The runner change must roll out BEFORE the service enables its `RejectInvalidSessions` flag.**

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/Runner.Common/BrokerServer.cs
+++ b/src/Runner.Common/BrokerServer.cs
@@ -108,7 +108,7 @@ namespace GitHub.Runner.Common
 
         public bool ShouldRetryException(Exception ex)
         {
-            if (ex is AccessDeniedException || ex is VssUnauthorizedException || ex is RunnerNotFoundException || ex is HostedRunnerDeprovisionedException)
+            if (ex is AccessDeniedException || ex is VssUnauthorizedException || ex is RunnerNotFoundException || ex is HostedRunnerDeprovisionedException || ex is TaskAgentSessionExpiredException)
             {
                 return false;
             }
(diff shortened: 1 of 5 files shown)


src/Runner.Common/BrokerServer.cs
```diff
@@ -108,7 +108,7 @@ public Task ForceRefreshConnection(VssCredentials credentials)
 
         public bool ShouldRetryException(Exception ex)
```
luketomlinson, line 109: maybe this layer is unnecessary, I'm not sure

aiqiaoy (approved):

PR #4565, opened 2026-07-22
Implement VSock secret notifier.

TingluoHuang:
Build based on https://github.com/actions/runner/pull/4537

On Linux runner, when `GITHUB_ACTIONS_RUNNER_VSOCK_CID_PORT` env is set to `2:9999`, the runner will try to connect to vsocket on CID:2 PORT:9999.

The runner will notify all secrets to the receiver of the vsocket.

For a regex secret:
`
{"type":"regex","pattern":"my-secret-pattern"}
`

For a variable secret:
`{"type":"variable","values":["password","password_base64encoded", "password_all_encoder"]}`

### DIFF
--- a/src/Runner.Common/VSockSecretNotifier.cs
+++ b/src/Runner.Common/VSockSecretNotifier.cs
@@ -60,6 +60,12 @@ namespace GitHub.Runner.Common
                 return false;
             }
 
+            var socketConnectTimeout = 5; // Default to 5 seconds
+            if (int.TryParse(Environment.GetEnvironmentVariable("GITHUB_ACTIONS_RUNNER_VSOCK_CONNECT_TIMEOUT"), out var timeout) && timeout > 0)
+            {
+                socketConnectTimeout = timeout;
+            }
+
             try
             {
                 SafeSocketHandle nativeSocket = NativeSocket((int)(AddressFamily)40, (int)SocketType.Stream, 0);
@@ -71,7 +77,11 @@ namespace GitHub.Runner.Common
                 }
 
                 _vsock = new Socket(nativeSocket);
-                await _vsock.ConnectAsync(new HostVsockEndPoint(cid, port), HostContext.RunnerShutdownToken);
+                using (var connectCancelTokenSource = CancellationTokenSource.CreateLinkedTokenSource(HostContext.RunnerShutdownToken))
+                {
+                    connectCancelTokenSource.CancelAfter(TimeSpan.FromSeconds(socketConnectTimeout));
+                    await _vsock.ConnectAsync(new HostVsockEndPoint(cid, port), connectCancelTokenSource.Token);
+                }
             }
             catch (Exception ex)
             {
--- a/src/Runner.Common/VSockSecretNotifier.cs
+++ b/src/Runner.Common/VSockSecretNotifier.cs
@@ -15,7 +15,7 @@ namespace GitHub.Runner.Common
     [ServiceLocator(Default = typeof(VSockSecretNotifier))]
     public interface IVSockSecretNotifier : IRunnerService, IAsyncDisposable
     {
-        bool TryStartNotifier();
+        Task<bool> TryStartNotifierAsync();
 
         void NotifyNewSecret(NewSecretEventArgs newSecret);
     }
@@ -30,7 +30,7 @@ namespace GitHub.Runner.Common
 
         private Channel<byte[]> _channel = Channel.CreateUnbounded<byte[]>(new UnboundedChannelOptions() { SingleReader = true });
 
-        public bool TryStartNotifier()
+        public async Task<bool> TryStartNotifierAsync()
         {
             if (_vsock != null)
             {
@@ -71,7 +71,7 @@ namespace GitHub.Runner.Common
                 }
 
                 _vsock = new Socket(nativeSocket);
-                _vsock.Connect(new HostVsockEndPoint(cid, port));
+                await _vsock.ConnectAsync(new HostVsockEndPoint(cid, port), HostContext.RunnerShutdownToken);
             }
             catch (Exception ex)
             {
--- a/src/Runner.Common/VSockSecretNotifier.cs
+++ b/src/Runner.Common/VSockSecretNotifier.cs
@@ -1,6 +1,4 @@
 using System;
-using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Net;
 using System.Net.Sockets;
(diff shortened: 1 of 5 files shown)


src/Sdk/DTLogging/Logging/SecretMasker.cs
```diff
@@ -66,6 +68,8 @@ public void AddRegex(String pattern)
                     m_lock.ExitWriteLock();
                 }
             }
+
+            NewSecretAdded?.Invoke(this, new NewRegexSecretEventArgs(pattern));
```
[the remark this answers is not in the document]
TingluoHuang: don't need to worry since `AddRegex` is only called in the beginning of the job.

src/Sdk/DTLogging/Logging/ISecretMasker.cs
```diff
@@ -11,5 +13,41 @@ public interface ISecretMasker
         void AddValueEncoder(ValueEncoder encoder);
         ISecretMasker Clone();
         String MaskSecrets(String input);
+        public event EventHandler<NewSecretEventArgs> NewSecretAdded;
```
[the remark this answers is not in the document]
TingluoHuang: not sure what do you mean.

src/Runner.Common/VSockSecretNotifier.cs
```diff
@@ -0,0 +1,201 @@
...
+        bool TryStartNotifier();
+
+        void NotifyNewSecret(NewSecretEventArgs newSecret);
+    }
+
+    public sealed class VSockSecretNotifier : RunnerService, IVSockSecretNotifier
+    {
+        private Socket _vsock = null;
+
+        private Task _secretNotificationTask = null;
```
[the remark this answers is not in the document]
TingluoHuang: fixed by adding dispose()

src/Runner.Common/VSockSecretNotifier.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        public void NotifyNewSecret(NewSecretEventArgs newSecret)
+        {
+            if (_vsock == null)
+            {
+                Trace.Verbose("VSocket is not connected, skipping secret notification.");
+                return;
+            }
+
+            byte[] payloadBytes = Encoding.UTF8.GetBytes(StringUtil.ConvertToJson(newSecret, Newtonsoft.Json.Formatting.None));
```
steiza:
```suggestion
            var wrappedSecret = new { RunnerSecrets = newSecret };
            byte[] payloadBytes = Encoding.UTF8.GetBytes(StringUtil.ConvertToJson(wrappedSecret, Newtonsoft.Json.Formatting.None));
```
Minor: the firewall is expecting the top-level JSON key to be `RunnerSecrets`

steiza (commented):
One minor comment, otherwise LGTM!

We'll have to adapt the Actions network firewall to understand this new message format, but it will safely ignore messages that it does not understand.

steiza (approved):
This works for me - thanks!

src/Runner.Common/VSockSecretNotifier.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            // we don't need to check return since unbounded channel will always accept the item.
+            _channel.Writer.TryWrite(fullPayload);
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            if (_vsock != null && _secretNotificationTask != null)
+            {
+                _cancellationTokenSource?.Cancel();
```
salmanmkc, line 114: Should we have Writer.Complete()` and drain with a timeout before cancelling/disposing the socket?

src/Runner.Common/VSockSecretNotifier.cs
```diff
@@ -0,0 +1,224 @@
...
+                            }
+                        }
+                        catch (Exception ex)
+                        {
+                            Trace.Error($"Failed to notify new secret over VSocket: {ex}");
+                        }
+                    }
+                }
+            }
+            catch (Exception ex)
```
salmanmkc, line 171: doesn't a normal shutdown get classed as an error here, might want `OperationCanceledException` specifically excluded?

src/Runner.Common/VSockSecretNotifier.cs
```diff
@@ -0,0 +1,224 @@
...
+                SafeSocketHandle nativeSocket = NativeSocket((int)(AddressFamily)40, (int)SocketType.Stream, 0);
+                if (nativeSocket.IsInvalid)
+                {
+                    int error = Marshal.GetLastPInvokeError();
+                    nativeSocket.Dispose();
+                    throw new SocketException(error);
+                }
+
+                _vsock = new Socket(nativeSocket);
+                _vsock.Connect(new HostVsockEndPoint(cid, port));
```
salmanmkc, line 76: should we have a timeout here as it could be blocking?

src/Test/L0/Worker/WorkerL0.cs
```diff
@@ -88,6 +90,7 @@ public async void DispatchRunNewJob()
                 var worker = new GitHub.Runner.Worker.Worker();
                 hc.EnqueueInstance<IProcessChannel>(_processChannel.Object);
                 hc.EnqueueInstance<IJobRunner>(_jobRunner.Object);
+                hc.SetSingleton<IVSockSecretNotifier>(_vsockSecretNotifier.Object);
```
salmanmkc, line 93: maybe we can add more tests for VSockSecretNotifier like connection failure returning false?

src/Runner.Common/VSockSecretNotifier.cs
salmanmkc: (on connect)

TingluoHuang: connectasync with timeout doesn't work for the vsock, it will blows up with not support protocol exception, i tried that first. 🫠

src/Runner.Common/VSockSecretNotifier.cs
TingluoHuang: `await _secretNotificationTask;` should do that for us, the last line of ProcessSecretChannel is `_channel.Writer.TryComplete();`

src/Test/L0/Worker/WorkerL0.cs
TingluoHuang: VSockSecretNotifier is best effort, so i think we should be fine here.

salmanmkc (approved):
lgtm thanks for addressing the comments!

PR #4618, opened 2026-08-07
Exit ephemeral runner when its job assignment is lost

### LINKED ISSUE #4617: Ephemeral runner survives a lost job assignment (409 Conflict) and becomes a ghost
### Describe the bug

An ephemeral runner that fails to acquire an assigned job with `HTTP 409 Conflict` calls `continue` and re-enters its message loop. By then the service has already removed its registration, because assigning a job to an ephemeral runner consumes it.

The result is a ghost runner: the process is alive, the container reports healthy, and the last line on stdout is `Listening for Jobs` — but `GET /orgs/{org}/actions/runners` does not list it, and it will never be assigned work again. Nothing surfaces the failure. Only a manual restart recovers it.

### To Reproduce

Hard to reproduce on demand — it is a race on job assignment. Observed on a pool of 5 org-scoped ephemeral runners handling two active monorepos. Each occurrence permanently removes one runner from the pool.

1. Run several ephemeral runners (`config.sh --ephemeral`) against one org.
2. Queue enough concurrent jobs that assignments race.
3. Wait. Runners drop out one by one, with no error on stdout.

### Expected behavior

An ephemeral runner whose job assignment is lost should exit. The supervising process (Docker restart policy, systemd, ARC) then replaces it with a freshly registered runner.

### Actual behavior

It stays in the listen loop forever, deregistered service-side.

### Runner Version and Platform

Runner: `2.334.0`
OS: Linux x64, Docker (`myoung34/github-runner:latest`)
Scope: organization runners, `--ephemeral --disableupdate`

### What's not working?

`_diag/Runner_*.log`, last lines before the runner went silent:

```
[11:55:53Z INFO Terminal] WRITE LINE: 2026-08-07 11:55:53Z: Listening for Jobs
[11:55:53Z INFO JobDispatcher] Set runner/worker IPC timeout to 30 seconds.
[11:59:45Z INFO BrokerMessageListener] Acknowledging runner request '5f7c2a52-****'.
[11:59:46Z ERR  GitHubActionsService] POST request to https://run-actions-1-azure-eastus.actions.githubusercontent.com/37/acquirejob failed. HTTP Status: Conflict
[11:59:46Z INFO Runner] Skipping message Job. Job message already acquired '5f7c2a52-****'. job assignment is invalid: MissingKey
```

Nothing after that. The process was still running 12 minutes later, and the runner was absent from the org runners API the whole time.

### Root cause

`src/Runner.Listener/Runner.cs` (main, ~L737):

```csharp
catch (Exception ex) when (
    ex is TaskOrchestrationJobNotFoundException ||          // HTTP status 404
    ex is TaskOrchestrationJobAlreadyAcquiredException ||   // HTTP status 409
    ex is TaskOrchestrationJobUnprocessableException)       // HTTP status 422
{
    Trace.Info($"Skipping message Job. {ex.Message}");
    await _acquireJobThrottler.IncrementAndWaitAsync(messageQueueLoopTokenSource.Token);
    continue;
}
```

The handler never checks whether the runner is ephemeral. Skipping is the right call for a persistent runner, but for an ephemeral one the registration is already gone, so `continue` loops on a dead session.

The analogous case ~30 lines above already handles ephemeral correctly:

```csharp
Trace.Info($"Acknowledge returned job-not-found for ephemeral runner request '{messageRef.RunnerRequestId}'. Exiting runner.");
runOnceJobCompleted = true;
return Constants.Runner.ReturnCode.Success;
```

So the intent exists for a lost acknowledge, but not for a lost acquire.

Still present on `main` as of today, i.e. after v2.336.0 — none of 2.335.0 / 2.335.1 / 2.336.0 touch this path.

### Related issues

- #4076 reports the same user-visible symptom (ephemeral runner alive, `Listening for Jobs`, never picks up work) but attributes it to a GitHub outage and has no `_diag` trace. The trace above may be one concrete cause behind it. That issue asks for a generic `--exit-after-idle` flag; the fix below is narrower and does not need a new flag.
- #4598 is the mirror case: the ephemeral runner exits and orphans the assigned job. Here it survives an assignment it never got.
- #2286 and #4498 involve a listener restart and a mid-job communication loss respectively — different paths.

### Impact

Silent, cumulative pool drain. Each hit costs one runner with no visible error — containers stay `Up` and `healthy`, stdout still reads `Listening for Jobs`. Our pool reached zero available runners with a job queued for 68 minutes before anyone noticed.

### Suggested fix

Mirror the existing ephemeral handling, but only on the two statuses that mean the assignment is gone:

```csharp
if (settings.Ephemeral &&
    (ex is TaskOrchestrationJobNotFoundException || ex is TaskOrchestrationJobAlreadyAcquiredException))
{
    _term.WriteLine("The job assigned to this ephemeral runner is no longer available. Cleaning up local configuration.");
    Trace.Info($"Ephemeral runner lost its job assignment. Exiting runner. {ex.Message}");
    skipSessionDeletion = true;
    runOnceJobCompleted = true;
    return Constants.Runner.ReturnCode.Success;
}

Trace.Info($"Skipping message Job. {ex.Message}");
await _acquireJobThrottler.IncrementAndWaitAsync(messageQueueLoopTokenSource.Token);
continue;
```

422 (`TaskOrchestrationJobUnprocessableException`) is excluded on purpose: it says the job cannot be processed, not that the assignment moved, and exiting would delete the local config of a runner that could still serve. `skipSessionDeletion` avoids a 30 s stall deleting a session whose registration the service already consumed. A persistent runner keeps the current skip-and-retry behaviour.

PR: #4618.

### Workaround

External watchdog: read `agentName` from `/actions-runner/.runner` in each running container, compare against `GET /orgs/{org}/actions/runners`, and restart any container whose name is absent for more than two minutes.

LouisCuvelier:
Fixes #4617

### Problem

When `GetJobMessageAsync` fails with 404 / 409 / 422, `Runner.cs` logs `Skipping message Job.` and `continue`s back into the message loop. That is correct for a persistent runner.

For an **ephemeral** runner it is not, on the two statuses that mean the assignment is gone: the service consumes the registration when it assigns the job, so there is no session left to listen on. The runner stays alive, deregistered, and is never assigned work again.

Nothing surfaces the failure. The process keeps printing `Listening for Jobs`, the container reports healthy, and the runner is simply absent from `GET /orgs/{org}/actions/runners`. On a pool of ephemeral runners each occurrence permanently removes one, so the pool drains to zero over hours and jobs queue with no error anywhere.

Observed on runner `2.334.0` (still present on `main`):

```
[11:55:53Z INFO Terminal] WRITE LINE: 2026-08-07 11:55:53Z: Listening for Jobs
[11:59:45Z INFO BrokerMessageListener] Acknowledging runner request '5f7c2a52-****'.
[11:59:46Z ERR  GitHubActionsService] POST .../acquirejob failed. HTTP Status: Conflict
[11:59:46Z INFO Runner] Skipping message Job. Job message already acquired '5f7c2a52-****'. job assignment is invalid: MissingKey
```

Nothing after that line. The process was still running 12 minutes later.

### Change

On **404 and 409** for an ephemeral runner, exit instead of skipping, mirroring the handling #4540 already put in place ~40 lines above for a lost acknowledge. Setting `runOnceJobCompleted` lets the existing `finally` delete the local config, so the supervising process (Docker restart policy, systemd, ARC) registers a fresh runner.

**422 is deliberately excluded.** `Unprocessable job` says the job cannot be processed, not that the assignment moved to another runner. Exiting there would delete `.runner` and `.credentials` (`ConfigurationManager.cs:496`) on a runner that could still serve, and `run-helper.sh` reads exit 0 as final — the host would never come back, and could not re-register if it did.

`skipSessionDeletion = true` is set before returning: the registration is already consumed, so `DeleteSessionAsync` would block up to 30 s (`MessageListener.cs:206`) on a call the service is bound to reject. This mirrors the registration-gone handler at line 838. It also removes a failure mode — `runOnce` (line 331) and `settings.Ephemeral` are read from different objects and can disagree after a server-pushed config migration, in which case the `catch (…) when (runOnce)` at line 878 would not swallow the resulting 401 and the exception would escape the `finally`.

The exit also writes to the terminal. A trace-only exit shows the operator `Removed .credentials`, `Removed .runner` and a bare exit 0, indistinguishable from a completed job.

**Why `settings.Ephemeral` and not `runOnce`.** `runOnce` also covers `--once` and `returnJobResultForHosted`. `--once` is a client-side flag — the service does not consume the registration, so skip-and-retry stays correct there. #4540 scoped its fix the same way. A test row pins this.

### Not covered here

Three adjacent gaps found while reviewing, each needing its own change:

- The legacy `IActionsRunServer` acquire path (empty `RunServiceUrl`, `Runner.cs:713-722`) has no handling at all; a lost assignment there escapes to `Program.MainAsync` and `run.sh` relaunches against a consumed registration.
- `RunServiceHttpClient.cs:100-121` only maps the three typed exceptions when `TryParseErrorBody` succeeds; a 409 with an unparsable body still falls through to the generic catch and the old skip.
- Exiting `Success` on a hosted runner reports 0 rather than a translated job result, so lost assignments raise no failure-code spike. The pre-existing acknowledge path at line 703 has the same shape.

### Tests

- `TestEphemeralRunnerJobRequestMessageFromRunServiceExitsOnLostJobAssignment` — a `[Theory]` over 404 and 409. Asserts the runner exits `Success`, never dispatches a job, never touches the acquire throttler, does **not** delete the session, and does delete the local config.
- `TestRunnerJobRequestMessageFromRunServiceContinuesOnLostJobAssignment` — the counterpart, over four rows: ephemeral + 422, persistent + 404, persistent + 409, and `--once` + 409. Each asserts the runner keeps listening, throttles once, drains the message, and keeps its local config.

Both are built on a shared arrange helper. Its message-loop mock parks on an empty queue instead of dequeuing, so a regression surfaces as the intended assertion rather than `InvalidOperationException: Queue empty`, and each theory row names its own `TestHostContext` so rows stop deleting each other's trace file.

Verified the tests fail against the regressions they guard:

```
# exit on all three statuses  -> the 422 row fails
# predicate widened to runOnce -> the --once row fails
# unpatched Runner.cs          -> both exit rows fail
```

`./dev.sh build` is clean and the full `./dev.sh test` suite passes, 1131/1131.


carlos2martinize (commented):
The assignment loses over the language more than the function we are mislead by

carlos2martinize (changes requested):

LouisCuvelier:
@carlos2martinize thanks for taking a look. I'm afraid I don't follow the comment — could you point at the specific line or behaviour you'd like changed?

For context, the change is deliberately narrow: on a 404 or 409 from `acquirejob`, an ephemeral runner now exits instead of returning to a message loop whose session the service has already torn down. 422 keeps the previous skip-and-retry behaviour, and a persistent runner is unaffected. Both paths are covered by tests, and the reasoning for each choice is in the PR description.

Happy to address a concrete concern if you can restate it.

PR #4636, opened 2026-08-14
Implement retry logic for HTTP requests with support for Retry-After header

TingluoHuang:
Prepare for supporting `Retry-After` header on 429 response.

### DIFF
--- a/src/Runner.Sdk/Util/UrlUtil.cs
+++ b/src/Runner.Sdk/Util/UrlUtil.cs
@@ -70,5 +70,15 @@ namespace GitHub.Runner.Sdk
             }
             return string.Empty;
         }
+
+        public static string GetRetryAfter(HttpResponseHeaders headers)
+        {
+            if (headers != null &&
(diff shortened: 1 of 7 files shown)


src/Runner.Sdk/Util/UrlUtil.cs
```diff
@@ -70,5 +70,15 @@ public static string GetVssRequestId(HttpResponseHeaders headers)
             }
             return string.Empty;
         }
+
+        public static string GetRetryAfter(HttpResponseHeaders headers)
+        {
+            if (headers != null &&
+                headers.TryGetValues("retry-after", out var headerValues))
```
luketomlinson, line 77: Is this case insensitive?

TingluoHuang: `HttpResponseHeaders` is case insensitive, so we should be fine.

luketomlinson (approved):

PR #4640, opened 2026-08-17
Install libatomic so Node.js 25+ works in the runner image and on self-hosted Linux runners

### LINKED ISSUE #4591: Node v26 not working in Docker image after installing Node due to missing libatomic
**Describe the bug**

Since Node v25 (now EOL), Node requires the `libatomic1` package. See also https://github.com/npm/cli/issues/8680

If I install Node **26** with `actions/setup-node`, I get errors such as this:

Warning: /home/runner/_work/_tool/node/26.5.1/x64/bin/node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory

**To Reproduce**

Use the latest [actions-runner](https://github.com/actions/runner/pkgs/container/actions-runner) (Linux) image, e.g. as part of ARC scale set.

Install node:
```
      - name: Setup Node.js
        uses: actions/setup-node@v6
        with:
          node-version: '26'
```

That step/action already shows the error when expanding "Environment details"

Image
 

**Expected behavior**

If you would preinstall `libatomic1` this fix the issue. It is to be expected that people install Node.js, so it should actually work.

Linux AMD64 2.336.0

yooknee:
Fixes #4591

## Problem

The official Node.js binaries link against `libatomic.so.1` starting with **Node.js 25**. Neither
the `mcr.microsoft.com/dotnet/runtime-deps:8.0-noble` base image nor a stock Debian/Ubuntu host
ships that library, so any workflow step that provisions a modern Node — most commonly
`actions/setup-node` — installs successfully and then fails the moment Node is executed:

Installing Node is a routine workflow step, so the runner should supply the library rather than
leaving every user to discover the workaround themselves.

## Verification

`readelf -d bin/node` on the official tarballs pins the cutover at Node 25, on both architectures:

| Node build | `libatomic.so.1` in `NEEDED` |
|---|---|
| v24.18.1 linux-x64 (currently bundled in `externals/`) | no |
| v24.19.0 linux-x64 | no |
| v25.9.0 linux-x64 | **yes** |
| v26.7.0 linux-x64 | **yes** |
| v26.7.0 linux-arm64 | **yes** |

The base image installs only `ca-certificates, libc6, libgcc-s1, libicu74, libssl3t64, libstdc++6,
tzdata, tzdata-legacy, zlib1g` — `libatomic1` is genuinely absent.

## Changes

**`images/Dockerfile`** — `libatomic1` added to the existing apt transaction in the runtime stage,
not a new `RUN`, so the layer count is unchanged and the existing `rm -rf /var/lib/apt/lists/*`
still applies.

**`.github/workflows/build.yml`** — regression test in the `docker` job so the library cannot
silently drop out of the image again. It runs on both the amd64 and arm64 matrix legs:

```bash
docker run --rm <image> bash -c 'set -euo pipefail; "$(command -v ldconfig || echo /sbin/ldconfig)" -p | grep -F -- "libatomic.so.1"'

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,4 +127,12 @@ jobs:
     - name: Test Docker image
       run: |
         docker run --rm ${{ github.sha }}:latest ./run.sh --version
+
+    # Regression test for https://github.com/actions/runner/issues/4591.
+    # Node.js 25+ links against libatomic.so.1, which the dotnet/runtime-deps base image does not ship,
+    # so actions/setup-node with a modern Node version breaks unless the image installs libatomic1.
+    - name: Test Docker image provides Node.js 25+ runtime dependencies
+      run: |
+        docker run --rm ${{ github.sha }}:latest \
+          bash -c 'set -euo pipefail; "$(command -v ldconfig || echo /sbin/ldconfig)" -p | grep -F -- "libatomic.so.1"'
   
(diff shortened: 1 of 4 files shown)


zxzinn:
facing the same issue

PR #4671, opened 2026-09-03
Fix run service job renewal retry

### LINKED ISSUE #4635: Run Service renewal throws NullReferenceException after a post-success failure and abandons the active job
**Describe the bug**

For Run Service jobs, if lease renewal succeeds at least once and a later renewal throws a non-terminal exception, the renewal task can fail with `NullReferenceException`. The active worker is then cancelled and the job is reported as `Abandoned`, even though the most recently acquired lease is still valid.

In the Run Service overload of `RenewJobRequestAsync`, `request` is initialized to `null` and never assigned:

```csharp
TaskAgentJobRequest request = null;

var renewResponse = await runServer.RenewJobAsync(planId, jobId, token);
```

After the first successful renewal, the exception path reads:

```csharp
remainingTime = request.LockedUntil.Value + TimeSpan.FromMinutes(5) - DateTime.UtcNow;
```

The successful response is stored in `renewResponse`, but the retry path reads `request.LockedUntil`. Therefore, a generic renewal exception after the first success causes a `NullReferenceException`.

Relevant source:

- https://github.com/actions/runner/blob/v2.336.0/src/Runner.Listener/JobDispatcher.cs#L748-L850
- https://github.com/actions/runner/blob/main/src/Runner.Listener/JobDispatcher.cs#L748-L850

The same implementation is present in `v2.336.0` and in `main` as of 2026-08-14.

**To Reproduce**

This can be reproduced at unit level:

1. Configure `IRunServer.RenewJobAsync` to return a successful `RenewJobResponse` with a future `LockedUntil`.
2. Allow the next renewal cycle to start.
3. Configure the next call to throw `HttpRequestException` or another generic exception.
4. Observe that `RenewJobRequestAsync` throws `NullReferenceException` instead of entering the lease-window retry path.
5. The completed/faulted renewal task causes `RunAsync` to cancel the worker and report the job as `Abandoned`.

**Expected behavior**

After a successful renewal, a later retriable exception should use the most recent `LockedUntil` value and continue retrying within the existing lease window, including the configured five-minute buffer. The active worker should not be cancelled while that retry window remains valid.

- Runner version: `2.336.0`
- Platform: Linux x64
- Environment: self-hosted runner in a Kubernetes Pod

## What's not working?

The issue was observed twice on the same runner during a period of TLS/network instability.

First occurrence:

```text
[2026-08-14 05:23:22Z] Successfully renew job <job-id>,
    job is valid till 08/14/2026 05:33:22

[2026-08-14 05:26:45Z] Back off ... before next retry. 1 attempt left.

[2026-08-14 05:26:55Z] Catch exception during renew runner job <job-id>.
System.Net.Http.HttpRequestException: The SSL connection could not be established.
System.IO.IOException: Received an unexpected EOF or 0 bytes from the transport stream.

[2026-08-14 05:26:55Z] Send job cancellation message to worker for job <job-id>.
[2026-08-14 05:29:31Z] Job completed with result: Abandoned
[2026-08-14 05:29:37Z] System.NullReferenceException:
    Object reference not set to an instance of an object.
    at GitHub.Runner.Listener.JobDispatcher.RenewJobRequestAsync(IRunServer runServer, ...)
```

The `RunServer` request layer made five total attempts—one initial attempt and four retries—before the exception reached `JobDispatcher`.

A second job showed the same sequence:

```text
05:51:45  Last successful renewal; valid until 06:01:45
05:55:27  Renewal exception and immediate worker cancellation
06:00:26  Job reported as Abandoned
06:08:48  NullReferenceException surfaced
```

The delayed `NullReferenceException` is surfaced when the already-faulted renewal task is later awaited; worker cancellation begins as soon as that task becomes completed.

## Job Log Output

The active workflow step was cancelled without a workflow cancellation request.

The worker received the cancellation at the same time as the renewal failure:

```text
[2026-08-14 05:26:55Z] Cancellation/Shutdown message received.
[2026-08-14 05:26:55Z] Cancel current running step.
[2026-08-14 05:26:55Z] Sending SIGINT to process <pid>.
[2026-08-14 05:27:03Z] Sending SIGTERM to process <pid>.
[2026-08-14 05:27:05Z] Kill entire process tree since both cancel and terminate signal have been ignored.
```

Full diagnostic logs contain environment-specific information and can be provided privately if needed.

**Possible fix direction**

One option would be to retain the most recent `renewResponse.LockedUntil` value and use it in the exception path instead of reading the unassigned `request`.

A regression test could cover the Run Service sequence:

```text
successful renewal -> generic exception -> retry/recovery
```

The existing Run Service test covers the terminal `TaskOrchestrationJobNotFoundException` case, but not a generic exception after a successful renewal:

https://github.com/actions/runner/blob/main/src/Test/L0/Listener/JobDispatcherL0.cs#L219-L285

The affected Run Service renewal overload was introduced in https://github.com/actions/runner/pull/2461.

ishu599:
Fixes #4635.

Run Service job renewal could throw a `NullReferenceException` when a renewal failed after a previous renewal had succeeded.

The retry path was using `request.LockedUntil`, but `request` is never initialized in the `IRunServer` renewal flow. This caused the renewal task to fail and the active job to be abandoned.

This change:
- Tracks the `LockedUntil` value from the last successful renewal.
- Uses the last successful lock expiration when calculating the retry time.
- Updates the retry logging to use the last successful lock expiration.
- Adds a regression test covering a successful renewal followed by a generic failure and a successful retry.

## Testing

- Added a regression test for Run Service job renewal retry.
- Verified the test passes with the fix.
- Verified the test fails with the previous implementation due to `NullReferenceException`.

### DIFF
--- a/src/Runner.Listener/JobDispatcher.cs
+++ b/src/Runner.Listener/JobDispatcher.cs
@@ -747,7 +747,7 @@ namespace GitHub.Runner.Listener
 
         private async Task RenewJobRequestAsync(IRunServer runServer, Guid planId, Guid jobId, TaskCompletionSource<int> firstJobRequestRenewed, CancellationToken token)
         {
-            TaskAgentJobRequest request = null;
+            DateTime? lastSuccessfulLockedUntil = null;
             int firstRenewRetryLimit = 5;
             int encounteringError = 0;
 
@@ -759,6 +759,7 @@ namespace GitHub.Runner.Listener
                 {
                     var renewResponse = await runServer.RenewJobAsync(planId, jobId, token);
                     Trace.Info($"Successfully renew job {jobId}, job is valid till {renewResponse.LockedUntil}");
+                    lastSuccessfulLockedUntil = renewResponse.LockedUntil;
 
                     if (!firstJobRequestRenewed.Task.IsCompleted)
                     {
@@ -807,7 +808,7 @@ namespace GitHub.Runner.Listener
                     else
                     {
                         // retry till reach lockeduntil + 5 mins extra buffer.
-                        remainingTime = request.LockedUntil.Value + TimeSpan.FromMinutes(5) - DateTime.UtcNow;
+                        remainingTime = lastSuccessfulLockedUntil.Value + TimeSpan.FromMinutes(5) - DateTime.UtcNow;
                     }
 
                     if (remainingTime > TimeSpan.Zero)
@@ -820,7 +821,7 @@ namespace GitHub.Runner.Listener
                         }
                         else
                         {
-                            Trace.Info($"Retrying lock renewal for job {jobId}. Job is valid until {request.LockedUntil.Value}.");
+                            Trace.Info($"Retrying lock renewal for job {jobId}. Job is valid until {lastSuccessfulLockedUntil.Value}.");
                             if (encounteringError > 5)
                             {
                                 delayTime = BackoffTimerHelper.GetRandomBackoff(TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(30));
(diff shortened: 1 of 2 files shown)


ishu599:
check plz
nullreference issue is resolved

PR #4707, opened 2026-09-11
fix(parser): handle self-repository references in comparison parser

### LINKED ISSUE #4706: Bug: self-repository references emit false validation errors in parser comparison
`$/<action-path>` self-repository references are accepted and executed successfully, but emit false `Expected format {org}/{repo}[/path]@ref` errors when action-manifest parser comparison is enabled.

## Environment

- GitHub-hosted Linux runner
- Runner version: `2.337.0`
- Composite action from a separate template repository

## Minimal example

```yaml
runs:
  using: composite
  steps:
    - name: Nested action
      uses: $/<nested-action-path>
```

## Observed behavior

```text
<org>/<template-repository>/<parent-action-path>/action.yml (Line: <line>, Col: <column>): Expected format {org}/{repo}[/path]@ref. Actual '$/<nested-action-path>'
The template is not valid. <same error>
```

Despite those annotations, the runner resolves the action as `__selfRepository`, executes the nested action successfully, and the job completes successfully. The annotation is repeated whenever the composite action is parsed.

## Expected behavior

`$/<action-path>` should not produce an `org/repo@ref` validation error.

## Likely cause

The legacy action-manifest parser recognizes `$/` before normal repository-reference validation. The newer `WorkflowTemplateConverter` parser recognizes `$/` when generating the action ID, but its later validation branch only excludes `docker://`, `./`, and `.\\`; it still validates `$/…` as `org/repo@ref`.

When `actions_runner_compare_workflow_parser` is enabled, the comparison wrapper runs both parsers. The newer parser emits the annotation and fails, but the wrapper returns the successful legacy-parser result. This explains why the job succeeds with error annotations.

## Suggested fix

Exclude valid self-repository references from the newer parser's normal reference validation:

```csharp
!GitHub.DistributedTask.Pipelines.PipelineConstants.TryParseSelfRepository(
    uses.Value, out _)
```

Please add a regression test that loads a composite action containing `uses: $/<action-path>` with parser comparison enabled and asserts that no validation annotation is emitted.

nobbs:
Accept `$/` self-repository references consistently when action-manifest parser comparison is enabled.

The newer parser previously emitted a false validation annotation and converted this reference differently from the legacy parser. This aligns validation and conversion with the legacy behavior, with a regression test covering comparison mode.

Fixes #4706

### DIFF
--- a/src/Runner.Worker/ActionManifestManagerWrapper.cs
+++ b/src/Runner.Worker/ActionManifestManagerWrapper.cs
@@ -274,6 +274,15 @@ namespace GitHub.Runner.Worker
                 };
             }
 
+            if (GitHub.DistributedTask.Pipelines.PipelineConstants.TryParseSelfRepository(uses, out var selfPath))
+            {
+                return new GitHub.DistributedTask.Pipelines.RepositoryPathReference
+                {
+                    RepositoryType = GitHub.DistributedTask.Pipelines.PipelineConstants.SelfRepositoryAlias,
+                    Path = selfPath
+                };
+            }
+
             // Repository reference: owner/repo@ref or owner/repo/path@ref
             var atIndex = uses.LastIndexOf('@');
             string refPart = null;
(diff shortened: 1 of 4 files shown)


nobbs:
Closing as a duplicate of #4669, which contains the same production fix and broader regression coverage for self-repository references in parser comparison.

PR #4718, opened 2026-09-14
Return session conflict for migrated session creation

luketomlinson:
## Summary

- Return the existing `SessionConflict` runner exit code when migrated session creation reports `CreateSessionResult.SessionConflict`.
- Preserve fallback to original `.runner` settings for generic migrated session creation failures.
- Keep shutdown cancellation, token-revoked, and hosted-deprovision paths flowing to the existing outer handlers instead of using backend fallback.

## Testing

- `cd src && ./dev.sh format`
- `cd src && PATH="$PWD/../_dotnetsdk/8.0.425:$PATH" dotnet test Test/Test.csproj --filter 'FullyQualifiedName~GitHub.Runner.Common.Tests.Listener.RunnerL0.TestRunWithMigratedSession' --no-restore`
- `cd src && PATH="$PWD/../_dotnetsdk/8.0.425:$PATH" dotnet test Test/Test.csproj --filter 'FullyQualifiedName~GitHub.Runner.Common.Tests.Listener.RunnerL0' --no-restore`

## Feature flag

No new feature flag. This is a narrow correctness fix in pre-job session creation that uses an existing public return code, and no supported job-variable feature gate is available before a job is acquired.


H-beep518:
{
  "id": "resp_1234567890",
  "object": "response",
  "created_at": 1721596428,
  "status": "completed",
  "completed_at": 1721596429,
  "error": null,
  "incomplete_details": null,
  "input": [],
  "instructions": null,
  "max_output_tokens": null,
  "model": "gpt-4o-2024-08-06",
  "output": [{
    "id": "msg_1234567890",
    "type": "message",
    "role": "assistant",
    "content": [
      {
        "type": "refusal",
        "refusal": "I'm sorry, I cannot assist with that request."
      }
    ]
  }],
  "usage": {
    "input_tokens": 81,
    "output_tokens": 11,
    "total_tokens": 92,
    "output_tokens_details": {
      "reasoning_tokens": 0,
    }
  },
}

PR #4729, opened 2026-09-19
Keep the Busy long-poll open when a job completes

### LINKED ISSUE #4728: Persistent runner drops the next job when it aborts its Busy long-poll at job completion ("was not acquired by Runner")
### Describe the bug

When a job finishes, `OnJobStatus` (`src/Runner.Listener/BrokerMessageListener.cs`, and the same code in `MessageListener.cs`) sets `_runnerStatus = Online` and cancels `_getMessagesTokenSource`, aborting the `GET /message?status=Busy` long-poll that is still open. The service assigns the next queued job to the runner at that same moment and delivers it on that open poll. Aborting the request drops the message on the client, while the service treats it as delivered.

The service redelivers roughly 61 s later. By then the assignment has expired: `acknowledge` returns 404, `acquirejob` returns 409 `MissingKey`, and the runner logs "Skipping message Job". After repeated misses the job is cancelled with the check-run annotation:

> The job was not acquired by Runner of type self-hosted even after multiple attempts

The job has no runner name and zero steps. Any aggregate job that `needs:` it then fails the workflow run.

### Evidence (runner 2.337.0, Linux x64, persistent runner, repo-scoped)

```
[2026-09-19 01:41:43Z INFO Terminal] WRITE LINE: Job Release Validation (prod) completed with result: Succeeded
[2026-09-19 01:41:43Z WARN GitHubActionsService] GET request to https://broker.actions.githubusercontent.com/message?sessionId=...&status=Busy&runnerVersion=2.337.0... has been cancelled
[2026-09-19 01:41:43Z INFO BrokerMessageListener] Get messages has been cancelled using local token source. Continue to get messages with new status.
[2026-09-19 01:42:44Z INFO BrokerMessageListener] Acknowledging runner request '07f09524-...'.
[2026-09-19 01:42:44Z ERR  GitHubActionsService] POST request to https://broker.actions.githubusercontent.com/acknowledge?...&status=Online... failed. HTTP Status: NotFound
[2026-09-19 01:42:45Z ERR  GitHubActionsService] POST request to https://run-actions-3-azure-eastus.actions.githubusercontent.com/104/acquirejob failed. HTTP Status: Conflict
[2026-09-19 01:42:45Z INFO Runner] Skipping message Job. Job message already acquired '07f09524-...'. job assignment is invalid: MissingKey
```

A fresh `status=Online` poll opens at 01:41:43 but receives nothing until the 61 s redelivery. If the job had been queued service-side, that poll would have returned it immediately, so the first delivery went out on the aborted Busy poll.

### Frequency

The client code is unchanged since 2.336.0 (our 2.336.0 logs show the same abort on every job completion). What changed is the rate, starting around 2026-09-13, which suggests a service-side change in when the next job is dispatched. `MissingKey` lines per day across two runners on one host:

| Date | `MissingKey` | Jobs run |
|---|---:|---:|
| 2026-08-26 to 2026-09-06 | 6 total | |
| 2026-09-13 | 9 | 148 |
| 2026-09-16 | 17 | 215 |
| 2026-09-17 | 31 | 229 |
| 2026-09-19 | 20 | 53 |

Most misses are redelivered to the other runner in time. The rest end as cancelled, runner-less jobs: 6 in the last three days, all in a matrix that queues about a dozen jobs against two runners, which is where job-completion handoffs are densest.

### To reproduce

Two persistent runners, one matrix of 10+ short jobs so the queue stays non-empty while jobs complete. Watch `_diag/Runner_*.log` for "cancelled using local token source" followed about 60 s later by `acquirejob` 409 `MissingKey`.

### Expected behavior

A job assigned at completion time is received and run by the runner it was assigned to.

### Suggested fix

Do not abort the in-flight poll on the Busy to Online transition. The service already knows the job completed, and it assigns work on the Busy poll, so letting that poll return delivers the job. `_runnerStatus` is still updated and goes out on the next poll. A draft PR follows.

### Related

- #4617: the ephemeral-runner side of a lost acquire (409 `MissingKey`), where the runner turns into a ghost. This issue is the persistent-runner side, where the job is lost instead.

### Runner version and platform

2.337.0 (behaviour present in 2.336.0), Linux x64, systemd-managed persistent runners, repository scope.

osazemeu:
Fixes #4728.

When a job completes, `OnJobStatus` cancels the open `GET /message?status=Busy` poll so the next poll reports `Online`. The service assigns the next queued job at that moment and delivers it on the Busy poll, so the abort drops the message. The redelivery arrives after the assignment expires (`acquirejob` 409 `MissingKey`) and the job is eventually cancelled as "not acquired by Runner".

This change keeps the poll open on the transition to `Online`. `_runnerStatus` is still updated and is sent on the next poll. The transition to `Busy` still cancels as before.

Applied to both `BrokerMessageListener` and `MessageListener`, which share the pattern.

Trade-off: the runner's displayed status can read Busy until the open poll returns (at most one long-poll interval). Job dispatch does not depend on that value, since the service already dispatches onto a Busy poll.

### Tests

New L0 test `OnJobStatus_OnlineKeepsInFlightPollOpen_BusyCancelsIt` in both `BrokerMessageListenerL0` and `MessageListenerL0`. The mocked `GetRunnerMessageAsync` / `GetAgentMessageAsync` returns a poll that stays open until the test answers it or its token is cancelled, and records the status and token of each poll. The test checks that:

- an `Online` status event leaves the in-flight Busy poll's token uncancelled;
- the next poll after that one returns reports `Online`;
- a `Busy` status event still cancels the in-flight poll, and the re-issued poll reports `Busy` and returns its message from `GetNextMessageAsync`.

Results on linux-x64 (`dotnet test` filtered to `MessageListenerL0` and `BrokerMessageListenerL0`): 19 passed, 0 failed. With the two listener files reverted to `main`, both new tests fail on the `Online` assertion, so they exercise the fix. `dotnet format --verify-no-changes` reports no changes for the four touched files.

### Feature flag

Not added. The `Features` flags in `Constants.cs` reach the runner in job messages, and `OnJobStatus` runs between jobs, so there is no flag the listener could read at this point. Happy to add an environment-variable switch if you would prefer one.


osazemeu:
Results from running this patch on two self-hosted runners (v2.337.0 + this change, Linux x64), swapped in at 2026-09-19 09:53Z.

**Before (stock 2.337.0):** `acquirejob` 409 `MissingKey` in the runners' `_diag` logs: 9 of 150 jobs on 09-13, 17 of 213 on 09-16, 31 of 231 on 09-17, and 63 of 139 on 09-19 before the swap. Six jobs were cancelled with "The job was not acquired by Runner of type self-hosted even after multiple attempts" in the three days before.

**After (patched):** 0 `MissingKey` across 349 jobs over about 2.1 days, and 0 "not acquired" cancellations across 47 workflow runs.

The only change between the two periods is this patch. Marking ready for review.

